Issue #123, opened 2026-05-27
Phase 6: Fused Metal paged-attention kernel


inureyes:
## Context

Gather-then-SDPA (Phase 2) re-materializes a contiguous K/V each decode step, so its cost grows with context length. Phase 0 (ADR 0001) confirmed that cost is material at long or batched context, so this issue replaces it with a fused Metal kernel that reads scattered blocks directly.

> **Phase 0 outcome (ADR 0001, #117):** the gather-then-SDPA cost is material at long or batched context. Layout-A overhead over contiguous SDPA is ~56% at 16k and ~67% at 32k for a single sequence, and already ~48% at 1k under batch 4 (2-3x past 4k), because batched SDPA stays cheap while the per-step gather scales with the sequence count. This kernel is therefore the planned optimization for the >=16k or batched-decode regime; until it lands, the #119 gather path is the active path and remains the correctness reference and fallback. See `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`.

## Tasks

- [ ] Implement a paged decode attention Metal kernel under `src/lib/mlx-cpp/turbo/`, modeled on `sparse_v_sdpa.metal`, consuming Q + block table + the pool tensors.
- [ ] Gate it behind `use_native_paged_kernel` plus a feature/env flag; keep gather-then-SDPA as the fallback path.
- [ ] Follow the kernel re-validation discipline in `CLAUDE.md`: compile on Apple Silicon and confirm RMS < 5e-3 vs the graph reference over 200 steps.

## Acceptance criteria

- [ ] The fused path matches the reference within RMS < 5e-3 and removes the per-step gather copy.
- [ ] Decode throughput at 16k context is within the target gap of the dense backend established in the Phase 0 ADR.

## Dependencies

Blocked by Phase 2; gated by the Phase 0 kernel-strategy decision.

Part of #116

Issue #120, opened 2026-05-27
Phase 3: Paged prefill into the block pool


inureyes:
## Context

Prefill currently writes into dense buffers and the paged path is decode-only (`is_paged_decode` requires `q_len == 1`). For an end-to-end paged sequence, prefill must allocate and write pool blocks, and it must be able to start after a shared prefix.

> **Phase 0 outcome (ADR 0001, #117):** prefill writes into the same **layout A** pool established in #118. Block writes must reassign the pool tensor (`pool = slice_update(pool, ...)`) so MLX donates the buffer and writes are O(block) in place rather than O(pool), the same append discipline #118 sets. See `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`.

## Tasks

- [x] Allocate blocks during prefill and write prefill K/V into the pool. (`PagedBlockPool::write_prefill`)
- [x] Support starting prefill after a shared prefix: reference the matched prefix's shared blocks and write only the divergent suffix into fresh blocks. (writes at the absolute tail; real copy-on-write of a shared partial tail block via `copy_on_write_block`)
- [x] Keep prefill attention numerically identical to the dense prefill path (no change to logits). (byte-identical pool storage; pooled vs dense decode parity max RMS = 0 over the prefill→decode test)
- [ ] Decide whether to keep the dense prefill fast path for single-stream / non-batched runs, or unify on the pool. (deferred to #121 — the live forward/scheduler wiring decision)

## Acceptance criteria

- [x] A cold request and a request that shares a prefix both produce logits identical to the dense path. (proven at the pool layer: cold round-trip byte-identical to dense; shared-prefix suffix gathers correctly; end-to-end prefill→decode parity RMS = 0)
- [x] The shared-prefix request allocates blocks only for its suffix (verified via `PagedCacheStats`).

> **Scope note (#120 is pool-layer only).** The model `forward(&self, caches: &mut [KVCache], ...)` only sees dense `KVCache`; the pool is not reachable from the live forward until **#121** wires a paged-aware cache mode + scheduler. So `write_prefill` is the additive pool-layer prefill-write capability + tests, not live-wired into the model forward — the same additive/deferred pattern as #118/#119. Removing dense placeholders and the dense-vs-pool fast-path decision are #121; the fused C++ kernel is #123.

## Dependencies

Blocked by Phase 1 and Phase 2.

Part of #116

Issue #119, opened 2026-05-27
Phase 2: Paged decode attention over real block tables


inureyes:
## Context

The decode kernel currently slices per-sequence dense buffers using an identity block table built by `PagedDecodeMetadata::from_visible_lengths` (`[0, 1, 2, ...]`). With real pool storage in place, attention must gather the sequence's actual, possibly non-contiguous, physical blocks.

> **Phase 0 outcome (ADR 0001, #117):** option A is **gather-then-SDPA** — `take(pool, block_ids, axis=0)` (layout A), then `reshape`, then `transpose [0,2,1,3]`, then fused `scaled_dot_product_attention`, reusing existing FFI with no new kernel. MLX fuses the gather into the SDPA read, so gather+SDPA stays within ~15% of contiguous SDPA at <=4k single-sequence context (it climbs with context and batch; the fused-kernel alternative is deferred to #123). Build the gather on the layout-A pool from #118. See `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`.

## Tasks

- [ ] Replace the identity tables with the sequence's real `PagedSequenceState` block tables in the decode dispatch (`src/models/model_owned.rs` and the per-model paths in `qwen3.rs` / `llama3.rs` / `gemma3.rs` / `llama4.rs`).
- [ ] Update `paged_decode_attention_dense_compat` (C++ in `cpp/mlx_cxx_bridge.cpp`) and the Rust fallback `paged_decode_attention_dense_fallback` (`layers.rs`) to gather by physical block id from the pool (Phase 0 option A).
- [ ] Apply the same change to the rotating / sliding-window variant `paged_decode_attention_rotating_compat`.

## Acceptance criteria

- [ ] Decode output with a fragmented (non-contiguous) block table matches the dense-backend reference within RMS < 5e-3 over 200 steps.
- [ ] Supported models (llama3/llama4/qwen3/qwen3.5/gemma3 and their VLM wrappers) are functionally unaffected.

## Dependencies

Blocked by Phase 1 (global block-pool tensor storage).

Part of #116

Issue #116, opened 2026-05-27
[Epic] Unified KV cache: radix-tree-indexed paged attention with shared physical blocks


inureyes:
## Summary

Merge mlxcel's two independent KV-cache subsystems, the server-side prompt-prefix **radix cache** and the core-side **paged block pool**, into one unified, refcounted, copy-on-write KV store in the style of vLLM PagedAttention and SGLang RadixAttention. End state: physical KV lives in a shared pool of fixed-size blocks; a radix tree indexes those blocks by token prefix; concurrent requests that share a prefix point at the same physical blocks (refcount, no copy); and a gathering attention path reads scattered blocks at decode time.

## Motivation

Today prefix reuse and paged batching cannot run at the same time. `BatchScheduler::try_adopt_cached_prefix` returns `None` whenever the active decode backend is `Paged` (`src/server/batch/scheduler.rs`), and `docs/en/prompt_cache.md` documents this as a hard limitation ("Paged backend adopt/donate disabled ... Dense backend only"). A batching server therefore gets either cross-request prefix reuse (dense backend) or paged memory management, never both. With paged batching on, a multi-tenant server pays full prefill for every concurrent request that shares a system prompt and stores a duplicate copy of that prefix KV per sequence.

A unified store fixes both: a shared prefix is stored once and reused across concurrent requests without re-prefill, and block-granular allocation removes per-sequence contiguous-buffer over-reservation.

Apple Silicon note: unified memory removes the CPU/GPU swap-out motivation behind parts of vLLM's design, so the win here is memory sharing and prefill avoidance, not host offload. The main risk is decode-time gather cost on MLX without a fused kernel; Phase 0 measures this before committing to the fused-kernel work.

## Current architecture

1. **Dense KVCache** (`cache.rs`, `KVCache`): pre-allocated contiguous per-sequence K/V buffers with `slice_update` appends. Default mode `KVCacheMode::Fp16`. Default storage for normal generation.
2. **Paged block pool** (`cache/paged.rs`: `PagedBlockPool` / `PagedSequenceState` / `PagedKvLayout`): refcounted physical-block allocator with free lists. In Fp16/Int8 modes it tracks block-table metadata only; the actual K/V tensors still live in dense `KVCache` placeholders (see the `PagedBlockPool` doc comment). Default block size 32.
[...]

PR #148, opened 2026-06-02
feat: add physical block-pool K/V tensor storage to PagedBlockPool

### LINKED ISSUE #118: Phase 1: Global block-pool tensor storage
## Context

Physical K/V tensors currently live in per-sequence dense `KVCache` placeholders; in Fp16/Int8 paged mode the `PagedBlockPool` tracks only block-table metadata (see its doc comment in `cache/paged.rs`). Move the actual tensor storage into the shared pool so a sequence's KV is a list of physical block ids rather than a contiguous buffer.

> **Phase 0 outcome (ADR 0001, #117):** the pool tensor layout is **layout A** `[num_blocks, block_size, n_kv_heads, head_dim]` (per-layer), chosen over the head-split layout (~2.1x faster on gather+SDPA in the microbench); keep the default block size 32. The block append must reassign the pool tensor (`pool = slice_update(pool, ...)`) so MLX donates the buffer and the write is O(block) in place. The microbench's O(pool) `slice_update` figures are the no-donation upper bound, not the real append cost. See `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`.

## Tasks

- [ ] Add per-layer pool tensors to `PagedBlockPool` (`cache/paged.rs`) using the layout chosen in Phase 0.
- [ ] Implement token writes into `pool[block_id, slot, ...]` via `slice_update` / `scatter`; extend `append_tokens` / `acquire_block` to write into pool storage instead of the dense placeholder.
- [ ] Implement visible-block gather (`take(pool, block_ids, axis=0)` reshaped to `[1, H, T, D]`).
- [ ] Route Fp16 and Int8 through pool storage; keep the Turbo4 per-page sidecars (`PagedTurboPageSidecars`) consistent with the new storage so detach/adopt still round-trips bit-identically.
- [ ] Update `nbytes` / `PagedCacheStats` accounting to read from pool tensors.

## Acceptance criteria

- [ ] A paged sequence stores KV only in the pool, with no dense placeholder duplication.
- [ ] For a contiguous block table, gathered visible K/V is byte-identical to the previous contiguous-slice result; existing paged decode tests pass.

## Dependencies

Blocked by Phase 0 (#116 child: feasibility spike).

Part of #116

inureyes:
## Summary

Phase 1 of epic #116 (unified paged KV cache). Adds the physical main K/V tensor storage to `PagedBlockPool` plus the write/gather primitives, byte accounting, and unit tests, following layout A and the gather-then-SDPA strategy locked by ADR 0001 (`docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`).

This is an additive capability that later phases consume. The live model/decode flow is intentionally not rewired here (decode-read is #119, prefill-write is #120, fused kernel is #123), per the project's integration rule that integration depending on later sub-issues is deferred to them. Net effect in live operation after this PR: pool tensors stay unallocated (no live writer wired yet), dense `KVCache` placeholders still carry live K/V, the existing paged decode path is byte-for-byte unchanged, and all existing paged tests pass. The new storage is exercised and proven only by the new unit tests, ready for #119/#120 to wire.

## What changed

- `src/lib/mlxcel-core/src/cache/paged.rs`
  - New per-layer physical pool state on `PagedBlockPool`: `pool_k` / `pool_v` (`Vec<Option<UniquePtr<MlxArray>>>`, layout A `[capacity_blocks, block_size, n_kv_heads, head_dim]`, separate K and V per layer), `pool_meta` (lazily-inferred `n_kv_heads`/`head_dim`/`dtype`/`capacity_blocks`), `block_rows` (per-layer `block_id -> row`), `free_rows`, `next_row`. `PagedBlockPool::new(layout)` signature is unchanged (tensors start `None`).
  - `write_block(block_id, layer_idx, slot_start, k_block, v_block)` — lazily allocates the layer pool on first write, assigns/looks up the layer-local row (reusing freed rows), grows in `POOL_GROW_CHUNK_BLOCKS` (32) chunks when a fresh row exceeds capacity, and writes via `slice_update` reassigning the pool tensor so MLX donates the buffer (O(block) append per ADR 0001). Accepts both `[1, n_kv_heads, n_slots, head_dim]` (SDPA-style, primary) and `[n_slots, n_kv_heads, head_dim]` (bare slab) block layouts; validates shape/dtype against captured geometry on every write.
  - `gather_visible(state, layer_idx)` — builds the physical-row index from the layer's block table (fragmented / out-of-order rows gather in block-table order), `take`s the rows, flattens, slices the visible `[logical_start, len)` window, and transposes into `[1, n_kv_heads, visible_len, head_dim]`, byte-identical to the equivalent dense contiguous buffer's visible slice. Returns `Ok(None)` for an empty layer.
  - `pool_tensor_bytes()` — sums `array_nbytes` over allocated `pool_k`/`pool_v`.
  - `release_block` now frees the main-K/V row (push to `free_rows`, remove from `block_rows`) in addition to the existing `turbo_sidecars.forget`.
  - No astype on the K/V path: dtype is preserved through `transpose_axes`/`reshape`/`slice_update`/`take`/`slice`. Fp16 stays Fp16, Int8 stays Int8. Int8 quantization scales stay in the dense path and are not touched.
  - Updated the `PagedBlockPool` doc comment to state the pool can now own physical main K/V storage (layout A) and that dense-placeholder removal lands with #119/#120.
- `src/lib/mlxcel-core/src/cache.rs`
  - Re-export the new `GatheredKv` type alias; wire the `paged_pool_tests` module.
  - `CachePool::memory_usage_bytes` adds `pool.pool_tensor_bytes()` additively next to `turbo_sidecar_bytes()`; the layout-derived `reserved_bytes`/`used_bytes`/`stats_for_sequences` budgets are unchanged.
- `src/lib/mlxcel-core/src/cache/paged_pool_tests.rs` (new) — 13 unit tests covering the acceptance criterion and all required cases.

## Test plan

- [x] `cargo check --lib -p mlxcel-core --features metal,accelerate` — clean
- [x] `cargo clippy --lib -p mlxcel-core --features metal,accelerate -- -D warnings` — clean
- [x] `cargo clippy -p mlxcel-core --tests --features metal,accelerate -- -D warnings` — clean
- [x] `cargo test --lib -p mlxcel-core cache::paged_pool --features metal,accelerate` — 13 passed
- [x] `cargo test --lib -p mlxcel-core cache::paged_turbo --features metal,accelerate` — 26 passed (existing `nbytes_*` budget assertions intact)
- [x] `cargo test --lib -p mlxcel-core cache::paged --features metal,accelerate` — 57 passed (detach round-trips intact)
- [x] `cargo fmt -p mlxcel-core --check` — clean

Closes #118

### DIFF
--- a/src/lib/mlxcel-core/src/cache.rs
+++ b/src/lib/mlxcel-core/src/cache.rs
@@ -80,6 +80,9 @@ mod detach;
 mod paged;
 mod paged_detach;
 #[cfg(test)]
+#[path = "cache/paged_pool_tests.rs"]
+mod paged_pool_tests;
+#[cfg(test)]
 #[path = "cache/paged_turbo_tests.rs"]
 mod paged_turbo_tests;
 #[cfg(test)]
@@ -96,7 +99,7 @@ pub use batch_quant::{
 };
 pub use detach::{DetachedCacheSet, DetachedHandle, DetachedKVCache, DetachedRotatingKVCache};
 pub use paged::{
-    PagedBlockId, PagedBlockPool, PagedCacheStats, PagedKvLayout, PagedLayerState,
+    GatheredKv, PagedBlockId, PagedBlockPool, PagedCacheStats, PagedKvLayout, PagedLayerState,
     PagedSequenceState,
 };
 pub use paged_detach::DetachedPagedCacheSet;
@@ -5365,17 +5368,20 @@ impl CachePool {
     pub fn memory_usage_bytes(&self) -> usize {
         let active_bytes: usize = self.active.values().map(|s| s.nbytes()).sum();
         let parked_bytes: usize = self.detached.values().map(|d| d.nbytes()).sum();
-        // Per-page Turbo4 sidecars in the paged pool are owned by the pool
-        // itself rather than by any individual `SequenceCacheSet`, so they
-        // are not visible to the per-sequence `nbytes()` walk above. Add
-        // them explicitly so admission-control sees the true KV footprint
-        // for paged Turbo4 deployments.
-        let pool_sidecar_bytes: usize = self
+        // Per-page Turbo4 sidecars and physical main-K/V pool tensors in the
+        // paged pool are owned by the pool itself rather than by any individual
+        // `SequenceCacheSet`, so they are not visible to the per-sequence
+        // `nbytes()` walk above. Add them explicitly so admission-control sees
+        // the true KV footprint for paged deployments. The main-K/V pool
+        // tensors are lazily allocated, so this contributes 0 until a writer is
+        // wired (#120) and never perturbs the layout-derived scheduling budgets
+        // (`reserved_bytes`/`used_bytes`).
+        let pool_bytes: usize = self
             .paged_pool
             .as_ref()
-            .map(|pool| pool.turbo_sidecar_bytes())
+            .map(|pool| pool.turbo_sidecar_bytes() + pool.pool_tensor_bytes())
             .unwrap_or(0);
-        active_bytes + parked_bytes + pool_sidecar_bytes
+        active_bytes + parked_bytes + pool_bytes
     }
 
     pub fn append_paged_tokens(
--- a/src/lib/mlxcel-core/src/cache/paged.rs
+++ b/src/lib/mlxcel-core/src/cache/paged.rs
@@ -390,14 +390,42 @@ impl PagedTurboPageSidecars {
     }
 }
 
+/// Inferred per-layer geometry of a physical main-K/V pool tensor.
+///
+/// `PagedKvLayout` only carries the per-block byte budget; the head count,
+/// head dim, and element dtype are not known until the first block is written.
+/// They are captured here on the first [`PagedBlockPool::write_block`] for the
+/// layer and then validated against on every subsequent write.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct PagedPoolMeta {
+    n_kv_heads: i32,
+    head_dim: i32,
+    /// MLX dtype code of the stored K/V (e.g. `dtype::FLOAT16`, `dtype::INT8`).
+    dtype: i32,
+    /// Number of physical block rows the current pool tensor can hold along
+    /// axis 0. Grown in chunks as new rows are assigned.
+    capacity_blocks: usize,
+}
+
 /// Physical block allocator shared across every active paged sequence.
 ///
-/// In `Fp16`/`Int8` modes the pool tracks block-table metadata only — the
-/// actual KV tensors live in dense [`super::KVCache`] placeholders attached to
-/// each [`super::SequenceCacheSet`]. In Turbo4 modes the pool also owns per-
-/// page sidecar MLX arrays (`v_packed`, `v_norms`, optionally `k_packed`,
-/// `k_norms`, `cold_keys`) so packed quantization state survives across
-/// detach/adopt round-trips and prefix-cache handoffs.
+/// The pool can own the physical main K and V tensors (layout A from
+/// ADR 0001: each block occupies `[block_size, n_kv_heads, head_dim]` and a
+/// layer's pool tensor is `[capacity_blocks, block_size, n_kv_heads, head_dim]`,
+/// separate tensors for K and V per layer). Storage is lazily allocated on the
+/// first [`PagedBlockPool::write_block`] for a layer, so an `Fp16`/`Int8`
+/// sequence that never writes (the current live flow, where dense
+/// [`super::KVCache`] placeholders still carry live K/V) adds zero tensor or
+/// row overhead. Removal of the dense placeholders from the live flow lands
+/// with the decode-read (#119) / prefill-write (#120) reader/writer switch;
+/// until then the pool storage added here is exercised only by unit tests.
+///
+/// In Turbo4 modes the pool also owns per-page sidecar MLX arrays (`v_packed`,
+/// `v_norms`, optionally `k_packed`, `k_norms`, `cold_keys`) so packed
+/// quantization state survives across detach/adopt round-trips and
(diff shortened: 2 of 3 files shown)


inureyes:
## Security & Performance Review — Phase 1 block-pool storage (#118)

Reviewed scope: the three changed files (`cache/paged.rs`, `cache.rs`, `cache/paged_pool_tests.rs`). In-memory single-owner (`&mut self`) cache code with no network/parse/injection surface, so the review focused on **memory-safety, donation correctness, integer-overflow/bounds, panic-safety, and stale-data on row recycle**. Live-flow wiring deferred to #119/#120 is treated as in-design, not a defect.

**Verdict: no CRITICAL or HIGH findings. Nothing auto-fixed (no fix warranted). PR stays at `status:review`.**

Warm verification (all clean, no new compile/lint/test surface introduced):
- `cargo check --lib -p mlxcel-core` — clean
- `cargo clippy --lib` and `--tests -- -D warnings` — clean
- `cargo fmt --check` — clean
- `cargo test --lib cache::paged` — 56 passed (12 new pool tests + 44 existing; detach round-trips and `nbytes_*` budgets intact)

### Findings

**1. Unbounded growth / retention — LOW (acceptable, not a leak).** The pool grows in 32-row chunks via `grow_pool` and never shrinks; `next_row` is monotonic. This is standard pool high-water retention, identical to the dense `KVCache` pre-allocate-and-keep discipline. Freed rows are recycled correctly: `release_block` pushes to `free_rows`, and `assign_row` pops `free_rows` **before** advancing `next_row`, so capacity is bounded by peak concurrent blocks (itself bounded by admission control, which now sees `pool_tensor_bytes()`). No `UniquePtr<MlxArray>` is leaked — every `.take()` is immediately followed by a reassign that consumes the old array into a `slice_update` graph node (freed at eval). Acceptable.

**2. Donation correctness / no hidden O(pool) copy — CORRECT.** `write_block` and `grow_pool` both follow the ADR 0001 donation pattern: `let old = pool[i].take(); pool[i] = Some(slice_update(&old, ...))`. After `.take()`, `old` is the sole Rust owner; the FFI passes `const MlxArray&` (no `mlx::core::array` copy), the lazy `Slice` primitive captures `src.inner`, and `old` drops before `eval`, so the array's refcount is 1 at eval and MLX donates the buffer in place (O(block) append). In `grow_pool` the freshly-`zeros`'d larger buffer is the donation target and `old` is the read-only update source — the one genuine data copy here is unavoidable on a real resize. Matches the proven dense-cache `slice_update` path. No lingering reference forces a full copy on the steady-state append.

**3. Integer overflow / bounds — LOW (defense-in-depth only; no memory-safety risk).** All hot-path arithmetic is sound: `next_row += 1` and `row + 1` need 2^64 blocks to wrap; `grow_pool` uses `div_ceil` + `saturating_mul` + `.max()`; growth trigger `row >= capacity ⇒ grow(row+1)` has correct 0-indexed boundary (no off-by-one). The only externally-arbitrary value not pre-bounded before an `as i32` cast is `write_block`'s `slot_start` (`slot_start as i32 + n_slots`): a pathological `slot_start ≥ ~2^31` would wrap (debug: panic; release: wrap) and could bypass the `> block_size` check. Likewise `n_blocks * block_size` in `gather_visible` is a theoretical i32 product overflow at absurd context lengths. **Neither can cause out-of-bounds memory access**: there is no `unsafe` and no raw indexing — every slice/index goes through MLX, which bounds-checks at the C++ layer and throws rather than corrupting memory. Realistic callers (#119/#120 prefill/decode) always pass `slot_start < block_size` (default 32). Optional future hardening: validate `slot_start` as `usize` against `block_size` before the cast. Not blocking.

**4. New `unsafe` — CLEAN.** Zero `unsafe` blocks in the new code (confirmed by scan). The K/V path uses only the safe FFI wrappers `take`/`slice_update`/`slice`/`reshape`/`transpose_axes`/`from_slice_i32`/`zeros`/`array_nbytes` — never the `unsafe` `gather`/`concatenate`/`stack` variants.

**5. Panic safety (`.expect()` in non-test code) — CORRECT.** The five new `.expect()` calls (`write_block` pool_k/pool_v takes, `gather_visible` pool_meta, `grow_pool` pool_k/pool_v takes) are each guarded by an invariant established immediately prior in the same `&mut self`/`&self` method: `pool_k`/`pool_v`/`pool_meta` are always allocated together (the `match pool_meta { None => set all three }` arm), so "meta present ⇒ tensors present" and "tensors present ⇒ meta present" both hold. No interior mutability and no await points, so no scheduler/attacker-influenced interleaving can violate them. Acceptable.

**6. Stale-data leak on row recycle — CORRECT (no leak path).** A recycled row is not zeroed, but it cannot serve stale data: `release_block` removes the old `block_id` from `block_rows`, and `gather_visible` returns an **error** for any block with no row entry, so the window between re-acquire and re-write cannot gather. After re-write, the row holds fresh data. Trailing padding slots of a partially-rewritten recycled row are never exposed because `gather_visible` slices exactly `[logical_start, len)` (verified by `partial_final_block_gathers_exactly_visible_len`). The `released_row_is_recycled_and_serves_fresh_data` test confirms the full free→recycle→rewrite→gather path serves fresh bytes only.

### Notes (informational, not findings)
- The write-before-read ordering for visible slots in `[logical_start, len)` is a contract for the #119/#120 writer (a row whose entry exists from a prior occupant but whose current-generation slots were not rewritten would gather stale bytes). Correctly out of #118 scope; the only #118 caller (unit tests) always writes every visible token first.
- `CachePool::memory_usage_bytes` adds `pool_tensor_bytes()` additively and contributes 0 until a live writer is wired (#120); it does not perturb the layout-derived `reserved_bytes`/`used_bytes` scheduling budgets. Accurate.
- PR body says "15 unit tests"; the file actually contains 12 `#[test]` functions (all passing). Cosmetic description mismatch only.

PR #162, opened 2026-06-04
perf: support variable-length prompts in B>1 batched MTP bursts

### LINKED ISSUE #161: perf: support variable-length prompts in B>1 batched MTP speculative bursts
## Summary

Support **variable-length prompts in a single B>1 batched MTP speculative burst**. Today the batched burst only groups concurrent requests whose prompt length is identical; a realistic mix of prompt lengths serializes into per-request B=1 bursts that head-of-line-block each other, which makes forced B>1 MTP *slower* than classic batched decode under realistic load.

## Evidence (measured, M5 Max, 31B + bf16 assistant, 4 concurrent, temperature 0)

| Case | classic agg tok/s | B>1 MTP agg tok/s | speedup | parity |
| ---- | ----------------: | ----------------: | ------: | ------ |
| same-length window (true B>1 burst) | 31.8 | 33.7 | 1.06x | identical |
| variable-length mix (realistic) | 32.0 | 24.8 | **0.78x** | identical |

In the variable-length run the four bursts complete at `burst_ms` ≈ 3800 / 8688 / 25680 / 25687 — i.e. serialized, each blocking the next. In the same-length run all four finish together (`burst_ms` ≈ 18.9 s each), confirming a single B>1 window formed. See `docs/benchmark_results/gemma4-mtp-speculative-decoding.md`.

Because of this, B>1 batched MTP stays off by default (`MLXCEL_ENABLE_MTP_BATCH=1` forces it). Resolving the variable-length case is the prerequisite to making batched MTP worthwhile.

## Root cause

The batched-burst window is formed in `src/server/batch/scheduler.rs` (~lines 1824–1852). Candidates may join the head's window only if every field matches, including an exact prompt-length equality:

```rust
self.prefill_queue.drain_matching_window(head_lane, max_extra, |candidate| {
    candidate.prompt_tokens.len() == head_prompt_len   // <-- the constraint
        && candidate.max_tokens == head_max_tokens
        && super::speculative_burst::sampling_config_eq(&candidate.sampling, &head_sampling)
        && super::speculative_burst::should_burst_for_sequence(dispatch, candidate)
        && super::speculative_burst::can_join_batched_burst_window(candidate)
});
```

A request whose prompt length differs from the head is left in the prefill queue. It later runs as its own B=1 burst, and (per the server-startup notice) "head-of-line-blocks concurrent classic-decode rows for its full duration." The batched MTP verify path assumes a rectangular `[B, prompt_len]` window, so ragged prompt lengths are not currently representable in one burst.

## Proposed approach

Allow a B>1 batched MTP burst to span rows with different prompt lengths by making the batched verify ragged-aware:

1. **Relax the window predicate** (`scheduler.rs` ~1835): drop the `prompt_tokens.len() == head_prompt_len` equality (keep `max_tokens` / sampling / lane matching, or relax those too as separate steps). Group up to `max_batch_size` burst-eligible rows regardless of prompt length.
2. **Left-pad the batched prefill/verify** to `max(prompt_len)` across the window, with a per-row attention mask so padding positions do not attend, and per-row valid lengths threaded into `prefill_and_seed`. Gemma 4's sliding-window + K=V + proportional-RoPE attention mask construction must honor the per-row offset.
3. **Per-row positions / RoPE anchors**: the batched `MtpTarget` already preserves per-row cache positions and RoPE anchors after mixed accepts (see `mtp_batched_burst_enabled` docs and `Gemma4MtpBatchedTargetAdapter`). Extend that so each row's *initial* position reflects its own (unpadded) prompt length, not a shared one.
4. **Per-row finish / rollback**: rows in the same window may accept different numbers of draft tokens per round and may hit EOS / `max_tokens` at different steps; the batched generator must retire finished rows and continue the rest without collapsing the window to B=1.

## Touchpoints

- `src/server/batch/scheduler.rs` — window formation (`drain_matching_window` predicate, ~1834) and the B>1 decline guard.
- `src/server/batch/speculative_burst.rs` — `can_join_batched_burst_window`, `sampling_config_eq`, the batched burst driver, and the startup notice describing the limitation.
- `src/models/gemma4_mtp_target.rs` — `Gemma4MtpBatchedTargetAdapter` / `Gemma4UnifiedMtpBatchedTargetAdapter` (batched verify forward, per-row positions).
- `src/lib/mlxcel-core/src/speculative/mtp/` — `MtpBatchedGenerator` (per-row accept/rollback/retire).
- Attention-mask construction for batched ragged prefill (Gemma 4 sliding-window / full-attention paths).

## Acceptance criteria

- Concurrent requests with **different prompt lengths** form a single B>1 batched MTP burst (no serialized head-of-line-blocking B=1 bursts), verified from server logs (one batched burst, rows finishing together rather than sequentially).
- On the 31B + bf16 assistant at `temperature 0`, a 4-request variable-length window is **at least as fast as classic batched decode** (target: clearly >1.0x, recovering the same-length ~1.06x or better), measured via the same harness in `docs/benchmark_results/gemma4-mtp-speculative-decoding.md`.
- **Greedy parity preserved**: every row's `temperature 0` output is byte-identical to its classic-decode output.
- Once variable-length batched MTP is consistently faster than classic batched decode, evaluate flipping `MLXCEL_ENABLE_MTP_BATCH` to default-on (mirroring the B=1 default-on change), keeping the env var as an opt-out.

## Notes

- This is the follow-up referenced as "Variable-length-prompt batched bursts are a documented follow-up" in the speculative server-startup notice.
- B=1 MTP already runs by default and is profitable (~1.2–1.4x on the 31B, ~1.87x on the 12B Unified); this issue is specifically about the batched (B>1) concurrent-serving path.

inureyes:
## Summary

Implements variable-length-prompt support for B>1 batched MTP speculative bursts (issue #161), behind a new `MLXCEL_ENABLE_MTP_BATCH_RAGGED` opt-in subordinate to `MLXCEL_ENABLE_MTP_BATCH`. Today the batched burst only groups concurrent requests whose prompt length is identical, so a variable-length mix serializes into head-of-line-blocking B=1 bursts (~0.78x). This lets burst-eligible rows of different prompt lengths form a single B>1 batched MTP burst, served via left-padding to `max_prompt_len`.

The production default path — including `MLXCEL_ENABLE_MTP_BATCH=1` alone — is unchanged: the validated same-length batched burst stays the default behaviour until the operator also sets the new ragged opt-in. This keeps the work strictly out of the production default while the orchestrator's real-model greedy-parity gate validates the on-hardware behaviour.

## Real-model validation (31B, M5 Max)

**Greedy parity: PASSED.** Four variable-length concurrent requests at temperature 0 form a single ragged B>1 burst and every row's output is byte-identical to classic decode. The previously-broken most-padded row (whose prompt padding was not masked on the single-token `l==1` verify path) now produces the correct output and hits EOS at the right position. The root fix was ensuring the resident prompt-padding mask is applied consistently on the single-token verify forward, not only on the multi-token path.

**Throughput: marginal and not a consistent win.** Measured ~0.94–1.13x vs classic batched decode across runs (confounded by short-row early-EOS). This recovers the prior variable-length serialization penalty (~0.78x) but is not a clear speedup, so `MLXCEL_ENABLE_MTP_BATCH_RAGGED` stays off by default. The issue's conditional default-on flip remains deferred pending a more definitive throughput measurement.

## Greedy-parity mechanism (the must-have)

Parity is preserved **by construction** via the left-padding *uniform per-row position shift*:

- Row `r` (length `L_r`) is left-padded to `max_len`, so its real tokens occupy padded indices `[lp_r, max_len)` with `lp_r = max_len - L_r`. Proportional RoPE is applied with a single scalar offset (`0` at prefill -> position == padded index), so every real token of row `r` is shifted right by the same constant `lp_r` versus a standalone B=1 run that would place it at `[0, L_r)`. Because attention depends only on the relative RoPE phase `q_pos - k_pos`, and both query and key of any in-row pair carry the same `+lp_r` shift, every in-row relative phase — and thus every attention score and every argmax — is preserved. The left-padding mask blocks every real query from attending to the leading padding keys.
- In subsequent verify rounds the shared physical cache offset is uniform for every row (unchanged from the same-length path), so each row's appended verify tokens carry the same `+lp_r` shift end-to-end.
- Seed/finalize metadata is unified through `seed_anchors_from_valid_len` / `build_seed_metadata` with the shifted-frame formula `kv_offset = left_padding + kv_valid_len`, `bonus_position = kv_offset - 1`. The drafter's `set_shared_kv_batched` (already per-row aware) normalizes each row's K/V to the prefix-valid layout via `kv_valid_len` + `left_padding`, and rotates its query at `position_per_row[r] = lp_r + kv_valid_len[r] - 1`. The `lp_r` cancels in every query-key difference, so the drafter's cross-attention relative phases are byte-identical to the same-length path. For equal-length rows (`left_padding == 0`) the formula reduces to the legacy `kv_offset == kv_valid_len` metadata exactly, so the same-length burst is byte-for-byte unchanged.

Net: the ragged path is parity-correct iff the existing same-length batched path is, modulo the prefill left-padding (proven correct) and a metadata shift (shown to cancel). It rides entirely on the same-length foundation the orchestrator already validates.

## CRITICAL fix: out-of-regime windows now fall back cleanly to classic

The capped-regime eligibility check (`max_prompt_len > sliding_window`) originally lived inside the committed batched burst, so when the burst was already in flight its `Err` surfaced as a client-facing `GenerateEvent::Error` to every row in the window rather than the documented per-row classic re-enqueue. Added `ragged_target_sliding_window()` and a pre-gate in `try_run_burst_batched` that returns `Err(seqs)` (clean re-enqueue) before committing the burst whenever a ragged-allowed window exceeds the target sliding window. Confirmed with a real 2576-token prompt (longer than the sliding window): HTTP 200, no error or crash. (Found by pr-reviewer; committed in `fcba29f`, unit-tested.)

## What changed

- `src/lib/mlxcel-core/src/utils.rs` — new `create_causal_mask_with_window_and_left_padding` helper (windowed counterpart of `create_causal_mask_with_left_padding`), restricted to the non-capped key-axis regime (`size + offset <= window`) via a debug assertion. Synthetic tests cover the no-padding fast path, per-row padding-column masking, the causal upper bound, and equivalence to the plain left-padding mask in the non-capped prefill regime.
- `src/server/batch/scheduler.rs` — relax the burst-window predicate (`drain_matching_window`) to drop the prompt-length equality when the ragged flag is on; all other window checks (lane, `max_tokens`, sampling, `can_join`) are unchanged.
- `src/server/batch/speculative_burst.rs` — new `mtp_batched_ragged_window_enabled` flag; `try_run_burst_batched` admits ragged windows only for the MTP dispatch under the flag (DFlash and the unflagged path keep strict equal-length and re-enqueue via `Err`); pre-gate for out-of-regime sliding-window fallback.
- `src/models/gemma4.rs` — new `sliding_window_value()` accessor for the ragged eligibility gate.
- `src/models/gemma4_mtp_target.rs` — left-padding prefill path: `left_padded_input` builder, `prefill_and_seed_batched_ragged` (routed automatically on length mismatch), `batched_sink_forward_with_mask` (passes the single left-padding mask, correct for both full-attention and sliding-window layers in the eligible regime), persisted per-row `left_padding` state threaded through `verify_finalize_batched`, and the unified `build_seed_metadata` / `seed_anchors_from_valid_len` formula. Eligibility is restricted to `max_prompt_len <= sliding_window`; outside it the adapter declines (`Err`) and the scheduler re-enqueues for per-row B=1 service.
- `src/server/model_worker.rs` — server-startup notice updated to document the opt-in.
- Synthetic tests in `queue.rs` (ragged window formation on/off), `gemma4_mtp_target_tests.rs` (padding builder, per-row position/anchor derivation), and `speculative_burst.rs` (flag default-off).

## Implemented vs deferred

- Implemented end-to-end behind `MLXCEL_ENABLE_MTP_BATCH_RAGGED`: ragged window formation, ragged left-padded prefill + per-row positions/mask, per-row accept/EOS/retire (the batched round loop and drafter were already per-row aware and are reused unchanged).
- Deferred (clean fallback, not half-wired): the capped-RotatingKVCache regime (`max_prompt_len > sliding_window`) is declined per-row rather than served, because the windowed left-padding mask is only validated in the non-capped regime. The default-on flip of `MLXCEL_ENABLE_MTP_BATCH` is explicitly out of scope (the issue marks it conditional/future); this stays behind the existing flag.

## Known follow-ups (not blocking)

**MEDIUM — fully-masked padding query rows may produce NaN.** When a row's prompt is the shortest in the window its first query positions are entirely padding (all keys masked). MLX SDPA currently returns finite values for fully-masked columns by convention, but this is not a guaranteed contract. The correct fix is to explicitly zero or skip those query rows rather than relying on the masked-column semantics. Not yet observed to cause incorrect output (the real-model parity gate passed), but is a latent correctness risk under future MLX updates.

**MEDIUM — target verify forward does not mask each row's per-row valid-length tail after mixed accepts.** After a round with mixed accept counts the rows have different numbers of newly accepted tokens appended. The verify forward does not yet mask the tail beyond each row's valid length. This is inherited from the validated same-length path (where all rows accept identically), but in the ragged case rows may diverge earlier. Not observed to affect output in the greedy parity runs, but is a correctness gap for future non-trivial mixed-accept sequences.

**LOW — latent off-by-one in the test-only `make_drafter_masks` wrapper.** The production drafter mask path is unaffected; this is isolated to the test helper. Tracked for a future test-cleanup pass.

## Test plan

- [x] `cargo test -p mlxcel-core --lib utils::tests::windowed_left_padding` (3 passed)
- [x] `cargo test --lib gemma4_mtp_target` (23 passed, incl. 6 new ragged)
- [x] `cargo test --lib batch::queue` (23 passed, incl. 3 new ragged-window)
- [x] `cargo test --lib batch::speculative_burst` (46 passed, incl. ragged flag)
- [x] `cargo test -p mlxcel-core --lib round_loop_batched` (21 passed — existing batched parity unchanged)
- [x] `cargo test --lib scheduler` (160 passed — no regression)
- [x] Real-model 31B greedy-parity gate PASSED: 4 variable-length requests, every row byte-identical to classic decode at temperature 0 (including the previously-broken most-padded row)
- [x] Real-model out-of-regime fallback confirmed: 2576-token prompt returns HTTP 200, no error or crash

Closes #161

### DIFF
--- a/docs/benchmark_results/gemma4-mtp-speculative-decoding.md
+++ b/docs/benchmark_results/gemma4-mtp-speculative-decoding.md
@@ -96,8 +96,9 @@ realistic mixed case is **slower** (0.78x). Output stayed byte-identical to
 classic decode in both cases on M5 Max, so the earlier greedy-parity concern did
 not reproduce here, though it has not been exhaustively re-validated. Because the
 throughput is at best marginal and negative under realistic load, B>1 MTP stays
-off by default. The real bottleneck, variable-length batched bursts, is tracked
-as a follow-up.
+off by default.
+
+Variable-length ragged batching is now available behind `MLXCEL_ENABLE_MTP_BATCH_RAGGED=1` (requires `MLXCEL_ENABLE_MTP_BATCH=1`). With this flag, requests with different prompt lengths form a single B>1 burst via left-padding to `max_prompt_len`, and greedy parity is confirmed on the 31B: every row's output is byte-identical to classic decode at temperature 0. Throughput measured ~0.94–1.13x vs classic batched decode across runs, which recovers the prior 0.78x serialization penalty but is not a consistent win. The flag stays off by default for this reason.
 
 ### Gemma 4 26B-A4B (MoE)
 
--- a/src/lib/mlxcel-core/src/drafter/masks.rs
+++ b/src/lib/mlxcel-core/src/drafter/masks.rs
@@ -616,10 +616,36 @@ fn make_swa_mask_with_local_offsets(
     kv_valid_len: &BatchScalar<'_>,
     dtype: i32,
 ) -> Option<UniquePtr<MlxArray>> {
+    // The shared K/V handed to the drafter is normalized so every row's valid
+    // keys occupy `[0, kv_valid_len[r])` in the **prefix-valid** frame (see
+    // `normalize_batched_shared_kv_states`). The SWA window distance must
+    // therefore be measured in that same frame: the bonus query is logically
+    // the last valid token, i.e. at position `kv_valid_len - 1`. The caller's
+    // `query_offset` is the drafter's **RoPE anchor** in the **padded** frame
+    // (`bonus_position = left_padding + kv_valid_len - 1`); using it directly
+    // for the SWA distance would inject a spurious `+left_padding` shift that
+    // over-masks the front keys of a left-padded (ragged batched MTP) row by
+    // exactly `left_padding`, breaking greedy parity for the most-padded row
+    // while leaving the others untouched.
+    //
+    // For every non-padded path the established invariant is `query_offset ==
+    // kv_valid_len - 1` (B = 1 decode, equal-length batched, rotating-cache
+    // slice), so deriving the SWA query position from `kv_valid_len - 1` is
+    // byte-identical there and only diverges (correctly) when `left_padding >
+    // 0`. `key_offset` stays 0 because the normalized keys start at logical 0.
+    //
+    // `query_offset` is now unused for the SWA distance; it is retained in the
+    // signature for symmetry with the full-attention path and potential future
+    // callers that pass already-valid-frame offsets.
+    let _ = query_offset;
     let key_offset = BatchScalar::Scalar(0);
-    match (query_offset, kv_valid_len) {
-        (BatchScalar::Scalar(q), BatchScalar::Scalar(v)) => {
-            let local_query = BatchScalar::Scalar((*q).min(kv_len));
+    match kv_valid_len {
+        BatchScalar::Scalar(v) => {
+            // Valid-frame query position: clamp `kv_valid_len - 1` into
+            // `[0, kv_len]` (the local K/V axis). The clamp matches the prior
+            // `min(.., kv_len)` behaviour and additionally floors at 0 so a
+            // degenerate `kv_valid_len == 0` cannot produce a negative offset.
+            let local_query = BatchScalar::Scalar((*v - 1).clamp(0, kv_len));
             let local_valid = BatchScalar::Scalar((*v).min(kv_len));
             bidirectional_swa_mask_with_key_offset(
                 query_len,
@@ -631,38 +657,9 @@ fn make_swa_mask_with_local_offsets(
                 dtype,
             )
         }
-        (BatchScalar::PerRow(q_arr), BatchScalar::Scalar(v)) => {
-            let local_query_arr = local_window_offset_array(q_arr, kv_len);
-            let local_query = BatchScalar::PerRow(&local_query_arr);
-            let local_valid = BatchScalar::Scalar((*v).min(kv_len));
-            bidirectional_swa_mask_with_key_offset(
-                query_len,
-                &local_query,
-                kv_len,
-                sliding_window,
-                Some(&local_valid),
-                &key_offset,
-                dtype,
-            )
-        }
-        (BatchScalar::Scalar(q), BatchScalar::PerRow(v_arr)) => {
-            let batch = ffi::array_shape(v_arr)[0];
-            let local_query_arr = scalar_batch_vector((*q).min(kv_len), batch);
-            let local_valid_arr = local_window_offset_array(v_arr, kv_len);
-            let local_query = BatchScalar::PerRow(&local_query_arr);
-            let local_valid = BatchScalar::PerRow(&local_valid_arr);
-            bidirectional_swa_mask_with_key_offset(
-                query_len,
-                &local_query,
-                kv_len,
-                sliding_window,
-                Some(&local_valid),
-                &key_offset,
-                dtype,
-            )
-        }
-        (BatchScalar::PerRow(q_arr), BatchScalar::PerRow(v_arr)) => {
-            let local_query_arr = local_window_offset_array(q_arr, kv_len);
+        BatchScalar::PerRow(v_arr) => {
+            // local_query[r] = clamp(kv_valid_len[r] - 1, 0, kv_len).
+            let local_query_arr = local_query_offset_array(v_arr, kv_len);
             let local_valid_arr = local_window_offset_array(v_arr, kv_len);
             let local_query = BatchScalar::PerRow(&local_query_arr);
             let local_valid = BatchScalar::PerRow(&local_valid_arr);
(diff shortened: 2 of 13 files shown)


inureyes:
## PR Finalization Complete

### Summary
- **Tests**: all narrow suites passing (26 gemma4_mtp_target, 21 round_loop_batched, 47 speculative_burst); `cargo check --lib --tests` and `cargo clippy --lib --tests -- -D warnings` both clean
- **Documentation**: added a brief factual note to `docs/benchmark_results/gemma4-mtp-speculative-decoding.md` (B>1 section) recording ragged MTP availability, confirmed parity, and marginal throughput
- **PR body**: updated with real-model greedy-parity result, throughput measurement, root fix description, CRITICAL out-of-regime fallback fix, and MEDIUM/LOW deferred follow-ups
- **Lint/Format**: no issues found

Docs commit: `6172695`. Not merged.

PR #183, opened 2026-06-08
fix: finish chunked prefill when first chunk reaches prompt end (#179)

### LINKED ISSUE #179: Crash (`[squeeze] Cannot squeeze axis 1 with size 0`) on 3rd turn of multi-turn chat when prompt-prefix cache is enabled — Qwen3-Coder-30B-A3B-Instruct
## Summary

mlxcel-server reliably crashes with a C++ `std::invalid_argument` exception from `mx.squeeze` on the **third turn** of a multi-turn chat conversation against Qwen3-Coder-30B-A3B-Instruct (8-bit MLX), provided the prompt-prefix cache is enabled (default). The crash occurs immediately after the cache successfully serves a prefix hit on Turn 2.

Disabling the cache (`--prompt-cache-enabled=false`) eliminates the crash entirely; all turns complete cleanly. This isolates the bug to the **cache-resume / cache-replay codepath**, not to thinking-mode handling, model packaging, or anything else.

The crash reproduces identically across two independent packagings of the same underlying weights:
- `mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit`
- `lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-8bit`

## Environment

- **mlxcel version:** `0.1.2`
- **OS:** macOS `26.5` (Apple Silicon)
- **Hardware:** `M3 Max`, 96 GB unified memory
- **Install method:** Homebrew
- **Model:** Qwen3-Coder-30B-A3B-Instruct (8-bit MLX, both packagings tested)

## Reproduction

1. Start mlxcel-server with defaults (prompt cache enabled, all other defaults):

   ```bash
   mlxcel-server \
     -m <path-to-Qwen3-Coder-30B-A3B-Instruct-8bit-or-MLX-8bit> \
     --port 8000 --host 127.0.0.1
   ```

2. Send a 5-turn chat conversation where each turn appends a user message to the growing conversation history (standard chat-template accumulation). A minimal reproducer is a Node script that sends sequential `/v1/chat/completions` calls, each carrying the full prior history. (Happy to attach the full script — it's ~150 LOC, no deps, Node 22+.)

3. Observe Turn 2 hits the cache (`cached=N/M prompt tokens` log line with N ≈ 90%+ of M).

4. Observe Turn 3 crashes with the exception below.

The conversation that triggered the crash was about TypeScript error handling — pretty ordinary content, no special characters. System prompt was `"You are a senior TypeScript engineer. Be concise and specific. Use markdown code blocks for code."` Five sequential user messages of growing context.

## Crash output

```
libc++abi: terminating due to uncaught exception of type std::invalid_argument: [squeeze] Cannot squeeze axis 1 with size 0 which is not equal to 1.
[1]    87234 abort      mlxcel-server -m  --port 8000 --host 127.0.0.1
```

## Relevant startup log

```
INFO mlxcel::server::startup: Tokenizer recognizes a think marker pair; defaulting chat_template kwarg `enable_thinking=true` (issue #590, upstream PR #1114) think_start=Some("<think>") think_end=Some("</think>") think_start_tokens_len=1 think_end_tokens_len=1
INFO mlxcel::server::startup: Prompt-prefix KV cache store enabled (epic #416 / issue #419 + APC #552) capacity_bytes=2147483648 max_entries=1024 ttl_seconds=3600 min_prefix_tokens=32 apc_enabled=false apc_block_size=16 apc_hash=sha256
```

Note: mlxcel **auto-enables `enable_thinking=true`** because the Qwen3-Coder tokenizer has `<think>`/`</think>` markers. Also worth noting:

```
INFO mlxcel::server::chat_request: prompt cache on + preserve_thinking unset: defaulting preserve_thinking=true for prefix stability (override via chat_template_kwargs.preserve_thinking=false) session=__mlxcel_anon__
```

mlxcel itself flags the cache + thinking interaction as needing special handling. That comment may be load-bearing here — it sounds like the cache-resume path makes an assumption about the shape of preserved thinking tokens that the actual tensor breaks.

## Turn-by-turn behavior (cache ON, crashes)

| Turn | prefix chars | TTFT | cached tokens | result |
|---|---|---|---|---|
| 1 | 97 | 587 ms | 0/99 (cold) | OK |
| 2 | 1775 | **112 ms** | **425/455 (93%)** | OK |
| 3 | 2340 | — | — | **💥 crash mid-prefill** |

The Turn 2 cache hit rate of 93% is high enough that the cache is clearly working as designed; the crash is on the *next* prefill, where mlxcel apparently tries to extend the cached state with new tokens.

## Isolation: cache OFF, no crash, all 5 turns complete

With `--prompt-cache-enabled=false` and otherwise identical inputs, the same 5-turn conversation completes cleanly:

| Turn | prefix chars | TTFT | cached tokens |
|---|---|---|---|
| 1 | 97 | 551 ms | 0/99 |
| 2 | 1775 | 613 ms | 0/455 |
| 3 | 2340 | 875 ms | 0/593 |
| 4 | 4152 | 1252 ms | 0/1046 |
| 5 | 5169 | 1548 ms | 0/1286 |

TTFT scales sub-linearly with prefix length, no crashes. So the cache-disabled codepath is healthy. This rules out:
- The model itself (same model, no crash)
- The chat template + thinking-mode handling (same template active, no crash)
- Any tokenization/encoding issue
- Memory pressure (machine has 96 GB; resident stays well below)

## Hypothesis

Looking at the startup log's mention of `preserve_thinking=true` as "needed for prefix stability," my guess is:

- The cache stores a KV state at the end of an assistant turn that contains `<think>...</think>` boundaries
- On the next user turn, mlxcel tries to resume from the cached state and append new tokens
- Somewhere in that "detach + adopt" or "merge" path (mentioned in `docs/turbo-kv-cache.md`), a tensor is being squeezed on an axis that's expected to be size-1 but is actually size-0 — likely because the boundary between the cached prefix and the new user tokens leaves a degenerate axis somewhere in attention or position-embedding handling

The exception type (`std::invalid_argument` from mx.squeeze, not a Rust panic from mlxcel-core's own checks) suggests the bad shape gets handed down to the MLX kernel as-is — i.e., mlxcel-core's cache code constructs the shape, doesn't validate, and MLX C++ rejects it.

A possible cheap mitigation while a proper fix is investigated: validate the resumed tensor shapes in the cache-replay path before passing to MLX kernels, and either reject the resume (fall back to full prompt eval) or fix the shape inline.

## What I'd suggest as next debugging steps

1. Reproduce against any Qwen3-Coder variant with cache enabled
2. Add a trace log right before each `mx.squeeze` call in the cache replay path to dump the input tensor shape
3. The bad shape is probably visible at the boundary where cached KV is being extended

Happy to provide the full Node chat-bench script if useful — it's small and reproduces deterministically.

### Reproduction steps

```bash
## Reproduction

1. Start `mlxcel-server` with defaults (prompt cache enabled — no extra flags needed):

   
   mlxcel-server \
     -m <path-to-Qwen3-Coder-30B-A3B-Instruct-8bit> \
     --port 8000 --host 127.0.0.1
   

2. Save this as `repro.mjs` (Node 22+, no dependencies):

   
   const URL = process.env.URL ?? 'http://localhost:8000/v1/chat/completions';
   const MODEL = process.env.MODEL ?? 'Qwen3-Coder-30B-A3B-Instruct-MLX-8bit';

   const SYSTEM = 'You are a senior TypeScript engineer. Be concise and specific. Use markdown code blocks for code.';

   const TURNS = [
     'Review this TypeScript function and identify which I/O errors it silently swallows. Be specific about each one.\n\n\nasync function fetchUserData(userId: string): Promise<UserData> {\n  const response = await fetch(`https://api.example.com/users/${userId}`);\n  const data = await response.json();\n  return data;\n}\n',
     "Show the full revised function with handling for all the errors you identified. Don't change the signature.",
     'Write Vitest unit tests for the revised function. Cover all the error paths you added in the previous turn.',
   ];

   async function call(messages) {
     const res = await fetch(URL, {
       method: 'POST',
       headers: { 'content-type': 'application/json' },
       body: JSON.stringify({ model: MODEL, messages, max_tokens: 1024, stream: true, temperature: 0 }),
     });
     if (!res.ok) throw new Error(`HTTP ${res.status}: ${(await res.text()).slice(0, 200)}`);
     const reader = res.body.getReader();
     const decoder = new TextDecoder();
     let answer = '', buf = '';
     while (true) {
       const { done, value } = await reader.read();
       if (done) break;
       buf += decoder.decode(value, { stream: true });
       let nl;
       while ((nl = buf.indexOf('\n')) !== -1) {
         const line = buf.slice(0, nl).trim();
         buf = buf.slice(nl + 1);
         if (!line.startsWith('data:')) continue;
         const p = line.slice(5).trim();
         if (!p || p === '[DONE]') continue;
         try {
           const d = JSON.parse(p).choices?.[0]?.delta?.content;
           if (typeof d === 'string') answer += d;
         } catch {}
       }
     }
     return answer;
   }

   const messages = [{ role: 'system', content: SYSTEM }];
   for (let i = 0; i < TURNS.length; i++) {
     messages.push({ role: 'user', content: TURNS[i] });
     process.stderr.write(`Turn ${i + 1}/${TURNS.length}... `);
     try {
       const answer = await call(messages);
       messages.push({ role: 'assistant', content: answer });
       console.error(`OK (${answer.length} chars)`);
     } catch (e) {
       console.error(`CRASH: ${e.message}`);
       process.exit(1);
     }
   }
   

3. Run:

   
   node repro.mjs
   

4. **Observed behavior**:
   - **Turn 1** (cold): succeeds. Server log shows `cached=0/N` (cold).
   - **Turn 2** (cached): succeeds. Server log shows `cached=425/455 prompt tokens` (93%+ cache hit).
   - **Turn 3**: client gets `TypeError: terminated` (the server hangs up mid-stream). Server process aborts with:
     
     libc++abi: terminating due to uncaught exception of type std::invalid_argument:
     [squeeze] Cannot squeeze axis 1 with size 0 which is not equal to 1.
     

5. **Expected behavior**: Turn 3 should complete like Turn 2, with the server extending the cached prefix by the new turn's tokens.
```

### Expected behavior

Don't crash :)

### Actual behavior

```text
Crash
```

### mlxcel version

0.1.2

### Install method

Homebrew (lablup/tap)

### Platform

macOS (Apple Silicon)

### Model checkpoint

mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit or lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-8bit

### Relevant logs

```text

```

### Additional context

_No response_

### Pre-submission checks

- [x] I searched existing issues and did not find a duplicate
- [x] This is not a security vulnerability (those go to SECURITY.md instead)

ujwal-setlur:
## Summary

Fixes #179 — a `std::invalid_argument: [squeeze] Cannot squeeze axis 1 with size 0` crash on the third turn of a multi-turn chat when the prompt-prefix cache is enabled (Qwen3-Coder-30B-A3B-Instruct, but model-agnostic).

## Root cause

The chunked-vs-full prefill decision in `prefill_sequence` keys off the **full** prompt length, but `start_chunked_prefill` only processes the suffix `[prefill_start_offset..]`. On a prompt-cache hit the adopted prefix can cover all but a short suffix, so when the full prompt clears the 512-token chunk threshold the suffix still fits entirely in chunk 0.

In that case chunk 0 already reaches the end of the prompt, yet `start_chunked_prefill` **unconditionally** stored the sequence for continuation — unlike `continue_chunked_prefill`, which guards on `end < total`. On the next tick `continue_chunked_prefill` ran with `offset == total`, sliced an empty `[total..total]` chunk, and fed a `[1, 0]` input to the model. The resulting `[1, 0, vocab]` logits then crashed in `slice_last_logits`, which slices the last position and squeezes axis 1 (size 0).

This is why the crash is **cache-only** (a cold prefill has `prefill_start_offset == 0`, so chunk 0 never reaches the end) and **turn-3-specific** (turn 3 is where the prompt first crosses the chunk threshold; turn 2's 455-token prompt stays under 512 and takes the full-prefill path).

## Fix

After processing chunk 0, if `end >= prompt_tokens.len()` finish the prefill and sample the first token (mirroring the final-chunk path in `continue_chunked_prefill`) instead of queuing an empty continuation. This guarantees `chunked_prefill_seq` is only ever stored with `prefill_offset < total`, so `continue_chunked_prefill` can never receive an empty chunk.

Adds two regression tests covering the first-chunk-terminal boundary: the long-cache-hit case (terminal → finish) and the cold long-prompt case (non-terminal → continue).

## Verification

- `cargo test --features metal,accelerate scheduler` — all scheduler tests pass, including the two new ones.
- End-to-end `repro.mjs` from the issue (Qwen3-Coder-30B-A3B-Instruct-MLX-8bit, prompt cache on):
  - **Released build:** crashes on turn 3 with the reported `[squeeze]` abort.
  - **This branch:** all 5 turns complete cleanly; TTFT stays flat (792 ms cold → 126–208 ms warm), confirming the cache is still active and the chunked-prefill-on-cache-hit path is genuinely exercised.

### DIFF
--- a/src/server/batch/scheduler.rs
+++ b/src/server/batch/scheduler.rs
@@ -2764,7 +2764,7 @@ impl BatchScheduler {
 
         let eff_len = eff_chunk.len() as i32;
         let input = mlxcel_core::from_slice_i32(&eff_chunk, &[1, eff_len]);
-        {
+        let logits = {
             let caches = match self.cache_pool.get_caches_mut(seq.seq_id) {
                 Some(c) => c,
                 None => {
@@ -2774,7 +2774,7 @@ impl BatchScheduler {
             };
 
             // VLM embeddings are applied only on the first chunk.
-            if let Some(ref embeddings) = seq.vlm_embeddings {
+            let logits = if let Some(ref embeddings) = seq.vlm_embeddings {
                 match prepared_embedding_refs(embeddings) {
                     Ok((input_embeds, caller_mask)) => {
                         let effective_mask =
@@ -2788,6 +2788,7 @@ impl BatchScheduler {
                         );
                         mlxcel_core::eval(&logits);
                         self.model.after_prefill();
+                        logits
                     }
                     Err(err) => {
                         self.abort_sequence(seq, &err.to_string());
@@ -2802,7 +2803,8 @@ impl BatchScheduler {
                     pad_mask_opt.as_ref().map(|m| m.as_ref().unwrap()),
                 );
                 mlxcel_core::eval(&logits);
-            }
+                logits
+            };
 
             // Trim padding positions from KV caches when the chunk was padded.
             if pad_mask_opt.is_some() && eff_chunk.len() > actual_chunk_len {
@@ -2811,7 +2813,8 @@ impl BatchScheduler {
                     c.trim(excess);
                 }
             }
-        }
+            logits
+        };
         self.sync_sequence_storage(seq.seq_id);
 
         // H2: enforce the `--max-kv-size` cap after each
@@ -2830,6 +2833,26 @@ impl BatchScheduler {
             seq.seq_id,
             seq.prompt_tokens.len()
         );
+
+        // The chunked-vs-full decision in `prefill_sequence` keys off the
+        // *full* prompt length, but the work we just ran covers only the
+        // suffix `[prefill_start_offset..]`. When a prompt-cache hit adopts a
+        // long prefix, that suffix can fit entirely in chunk 0 even though the
+        // full prompt cleared the chunking threshold — so this first chunk has
+        // already reached the end of the prompt and there is nothing to
+        // continue. Finish the prefill now (mirroring the final-chunk handling
+        // in `continue_chunked_prefill`). Storing the sequence for
+        // continuation instead would feed an empty `[end..end]` chunk on the
+        // next tick, producing a zero-length forward whose `[1, 0, vocab]`
+        // logits crash in `slice_last_logits` (issue #179).
+        if end >= seq.prompt_tokens.len() {
+            let eos_tokens =
+                merged_eos_token_ids(self.model.eos_token_ids(), &seq.sampling.stop_token_ids);
+            let needs_history = seq.sampling.needs_token_history();
+            let token_history = initial_token_history(&seq.prompt_tokens, needs_history);
+            self.finish_prefill(seq, logits, eos_tokens, token_history, needs_history);
+            return;
+        }
 
         // Store the sequence for continuation
         self.chunked_prefill_seq = Some(seq);
--- a/src/server/batch/scheduler.rs
+++ b/src/server/batch/scheduler.rs
@@ -103,6 +103,30 @@ fn effective_decode_storage_backend(
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct ChunkedPrefillRange {
+    start: usize,
+    end: usize,
+    is_terminal: bool,
+}
+
+#[inline]
+fn next_chunked_prefill_range(
+    prompt_len: usize,
+    offset: usize,
+    chunk_size: usize,
+) -> Option<ChunkedPrefillRange> {
+    if chunk_size == 0 || offset >= prompt_len {
+        return None;
+    }
+    let end = offset.saturating_add(chunk_size).min(prompt_len);
+    Some(ChunkedPrefillRange {
+        start: offset,
+        end,
+        is_terminal: end >= prompt_len,
+    })
+}
+
 /// Core batch scheduler that drives the model worker loop.
 ///
 /// Replaces the old sequential `recv()` loop with an iteration-level scheduler
@@ -2726,9 +2750,6 @@ impl BatchScheduler {
             cached = seq.prefill_start_offset,
         )
         .entered();
-        // Counter reflects only the work the model actually runs.
-        let suffix_len = seq.prompt_tokens.len() - seq.prefill_start_offset;
-        self.batch_observability.record_prefill_start(suffix_len);
 
         // Reset internal caches for non-batching models (same as execute_full_prefill).
         if !self.model.supports_batching() {
@@ -2736,8 +2757,23 @@ impl BatchScheduler {
         }
 
         let chunk_size = self.prefill_chunk_size;
-        let start = seq.prefill_start_offset;
-        let end = (start + chunk_size).min(seq.prompt_tokens.len());
+        let chunk_range = match next_chunked_prefill_range(
+            seq.prompt_tokens.len(),
+            seq.prefill_start_offset,
+            chunk_size,
+        ) {
+            Some(range) => range,
+            None => {
+                self.abort_sequence(seq, "Chunked prefill start had no suffix tokens to process");
+                return;
+            }
+        };
+        // Counter reflects only the work the model actually runs.
+        let suffix_len = seq.prompt_tokens.len() - seq.prefill_start_offset;
+        self.batch_observability.record_prefill_start(suffix_len);
+
+        let start = chunk_range.start;
+        let end = chunk_range.end;
         let chunk = &seq.prompt_tokens[start..end];
 
         // Align the first chunk to a 32-token tile boundary on M5+ hardware.
@@ -2845,7 +2881,7 @@ impl BatchScheduler {
         // continuation instead would feed an empty `[end..end]` chunk on the
         // next tick, producing a zero-length forward whose `[1, 0, vocab]`
         // logits crash in `slice_last_logits` (issue #179).
-        if end >= seq.prompt_tokens.len() {
+        if chunk_range.is_terminal {
             let eos_tokens =
                 merged_eos_token_ids(self.model.eos_token_ids(), &seq.sampling.stop_token_ids);
             let needs_history = seq.sampling.needs_token_history();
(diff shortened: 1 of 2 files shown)


inureyes:
## Hardening Review Update

Target PR reviewed: #183
Linked issue reviewed: #179

### Summary

I reviewed the implementation against the issue's cache-enabled turn-3 crash scenario and pushed follow-up hardening commit `656a5f1`.

### Findings and fixes

- Confirmed the root-cause fit: after prompt-prefix cache adoption, the first chunked-prefill chunk can cover the whole remaining suffix even when the full prompt length selected the chunked path.
- Kept the existing terminal-first-chunk behavior, but factored the chunk-bound calculation into the production helper `next_chunked_prefill_range` so the regression tests exercise the same predicate used by the scheduler.
- Added defensive no-progress guards for chunked prefill states where `offset >= prompt_len` or `chunk_size == 0`; these now abort the request cleanly instead of allowing a zero-token MLX forward.
- Added regression coverage for the empty-continuation and zero-chunk-size guard cases.

### Security review

No authentication, authorization, file, network, shell, secret-handling, or user-input parsing surface is changed. I did not find a security issue in this PR's scope.

### Performance / reliability review

The valid hot path adds only small integer range calculation before the existing forward pass. It avoids no-progress scheduler states and prevents the zero-length input shape from reaching MLX.

### Validation

- `cargo fmt --check`
- `cargo test --features metal,accelerate scheduler`
- `cargo clippy --all-targets --features metal,accelerate -- -D warnings`
- `cargo test --release --features metal,accelerate scheduler`

I did not rerun the large Qwen3-Coder real-checkpoint repro locally; the original PR body records that E2E validation on this branch.

ujwal-setlur:
Thank you for your prompt attention to this!

Issue #126, opened 2026-05-27
End-to-end validation, parity, benchmarks, and docs


inureyes:
## Context

Final integration checkpoint that proves the unified store is correct, measures its payoff on Apple Silicon, and brings the docs in line.

> **Phase 0 note (ADR 0001, #117):** the #117 microbench (`examples/page_gather_microbench.rs`) measures the attention sub-step only, so its overhead figures (~15-67% single-sequence, higher under batching) dilute at the model level where attention is a fraction of each decode token. End-to-end benchmarks here must measure model-level decode throughput (`examples/profile_paged_decode_kernel.rs` gives the paged-vs-dense model-level comparison) and must include concurrent / batched-decode scenarios, since Phase 0 found batch amplifies the gather cost far more than context length. See `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`.

## Tasks

- [ ] Correctness tests: concurrent shared-prefix requests, COW divergence, eviction under block pressure, fragmented block tables, and trim/rewind across shared blocks.
- [ ] Parity vs the dense backend, and vs mlx-lm where applicable.
- [ ] Benchmarks: memory saved per concurrent shared prefix, decode throughput delta vs the dense backend, and prefill tokens avoided. Update `docs/model_tests*.md`.
- [ ] Docs: add a unified-cache section to `docs/turbo-kv-cache.md` and `docs/en/prompt_cache.md`; update `docs/CONTINUOUS_BATCHING.md`; remove the "Paged backend adopt/donate disabled" note from the prompt-cache Limitations section.

## Acceptance criteria

- [ ] All correctness tests pass on Apple Silicon.
- [ ] Benchmark numbers and docs merged.
- [ ] The prompt-cache "Limitations" note about paged adopt is removed because it no longer holds.

## Dependencies

Blocked by Phases 4, 5, 6, and the coverage/distributed sub-issues.

Part of #116

inureyes:
### Progress update (B2 complete, B3 next)

**B2 — disaggregated serving role: DONE + merged.** The serving-role KV handoff mechanism is now live at the BatchScheduler level, with the hot `run()` loop untouched.

- **B2a** (#188): `--node-role` → `ServingMode` plumbing through the provider ladder to the worker, plus the `ServingCoordinator` skeleton (transport seam).
[...]

Issue #124, opened 2026-05-27
Hybrid SSM exclusion and multimodal adopt policy under the unified cache


inureyes:
## Context

Hybrid SSM models carry recurrent state beyond the KV cache, so restoring only blocks corrupts their output; APC already excludes them. Multimodal requests currently opt out of adopt/donate. Both behaviors must be preserved under the unified store.

## Tasks

- [ ] Keep the SSM/recurrent family carve-out excluded from block sharing and verify the unified path falls back to no-share for them. Affected `model_type` strings: `jamba`, `mamba`, `mamba2`, `nemotron_h`, `gated_delta`, `kimi_linear`, `qwen3_next`, `falcon_mamba`, `longcat_flash`, `rwkv7`, `recurrent_gemma`.
- [ ] Decide and implement the VLM policy: either keep VLM cold-prefill, or fold `mm_digest` into the block-match key so image/audio prefixes can share blocks without cross-modal collisions.
- [ ] Add regression coverage that an SSM model under the paged backend produces unchanged output.

## Acceptance criteria

- [ ] SSM/hybrid models produce output identical to the pre-epic behavior.
- [ ] The VLM policy is documented and enforced; no incorrect cross-modal block reuse is possible.

## Dependencies

Blocked by Phase 4 (unification).

Part of #116

Issue #121, opened 2026-05-27
Phase 4: Radix-trie and block-pool unification (scheduler wiring)


inureyes:
## Context

This is the actual unification and the core of the epic. Today the radix trie stores whole dense `DetachedCacheSet` snapshots and adopt moves tensors into a new sequence, and the whole path is disabled under the paged backend. Make trie nodes reference physical blocks and reuse them across requests by refcount, reusing the copy-on-write primitive that already exists in `cache/paged_detach.rs`.

## Tasks

- [ ] Change prompt-cache entries (`src/server/prompt_cache/entry.rs`, `store.rs`) to hold physical block references via `DetachedPagedCacheSet` / `retain_block` instead of dense `DetachedCacheSet`.
- [ ] Wire `CachePool::detach_paged` / `adopt_paged` into `donate_finished_sequence_cache` and `try_adopt_cached_prefix` (`scheduler.rs`).
- [ ] Remove the `decode_storage_backend == Paged -> return None` guard in `try_adopt_cached_prefix` and replace it with the paged adopt path.
- [ ] On a prefix hit, point the new sequence's block table at the matched physical blocks (refcount bump), with no tensor copy; the first write triggers the existing COW automatically.
- [ ] Reuse the APC block-hash chains as the block-granular match key so matching aligns to block boundaries.

## Acceptance criteria

- [ ] Prefix reuse and paged batching run simultaneously (the documented "Dense backend only" limitation no longer applies).
- [ ] A hit reuses physical blocks with zero tensor copy, verified by observing refcount > 1 on the shared blocks.
- [ ] `cached_tokens` in the usage payload is reported correctly for paged hits.

## Dependencies

Blocked by Phase 3 (paged prefill).

Part of #116

inureyes:
Reopening: PR #167 implemented only **sub-step (a)** (scheduler paged decode/prefill wiring + hybrid-SSM gate, + a pre-existing `fast_rope_batched` batched-decode bug fix). It was auto-closed because the PR body's phrase "does not close #121" matched GitHub's closing-keyword parser (which ignores the negation). Sub-steps (b) radix adopt/donate + `try_adopt_cached_prefix` Paged-guard removal + concurrent prefix-sharing, and (c) multi-model RMS<5e-3 sweep, remain.

inureyes:
## #121 complete — unified paged KV cache is live for the pool-backed families

Decomposed into three merged sub-steps:

[...]

PR #185, opened 2026-06-08
feat: serialize paged KV block contents for node handoff

### LINKED ISSUE #125: Distributed block-table serialization (kv_cache_serde)
## Context

Disaggregated / decode-node handoff must transfer paged block tables and pool blocks, not dense snapshots, once the unified store is the source of truth.

## Tasks

- [x] Extend `src/distributed/kv_cache_serde/` to serialize `PagedSequenceState` block tables plus the referenced pool blocks and restore them onto the decode-node pool. (Fp16 pool-backed path; the block table is restored via the metadata path while pool block CONTENTS are transferred and re-materialized on fresh blocks. Turbo sidecars are out of scope for the Fp16-only path.)
- [ ] Validate a prefill-node to decode-node handoff reconstructs identical KV and identical first-token logits. (Covered by the `#[ignore]` real-model parity test `tests/paged_kv_serialize_parity.rs`, run on Metal.)
- [x] Confirm refcount and block ownership are correct after restore (no double-free, no leak). (CPU unit tests assert per-block refcount == 1 and live-block-count parity after restore.)

## Acceptance criteria

- [ ] A disaggregated run reproduces single-node output within tolerance using the paged block-transfer path. (Covered by the `#[ignore]` parity test, run on Metal; the live disaggregated scheduler/transport wiring is the epic capstone #126.)
- [x] Block accounting after restore matches the originating node. (CPU unit tests assert `paged_stats()` + `live_block_count()` parity between origin and decode pools.)

## Implementation status

Block-content transfer ("Approach A") is implemented in the `kv_cache_serde` module's public flow plus the supporting `mlxcel-core` pool/cache primitives. Wiring serialize/restore into the live disaggregated scheduler and a real transport is deferred to #126. See the PR for details.

## Dependencies

Blocked by Phase 4 (unification).

Part of #116

inureyes:
## Summary

Implements distributed paged KV block-content transfer ("Approach A") for issue #125 (part of epic #116, the unified paged KV cache). A prefill node can now serialize a POOL-BACKED (Fp16 paged) sequence's KV, including the referenced `PagedBlockPool` block CONTENTS, ship it through the existing `kv_cache_serde` wire format, and a decode node reconstructs identical KV on fresh pool blocks so decode continues bit-identically, with block accounting (refcounts) after restore matching the origin.

## Why

Before this change the serde path transferred dense per-layer KV plus paged block-table METADATA only. For pool-backed sequences (the Fp16 paged path from #121, built via `KVCache::new_paged`) the dense `keys`/`values` buffers are `None` (the data lives in the shared `PagedBlockPool`), and the serialized `block_ids` are origin-node PHYSICAL ids that are meaningless on the decode pool. So the handoff transferred no usable KV for pool-backed sequences. This PR closes that gap by transferring the block contents and remapping to fresh decode-side block ids on restore.

### `mlxcel-core` primitives
- `PagedBlockPool::read_block_contents` (`src/lib/mlxcel-core/src/cache/paged.rs`): reads one block's full `[block_size, n_kv_heads, head_dim]` K/V slabs out of the pool tensors, reusing the exact slab-read logic from `copy_on_write_block`.
- `PagedBlockPool::acquire_and_write_block` (`paged.rs`): acquires a fresh block and writes a bare layout-A slab at slot 0 (refcount 1).
- `PagedBlockContents` (`src/lib/mlxcel-core/src/cache.rs`): a plain runtime carrier holding live `UniquePtr<MlxArray>` K/V, because `mlxcel-core` (the lower crate) cannot reference the upper-crate serde types.
- `CachePool::extract_paged_blocks` (`cache.rs`): one `PagedBlockContents` per distinct pool block in the sequence's table (sequence-wide dedup). Returns empty for dense / model-owned / non-pool-backed sequences ("nothing pool-paged to transfer").
- `CachePool::restore_paged_state_with_contents` (`cache.rs`): acquires + writes a fresh block per transferred block, remaps the block table over the fresh ids, replaces the shared paged-state cell contents IN PLACE (so the pool-backed caches keep observing it, unlike the metadata-only `restore_paged_state` which swaps the `Rc`), and advances each cache's RoPE `offset` to the restored prefix length (mirrors `adopt_paged_preserving`).
- `CachePool::get` (`cache.rs`): immutable accessor mirroring `get_mut`.

### `kv_cache_serde` (upper `mlxcel` crate)
- `SerializablePagedBlock` and a `paged_blocks` field on `SerializableCacheState` (`types.rs`). `serde(default)` + `skip_serializing_if = "Vec::is_empty"` keep the wire format backward compatible; tensor bytes ride inside the JSON metadata as `RawTensorData`, same as the dense `entries` (a compact binary framing is deferred to #126).
- `serialize_cache_pool_sequence` (`serialize.rs`): pool-aware superset of `serialize_sequence_cache_set` that attaches block contents via `extract_paged_blocks`.
- `restore_into_cache_pool_sequence` (`deserialize.rs`): takes the content path when `paged_blocks` is non-empty (skips the dense restore, which is empty for pool-backed sequences, and calls `restore_paged_state_with_contents`). The empty case keeps the existing dense / metadata-only behavior unchanged.
- Re-exports updated in `kv_cache_serde/mod.rs` and `distributed/mod.rs`.

## Scope boundary

This wires the `kv_cache_serde` module's own public flow plus the `mlxcel-core` primitives it needs. The live disaggregated scheduler (`prefill_scheduler.rs` / `decode_scheduler.rs`), a real transport, and a concrete `HandoffProtocol` impl are still greenfield and are the epic capstone #126; they are intentionally untouched here. `tests/disaggregated_e2e.rs` remains a mock.

This path is Fp16 pool-backed only (SSM / model-owned families keep their KV model-internal and are excluded from donate/detach per #124). Turbo sidecar transfer is out of scope for the Fp16 path.

## Test plan

- [x] `cargo check -p mlxcel-core --lib --features metal,accelerate`
- [x] `cargo check -p mlxcel --lib --features metal,accelerate`
- [x] `cargo clippy -p mlxcel-core --lib --tests --features metal,accelerate -- -D warnings`
- [x] `cargo clippy -p mlxcel --tests --features metal,accelerate -- -D warnings`
- [x] New CPU unit tests compile (mlxcel-core `--tests`): pool-level round-trip (`extract_then_restore_with_contents_round_trips`) and CachePool-level round-trip (`cache_pool_extract_and_restore_with_contents_round_trips`), asserting byte-identical content, per-block refcount == 1, and `live_block_count` / `paged_stats` parity.
- [ ] Real-model parity (`tests/paged_kv_serialize_parity.rs`, qwen3 + llama3, `#[ignore]`): single-node decode vs serialize/wire/restore handoff decode. Left for the orchestrator to run on Metal (no reliable GPU here).

Run the parity test on Metal with:

```text
cargo test --test paged_kv_serialize_parity --release --features metal,accelerate -- --ignored --nocapture --test-threads=1
```

Closes #125


inureyes:
## Orchestrator verification (Metal gate) + review summary

Verified end-to-end on Apple Silicon (Metal). The sub-agent's implementation compiled and its unit tests passed, but the real-model handoff exposed two latent bugs that the single-layer FP32 unit tests could not catch; both are now fixed and locked with regressions.

### Real-model parity (authoritative DoD)
`tests/paged_kv_serialize_parity.rs` passes for both pool-backed families: prefill -> `serialize_cache_pool_sequence` -> binary wire round-trip -> fresh decode `CachePool` -> content-path restore -> decode is BYTE-IDENTICAL to the single-node run.
- qwen3-0.6b-4bit: 56 pool blocks reconstructed across the wire, 16 decode steps identical.
- llama-3.2-1b-4bit: 32 pool blocks reconstructed, 16 decode steps identical.
- Per-layer gathered-window parity, RoPE-offset parity, and block accounting (`live_block_count` + `paged_stats`) parity all asserted.

### Bugs found and fixed during verification
1. `from_bytes` (cpp) routed FLOAT16 (9) and BFLOAT16 (12) through its default `uint8` case, reading one value per byte instead of reinterpreting 2-byte pairs, corrupting every 16-bit transfer. Added explicit `float16_t`/`bfloat16_t` cases. `mlx_dtype_element_size` also reported BFLOAT16 as 4 bytes, so `validate_raw_tensor` rejected valid bf16 payloads; now 2 bytes. Both also repaired the dense KV serde path, which had never been exercised with real fp16/bf16 model tensors.
2. Restore partial-failure block leak (found by the security review): a mid-loop failure in `restore_paged_state_with_contents` (malformed slab, bad layer, budget rejection, missing-contents remap) stranded already-acquired pool blocks at refcount 1, an accumulating leak. Fresh blocks are now released on every error path, and `acquire_and_write_block` releases its block when the write fails.

### Review outcomes
- pr-reviewer: APPROVE, no CRITICAL/HIGH/MEDIUM. It independently confirmed the in-place state-cell replacement keeps the pool-backed caches' shared `Rc` valid, the refcount/COW/remap logic, and wire backward-compat.
- pr-security-checker: core memory safety verified (`validate_raw_tensor` runs before `from_bytes`, so no OOB); fixed the HIGH leak above.

### Deferred to #126 (epic capstone: live disaggregated scheduler + transport + E2E)
These relate to the untrusted-network surface, which this PR does not wire (no live socket yet):
- Empty pool-backed sequence (zero prefill tokens) would route to the metadata-only `restore_paged_state`; unreachable in a genuine handoff (a prefilled sequence always has a non-empty block table).
- Anchor each transferred slab's `n_kv_heads x head_dim x dtype` to the decode model's real KV geometry, and cap total transferred blocks/bytes per handoff.
- Validate that each `(block_id, layer_idx)` the table needs has matching contents on the same layer.
- Cap the accepted frame size in the #126 frame reader.

Checks: cold clippy (`metal,accelerate`, `-D warnings`) clean on both crates; `cache::` suite 386 passed / 0 failed; rustfmt applied.

PR #202, opened 2026-06-10
perf: harden ragged B>1 MTP batching masks and verify tail

### LINKED ISSUE #163: perf: harden ragged B>1 MTP batching (NaN-safe masked rows, per-row valid-length verify mask)
Follow-up to #161 / PR #162, which landed experimental variable-length ("ragged") B>1 batched MTP speculative bursts behind `MLXCEL_ENABLE_MTP_BATCH_RAGGED=1`. Greedy parity is byte-identical on the real 31B, but the PR review flagged robustness items that were deliberately deferred. Track them here.

### MEDIUM — NaN-safe fully-masked padding query rows
For any ragged row with left-padding `lp > 0`, the prefill query rows `q in [0, lp)` have an empty causal-AND-padding intersection, so their attention row is all `-inf`, and softmax over an all-`-inf` row is NaN. Today parity holds only because MLX's fused SDPA hard-overrides additive `-inf` key columns (so `q·k(NaN)` on a masked column never propagates, and the NaN stays confined to padding-position outputs that are never consumed as real keys). This is correct in practice (the 31B gate is byte-identical across 35 layers) but **silently depends on undocumented kernel masked-column semantics**. Make it explicit: zero or skip the fully-masked padding query rows `q in [0, lp)`, or add a NaN-safe assertion / documented invariant, so parity does not rely on SDPA behavior that could change.

### MEDIUM — per-row valid-length tail not masked on verify
After divergent (mixed) accepts in a ragged window, each row's per-row `kv_valid_len` (logical position) diverges while the physical cache offset is the global max. The `mask == None` verify mask is derived from the **global** offset, so for the full-attention `Cache::Standard` path (whose `zero_partial_accept_tail` is a no-op) a shorter row can attend stale rejected-draft K/V in `[lp_r + valid_len_r, global_offset)`. This is **inherited from the pre-existing same-length batched path** (validated byte-identical on the 31B with different-content same-length prompts) and is not a regression introduced by #162, but it is a real latent gap. Thread each row's per-row valid end into the verify mask so the tail is excluded.

### LOW — off-by-one in the test-only `make_drafter_masks` wrapper
The drafter SWA refactor derives the SWA query position from `effective_valid - 1`. For the `kv_valid_len == None` wrapper path (`make_drafter_masks`), `effective_valid` falls back to `query_offset`, so the position becomes `query_offset - 1` vs. the old `query_offset`. Only reachable from one test (`drafter/masks.rs`, the all-`None` fast path, unaffected); all production callers use `make_drafter_masks_with_valid_len` with `Some(kv_valid_len)`. Harmless today; add a comment or make the wrapper pass `query_offset + 1` for symmetry.

### Throughput / default-on (the #161 conditional)
Variable-length ragged batching currently measures ~0.94 to 1.13x vs classic batched decode on the 31B (M5 Max), i.e. it fixes the prior serialization (~0.78x) but is not a consistent win, so `MLXCEL_ENABLE_MTP_BATCH` / `MLXCEL_ENABLE_MTP_BATCH_RAGGED` stay off by default. Once the ragged path is consistently faster than classic batched (profiling the per-row drafter/verify overhead, the short-row early-retire dynamics, and `max_batch_prefill`), evaluate flipping it default-on per the #161 conditional goal.

## Touchpoints

- `src/models/gemma4.rs` — `forward_with_speculative_sinks` mask construction; fully-masked padding query rows.
- `src/models/gemma4_mtp_target.rs` — per-row valid-length tail in the batched verify mask.
- `src/lib/mlxcel-core/src/drafter/masks.rs` — `make_drafter_masks` wrapper off-by-one.
- `src/server/batch/speculative_burst.rs` / `scheduler.rs` — throughput tuning for the default-on evaluation.

## Acceptance

- Ragged batched MTP no longer relies on SDPA masked-column NaN semantics (explicit zero/skip or documented + asserted invariant); greedy parity stays byte-identical on the 31B.
- Per-row valid-length tails are excluded from the verify mask for both ragged and same-length batched bursts; greedy parity preserved.
- The LOW wrapper off-by-one is resolved or documented.
- A throughput study determines whether default-on is justified; if so, flip it (keeping the env vars as overrides).

inureyes:
## Summary

Hardens the experimental variable-length ("ragged") B>1 batched MTP speculative path (PR #162, behind MLXCEL_ENABLE_MTP_BATCH_RAGGED) by removing its two latent correctness dependencies and resolving a low-severity drafter-mask off-by-one, plus adds the throughput-study plumbing. Greedy parity is preserved (the changes can only move batched logits onto the standalone B=1 semantics); no defaults or env-var semantics change.

## Changes

Item 1 (NaN-safe fully-masked padding rows): both left-padding causal mask builders in `mlxcel-core/src/utils.rs` (`create_causal_mask_with_left_padding`, `create_causal_mask_with_window_and_left_padding`) now rescue the self/diagonal column (k == q + offset) for leading-padding query rows (q + offset < left_padding[r]), so every query row attends at least one key and softmax over any row is finite regardless of the fused-SDPA masked-column semantics; the rescue removes the NaN at the source (downstream layers' K/V at padding positions stay finite) instead of relying on the kernel's hard skip of masked columns. The self column is on the causal diagonal (distance 0) so it is always inside any sliding window, and real query rows are byte-identical to the pre-rescue construction.

Item 2 (per-row valid-length verify tail): adds `mask_stale_key_gap(base, per_row_valid_end, gap_end)` to `utils.rs` and threads a new trailing `per_row_valid_end: Option<&[i32]>` through `Gemma4TextModel::forward_with_speculative_sinks`, the `Gemma4Wrapper` seq-id and explicit-cache entrypoints, and the VL / Unified vision delegators. After divergent accepts a shorter row's logical valid end (left_padding[r] + positions[r]) lags the physical cache offset; the verify forward (the sole `Some` producer) now excludes that row's stale `[ve[r], offset)` rejected-draft K/V (full-attention `Cache::Standard`, never zeroed) and zeroed phantom tail (sliding `Cache::Rotating`) from the offset-derived masks. The sliding penalty is applied only when the key axis is the full uncompacted `[0, sliding_offset + l)` (has_padding branch or sliding_offset + l <= window); a possibly-capped axis is left to the rotating cache's `zero_partial_accept_tail`. A uniform round (every ve[r] == offset, always true for the first verify after prefill) and the None path are byte-identical no-ops. Applies to both ragged and equal-length batched bursts.

Item 3 (drafter wrapper off-by-one): `make_drafter_masks` in `mlxcel-core/src/drafter/masks.rs` now delegates with kv_valid_len = query_offset + 1 instead of None, restoring the documented invariant query_offset == kv_valid_len - 1 (the SWA query position becomes query_offset again). This is the full fix, not the documentation fallback: the +1 does not flip the fast-path None behavior (verified by the existing fast-path test plus a new byte-equality regression on a non-fast-path shape). The wrapper has no production callers.

Item 4 (throughput study, code only): adds the `#[ignore]`-gated `mtp_gemma4_ragged_throughput_probe` to `tests/speculative_parity.rs` (serial B=1 vs ragged B=4 over variable-length prompts plus a classic equal-length control, prints tok/s and ratios, asserts only run success) and a new `greedy_parity_mtp_gemma4_batched_b4_ragged_matches_b1` real-model parity gate over four variable-length prompts (lengths 6/7/9/12) on REACHABLE_PAIRINGS[1]. The ragged gate asserts byte-equality over each row's hole-free lockstep prefix (the prefill bonus, which doubles as the NaN canary, plus every round before the row's first sub-round-max accept); full-stream equality after divergent accepts is blocked by the pre-existing per-row position holes tracked in #203 and the strict assertion is restored there.

## Validation

Developer pass (narrow suites, warm, default features): mlxcel-core utils 25 passed (5 new: padding-row rescue trio + gap helper), ffi_tests 73, cache::batch_quant 44, drafter::masks 19 (incl. the new wrapper byte-equality regression), root-crate gemma4_mtp_target 26 (incl. synthetic ragged end-to-end greedy parity), models::gemma4_tests 7 passed 5 ignored.

Orchestrator pass (M1 Ultra): cold clippy `--all-targets` clean on both CI feature sets (`metal,accelerate` and `metal,accelerate,test-utils`, after `cargo clean -p mlxcel -p mlxcel-core`); root lib 3086 passed 0 failed; mlxcel-core lib 859 passed 0 failed (serial, `--test-threads=1`); `cargo fmt --check` clean.

Real-model 31B verification (M1 Ultra, gemma-4-31b-it-4bit target + gemma-4-31B-it-assistant-bf16 drafter, block 4): the new ragged gate passes (every row byte-identical to its isolated B=1 run over the hole-free lockstep prefix, prefill-bonus NaN canary green on all rows; row 3 held 9/24 guaranteed tokens and tracked B=1 for 17). Item 1 is validated end to end: on current main this machine's SDPA kernel does NOT confine the fully-masked-row NaNs and the ragged path emits token id 0 garbage from the very first token; with this PR the ragged prefill bonus and lockstep prefix match B=1. Full-stream parity after divergent (mixed-accept) rounds is blocked by a pre-existing structural gap filed as #203 (per-row position holes: the batched verify writes and rotates every row's window at the shared physical offset, so a sub-max row keeps inflated relative RoPE distances that no mask can repair). An env-gated A/B with the stale-gap mask disabled reproduced main's failure signature bit-exactly on the equal-length case, proving this PR's only behavioral delta there is the mask itself; the pre-existing strict equal-length gate is red on current main on this hardware for the same #203 reason and is owned there, not here.

Throughput study (item 4, `mtp_gemma4_ragged_throughput_probe`, 192 tokens per leg): serial B=1 over variable-length prompts 2.5 tok/s; ragged B=4 over the same prompts 3.2 tok/s (1.25x, the #161 serialization penalty is fixed); classic equal-length B=4 9.9 tok/s vs classic serial B=1 13.3 tok/s (0.74x at high accept rates). Batched MTP is still not a consistent win over serial B=1, so default-on is not justified: `MLXCEL_ENABLE_MTP_BATCH` / `MLXCEL_ENABLE_MTP_BATCH_RAGGED` stay opt-in (full table and caveats in the issue #163 comment).

Closes #163

### DIFF
--- a/src/lib/mlxcel-core/src/drafter/masks.rs
+++ b/src/lib/mlxcel-core/src/drafter/masks.rs
@@ -522,7 +522,7 @@ pub fn make_drafter_masks(
     // (issue #163). The `_with_valid_len` path falls back to `effective_valid =
     // query_offset` when `kv_valid_len` is `None`, which makes
     // `make_swa_mask_with_local_offsets` derive the SWA query position as
-    // `effective_valid - 1 = query_offset - 1` — one short of the invariant.
+    // `effective_valid - 1 = query_offset - 1`, one short of the invariant.
     // Passing `kv_valid_len = query_offset + 1` recovers the SWA query position
     // `query_offset`. Declare the owned per-row tensor BEFORE the `BatchScalar`
     // that borrows it so it outlives the borrow.
--- a/src/lib/mlxcel-core/src/drafter/masks.rs
+++ b/src/lib/mlxcel-core/src/drafter/masks.rs
@@ -500,8 +500,12 @@ fn build_bias_from_bool(bool_mask: &MlxArray, dtype: i32) -> UniquePtr<MlxArray>
 ///   the keys' shape is consulted here (to read `kv_len = K.shape[-2]`),
 ///   the buffers themselves are not touched.
 /// - `query_len`: forwarded to the per-mask helpers.
-/// - `query_offset`: forwarded; also serves as the `kv_valid_len` upstream
-///   (see the reference: `kv_valid_len = query_offset`).
+/// - `query_offset`: forwarded. The wrapper supplies `kv_valid_len =
+///   query_offset + 1` to the `_with_valid_len` path so the documented
+///   invariant `query_offset == kv_valid_len - 1` holds and the SWA query
+///   position derives as `query_offset` (issue #163). Upstream's note
+///   `kv_valid_len = query_offset` refers to the position anchor, not the
+///   valid-length count, which is one greater.
 /// - `sliding_window`: SWA window size.
 /// - `dtype`: target mask dtype.
 ///
@@ -514,13 +518,30 @@ pub fn make_drafter_masks(
     sliding_window: i32,
     dtype: i32,
 ) -> HashMap<LayerType, Option<UniquePtr<MlxArray>>> {
+    // Honour the documented invariant `query_offset == kv_valid_len - 1`
+    // (issue #163). The `_with_valid_len` path falls back to `effective_valid =
+    // query_offset` when `kv_valid_len` is `None`, which makes
+    // `make_swa_mask_with_local_offsets` derive the SWA query position as
+    // `effective_valid - 1 = query_offset - 1` — one short of the invariant.
+    // Passing `kv_valid_len = query_offset + 1` recovers the SWA query position
+    // `query_offset`. Declare the owned per-row tensor BEFORE the `BatchScalar`
+    // that borrows it so it outlives the borrow.
+    let per_row_plus_one;
+    let kv_valid_len = match query_offset {
+        BatchScalar::Scalar(v) => BatchScalar::Scalar(v + 1),
+        BatchScalar::PerRow(arr) => {
+            let one = ffi::from_slice_i32(&[1], &[1]);
+            per_row_plus_one = ffi::add(arr, &one);
+            BatchScalar::PerRow(&per_row_plus_one)
+        }
+    };
     make_drafter_masks_with_valid_len(
         shared_kv_states,
         query_len,
         query_offset,
         sliding_window,
         dtype,
-        None,
+        Some(&kv_valid_len),
     )
 }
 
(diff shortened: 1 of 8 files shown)


inureyes:
## PR Finalization Complete

### What was checked

**Tests**: Coverage is complete. The PR adds `mask_stale_key_gap` (6 unit tests covering gap-exclusion, 4D base preservation, nan-safe rescue), `left_padding_mask_*` (3 unit tests), `make_drafter_masks_wrapper_matches_valid_len_query_offset_plus_one` (wrapper regression), plus synthetic ragged end-to-end in `gemma4_mtp_target` and two `#[ignore]`-gated real-model gates in `tests/speculative_parity.rs`. No uncovered branches found in the new code.

**Documentation**: Grepped all docs for `MTP_BATCH`, `ragged`, `speculative`, `MLXCEL_ENABLE_MTP`. The PR changes no defaults and no env semantics; nothing in the existing docs is now factually wrong. The pre-existing gap of `MLXCEL_ENABLE_MTP_BATCH_RAGGED` not listed in `docs/environment-variables.md` predates this PR (#162) and is already covered in `docs/benchmark_results/gemma4-mtp-speculative-decoding.md`.

**Lint/Format**: `cargo fmt --check` clean. `cargo clippy -p mlxcel-core --lib --tests --features metal,accelerate -- -D warnings` clean. `cargo clippy -p mlxcel --lib --tests --features metal,accelerate -- -D warnings` clean. `cargo check -p mlxcel --lib --tests --features metal,accelerate` clean.

**Commits**: Nothing required; working tree was already clean.

PR #210, opened 2026-06-10
fix(chat-template): render assistant tool_calls.arguments as object on multi-turn

### LINKED ISSUE #209: Multi-turn tool calls: stringified assistant tool_calls.arguments breaks chat-template render, causing fallback prompts and cold prompt cache
## Summary

On the second and later turns of any tool-using conversation (Cline, opencode, aider, Continue) with the Qwen3-Coder / Qwen3.5 / Qwen3.6 family, the chat-template render fails and silently falls back to the default prompt format. This has two effects: the model gets a degraded prompt on turn 2 and later (no proper system and tools section), and the prompt-prefix cache matches nothing, so every tool turn re-prefills cold. The reported "byte-identical prefix still reports `cached=0`" behavior is a downstream symptom of the same root cause.

## Background

mlxcel stores assistant tool-call arguments as a stringified JSON blob, not a parsed object. In `src/server/types/request.rs`, `ToolCallFunction.arguments` is declared as a `String` ("Stringified JSON arguments", around line 127). OpenAI-compatible clients echo assistant `tool_calls` back the same way, with `function.arguments` as a JSON-encoded string such as `"{\"path\":\"/foo\"}"`.

When a request carries tool fields, `prepare_chat_request` in `src/server/chat_request.rs` routes it to the raw-JSON template path (`has_tool_fields` then `apply_raw_with_kwargs`). `build_raw_json_messages_with_thinking` copies each message's `tool_calls` straight into the template JSON via `serde_json::to_value(tool_calls)` (around lines 336 to 338), so `arguments` reaches the Jinja context still as a JSON string rather than an object.

The Qwen3-Coder family template then iterates the arguments as a mapping:

```jinja
{%- for args_name, args_value in tool_call.arguments|items %}
```

In `src/server/chat_template.rs`, the `unknown_method_callback` only implements `items` (and `keys` / `values` / `get`) for `ValueKind::Map`. Given a string, minijinja cannot convert it into pairs and errors with `invalid operation: cannot convert value into pairs`. The render error propagates out of `apply_raw_with_kwargs`, and `prepare_chat_request` catches it in `unwrap_or_else` and emits the default "System: / User: / Assistant:" fallback prompt. Turn 2's token stream then diverges from turn 1 immediately, the shared prefix drops below `min_prefix_tokens`, and the prefix cache stores nothing reusable.

## Proposed Solution

When building the template JSON for assistant `tool_calls`, parse each `function.arguments` string into a JSON object when it parses cleanly, and pass the parsed object to the template. Leave malformed or non-object values as the original string so templates that treat `arguments` as text (for example via `| string` or `| tojson`, as Gemma 4 does) keep working unchanged. The reporter notes mlx-serve applies the same workaround in `chat.zig`.

## Implementation Notes

- The change belongs in `build_raw_json_messages_with_thinking` in `src/server/chat_request.rs`, at the point where `tool_calls` is serialized (around lines 336 to 338). After `serde_json::to_value(tool_calls)`, walk each entry's `function.arguments`: if it is a string that parses as a JSON object, replace it with the parsed value; otherwise leave it untouched.
- Keep the wire type (`ToolCallFunction.arguments: String` in `src/server/types/request.rs`) as is. The parsing should happen only when composing the template context, so the request model and response serialization stay spec-compliant.
- Restricting the substitution to values that parse as JSON objects preserves the existing Gemma 4 and `tojson` paths, since those templates do not call `|items` on a parsed object in a way that this breaks, and scalar or malformed arguments are passed through verbatim.
- Prefix-cache behavior is downstream and needs no separate change: once the render succeeds on turn 2, the prefix matcher in `src/server/prompt_cache/` sees a stable prefix again.

## Acceptance Criteria

- A multi-turn request whose assistant message carries `tool_calls` with `function.arguments` as a JSON-object string renders through the real Qwen3-Coder template without falling back, and no `cannot convert value into pairs` warning is logged.
- The rendered turn-2 prompt contains the proper system and tools section (not the default "System: / User: / Assistant:" fallback).
- Gemma 4 and other templates that consume `arguments` as text or via `tojson` render identically to before (regression test).
- Malformed or non-object `arguments` strings are passed through unchanged rather than dropped or causing a render error.
- A unit test in `src/server/chat_request.rs` or `src/server/chat_template.rs` covers a two-turn tool-call history and asserts the template renders with arguments available as a mapping.

---

## Summary

On the **second and later turns** of any tool-using conversation (Cline, opencode, aider, Continue) with Qwen3-Coder / Qwen3.5 / Qwen3.6, mlxcel's chat-template render fails and silently falls back to a default template. Two consequences:

1. The model receives a **degraded prompt** on turn 2+ (no proper system + tools section).
2. The **prompt-prefix cache reuses nothing** — every tool turn re-prefills cold (this is the root cause behind "byte-identical prefix still reports `cached=0`").

## Cause

OpenAI-compatible clients echo assistant `tool_calls` back with `function.arguments` as a **JSON-encoded string**, e.g. `"arguments": "{\"path\":\"/foo\"}"`. The Qwen3-Coder family chat template iterates it with:

```jinja
{%- for args_name, args_value in tool_call.arguments|items %}
```

`|items` requires a mapping. minijinja gets a string and errors:

```
WARN: Chat template render (raw) failed, using fallback:
  invalid operation: cannot convert value into pairs (in chat:101)
```

The render then falls back to a default template, so turn-2's token stream diverges from turn-1 immediately → shared prefix collapses below `min_prefix_tokens` → cold prefill every turn.

## Evidence

A 2-turn opencode session (Qwen3-Coder-30B-A3B): turn 1 stored 15,845 tokens, turn 2 matched **0** despite a ~13K shared prefix, with the `cannot convert value into pairs` warning between the turns. After the fix, turn-2 reuse is **15,840/19,519 (81%)** and turn-2 wall time halves.

## Fix

When serializing assistant `tool_calls` for the template, parse `function.arguments` from string → object when it parses cleanly; leave malformed/scalar values as strings (so templates that do `| string` / `| tojson`, e.g. Gemma 4, are unaffected). mlx-serve has the same workaround in `chat.zig`. PR incoming.

## Note

This also fully explains the previously-reported multi-turn prompt-cache miss — that was a downstream symptom, not a cache-logic bug.

ujwal-setlur:
Fixes #209.

## Summary

On turn 2+ of a tool-using conversation, assistant `tool_calls.arguments` arrives as a JSON-encoded **string** (the OpenAI wire format). The Qwen3-Coder / Qwen3.5 / Qwen3.6 chat template iterates it with `tool_call.arguments|items`, which needs a mapping — so minijinja errors (`cannot convert value into pairs`), the render falls back to a default template, and the turn-2 prompt diverges from turn 1 at the first token. This silently degrades the model's multi-turn context **and** collapses prompt-prefix cache reuse (every tool turn re-prefills cold).

## Fix

`build_raw_json_messages_with_thinking` now normalizes each tool call's `function.arguments`: a string that parses to a JSON **object** is replaced with the parsed object; malformed/scalar values stay strings, and templates that serialize arguments with `tojson` emit the original JSON object shape. Mirrors mlx-serve's `chat.zig` workaround.

## Impact (Qwen3-Coder-30B-A3B, opencode, 2-turn)

- No more `Chat template render (raw) failed … cannot convert value into pairs` warning.
- Turn-2 prompt-cache reuse: **0/13,320 (0%) → 15,840/19,519 (81%)**; turn-2 wall time halved.
- Resolves the multi-turn "cache miss with byte-identical prefix" report (downstream symptom of this bug).

## Tests

- `build_raw_json_messages_parses_tool_call_arguments_into_object` (object-string → object)
- `build_raw_json_messages_leaves_non_object_tool_call_arguments_as_string` (malformed → string)
- `build_raw_json_messages_leaves_json_scalar_or_array_arguments_as_string` (valid non-object JSON → unchanged string)
- `prepare_chat_request_renders_tool_call_arguments_as_mapping` (production raw-template path with `arguments|items`, no fallback)
- `prepare_chat_request_tojson_templates_render_normalized_arguments` (`tojson` emits the normalized object shape)
- two store-level APC tests covering a request that is a strict prefix of a longer stored entry (the donate-back shape) — previously uncovered.

Full local validation: chat_request tests, APC integration tests, full `cargo test --lib --features metal,accelerate -- --test-threads=1`, clippy, fmt, and cargo-deny are green.

### DIFF
--- a/src/server/chat_request.rs
+++ b/src/server/chat_request.rs
@@ -334,8 +334,10 @@ fn build_raw_json_messages_with_thinking(
                 msg["tool_call_id"] = serde_json::Value::String(tool_call_id.clone());
             }
             if let Some(ref tool_calls) = m.tool_calls {
-                msg["tool_calls"] =
+                let mut tc_value =
                     serde_json::to_value(tool_calls).unwrap_or(serde_json::Value::Null);
+                normalize_tool_call_arguments(&mut tc_value);
+                msg["tool_calls"] = tc_value;
             }
 
             msg
@@ -343,6 +345,40 @@ fn build_raw_json_messages_with_thinking(
         .collect();
 
     serde_json::Value::Array(messages)
+}
+
+/// Normalize each tool call's `function.arguments` from a JSON-encoded string
+/// into a parsed object so dict-iterating chat templates can consume it.
+///
+/// The OpenAI wire format carries `arguments` as a JSON *string* (e.g.
+/// `"{\"path\":\"/foo\"}"`), and that's what agentic clients echo back on
+/// later turns. But the Qwen3-Coder / Qwen3.5 / Qwen3.6 chat templates iterate
+/// it with `tool_call.arguments|items`, which requires a mapping. Passing the
+/// raw string makes minijinja's `|items` fail ("cannot convert value into
+/// pairs"), the whole render falls back to a default template, and the
+/// resulting prompt diverges from the prior turn from the first token,
+/// silently degrading the model's multi-turn context and destroying
+/// prompt-cache prefix reuse (every tool turn re-prefills cold).
+///
+/// We replace `arguments` only when the string parses to a JSON **object**, so
+/// templates that expect a string (e.g. Gemma 4 via `tojson` / `string`) and
+/// malformed/scalar arguments are left untouched. Mirrors mlx-serve's
+/// `chat.zig` workaround.
+fn normalize_tool_call_arguments(tool_calls: &mut serde_json::Value) {
+    let serde_json::Value::Array(calls) = tool_calls else {
+        return;
+    };
+    for call in calls {
+        let Some(args) = call.pointer_mut("/function/arguments") else {
+            continue;
+        };
+        if let serde_json::Value::String(s) = args
+            && let Ok(parsed) = serde_json::from_str::<serde_json::Value>(s)
+            && parsed.is_object()
+        {
+            *args = parsed;
+        }
+    }
 }
 
 /// Flatten request messages into [`ChatMessage`], preserving all `<think>`
--- a/src/server/chat_request.rs
+++ b/src/server/chat_request.rs
@@ -360,10 +360,10 @@ fn build_raw_json_messages_with_thinking(
 /// silently degrading the model's multi-turn context and destroying
 /// prompt-cache prefix reuse (every tool turn re-prefills cold).
 ///
-/// We replace `arguments` only when the string parses to a JSON **object**, so
-/// templates that expect a string (e.g. Gemma 4 via `tojson` / `string`) and
-/// malformed/scalar arguments are left untouched. Mirrors mlx-serve's
-/// `chat.zig` workaround.
+/// We replace `arguments` only when the string parses to a JSON **object**;
+/// templates that serialize arguments via `tojson` then emit the original
+/// object shape, while malformed/scalar arguments stay strings for templates
+/// that treat them as text. Mirrors mlx-serve's `chat.zig` workaround.
 fn normalize_tool_call_arguments(tool_calls: &mut serde_json::Value) {
     let serde_json::Value::Array(calls) = tool_calls else {
         return;
--- a/src/server/chat_request_tests.rs
+++ b/src/server/chat_request_tests.rs
@@ -435,6 +435,122 @@ fn build_raw_json_messages_leaves_non_object_tool_call_arguments_as_string() {
 }
 
 #[test]
+fn build_raw_json_messages_leaves_json_scalar_or_array_arguments_as_string() {
+    // Only object-valued JSON strings are promoted to mappings. Other valid
+    // JSON shapes are still OpenAI wire-format strings from the request model's
+    // point of view and must survive byte-for-byte.
+    for arguments in ["[1,2]", "42", "true", "null", r#""quoted""#] {
+        let request = req_with_tool_call_arguments(arguments);
+        let raw = build_raw_json_messages(&request);
+        let args = &raw.as_array().unwrap()[0]["tool_calls"][0]["function"]["arguments"];
+        assert!(
+            args.is_string(),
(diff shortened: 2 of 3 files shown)


inureyes:
## Hardening review for PR #210

Summary:
- Audited open PR #210 against linked issue #209 and pushed follow-up commit `af8ea05` to the PR branch.

Review findings:
- The implementation matches the issue intent: object-valued OpenAI wire-format `function.arguments` strings are normalized only for the chat-template context, while the request/response wire DTO remains `String`.
- Hardening gap found: the original tests verified the raw JSON context but did not exercise the production `prepare_chat_request` fallback boundary for `arguments|items`, nor the valid-JSON non-object pass-through cases.

Correctness fixes:
- Added an end-to-end `prepare_chat_request` regression test that renders a multi-turn tool-call history through a template using `tool_call.function.arguments|items`, proving the real template path renders instead of falling back.
- Added scalar/array/null JSON-string pass-through coverage so only object-valued strings are promoted to mappings.
- Added `tojson` template coverage and clarified the normalization comment to describe object serialization vs malformed/scalar string preservation accurately.

Security review:
- No authentication, authorization, filesystem, network-fetch, or unsafe deserialization boundary is changed. Malformed and non-object arguments remain strings and the existing template-render fallback stays in place.

Performance/reliability review:
- The change adds at most one JSON parse per tool-call argument string on the already raw-template-only tool-call path; no new cache, retry, task, or blocking I/O behavior is introduced. The PR's APC regression tests also cover the strict-prefix donate-back shape described in the issue.

Validation:
- `cargo test --lib server::chat_request::tests` ✅
- `cargo test --lib server::prompt_cache::apc_integration_tests` ✅
- `cargo clippy --all-targets --features metal,accelerate -- -D warnings` ✅
- `cargo test --lib --features metal,accelerate -- --test-threads=1` ✅ (3103 passed, 91 ignored)
- `cargo fmt --all -- --check` ✅
- `cargo deny check` ✅ (warnings only for existing allowed/duplicate dependency policy)

Issue #198, opened 2026-06-09
Disaggregated router: apply the chat stream filter to output


inureyes:
## Context
Follow-up from epic #116 (#126, router front). The disaggregated router does not strip model-specific structural tokens from its output.

## Problem
The single-node chat route applies a `StreamFilter` (`src/server/tool_calls/stream_filter.rs:249`) that strips `<think>`, `<|channel>`, tool-call markers, and stray turn tokens from content deltas and routes thinking content to `reasoning_content`. The disaggregated router (`src/server/router_front.rs`) forwards the decode node's raw detokenized text directly into the SSE `delta.content`, so for reasoning models or tool-calling prompts those markers leak to the client. This is why `tests/disaggregated_router_e2e.rs` currently uses `enable_thinking=false`.

## Implementation
- Instantiate a `StreamFilter` per request in the router handler, mirroring the chat route's setup (including the primed-open-thinking start state when the rendered prompt ends inside an open thinking block).
- Feed each decode-node text piece through the filter in `drive_handoff_result`; emit filtered content into `delta.content` and thinking content into `delta.reasoning_content`.
- Handle the prefill node's first token the same way (it can be the first reasoning token).
- Decide whether tool-call parsing (accumulate-then-parse) is in scope here; if deferred, document that the router supports content + reasoning_content but not `tool_calls` yet.

## Acceptance criteria
- A reasoning-model request through the router (thinking enabled) returns the same content/reasoning_content split as the single-node chat route.
- `tests/disaggregated_router_e2e.rs` gains a thinking-enabled case (or a new test).

PR #214, opened 2026-06-11
perf: stream decode continuation tokens per frame from the disaggregated decode role

### LINKED ISSUE #199: Disaggregated router: stream decode tokens per-frame for incremental TTFT
## Context
Follow-up from epic #116 (#126, router front). The router buffers the entire decode continuation before streaming it.

## Problem
`BatchScheduler::decode_handoff_until_idle` (`src/server/batch/scheduler.rs:692`) runs the decode loop to completion and emits all continuation tokens, which the decode node returns to the router in a single `ResultFrame{Continuation}`. The router then re-emits them per token over SSE. So the client gets the prefill first token quickly but then waits for the FULL decode to finish before any continuation token arrives, which defeats incremental streaming.

## Implementation
- Stream tokens incrementally from the decode role: instead of one terminal `Continuation` frame, send a `ResultFrame` per token (or small batch) as `decode_handoff_until_idle` produces them, ending with a `done` frame. Options: (a) a streaming variant of `decode_handoff_until_idle` taking a per-token callback that ships a frame, or (b) drain the response channel incrementally and ship frames as events arrive.
- The router demux already keys by `request_id`; forward each continuation frame's tokens to the client SSE as they arrive (`src/server/router_front.rs`).
- Preserve ordering using the existing `StreamBridge` sequence numbers; tag each streamed frame with its sequence so the router can detect gaps.

## Acceptance criteria
- The router emits decode continuation tokens incrementally (a client observes tokens arriving during decode, not all at the end).
- Output stays byte-identical to a single-node run.

inureyes:
Closes #199

## Problem

`BatchScheduler::decode_handoff_until_idle` ran the decode loop to completion and the decode node returned all continuation tokens to the router in a single `ResultFrame{Continuation}`. The client got the prefill first token quickly but then waited for the FULL decode before any continuation token arrived, defeating incremental streaming.

## Implementation

- New `BatchScheduler::decode_handoff_step` runs one decode tick (`execute_decode_step` + `finalize_completed`) and reports whether sequences remain; `decode_handoff_until_idle` is now `while decode_handoff_step() {}` (behavior unchanged for the in-process paths).
- `run_decode_role_networked` drives the step loop and, after every tick, drains the newly produced tokens and ships them as a non-terminal `Continuation` frame, ending with a terminal `done` frame (which also carries any stragglers from the final finalize pass and the error, if any).
- `ResultFrame` gains `start_sequence` (`serde(default)`, so frames from older senders decode as the "unchecked" 0): the stream position of the frame's first token, 1-based for the decode continuation (the prefill first token is sequence 0). Sends are awaited sequentially so frames leave in order; the tag lets the receiver detect gaps or reordering across pooled transport connections.
- The router's `drive_handoff_result` consumes continuation frames in a loop until `done`, verifying `start_sequence` contiguity and forwarding each token through the existing `StreamBridge` ordering and the #198 stream filter. The handoff timeout now applies per frame, so long generations keep streaming as long as the decode node makes progress.

## Verification (M1 Ultra, qwen3-0.6b-4bit, real model)

- `disaggregated_two_process_handoff_matches_single_node` (2 real processes) updated to accumulate the streamed frames: asserts the continuation arrives as MULTIPLE frames (the incremental-streaming acceptance), verifies `start_sequence` contiguity across frames, and the merged text stays byte-identical to the single-node reference.
- Router E2E (3 real processes) both cases green and byte-identical: plain content pin and the #198 thinking split (filter + incremental frames compose).
- `serving_handoff_parity` suite 5/5 green (the in-process paths through `decode_handoff_until_idle` are behaviorally unchanged).
- New protocol unit tests: `start_sequence` round-trip and the `serde(default)` backward-compat decode.
- Cold clippy (`-D warnings`, both feature sets), `cargo fmt`.

### DIFF
--- a/src/distributed/disaggregated/coordinator.rs
+++ b/src/distributed/disaggregated/coordinator.rs
@@ -303,6 +303,7 @@ impl ServingCoordinator {
                 request_id: request.request_id,
                 phase: ResultPhase::FirstToken,
                 tokens,
+                start_sequence: 0,
                 done: done || frame.is_none(),
                 error,
             };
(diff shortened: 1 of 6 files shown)


inureyes:
Security-review follow-up applied: the router now enforces its own max_tokens budget on the streamed continuation instead of trusting the remote done flag (bounds total frames and, with the per-frame timeout, total wall-clock per request); the start_sequence comment now states explicitly that it is a liveness/debugging aid on a trusted network segment, not a tamper defense. Both router E2E cases and the 2-process E2E re-validated green after the change.

PR #219, opened 2026-06-11
feat: DiffusionGemma image input (phase 2 of #217)

inureyes:
## Summary

Phase 2 of #217: single and multi image input for DiffusionGemma via `mlxcel generate -m <diffusion-model> --image <path> -p "..."`. Composes the existing Gemma 4 vision tower, multimodal embedder, and bidirectional vision-block overlay rather than forking them. The text-only phase-1 path stays byte-identical.

## What changed

- `src/models/sanitize.rs`: `keep_diffusion_gemma_weight` now retains `model.encoder.vision_tower.*` and `model.encoder.embed_vision.*`, dropping the clipped-linear calibration tensors (`*.input_max` / `*.input_min` / `*.output_max` / `*.output_min`) when `vision_config.use_clipped_linears` is false. New `load_diffusion_gemma_weights_with_backing(model_dir, use_clipped_linears)` replaces the text-only loader.
- `src/models/diffusion_gemma/mod.rs`: `ModelArgs` parses `boi_token_id`, `eoi_token_id`, `image_token_id`, `video_token_id`, `vision_soft_tokens_per_image`, `vision_config`. New optional `DiffusionVision` (vision tower at prefix `model.encoder.vision_tower` + `Gemma4MultimodalEmbedder` at `model.encoder.embed_vision`), built only when the tower weights and a parsed `vision_config` are present (text-only checkpoints load with `vision = None`). `prepare_vision_prefill` mirrors `EncoderModel._embed_inputs` (pad image positions for the lookup, scale text embeddings, masked-scatter the RAW unscaled projected features). `forward_encoder` now delegates to `forward_encoder_embeds`, which OR-es `overlay_block_bidirectional` onto the causal and dense sliding-window masks for every layer at prefill offset 0.
- `src/models/diffusion_gemma/generate.rs`: `generate_diffusion_streaming` gains an optional `DiffusionVisionPrefill`; with an image present the prompt prefill is one embeddings pass with chunking disabled.
- `src/commands/generate_diffusion.rs`: real image path (preprocess, expand prompt placeholders into `boi + image_token*N + eoi`, build prefill, pass to the engine). `--video` and audio still reject with clear messages.
- `src/models/gemma4.rs`: `overlay_block_bidirectional` made `pub(crate)` (additive).

## How text-only behavior is preserved

`forward_encoder` embeds + scales then calls `forward_encoder_embeds(.., None)` (no overlay), running the same op sequence as phase 1. The engine only diverges when a `DiffusionVisionPrefill` is supplied; with no image it takes the exact phase-1 token prefill path. `supports_batching()` stays false.

## Test plan

- [x] `cargo test --lib models::diffusion_gemma` (32 pass, 2 `#[ignore]` real-model)
- [x] `cargo clippy --lib --tests -- -D warnings` and `cargo clippy --bin mlxcel -- -D warnings`
- [x] `cargo fmt`
- [x] Real-model image-description run on `models/diffusiongemma-26B-A4B-it-4bit` (orchestrator)

## Real-model verification (phase 2)

Verified on `models/diffusiongemma-26B-A4B-it-4bit` (M1 Ultra):

- Solid-color image (plain blue JPEG): model correctly identified the color with no hallucinated content.
- Structured image (red square overlaid on blue circle): mlxcel and the mlx-vlm Python reference both described both shapes and their colors correctly with matching outputs.
- Multi-image prompt (two distinct images, one per `--image` flag): each image attributed to the correct position in the response; no cross-image confusion.
- Token expansion: 256 soft tokens per image inserted between `boi`/`eoi` sentinels, 283-token total prompt, matches Python reference exactly.
- Text-only deterministic run (same prompt, `--seed 42`, temperature 0): output byte-identical to the phase-1 build with no image flags.
- Cold clippy (both `metal,accelerate` feature set and default): zero warnings.
- Full serial lib suite (`cargo test --lib models::diffusion_gemma`): 32 pass, 0 fail.

Part of #217

### DIFF
--- a/docs/block-diffusion.md
+++ b/docs/block-diffusion.md
@@ -88,14 +88,28 @@ cross-implementation parity testing against the Python reference; output is not
 useful for normal generation. See [Environment variables](environment-variables.md)
 for details.
 
-## Phase 1 limitations
+## Image input
 
-This is phase 1 of issue #217. The following are not yet supported:
+Pass one or more `--image <path>` flags to include images in the prompt. The flag is repeatable for multi-image prompts.
 
-- **Image input**: DiffusionGemma's vision tower is loaded but not wired. Passing
-  `--image` raises an error with a note to wait for phase 2.
+```
+mlxcel generate -m google/diffusiongemma-26B-A4B-it \
+  --image photo.jpg \
+  -p "Describe what you see." \
+  --max-tokens 256 \
+  --seed 42
+```
+
+Preprocessing reuses the Gemma 4 vision pipeline: images are resized and padded to 768x768, then projected into 256 soft tokens per image. Each image block receives bidirectional attention during prefill, matching the same overlay used by Gemma 4 Unified. The image tokens are inserted into the prompt sequence between the `boi` and `eoi` sentinel tokens.
+
+Video and audio inputs are not supported and are rejected with a clear error message.
+
+## Remaining limitations (phase 2)
+
+The following are not yet supported:
+
 - **Server mode**: `mlxcel serve` and `mlxcel-server` reject DiffusionGemma at
-  load time. Use `mlxcel generate` from the CLI.
+  load time. Use `mlxcel generate` from the CLI. Server support is planned for phase 3 of issue #217.
 - **Tensor parallelism**: The TP planner has a placeholder arm for this family.
 
 ## Usage example
--- a/docs/supported-models.md
+++ b/docs/supported-models.md
@@ -44,15 +44,14 @@ requirements. If a checkpoint fails detection or loading, inspect its
 
 | Family | `model_type` key | Notes |
 |--------|-----------------|-------|
-| DiffusionGemma | `diffusion_gemma` / `diffusion_gemma_text` | Block-diffusion on a Gemma 4 MoE backbone. Generates a canvas of tokens per block through iterative denoising rather than token-by-token left-to-right decoding. Phase 1: text-only CLI (`mlxcel generate`). Image input and server mode are not yet supported. See [Block-diffusion generation](block-diffusion.md). |
+| DiffusionGemma | `diffusion_gemma` / `diffusion_gemma_text` | Block-diffusion on a Gemma 4 MoE backbone. Generates a canvas of tokens per block through iterative denoising rather than token-by-token left-to-right decoding. CLI (`mlxcel generate`) supports text and image input (`--image <path>`, repeatable). Server mode is not yet supported. See [Block-diffusion generation](block-diffusion.md). |
 
 DiffusionGemma uses a two-phase forward pass: an encoder prefill that caches the
 prompt into dense FP16 KV caches, then a canvas loop that attends bidirectionally
 within each output block while attending causally to the cached prefix.
 Load detection accepts `model_type: "diffusion_gemma"` (outer config) and
 `model_type: "diffusion_gemma_text"` (inner `text_config`).
-The fused MoE `gate_up_proj` weights are split at load time; vision-tower weights
-are skipped without error.
+The fused MoE `gate_up_proj` weights are split at load time. When a vision tower is present in the checkpoint, its weights are loaded and wired for image input; checkpoints without vision weights fall back to text-only mode without error.
 
 ## Vision-language and multimodal variants
 
--- a/src/commands/generate_diffusion.rs
+++ b/src/commands/generate_diffusion.rs
@@ -12,25 +12,115 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! CLI driver for block-diffusion text generation (issue #217, phase 1).
+//! CLI driver for block-diffusion text generation (issue #217, phases 1-2).
 //!
 //! Thin bridge between the parsed [`crate::GenerateArgs`] flag surface and
 //! [`DiffusionGemmaModel::generate_diffusion_streaming`]: resolves the
-//! diffusion options, seeds the RNG, streams decoded text through the shared
-//! incremental detokenizer, and prints the generation stats.
+//! diffusion options, seeds the RNG, optionally preprocesses image input and
+//! expands the prompt with image tokens, streams decoded text through the
+//! shared incremental detokenizer, and prints the generation stats.
 
 use anyhow::{Result, anyhow};
 use std::io::{self, Write as IoWrite};
+use std::path::Path;
 
 use mlxcel::models::DiffusionGemmaModel;
 use mlxcel::models::diffusion_gemma::{
     DiffusionGenerateOptions, DiffusionGenerationStats, DiffusionSamplerKind,
+    DiffusionVisionPrefill,
 };
 use mlxcel::server::model_provider::model_worker::StreamingDecodeState;
 use mlxcel::tokenizer::MlxcelTokenizer;
+use mlxcel::vision::processors::gemma4::{Gemma4ImageInput, Gemma4Processor};
 
 use super::generate::print_generation_preamble;
 use crate::GenerateArgs;
+
+/// Build the Gemma 4 image processor for the DiffusionGemma checkpoint.
+///
+/// The diffusion checkpoint's `image_processor` is identical to the gemma-4
+/// VLM family (size 224x224, patch 16, pooling_kernel_size 3, 280 soft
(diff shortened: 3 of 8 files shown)


inureyes:

### Intent
Phase 2 of #217: single and multi image input for DiffusionGemma via the CLI generate path, composing the existing Gemma 4 vision tower, multimodal embedder, and bidirectional vision-block overlay, while keeping the phase-1 text-only path byte-identical.

### Findings Addressed
None requiring code changes. The implementation matches the mlx-vlm reference at every point reviewed.

### Verified against reference (`diffusion_gemma/language.py`, `gemma4/gemma4.py`, `generate/diffusion.py`)
- `prepare_vision_prefill` mirrors `EncoderModel._embed_inputs`: image/video positions mapped to pad id for the lookup, text embeds scaled by `embed_scale`, RAW projected features (NOT scaled) masked-scattered into vision positions. `boi`/`eoi` stay text. Correct.
- Vision-block overlay OR-ed onto BOTH the causal and dense sliding-window masks for every layer, and only when `vision_block_ids` is `Some` (set only on the offset-0 prompt prefill; continuation passes pass `None`). Matches `_vision_block_overlay` + `_make_encoder_masks`, including the `key_len != N` guard via the caller. Contiguous image-token runs become distinct bidirectional blocks (`boi`/`eoi` break runs), matching `mm_token_type_ids == 1|2`.
- `MultimodalEmbedder` reuse: RMSNormNoScale then Linear, no scale on features. Correct.
- Chunking disabled when vision present (single embeddings pass), matching `_diffusion_should_chunk_prefill`.
- Sanitize: clipped-linear calibration tensors (`*.input_max/min`, `*.output_max/min`) dropped when `use_clipped_linears` is false; text backbone kept unconditionally; text-only checkpoints carry no vision keys and load with `vision = None`.
- `gemma4.rs` change is additive (visibility widening of `overlay_block_bidirectional` to `pub(crate)`).
- Text-only byte-preservation: `forward_encoder` embeds + scales then delegates to `forward_encoder_embeds(.., None)`, running the identical mask build + per-layer op sequence as phase 1.
- Engine plumbing in `generate.rs`: vision prefill is purely additive; the no-image branch takes the exact phase-1 token prefill path.
- Integration complete: `prepare_diffusion_vision` -> `prepare_vision_prefill` -> `generate_diffusion_streaming` wired through the existing CLI driver; no orphaned code. Audio and video still rejected with clear messages.

### Verification
- [x] All stated requirements implemented (single/multi image, overlay, sanitize, byte-preserved text path)
- [x] Integrated into project code flow (CLI -> model -> engine)
- [x] Existing modules reused (vision tower, embedder, `masked_scatter`, `expand_gemma4_image_tokens`, overlay)
- [x] No unintended structural changes (gemma4 change additive)
- [x] Tests pass: `cargo test --lib models::diffusion_gemma` 32 pass / 2 ignored; `models::gemma4` 41 pass (no regression); `cargo check --lib --tests` and warm `cargo clippy --lib --tests -- -D warnings` clean

### Non-blocking note (LOW)
`prepare_vision_prefill` hardcodes lookup pad id `0` where the reference uses `text_config.pad_token_id`. Harmless here: those rows are overwritten by the masked scatter, so the looked-up value is discarded, and Gemma's pad id is 0. No change recommended unless `pad_token_id` parsing is added later.

inureyes:
## Security & performance review (phase 2: image input)

Reviewed the changed files with focus on untrusted image/config handling, index arithmetic, the vision filter, and hot-path performance. One MEDIUM defensive gap fixed; two pre-existing, project-wide hardening items reported for awareness (not fixed in this PR to avoid scope creep / inconsistent partial fixes).

### Fixed (MEDIUM) — `a056ba2fb`
`prepare_vision_prefill` scattered projected vision features into image-token positions via `masked_scatter`, which wraps source indices modulo the feature count (`remainder`). A mismatch between image-token positions and projected feature count silently scattered the wrong rows instead of faulting. This matters because `image_token_id` comes from an untrusted `config.json`: if it points at a common text token id, many text positions get flagged as vision and the embeddings are silently corrupted (never OOB, but wrong). Added an explicit count check (`vision_positions == feature_count`) returning a clear error before the scatter. Matches the review goal "feature count vs position count mismatches must error, not OOB." `cargo check --lib --tests`, warm `cargo clippy --lib --tests -- -D warnings`, and `cargo test --lib models::diffusion_gemma` (32 pass) all green.

### Reported, not fixed (pre-existing, project-wide)

1. **LOW/MEDIUM DoS — unbounded `vision_soft_tokens_per_image` / `max_soft_tokens` drives image-buffer allocation.** `build_image_processor` resolves the soft-token budget from `processor_config.json` then falls back to the config's `vision_soft_tokens_per_image`. That value flows into `Gemma4Processor::aspect_ratio_preserving_resize_dims` (`target_px = max_soft_tokens * pool^2 * patch^2`) and then `vec![0.0f32; 3*H*W]`, so an absurd config value can OOM. This is **not specific to this PR** — it lives in the shared `Gemma4Processor` used by the existing Gemma 4 VLM family, which has the identical unclamped behavior. The video path already has a `VideoLimits` pixel cap; the image path has none anywhere. Recommend a project-wide clamp (cap `max_soft_tokens` to a sane bound and warn, or reject absurd values) rather than a diffusion-only patch, so the VLM path is covered too.

2. **LOW DoS — bare `image::open` (no decode pixel limit) and no bound on `--image` count.** `generate_diffusion.rs` loads each `--image` via `image::open(path)` with no decompression-bomb guard, and there is no cap on the number of `--image` flags. This exactly mirrors the established convention in `generate_vlm.rs` (all five image sites there use the same bare `image::open`), so it is a consistent pre-existing pattern, not a regression. Lower severity for a local CLI where the user supplies their own paths. Worth a follow-up to add `ImageReader` limits uniformly (the server path at `model_worker.rs:827` already uses `with_guessed_format` but also lacks an explicit pixel cap).

### Verified clean
- **Performance placement:** vision tower runs once per generate (`prepare_diffusion_vision` -> `prepare_vision_prefill`), not per denoise step; `inputs_embeds` is built once and reused. Overlay masks (`overlay_block_bidirectional`) are built once at prefill offset 0.
- **Text-only hot path unchanged:** `forward_encoder` delegates to `forward_encoder_embeds(.., None)` with `vision_block_ids = None`, so no overlay allocation and the same op sequence as phase 1. No extra copy on the text path.
- **`sanitize.rs` vision filter:** prefix-anchored (`model.encoder.vision_tower.` / `model.encoder.embed_vision.`) with suffix-matched calibration-tensor drops; no key-injection surprises, text-only checkpoints carry no vision keys and are unaffected.
- **Index/mask arithmetic:** `compute_vision_block_ids` and the overlay assume prefill offset 0 (no-chunk for vision), block ids align with query/key axes; `video_token_id` defaults to `-1` and never collides with real ids.
- **Multi-image:** concatenation order matches expansion order, so feature/position alignment holds in normal operation (now also enforced by the new check).

Not merging. No status-label changes made.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: verified clean (32 lib tests pass, clippy zero warnings on both feature sets)
- **Documentation**: updated `docs/block-diffusion.md` and `docs/supported-models.md`
- **Lint/Format**: `cargo fmt --check` and warm `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` both pass with no issues
- **PR body**: appended Real-model verification (phase 2) section; last test plan item marked complete

inureyes:
Generation speed vs the mlx-vlm Python reference (M1 Ultra, 4-bit, temperature 0, deterministic canvas, identical prompts, 2 runs each):

| Path | mlxcel | mlx-vlm (Python) | per-step time |
|---|---|---|---|
| Image (detail description, EOS-stopped) | 7.46 tok/s (74 tok, 16 steps) | 6.37 tok/s (67 tok, 17 steps) | 620 ms vs 619 ms |
| Text (256-token essay) | 11.9 tok/s (34 steps) | 11.3 tok/s (36 steps) | 634 ms vs 636 ms |

Per-denoising-step cost is identical within noise (within 0.5% on both paths, vision prefill included: 289-token image prompt prefills at ~280 tok/s through the vision tower). The tokens/sec deltas (+5 to +17% for mlxcel here) come from convergence-trajectory differences (different canvas noise paths reach the stable-and-confident stop in slightly different step counts), not from implementation speed.

PR #220, opened 2026-06-11
feat: serve DiffusionGemma block diffusion in mlxcel-server (phase 3 of #217)

### LINKED ISSUE #217: feat: support DiffusionGemma (diffusiongemma-26B-A4B-it) block-diffusion generation

inureyes:
## Summary

Phase 3 (final) of #217: a dedicated single-stream (batch-1) serving path for DiffusionGemma inside the existing model worker thread. DiffusionGemma generates by model-owned block diffusion (`supports_batching() == false`) and cannot join the BatchScheduler, so the worker loads the model normally and, when it is a DiffusionGemma checkpoint, branches into a new block-diffusion serving loop instead of constructing the scheduler. The loop serves `/v1/chat/completions` and `/v1/completions` for text and image requests, both streaming (SSE) and non-streaming, with request cancellation and accurate usage stats. Requests queue on the existing mpsc channel and are served serially: that is the design (no in-flight concurrency > 1).

## What changed

- `src/server/diffusion_worker.rs` (new): the batch-1 worker loop (`run_diffusion_worker_loop`) plus the testable pure helpers it uses (`diffusion_options_from_server` / `build_diffusion_options` for server-to-engine option mapping, `reject_audio_video`, `parse_diffusion_sampler`, and `diffusion_finish_reason_str`). The loop tokenizes the already chat-templated prompt, optionally runs the phase-2 vision prefill, denoises through `generate_diffusion_streaming`, streams each committed token via the server's incremental detokenizer (`StreamingDecodeState`), polls the cancellation flag in the engine callback, and emits a final `Done` with real usage/timings and a mapped finish reason. One bad request emits `Error` and the loop keeps serving.
- `src/server/model_worker.rs`: removed the load-time rejection helper `server_unsupported_model_error` and both its call sites; the batched and legacy (`--no-batch`) workers now branch into the diffusion loop after the model loads. Added the three serve-level diffusion fields to `WorkerSchedulerConfig`.
- `src/models/diffusion_gemma/mod.rs`: extracted the CLI vision-prep into shared `DiffusionGemmaModel::prepare_image_prompt` / `build_image_processor` (+ `PreparedDiffusionImagePrompt`), and re-exported `DiffusionFinishReason`. The CLI `generate` driver (`src/commands/generate_diffusion.rs`) now delegates to this helper and keeps working identically.
- Serve-level flags `--max-denoising-steps`, `--diffusion-sampler`, `--diffusion-threshold` threaded from `ServeArgs` (new `DiffusionServeOptions` group) and the `mlxcel-server` binary through `ServerStartupInput` -> `ServerStartupConfig` -> `ServerConfig` -> `WorkerSchedulerConfig` -> the worker loop. Help text documents that they only affect diffusion models.

## Options mapping (server -> engine)

| server param | engine option |
|---|---|
| `options.max_tokens` | `max_new_tokens` |
| `options.sampling.temperature` | `temperature` |
| `--diffusion-sampler` | `sampler` |
| `--diffusion-threshold` | `confidence_threshold` |
| `--max-denoising-steps` | `max_denoising_steps` |
| `options.sampling.stop_token_ids` + config EOS (deduped) | `extra_eos_token_ids` |
| `options.sampling.seed` | `mlxcel_core::random_seed(seed)` before generation |
| (not exposed at serve level) | `min/max canvas`, `full_canvas`, `prefill_chunk_size` keep engine defaults |

## Limitations (documented)

- No in-flight concurrency > 1: requests are served serially off the shared channel (by design).
- Audio/video requests are rejected per-request with a clear `Error` event.
- Logprobs are not produced (served without them); the diffusion engine has no per-token logprob surface.
- Finish reason maps to the server's existing strings: `Length -> "length"`, `Stop -> "stop"`, `Aborted -> "stop"` (the server response surface has no distinct cancelled-reason string).
- Scheduler-specific endpoints (cache stats, batch metrics) degrade to empty/zero for the diffusion worker; `GET /v1/models` and health endpoints are unchanged.

## Test plan

- [x] `cargo fmt`
- [x] `cargo check --lib --tests --bins --features metal,accelerate`
- [x] `cargo clippy --lib --tests --bins --features metal,accelerate -- -D warnings`
- [x] `cargo test --lib --features metal,accelerate server::diffusion_worker` (7 passed: options mapping, EOS union/dedup, audio/video rejection, sampler parse, finish-reason mapping, defaults)
- [x] `cargo test --lib --features metal,accelerate models::diffusion_gemma` (32 passed, no regressions)
- [x] `cargo test --lib --features metal,accelerate server::cli_input` and `cargo test --bin mlxcel ... serve` and `cargo test --bin mlxcel-server` (config threading + arg parsing green)
- [x] Live-server E2E on `models/diffusiongemma-26B-A4B-it-4bit` (orchestrator runs this; see Real-model verification below)

## Real-model verification (phase 3)

Live server on `models/diffusiongemma-26B-A4B-it-4bit`:

- `GET /v1/models` returns the model entry correctly.
- Chat completions non-streaming: reasoning to content channel transition confirmed, `finish_reason: stop`, usage token counts correct.
- Chat completions streaming (SSE): `delta.reasoning_content` carries the thinking channel; `delta.content` carries the final reply; `finish_reason: stop` on the closing chunk; usage counts match non-streaming.
- Base64 `image_url` request: 256 soft tokens expanded, model answered "a red square and a blue circle" correctly.
- Serve-level flags: `--diffusion-sampler confidence-threshold --diffusion-threshold 0.6` verified effective (178 tokens generated in 7.8 s wall clock).
- `POST /v1/completions`: mechanical-OK; raw un-templated prompts degrade as documented (the model depends on its chat structure), so chat completions are the recommended path.
- CLI deterministic text regression: output byte-identical across all three phases (md5 `c38559cc`).

Closes #217

### DIFF
--- a/docs/block-diffusion.md
+++ b/docs/block-diffusion.md
@@ -104,12 +104,31 @@ Preprocessing reuses the Gemma 4 vision pipeline: images are resized and padded 
 
 Video and audio inputs are not supported and are rejected with a clear error message.
 
-## Remaining limitations (phase 2)
+## Server mode
 
-The following are not yet supported:
+`mlxcel-server --model <diffusion-model>` (or `mlxcel serve -m <diffusion-model>`) loads the checkpoint and serves `/v1/chat/completions` and `/v1/completions`. Both streaming (SSE) and non-streaming responses are supported. Requests are served serially: DiffusionGemma cannot join the batch scheduler, so the server queues them in arrival order and processes one at a time. That is the intended design, not a temporary limitation.
 
-- **Server mode**: `mlxcel serve` and `mlxcel-server` reject DiffusionGemma at
-  load time. Use `mlxcel generate` from the CLI. Server support is planned for phase 3 of issue #217.
+Serve-level flags that affect diffusion models:
+
+| Flag | Default | Notes |
+|------|---------|-------|
+| `--diffusion-sampler entropy-bound\|confidence-threshold` | `entropy-bound` | Per-step token acceptance strategy. |
+| `--diffusion-threshold FLOAT` | `0.9` | Confidence threshold; validated to `[0, 1]` at startup. |
+| `--max-denoising-steps N` | checkpoint default (typically 48) | Maximum denoising iterations per canvas block. |
+
+Image input follows the standard OpenAI `image_url` content part format. Requests that include video or audio are rejected with a clear error; the connection stays open and the next request is served normally.
+
+### Operational caveats
+
+The checkpoint is a thinking-channel model. The assistant fills a `reasoning_content` channel before writing the visible reply. A few practical consequences:
+
+- **Give requests enough `max_tokens`.** With a budget of 64 tokens a simple question may produce only reasoning, leaving `content` empty with `finish_reason: "length"`. A budget of 256 produces the full answer for most short prompts.
+- **Non-streaming responses omit reasoning.** The server's existing convention is to strip `reasoning_content` from non-streaming chat completions. Streaming responses carry it as `delta.reasoning_content`.
+- **`/v1/completions` works but raw prompts degrade.** The endpoint accepts un-templated text mechanically, but the model depends on its chat structure, so output quality degrades significantly. Prefer `/v1/chat/completions`.
+- **Cancellation resolution.** A cancelled request is aborted within at most one denoising step, not instantaneously.
+
+## Remaining limitations
+
 - **Tensor parallelism**: The TP planner has a placeholder arm for this family.
 
 ## Usage example
--- a/docs/supported-models.md
+++ b/docs/supported-models.md
@@ -44,7 +44,7 @@ requirements. If a checkpoint fails detection or loading, inspect its
 
 | Family | `model_type` key | Notes |
 |--------|-----------------|-------|
-| DiffusionGemma | `diffusion_gemma` / `diffusion_gemma_text` | Block-diffusion on a Gemma 4 MoE backbone. Generates a canvas of tokens per block through iterative denoising rather than token-by-token left-to-right decoding. CLI (`mlxcel generate`) supports text and image input (`--image <path>`, repeatable). Server mode is not yet supported. See [Block-diffusion generation](block-diffusion.md). |
+| DiffusionGemma | `diffusion_gemma` / `diffusion_gemma_text` | Block-diffusion on a Gemma 4 MoE backbone. Generates a canvas of tokens per block through iterative denoising rather than token-by-token left-to-right decoding. CLI (`mlxcel generate`) supports text and image input (`--image <path>`, repeatable). Served in `mlxcel-server` (serial, batch-1 by design) via `/v1/chat/completions` and `/v1/completions`; image input follows the standard `image_url` content part format. See [Block-diffusion generation](block-diffusion.md). |
 
 DiffusionGemma uses a two-phase forward pass: an encoder prefill that caches the
 prompt into dense FP16 KV caches, then a canvas loop that attends bidirectionally
--- a/src/bin/mlx_server.rs
+++ b/src/bin/mlx_server.rs
@@ -56,19 +56,6 @@ use mlxcel::server::{
 ///    `download` fetches a HuggingFace model snapshot using the same
 ///    downloader the `mlxcel` CLI uses. Server flags are
 ///    rejected when a subcommand is supplied.
-/// Clap value parser: an f32 in the closed interval [0, 1].
-///
-/// Used by: `--diffusion-threshold` (fail fast at startup instead of
-/// surfacing a per-request engine error under the confidence sampler).
-fn parse_unit_interval(s: &str) -> Result<f32, String> {
-    let v: f32 = s.parse().map_err(|e| format!("not a number: {e}"))?;
-    if (0.0..=1.0).contains(&v) {
-        Ok(v)
-    } else {
-        Err(format!("must be between 0 and 1, got {v}"))
-    }
-}
-
 #[derive(Parser, Debug)]
 #[command(
     name = "mlxcel-server",
@@ -140,6 +127,19 @@ struct Cli {
     /// `args_conflicts_with_subcommands = true` on the parent command).
     #[command(flatten)]
     server: ServerArgs,
+}
+
+/// Clap value parser: an f32 in the closed interval [0, 1].
+///
+/// Used by: `--diffusion-threshold` (fail fast at startup instead of
+/// surfacing a per-request engine error under the confidence sampler).
+fn parse_unit_interval(s: &str) -> Result<f32, String> {
+    let v: f32 = s.parse().map_err(|e| format!("not a number: {e}"))?;
+    if (0.0..=1.0).contains(&v) {
+        Ok(v)
+    } else {
+        Err(format!("must be between 0 and 1, got {v}"))
+    }
 }
 
 /// Subcommands supported by `mlxcel-server`.
--- a/src/bin/mlx_server.rs
+++ b/src/bin/mlx_server.rs
@@ -297,6 +297,29 @@ struct ServerArgs {
     )]
     enable_vlm_prefix_cache: bool,
 
+    /// Maximum denoising steps per canvas block (diffusion models only;
+    /// default: the checkpoint's generation_config, typically 48)
+    #[arg(long = "max-denoising-steps", value_name = "N")]
+    max_denoising_steps: Option<usize>,
+
+    /// Per-step acceptance sampler for diffusion models (diffusion models only)
+    #[arg(
+        long = "diffusion-sampler",
+        value_name = "SAMPLER",
+        default_value = "entropy-bound",
+        value_parser = ["entropy-bound", "confidence-threshold"]
+    )]
+    diffusion_sampler: String,
+
+    /// Confidence threshold for `--diffusion-sampler confidence-threshold`
+    /// (diffusion models only)
+    #[arg(
+        long = "diffusion-threshold",
+        value_name = "FLOAT",
+        default_value_t = 0.9
+    )]
+    diffusion_threshold: f32,
+
     /// Preemption policy: "longest-first" (default) or "lowest-priority"
     #[arg(long = "preemption-policy", default_value = "longest-first")]
     preemption_policy: String,
@@ -1273,6 +1296,11 @@ fn build_startup_input(mut args: ServerArgs) -> anyhow::Result<ServerStartupInpu
         // the flag, so no separate env-fallback helper is needed.
         #[cfg(feature = "surgery")]
         surgery_config_path: args.surgery,
+        // serve-level block-diffusion knobs (#217 phase 3); diffusion models
+        // only.
+        max_denoising_steps: args.max_denoising_steps,
+        diffusion_sampler: args.diffusion_sampler.clone(),
+        diffusion_threshold: args.diffusion_threshold,
     })
 }
 
--- a/src/bin/mlx_server.rs
+++ b/src/bin/mlx_server.rs
@@ -56,6 +56,19 @@ use mlxcel::server::{
 ///    `download` fetches a HuggingFace model snapshot using the same
 ///    downloader the `mlxcel` CLI uses. Server flags are
 ///    rejected when a subcommand is supplied.
+/// Clap value parser: an f32 in the closed interval [0, 1].
+///
+/// Used by: `--diffusion-threshold` (fail fast at startup instead of
+/// surfacing a per-request engine error under the confidence sampler).
+fn parse_unit_interval(s: &str) -> Result<f32, String> {
+    let v: f32 = s.parse().map_err(|e| format!("not a number: {e}"))?;
+    if (0.0..=1.0).contains(&v) {
+        Ok(v)
+    } else {
+        Err(format!("must be between 0 and 1, got {v}"))
+    }
+}
+
 #[derive(Parser, Debug)]
 #[command(
     name = "mlxcel-server",
@@ -316,7 +329,8 @@ struct ServerArgs {
     #[arg(
         long = "diffusion-threshold",
         value_name = "FLOAT",
-        default_value_t = 0.9
+        default_value_t = 0.9,
+        value_parser = parse_unit_interval
     )]
     diffusion_threshold: f32,
 
--- a/src/commands/generate_diffusion.rs
+++ b/src/commands/generate_diffusion.rs
@@ -27,63 +27,27 @@ use std::path::Path;
 use mlxcel::models::DiffusionGemmaModel;
 use mlxcel::models::diffusion_gemma::{
     DiffusionGenerateOptions, DiffusionGenerationStats, DiffusionSamplerKind,
-    DiffusionVisionPrefill,
+    DiffusionVisionPrefill, PreparedDiffusionImagePrompt,
 };
 use mlxcel::server::model_provider::model_worker::StreamingDecodeState;
 use mlxcel::tokenizer::MlxcelTokenizer;
-use mlxcel::vision::processors::gemma4::{Gemma4ImageInput, Gemma4Processor};
 
 use super::generate::print_generation_preamble;
 use crate::GenerateArgs;
 
-/// Build the Gemma 4 image processor for the DiffusionGemma checkpoint.
-///
-/// The diffusion checkpoint's `image_processor` is identical to the gemma-4
-/// VLM family (size 224x224, patch 16, pooling_kernel_size 3, 280 soft
-/// tokens). Values are read from `processor_config.json` when present and
-/// fall back to those defaults otherwise.
-fn build_image_processor(model_dir: &Path, default_soft_tokens: usize) -> Gemma4Processor {
-    let image_cfg = std::fs::read_to_string(model_dir.join("processor_config.json"))
-        .ok()
-        .and_then(|s| serde_json::from_str::<serde_json::Value>(&s).ok())
-        .and_then(|cfg| cfg.get("image_processor").cloned());
-    let get = |key: &str, default: usize| -> usize {
-        image_cfg
-            .as_ref()
-            .and_then(|cfg| cfg.get(key))
-            .and_then(|v| v.as_u64())
-            .map(|v| v as usize)
-            .unwrap_or(default)
-    };
-    Gemma4Processor::new(
-        get("patch_size", 16),
-        get("max_soft_tokens", default_soft_tokens),
-        get("pooling_kernel_size", 3),
-    )
-}
-
-/// Result of preparing image input: the expanded prompt ids (with each image
-/// placeholder rewritten to `boi + image_token * N + eoi`) and the prefill.
-struct PreparedDiffusionVision {
-    expanded_ids: Vec<i32>,
-    prefill: DiffusionVisionPrefill,
-}
-
 /// Preprocess `--image` paths, expand the prompt, and build the vision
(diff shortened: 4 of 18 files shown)


inureyes:

### Intent
Phase 3 (final) of #217: serve DiffusionGemma block-diffusion in mlxcel-server via a dedicated single-stream (batch-1) worker loop, with serve-level diffusion flags threaded through, while leaving autoregressive serving byte-identical.

### Findings Addressed
None requiring changes. Implementation is correct, complete, and well integrated.

### Verification
- [x] All stated requirements implemented: diffusion worker loop, options mapping, flag threading, model_worker branching, CLI vision refactor all present and functional
- [x] Integrated into project code flow: both spawn paths (`spawn_model_worker_with_batch_config` + `spawn_legacy_model_worker`) branch into `run_diffusion_worker_loop` after load; flags thread `ServeArgs.diffusion` -> `ServerStartupInput` -> `ServerStartupConfig` -> `ServerConfig` -> `WorkerSchedulerConfig` -> worker, plus the standalone `mlx_server` binary
- [x] Project conventions followed: pure helpers split out for unit testing, doc comments, no em dashes, Apache headers
- [x] Existing modules reused: `StreamingDecodeState`, `decode_request_images`, `build_generation_result_with_cache`, shared `prepare_image_prompt`
- [x] No unintended structural changes: the AR path is preserved byte-identical (`match model { DiffusionGemma => {...; return}, other => other }` rebinds the same value for every non-diffusion variant; the removed `server_unsupported_model_error` had no other callers)
- [x] Tests pass

### Detail
- **Regression surface (model_worker.rs):** the match/return restructuring keeps the existing flow unchanged for AR models. `model` is rebound to the identical value; nothing between load and the branch consumes it differently. `config_eos` is read before the branch and passed through. Verified.
- **GenerateEvent sequencing:** every path emits exactly one terminal event. Early failures (audio/video, tokenize, image decode, no-vision, empty prompt) send a single `Error`. Success streams `Token`s then one `Done`. A mid-run engine `Err` yields `Token`s then one `Error`. Send failures to a gone receiver are harmless.
- **Cancellation:** the callback polls `cancelled` even when the byte-fallback decoder emits no text this token, and aborts on a closed receiver. Granularity is per committed token (block boundary), which is the engine's natural unit.
- **Stats accuracy:** `completion_tokens` counts actually-streamed tokens (incremented in `on_token`), `prompt_tokens = engine_prompt.len()` (post image-expansion), timing from `first_token_time`/`start`. The `build_generation_result` finish_reason is overridden with the diffusion engine verdict.
- **Options mapping:** max_tokens/temperature/sampler/threshold/max_denoising_steps map correctly; `extra_eos_token_ids` is the dedup union of request stop ids then config EOS (engine unions embedded EOS on top); seed applied via `random_seed` before generation. Serve default threshold 0.9 is in the engine-validated [0,1] range and only used by the confidence-threshold sampler.
- **CLI refactor:** `prepare_image_prompt`/`build_image_processor` moved onto `DiffusionGemmaModel`; `generate_diffusion.rs` delegates and prints the same output. Behavior unchanged.

### Checks run (no real model / no server, per constraints)
- `cargo check --lib --tests --bins --features metal,accelerate`: clean
- `cargo clippy --lib --tests --features metal,accelerate -- -D warnings`: clean
- `cargo test --lib server::diffusion_worker`: 7 passed
- `cargo test --lib models::diffusion_gemma`: 32 passed (2 ignored real-model)
- `cargo test --lib server::cli_input`: 93 passed
- `cargo test --bin mlxcel serve`: 13 passed

### Non-blocking observations (no change made)
- The legacy `--no-batch` worker uses `DiffusionServeDefaults::default()` and ignores the serve-level diffusion flags. This is documented in the code comment and is an intentional limitation.
- `--diffusion-threshold` is not range-validated at the CLI layer; an out-of-range value surfaces as a per-request `Error` only when a confidence-threshold request arrives. The default is valid, so this is a minor UX nit, not a defect.

inureyes:
## Security and performance review (network-facing surface)

Reviewed the phase-3 diffusion serving path with a focus on untrusted request fields, DoS/queue behaviour, channel hygiene, AR-model regression vectors, and per-request overhead. Net result: one fixed regression, two non-blocking observations, no CRITICAL/HIGH.

### Fixed (committed `4f96efe6f`)
- **MEDIUM (regression):** the new `parse_unit_interval` and its doc comment were inserted in `src/bin/mlx_server.rs` directly between the top-level doc block and the `#[derive(Parser)] Cli` struct. With no intervening item the whole "Usage modes" long-help block merged onto the function and detached from `Cli`, so `mlxcel-server --help` lost its long help, and clippy `doc_lazy_continuation` failed under `--tests --bins`. Moved the function below `Cli` (matching `src/main.rs`); parser behaviour unchanged. Warm clippy `--lib --tests --bins` and the `server::diffusion_worker` / `models::diffusion_gemma` suites are green.

### Verified safe (no action)
- **Image decode bombs:** `decode_request_images` applies the shared `current_image_input_limits()` (max alloc 512 MiB, max dims, payload/count caps) identical to the AR VLM path. No bypass.
- **max_tokens / request size:** the AR path itself does not clamp `max_tokens` upward (`overrides.max_tokens.unwrap_or(default)`); the diffusion loop bounds work by `max_new_tokens` the same way. No new uncapped vector and no AR cap is bypassed.
- **Channel hygiene:** every `response_tx.send` is `let _ =`; a dropped receiver makes the streaming callback return `false`, which aborts the denoise loop cleanly. No panic/wedge on client disconnect.
- **Shutdown / loaded flag:** `ModelRequest::Shutdown` breaks the loop, channel close ends it; `loaded.store(true)` runs before both the batched and legacy diffusion branches, so readiness is still signalled. The non-diffusion `match { other => other }` rebind is behaviour-neutral for AR models.
- **Per-request error paths:** all failure paths emit a single `Error` and continue. No `catch_unwind` exists anywhere in the worker layer today, so the diffusion loop correctly matches existing behaviour (an engine panic kills the worker thread regardless of model type).
- **Usage accounting:** `completion_tokens` is counted in `StreamingDecodeState::on_token` for every streamed token; EOS is not streamed/counted (standard). Prompt count uses the expanded (image-inclusive) prompt. Finish reason is overridden by the engine verdict.
- **Memory:** the ~1.4 GiB dequantized embedding table is built once per generate call (not per step) and `clear_memory_cache()` runs after each request and between blocks.

### Non-blocking observations (no fix; design-level follow-ups)
- **MEDIUM — cancellation granularity:** the cancel flag and client-disconnect are polled only per committed token, i.e. at block boundaries. `denoise_block` takes no cancel hook, so a cancelled request still finishes its current block (up to `max_denoising_steps`, default 48, forward passes over a 256-token canvas) before stopping. Bounded and consistent with the documented "no in-flight concurrency > 1" design, but on a 26B MoE that is a few seconds of wasted GPU work after cancel. Optional follow-up: thread a lightweight cancel poll into the denoise step loop.
- **LOW — per-request processor rebuild:** `prepare_image_prompt` -> `build_image_processor` re-reads and parses `processor_config.json` from disk on every image request. Negligible next to the diffusion forward passes; could be cached on the model if image throughput ever matters.

No merge performed.

PR #253, opened 2026-06-13
feat(server): add configurable allowed-origins for CORS

### LINKED ISSUE #244: feat(server): add configurable allowed-origins for CORS instead of permissive default
## Summary
`mlxcel-server` hardcodes `CorsLayer::permissive()` (`src/server/app.rs:158`), so it accepts cross-origin requests from any origin. Add an opt-in way to restrict the allowed origins, defaulting to today's behavior for backward compatibility.

## Why it matters
A permissive CORS policy lets any web page open in a browser issue cross-origin requests to a locally or network-exposed `mlxcel-server`. For any browser-reachable deployment this is a CSRF / origin-spoofing surface. Operators should be able to pin the server to a known set of front-end origins.

## Current behavior
- `src/server/app.rs:158` builds the router with `.layer(CorsLayer::permissive())` unconditionally.
- There is no CLI flag or config field to constrain origins.

## Proposed change
This is a mlxcel-native server change built on `tower-http`. There is no upstream code worth porting here: the reference implementations live on a different web stack, so their logic does not translate to the axum / tower-http layer mlxcel uses. Implement directly against the Rust types.

- Add an `--allowed-origins <ORIGINS>` CLI flag (comma-separated) plus a matching environment variable consistent with the server's existing env conventions, plumbed through the existing ServeArgs to ServerStartupInput to ServerConfig chain.
- When set, build a restrictive layer with `tower_http::cors::AllowOrigin::list(...)` from the parsed origins (preserving the methods and headers currently allowed), instead of `CorsLayer::permissive()`.
- When unset, keep the current permissive behavior so existing deployments are unaffected. Document the default explicitly.

## Scope
This hardening targets the network TCP HTTP listener (`serve_tcp`, used when `port != 0`), which is the browser-reachable surface CORS governs. Out of scope:
- Unix domain socket serving (`serve_unix_socket`, used when `port == 0`): browsers cannot connect to a UDS, so the cross-origin / CSRF threat does not reach this transport. The shared `app` still carries the CORS layer, so the restrictive layer must not break local non-browser clients over the socket (they send no `Origin` header). Keeping permissive on the socket path is acceptable.
- Disaggregated node-to-node serving: the prefill/decode control frames (`serving_protocol.rs` over raw `TcpTransport`) are not HTTP and never touch the CORS layer.

## Acceptance criteria
- [x] `--allowed-origins` CLI flag and matching env var parsed and threaded to the server config.
- [x] Restrictive `CorsLayer` (`AllowOrigin::list`) built when origins are provided; permissive retained when absent.
- [x] Invalid origin values are rejected or surfaced clearly at startup rather than silently dropped.
- [x] Unit test covering origin parsing and layer selection (allowed vs disallowed origin).
- [x] Wired into the actual server startup path in `src/server/app.rs`, not a standalone helper.
- [x] Documented in the server flags documentation.

## Effort
Medium.

inureyes:
## Summary

`mlxcel-server` applied `tower_http::cors::CorsLayer::permissive()` unconditionally (`src/server/app.rs`), reflecting any `Origin` and exposing every browser-reachable deployment to a CSRF / origin-spoofing surface. This adds an opt-in `--allowed-origins` flag (comma-separated) plus the matching `MLXCEL_ALLOWED_ORIGINS` env var so operators can pin the server to a known set of front-end origins, while keeping the permissive default when the flag is unset so existing deployments are unaffected.

## What changed

- New `src/server/cors.rs`: `parse_allowed_origins` validates each comma-split entry as a bare `scheme://host[:port]` origin (`http`/`https`, no path or query, no control characters), and `build_cors_layer` builds a restrictive layer with `AllowOrigin::list(...)` that narrows only the origin while preserving the methods and headers `permissive()` allowed; an unset/empty list falls back to `CorsLayer::permissive()`.
- Validation runs at the fallible `ServerStartupInput::into_startup_config` startup boundary, so a malformed origin fails startup with a message naming the offending value (for example `invalid --allowed-origins value 'app.example.com': origins must be a scheme://host[:port] with no path, e.g. https://app.example.com`) instead of being silently dropped.
- The value is threaded through `ServeArgs` / `mlxcel-server` args, `ServerStartupInput`, `ServerStartupConfig`, and `ServerConfig`, mirroring the existing `enable_vlm_prefix_cache` plumbing; `create_app` builds the layer from `state.config.cors_allowed_origins`.
- Unit tests in `src/server/cors_tests.rs` cover origin parsing (valid single/multiple, blank-only, missing scheme, path, query, non-http scheme, control characters) and layer behavior driven through `tower::ServiceExt::oneshot` (allowed origin reflected in `access-control-allow-origin`, disallowed origin gets no ACAO header, permissive default reflects `*`).
- Documented `MLXCEL_ALLOWED_ORIGINS` in `docs/environment-variables.md` and added a `--allowed-origins` example to the README server quickstart.

## Scope

The hardening targets the browser-reachable TCP HTTP listener that CORS governs. The Unix-domain-socket transport sends no `Origin` header, so the restrictive layer is a no-op there and local non-browser clients are unaffected; the shared `app` can carry the same layer on every transport without branching. The disaggregated node-to-node serving path (`serving_protocol.rs` over raw `TcpTransport`) is not HTTP and never touches the CORS layer.

## Test plan

- [x] `cargo check --lib --bin mlxcel-server`
- [x] `cargo check --bin mlxcel` and `cargo test --bin mlxcel build_startup_input --no-run`
- [x] `cargo test --lib server::cors` (13 passed)
- [x] `cargo clippy --bin mlxcel --bin mlxcel-server -- -D warnings`
- [x] `cargo fmt --check`

Closes #244


inureyes:
## Security & performance review (#244 CORS allow-list)

Reviewed the CORS hardening as a security control. No CRITICAL/HIGH findings: there is no allow-list bypass and no credentials-plus-wildcard hazard.

**CORS correctness (security):** the restrictive branch uses `AllowOrigin::list(...)` and never enables `allow_credentials`, so there is no `*`/reflected-origin-with-credentials combination (no tower-http panic, no credentialed wildcard). The default-unset path stays exactly `CorsLayer::permissive()`, so existing clients (including those sending no `Origin`) are unaffected.

**Validation soundness:** `parse_allowed_origins` / `validate_origin` cannot panic (no `unwrap`/`expect` on the operator input) and cannot inject CRLF/control characters (rejected by both `Uri::parse` and `HeaderValue::from_str`).

**Fixed (LOW, fail-closed misconfiguration):** the validator accepted a bare trailing slash (`https://app.example.com/`) and authority userinfo (`http://user@host`). Both validate at startup but can never match a real browser `Origin`, so the entry would silently never apply. The previous check used `Uri::path()`, which normalizes an empty path to `/` and so could not tell `https://host` from `https://host/`. `validate_origin` now inspects the raw authority and rejects `/`, `?`, `#`, `@` before the structural parse, so a misconfiguration now fails startup with a clear message (issue AC: "surfaced clearly at startup rather than silently dropped"). Added `cors_tests` for both cases; valid-origin tests still pass.

**Performance:** the layer is built once in `create_app`; the only clones of the origin list happen at startup (`build_server_config`, `AllowOrigin::list`). No per-request allocation. No DoS surface (operator-supplied, comma-split, bounded).

Pushed as `f2e584a6b`.

Issue #307, opened 2026-06-16
epic(moe): wire the remaining unwired MoE families to the fused decode-MoE kernel


inureyes:
## Background

The fused single-token decode-MoE kernel (#268 series, steps 2b-3b) and the default-on flip (#282, PR #285) are wired into only four model paths: `qwen3_moe`, `qwen3_next`, `dots1`, `gemma4`. #286 wires `qwen2_moe`. Seven more MoE families reuse a `SwitchGLU` for the expert matmul but never call `forward_fused_kernel`, so they always fall back to `gather_qmm` regardless of `MLXCEL_FUSED_MOE`. This epic tracks wiring those remaining families.

The canonical one-branch pattern lives in `src/models/qwen3_moe.rs`, `src/models/dots1.rs`, and `src/models/qwen3_next.rs`: gate on single-token decode plus `fused_moe_enabled()` (`src/models/switch_layers.rs:42`), call `experts.forward_fused_kernel(&x_flat, &topk_indices, &scores)` (`src/models/switch_layers.rs:276`), and fall back to `experts.forward(...)` plus `moe_weighted_sum(...)` (`src/models/switch_layers.rs:419`). All target families use SwiGLU (SiLU) experts, so the kernel runs with the default swiglu activation (act=0); no GeGLU/relu2 template is needed. Kernel gating: affine 4/8-bit gate/up, 4/6/8-bit down, matching `group_size`, single-token decode only; it auto-falls-back for any unsupported config.

## Two implementation classes

This is the key technical finding, verified in-tree.

**Group A (direct one-branch wiring):** the model's experts already use the shared `crate::models::switch_layers::SwitchGLU`, which exposes `forward_fused_kernel`. Wiring is the small one-branch change from #286/dots1.

- `minimax` (imports `crate::models::switch_layers::SwitchGLU`).
- `lfm2` (imports `super::switch_layers::SwitchGLU`; only the LFM2-MoE variant has the MoE block).
- `qwen3_vl_moe` (imports `super::qwen3_moe::SwitchGLU`, which has `forward_fused_kernel`).

**Group B (needs SwitchGLU consolidation first):** the model defines its OWN local `SwitchGLU` struct that does NOT expose `forward_fused_kernel`. These must migrate to the shared `switch_layers::SwitchGLU` (preferred: removes duplication) OR have a `forward_fused_kernel` method added to the local struct, BEFORE the one-branch dispatch works.

[...]

Issue #286, opened 2026-06-14
perf(moe): wire qwen2_moe to the fused decode-MoE kernel


inureyes:
## Problem / Background

PR #285 (issue #282) flipped `MLXCEL_FUSED_MOE` to on by default and, during M5 validation, found the fused single-token decode-MoE kernel (#268) is only wired into four model paths: `qwen3_moe`, `qwen3_next`, `dots1`, and `gemma4`. Several other MoE families reuse the shared `SwitchGLU` for the expert matmul but never call `forward_fused_kernel`, so they always fall back to `gather_qmm` regardless of the flag.

`qwen2_moe` (which serves qwen1.5-moe / Qwen2-MoE) is one of them. It was listed in #282's validation set but could not be validated, because it never dispatches the kernel (0 dispatches confirmed via a dispatch trace during #282). The doc's "Models covered" list overstated coverage and was corrected in #285.

## Scope

Wire `qwen2_moe`'s `SparseMoeBlock` to dispatch the fused kernel for single-token decode, mirroring `qwen3_moe` / `dots1`. This is low-risk and small:

- `qwen2_moe::SparseMoeBlock` already holds `experts: SwitchGLU` (`src/models/qwen2_moe.rs:398`), and `SwitchGLU` already exposes `forward_fused_kernel`.
- The block's `forward` (`src/models/qwen2_moe.rs:436-443`) currently always calls the non-fused `self.experts.forward(...)` + `moe_weighted_sum`. Replace only the routed-expert computation with the standard fused-gated branch, leaving the shared-expert addition (lines 445-456) intact.

Concrete change (mirrors `dots1.rs` / `qwen3_moe.rs`):

```rust
let expert_out_sum = {
    let fused = if mlxcel_core::array_shape(&x_flat)[0] == 1
        && crate::models::switch_layers::fused_moe_enabled()
    {
        self.experts
            .forward_fused_kernel(&x_flat, &topk_indices, &scores)
            .map(|out| mlxcel_core::reshape(&out, &[1, hidden_dim]))
    } else {
        None
    };
    match fused {
        Some(out) => out,
        None => {
            let expert_out = self.experts.forward(&x_flat, &topk_indices);
            crate::models::switch_layers::moe_weighted_sum(
                &expert_out, &scores, mlxcel_core::array_dtype(&x_flat),
            )
        }
    }
};
```

[...]

Issue #282, opened 2026-06-14
perf(moe): validate fused decode-MoE on M5, then enable MLXCEL_FUSED_MOE by default


inureyes:
## Problem / Background

The fused single-token decode-MoE kernel effort (#268) is implemented and merged across PRs #276 (two-kernel SwiGLU), #278 (6-bit/mixed-bit, dots.llm1), #279 (qwen3_next / qwen3.5/3.6), #280 (squared-ReLU kernel preserved behind `MLXCEL_FUSED_MOE_RELU2`), and #281 (GeGLU, gemma4). The feature is currently **off by default**, opt-in via the `MLXCEL_FUSED_MOE` env flag, scoped to single-token decode only, with 4/8-bit gate-up + 4/6/8-bit down affine quantization.

Measured on **M1 Ultra** (the only hardware tested so far), decode throughput with the flag on, all with no regression:

| Model | Activation / quant | Gain | Numerical status |
|-------|--------------------|------|------------------|
| gemma-4-26b-a4b-it | GeGLU 4-bit | +13% (73.8 -> 83.2 tok/s) | greedy byte-identical (with chat template) |
| qwen3.5/3.6-35b-a3b | SwiGLU 4-bit | +8.7% (68.7 -> 74.7 tok/s) | within f16 jitter class (greedy may flip near-tie tokens; abs RMS ~1e-4 < 5e-3) |
| dots.llm1 | SwiGLU mixed 4/6-bit | +4.7% (13.1 -> 13.7 tok/s) | byte-identical |
| qwen3-30b-a3b | SwiGLU 4-bit | +3.5% (47.3 -> 49.0 tok/s) | byte-identical |
| qwen1.5-moe-a2.7b + other SwiGLU 4-bit | SwiGLU 4-bit | ~parity, no regression | byte-identical |
| nemotron-h-30b | squared-ReLU | ~0% | reverted from default path, kept behind `MLXCEL_FUSED_MOE_RELU2` (MoE is a small slice of its Mamba-hybrid decode) |

The f16 jitter on deep hybrids (qwen3.5/3.6) is the expected numerical consequence of kernel fusion, not a bug. The design harness accepts "byte-identical OR within f16 jitter class" (abs RMS ~1e-4 < 5e-3 threshold).

## M5-specific risk (core reason for this issue)

[...]

Issue #268, opened 2026-06-13
perf(models): general MoE decode gap vs mlx-lm (dots.llm1 0.51x, qwen3_moe 0.69x, nemotron MoE)


inureyes:
## Problem / Background

On M1 Ultra, `dots.llm1.inst-mixed-4-6bit` decodes at ~13.2 tok/s in mlxcel versus ~25.7 tok/s in mlx-lm (0.51x, a ~2x gap). This is NOT a Mamba2 hybrid. dots.llm1 is a DeepSeek-V3-style MoE (sigmoid router with `e_score_correction_bias` selection-only bias, unbiased gather, norm_topk_prob, routed_scaling_factor) with mixed 4-bit/6-bit expert quantization (`v_proj`/`down_proj` at 6-bit). For comparison on the same host and harness, the other MoE family (`LFM2-8B-A1B`, lfm2-moe) is at parity (~1.00x) and dense `granite-3.3-2b` is at ~0.99x, so the gap is specific to this model's routing / attention / mixed-precision-gather path, not a general MoE problem.

Measured on an Apple M1 Ultra (128 GB, macOS 26.5.1, MLX commit a6ec712) against mlx-lm 0.31.3. Decode throughput was measured with `mlxcel-bench-decode` (warmup pass + measured pass in one process, `--no-chat-template`) versus `scripts/bench_mlxlm.py` (mlx-lm's own prompt_tps/generation_tps), identical checkpoint, raw continuation prompt, 100 decode tokens, best-of-3 with cooldown, temp-0 greedy.

## Proposed Solution

Investigate and root-cause the ~2x decode gap. Suspects to profile:

- SwitchGLU / `gather_qmm` with mixed 4/6-bit experts (the per-tensor bit inference added in PR #263 / `switch_layers.rs`).
- The attention path (whether dots.llm1 uses MLA and how the absorbed/decompressed projections are implemented vs mlx-lm).
- Router overhead (group_mask_scores, sigmoid + correction-bias selection).
- Any extra `eval()` / synchronization points or avoidable astype/graph nodes in the decode step.

Use `scripts/bench_decode.sh` / `mlxcel-bench-decode` and `scripts/bench_mlxlm.py`; the `mlxcel-gpu-profiling` skill may help localize the hotspot.

## Acceptance Criteria

- [x] Dominant cause of the ~2x decode gap identified and documented.
- [x] Gap closed meaningfully (target parity-ish with mlx-lm).
[...]

PR #311, opened 2026-06-16
perf(moe): wire mixtral to the fused decode-MoE kernel

### LINKED ISSUE #301: perf(moe): wire mixtral to the fused decode-MoE kernel
Parent epic: #307. Sibling: #286 (qwen2_moe, same pattern).

## Background

The fused single-token decode-MoE kernel (#268 series) and the default-on flip (#282, PR #285) are wired into only four model paths: `qwen3_moe`, `qwen3_next`, `dots1`, `gemma4`. #286 wires `qwen2_moe`. Several more MoE families reuse a `SwitchGLU` for the expert matmul but never call `forward_fused_kernel`, so they always fall back to `gather_qmm` regardless of `MLXCEL_FUSED_MOE`. This issue wires `mixtral`.

The canonical one-branch pattern lives in `src/models/qwen3_moe.rs`, `src/models/dots1.rs`, and `src/models/qwen3_next.rs`: gate on single-token decode plus `fused_moe_enabled()` (`src/models/switch_layers.rs:42`), call `experts.forward_fused_kernel(&x_flat, &topk_indices, &scores)` (`src/models/switch_layers.rs:276`), and fall back to `experts.forward(...)` plus `moe_weighted_sum(...)` (`src/models/switch_layers.rs:419`).

## Scope

`src/models/mixtral.rs`. `SparseMoeBlock::forward` is at line 281 (expert call at 313, `moe_weighted_sum` at 315). Routing: softmax over top-k, no shared expert. mixtral experts are SwiGLU (SiLU), so the kernel runs with the default swiglu activation (act=0); no GeGLU/relu2 template is needed.

Implementation class: Group B (needs SwitchGLU consolidation first). mixtral defines its OWN local `SwitchGLU` (line 156) and a local `SwitchLinear` that do NOT expose `forward_fused_kernel`. Before the one-branch dispatch can work, migrate the experts to the shared `crate::models::switch_layers::SwitchGLU` (preferred: removes duplication) OR add a `forward_fused_kernel` method to the local struct. Prefer migration. Verify there is no mixtral-specific behavior in the local struct before removing it.

Validation checkpoint: `mlx-community/Mixtral-8x7B-Instruct-v0.1-4bit` (in the project regression set; about 24 GB at 4-bit, fits 128 GB).

## Acceptance Criteria

- [ ] mixtral's MoE block dispatches the fused kernel for single-token decode when `MLXCEL_FUSED_MOE` is enabled (default) and the config qualifies; the non-fused fallback path is unchanged.
- [ ] The local `SwitchGLU`/`SwitchLinear` is migrated to the shared `switch_layers::SwitchGLU` (or a `forward_fused_kernel` method is added), the duplicate is removed where migrated, and the existing model forward stays correct.
- [ ] INTEGRATION: the change is wired into the actual mixtral `forward` decode path (not an orphaned helper). Confirm the kernel actually dispatches at runtime via a dispatch trace, not inferred from byte-identical output.
- [ ] Validate on `mlx-community/Mixtral-8x7B-Instruct-v0.1-4bit`: greedy temp-0 output is byte-identical or within the documented f16 jitter class vs the `MLXCEL_FUSED_MOE=0` (gather_qmm) baseline, with no NaN and no garbage.
- [ ] Measure decode tok/s fused-on vs off; record the gain; confirm no regression.
- [ ] Update the "Models covered" list in `docs/benchmark_results/fused-moe-decode-kernel-design.md` to move mixtral to the wired list.
- [ ] Add or extend tests where practical.

## Technical Considerations / References

- Pattern files: `src/models/qwen3_moe.rs`, `src/models/dots1.rs`, `src/models/qwen3_next.rs`. Shared helpers: `src/models/switch_layers.rs` (`fused_moe_enabled` at 42, `forward_fused_kernel` at 276, `moe_weighted_sum` at 419).
- Kernel gating: affine 4/8-bit gate/up, 4/6/8-bit down, matching `group_size`, single-token decode only; the kernel auto-falls-back for any unsupported config.
- f16 jitter caveat: the fused kernel accumulates the GEMV in f32 with a different reduction order than `gather_qmm`, so within-jitter greedy token flips are expected and accepted, not a bug.
- Related: #268 (umbrella MoE-gap issue), #282 and #285 (default-on), #286 (qwen2_moe, same pattern).

inureyes:
## Summary

Group B migration for the fused decode-MoE kernel (#268 series, default-on since #282), chain 4/6 under epic #307 and sibling to #286 (qwen2_moe). mixtral's MoE experts move from its local `SwitchGLU`/`SwitchLinear` to the shared `crate::models::switch_layers::SwitchGLU`, and `SparseMoeBlock::forward` gains the single-token decode fused-kernel dispatch.

## What changed

- `src/models/switch_layers.rs`: add `SwitchGLU::from_weights_with_proj_names(weights, prefix, group_size, bits, [gate, up, down])`. The default `from_weights` now delegates to it with the standard `["gate_proj", "up_proj", "down_proj"]` names, so every existing shared caller (qwen2_moe, mistral4, dots1, kimi_linear, minimax, longcat) is byte-for-byte unchanged. The per-expert stacker already derives the expert leaf name from the projection segment of the prefix, so overriding the leaf names is all that is needed; no checkpoint-specific strings live in the shared loader. Doc comment on `stack_individual_experts` updated to note the override path; new unit test `switch_glu_loads_overridden_proj_leaf_names` exercises the `w1`/`w2`/`w3` path.
- `src/models/mixtral.rs`: remove the local `SwitchGLU`/`SwitchLinear` (nothing else referenced them), point `SparseMoeBlock.experts` at the shared `SwitchGLU`, and load via `from_weights_with_proj_names(.., "{prefix}.switch_mlp", .., ["w1", "w3", "w2"])` (gate=w1, up=w3, down=w2). `SparseMoeBlock::forward` now tries `forward_fused_kernel` for single-token decode when `fused_moe_enabled()`, falling back to `experts.forward` + `moe_weighted_sum` otherwise. mixtral uses softmax over the top-k logits with no shared expert and no renorm, so `scores` already carries the full combine weights; everything else in the forward path is unchanged.
- `docs/benchmark_results/fused-moe-decode-kernel-design.md`: move mixtral into the wired "Models covered" list and document the `w1`/`w2`/`w3` mapping and the overridable-leaf-name loader.

## Weight-loading parity

The old local loader read `block_sparse_moe.experts.{idx}.{w1,w2,w3}.{weight,scales,biases}` for idx ascending and stacked them along axis 0 with `utils::stack_arrays`, with `gate_proj=w1`, `up_proj=w3`, `down_proj=w2`. The shared per-expert stacker reads the same keys (`{root}.experts.{idx}.{leaf}` where the `.switch_mlp` virtual prefix resolves `root` to `block_sparse_moe`, identical to the qwen2_moe path), in the same ascending index order, and stacks along axis 0 with `stack_owned`. Both stackers reduce to the same `ffi::stack(&ptrs, 0)` call, so the resulting `[num_experts, out, in_packed]` tensors are identical. Bits: the local loader hard-set `args.bits()`; the shared loader infers per-tensor bits from `(packed_in * 32) / (num_groups * group_size)`, which for Mixtral-8x7B-4bit (group_size 64) is `(512*32)/(64*64) = 4` for w1/w3 and `(1792*32)/(224*64) = 4` for w2, matching `args.bits()`. mode is `affine` (default), biases are present, so `quantized_parts()` is `Some` and the fused kernel qualifies.

## Test plan

- [x] `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` (clean; dead-code from the removed local structs would have failed `-D warnings`)
- [x] `cargo fmt -- --check`
- [ ] Real-checkpoint validation pending the orchestrator on `models/mixtral-8x7b-4bit`: greedy temp-0 output coherent and byte-identical or within the documented f16 jitter class vs `MLXCEL_FUSED_MOE=0`, no NaN, no garbage; decode tok/s fused-on vs off recorded with no regression. A wrong w1/w2/w3 mapping or a bits/shape mismatch would surface as garbage decode here.

Closes #301

### DIFF
--- a/docs/benchmark_results/fused-moe-decode-kernel-design.md
+++ b/docs/benchmark_results/fused-moe-decode-kernel-design.md
@@ -207,14 +207,17 @@ the expected numerical consequence of the fusion.
 
 ### Models covered
 
-The fused single-token decode dispatch is wired into seven model paths: qwen3_moe
+The fused single-token decode dispatch is wired into eight model paths: qwen3_moe
 (Qwen3 MoE), qwen3_next (qwen3.5/3.6), dots.llm1 (mixed 4/6-bit), gemma4 (GeGLU),
 qwen2_moe (qwen1.5-moe / Qwen2-MoE; migrated from its local `SwitchGLU` to the
 shared one, which gained a per-expert stacking loader for the `experts.{idx}`
-checkpoint layout), lfm2 (LFM2-MoE; sigmoid-routed, optional expert_bias and
-norm_topk_prob), and qwen3_vl_moe (Qwen3-VL MoE; imports `SwitchGLU` from qwen3_moe,
-SwiGLU activation, text-only decode path). Other MoE families reuse the shared
-`SwitchGLU` for the expert matmul but were not wired with the fused decode dispatch,
-so they stay on `gather_qmm` regardless of `MLXCEL_FUSED_MOE`: mixtral, olmoe,
-minimax, and phimoe. nemotron-h's MoE runs through the separate C++
-`fused_moe_forward` and is wired behind `MLXCEL_FUSED_MOE_RELU2` only.
+checkpoint layout), mixtral (Mixtral 8x7B; SwiGLU, softmax-routed with no shared
+expert; migrated from its local `SwitchGLU` to the shared one, which gained an
+overridable projection-leaf-name loader for the `w1`/`w2`/`w3` checkpoint
+convention, mapping gate=w1, up=w3, down=w2), lfm2 (LFM2-MoE; sigmoid-routed,
+optional expert_bias and norm_topk_prob), and qwen3_vl_moe (Qwen3-VL MoE; imports
+`SwitchGLU` from qwen3_moe, SwiGLU activation, text-only decode path). Other MoE
+families reuse the shared `SwitchGLU` for the expert matmul but were not wired with
+the fused decode dispatch, so they stay on `gather_qmm` regardless of
+`MLXCEL_FUSED_MOE`: olmoe, minimax, and phimoe. nemotron-h's MoE runs through the
+separate C++ `fused_moe_forward` and is wired behind `MLXCEL_FUSED_MOE_RELU2` only.
--- a/docs/benchmark_results/fused-moe-decode-kernel-design.md
+++ b/docs/benchmark_results/fused-moe-decode-kernel-design.md
@@ -160,6 +160,7 @@ falls back to the proven `gather_qmm` / `SwitchGLU` path automatically. Set
 | `MLXCEL_FUSED_MOE` | unset (on) | On by default. Set to `0` (also `false`/`off`/`no`, case-insensitive) to force the proven `gather_qmm` / `SwitchGLU` path; any other value, or leaving it unset, keeps the kernel on. |
 | `MLXCEL_FUSED_MOE_SGY` | 8 | Simdgroups per threadgroup (one output row each). Tune per hardware. |
 | `MLXCEL_FUSED_MOE_RELU2` | unset (off) | Enable the squared-ReLU (fc1/relu²/fc2) fused path used by nemotron-class experts. Correct but measured performance-neutral on nemotron-h; kept for a future MoE-dominated squared-ReLU model. |
+| `MLXCEL_FUSED_MOE_MAX_DFF` | 4096 | Expert-intermediate (Dff) upper bound. Above it, `forward_fused_kernel` declines and the caller falls back to `gather_qmm`. The fused path wins only while `gather_qmm` underutilizes the GPU (small experts); for large experts `gather_qmm` already saturates and the extra dispatch plus global-memory activation staging is a net loss. M1 Ultra measurements: Dff 704..2560 gain +3.5% to +15.4%, Dff 6400 (phi-3.5-moe) loses 5.9%, Dff 14336 (mixtral) loses 21%. The break-even is hardware-dependent, so the bound is tunable. |
 
 ### Measured decode gains (M1 Ultra, `MLXCEL_FUSED_MOE=1`)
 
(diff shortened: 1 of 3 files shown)


inureyes:
## Orchestrator validation + Dff guard (M1 Ultra)

Real-checkpoint validation on `mlx-community/Mixtral-8x7B-Instruct-v0.1-4bit` surfaced a regression: the fused decode kernel is ~21% slower than `gather_qmm` for mixtral (expert intermediate Dff=14336). The output is correct (coherent, w1/w2/w3 mapping verified), but the two-kernel fused path only wins while `gather_qmm` underutilizes the GPU (small experts); for large experts `gather_qmm` already saturates and the extra dispatch plus global-memory activation staging is a net loss.

To resolve this without regressing the small-expert winners, this PR adds a Dff upper-bound guard to `forward_fused_kernel` (`MLXCEL_FUSED_MOE_MAX_DFF`, default 4096; env-tunable since the break-even is hardware-dependent). The threshold was calibrated against measured break-even data:

| Dff | model | fused vs gather_qmm |
|----:|-------|---------------------|
| 704 | gemma4-26b-a4b | +13% |
| 768 | qwen3-30b-a3b / qwen3-vl-30b-a3b | +3.5% / +15.4% |
| 1408 | qwen2_moe / dots.llm1 | +5.8% |
| 1792 | lfm2 | +6.6% |
| 2560 | qwen3_next | +8.7% |
| 6400 | phi-3.5-moe | -5.9% |
| 14336 | mixtral | -21% |

4096 keeps every measured winner (including qwen3_next at 2560) and excludes both losers (6400, 14336).

Validation with the guard:
- mixtral default (guard on): 51.8-52.5 tok/s == `MLXCEL_FUSED_MOE=0` baseline 52.1 tok/s (falls back, no regression).
- mixtral `MLXCEL_FUSED_MOE_MAX_DFF=99999`: 41.2 tok/s (fused path re-arms, the -21% path returns, proving the guard is the gate).
- qwen2_moe (Dff 1408, below the bound): fused on 146.6 vs off 134.3 tok/s (still gains; guard does not disable small winners).

Net for mixtral: the local `SwitchGLU`/`SwitchLinear` migration to the shared struct removes duplication, and decode stays on the proven `gather_qmm` path.

PR #341, opened 2026-06-17
perf(models): generalize fused QKV+RMSNorm+RoPE to standard RMSNorm

### LINKED ISSUE #326: perf(models): generalize fused QKV+RMSNorm+RoPE to standard-RMSNorm QK-norm models
## Summary

Generalize the existing fused QKV-projection + split + Q/K-norm + RoPE kernel so models using a standard `RMSNorm` (not only `GemmaRMSNorm`) can use it, then wire it into Qwen3 and Qwen3-MoE first, with a staged rollout to the other QK-norm families.

Source: idea.md "2. Generic Fused QKV + RMSNorm + RoPE For QK-Norm Models" (P0, expected payoff medium, risk medium-high).

## Current state

- `FusedQKVLinear::forward_split_norm_rope_quantized()` exists but is specialized to `GemmaRMSNorm` and is documented as "Used by: Gemma3 dense attention path" (`src/lib/mlxcel-core/src/layers.rs:1282`).
- Qwen3 attention uses a fused QKV projection, then separate reshape, `q_norm`, `k_norm`, transpose, and RoPE ops (`src/models/qwen3.rs`, attention forward around line 115).
- Many models still apply Q/K `RMSNorm` as separate ops: Qwen3-MoE, Qwen3-VL, Qwen3-VL-MoE, OLMoE, OLMo2/3, Apertus, Exaone4, Exaone-MoE, and MiniMax.
- The decode-gap investigation reports dense Llama/Granite near parity, while the gap grows with extra small ops such as QK-norm and xIELU (`docs/benchmark_results/moe-decode-gap-investigation.md`).

## Proposal

- Generalize the fused projection/split/norm/rope primitive so standard `RMSNorm` weights can use it, not only `GemmaRMSNorm` (for example, accept a norm trait/enum or a generic norm parameter).
- Wire it first into Qwen3 and Qwen3-MoE, which already have a fused QKV layout and good benchmark coverage.
- Then port compatible QK-norm families in small batches: Qwen3-VL text path, OLMoE/OLMo2/OLMo3, Apertus, Exaone4/Exaone-MoE, MiniMax. These follow-up family rollouts can be split into sub-issues once the primitive is proven.

## Scope

- Generalize `forward_split_norm_rope_quantized` and update its `// Used by:` comment per the shared-function convention in `docs/code-guidelines.md`.
- Wire Qwen3 and Qwen3-MoE in this issue.

## Validation

- Correctness: RMS < 5e-3 against the existing graph path, plus greedy parity for fixed prompts where deterministic parity is expected.
- Performance: `mlxcel-bench-decode` on `qwen3-0.6b-4bit`, `qwen3-8b-4bit`, `qwen3-30b-a3b-4bit`, and `Apertus-8B-2509-4bit`.
- Regression guard: include a non-QK-norm dense model (for example Llama) to prove common attention behavior did not change.
- Suggested first experiment (idea.md): prototype generic RMSNorm support on Qwen3 only first, to isolate the QK-norm hypothesis with minimal model spread.

## Acceptance criteria

- [x] The fused primitive accepts standard `RMSNorm` Q/K norms, not just `GemmaRMSNorm`.
- [x] Qwen3 and Qwen3-MoE attention dispatch the generalized fused path on decode; the non-fused path remains as fallback.
- [x] RMS < 5e-3 vs the graph reference; greedy temp-0 parity (or documented jitter) on the listed checkpoints.
- [x] `mlxcel-bench-decode` numbers recorded; the non-QK-norm dense regression guard shows no change.
- [x] `// Used by:` comment updated; INTEGRATION into the real Qwen3/Qwen3-MoE forward path verified.

## Risk and priority

idea.md priority P0; risk medium-high. Directly targets the "extra small ops per layer" decode gap and yields one shared primitive reusable by many recently added QK-norm families.

## References

- idea.md "2. Generic Fused QKV + RMSNorm + RoPE For QK-Norm Models".
- Code: `src/lib/mlxcel-core/src/layers.rs` (`forward_split_norm_rope_quantized`), `src/models/qwen3.rs`.
- `docs/benchmark_results/moe-decode-gap-investigation.md`, `docs/code-guidelines.md` (shared-function rule).

inureyes:
## Summary

Generalize `FusedQKVLinear::forward_split_norm_rope_quantized` so models with a standard `RMSNorm` Q/K norm can use the fused projection + split + Q/K norm + RoPE kernel, not only Gemma's `(1 + weight)` RMSNorm, then wire it into the Qwen3 and Qwen3-MoE decode attention paths. This targets the per-layer "extra small ops" decode gap that the MoE decode-gap investigation attributes to QK-norm models (`docs/benchmark_results/moe-decode-gap-investigation.md`).

## What changed

- `src/lib/mlxcel-core/src/layers.rs`: new `FusedQkNorm` trait (raw weight for `RMSNorm`, pre-computed `(1 + weight)` for `GemmaRMSNorm`); `forward_split_norm_rope_quantized` is now generic over it. The C++ kernel already applies the plain `x * weight * rsqrt(mean(x^2) + eps)` form, so the Gemma-vs-standard distinction is just which weight the trait hands back, never a kernel branch. The `// Used by:` comment is updated to Gemma3 + Qwen3 + Qwen3-MoE.
- `src/lib/mlxcel-core/src/lib.rs`: FFI doc comment for `fused_qkv_project_split_norm_rope` updated to the generalized usage.
- `src/models/qwen3.rs`, `src/models/qwen3_moe.rs`: each (separate but identical) attention module dispatches the fused path on decode (`l == 1`) and keeps the existing graph sequence as the prefill and non-quantized fallback. The kernel norms after the head transpose while the fallback norms before it; RMSNorm reduces over `head_dim`, which the transpose leaves untouched, so the two orders match per element. Gemma3 keeps passing `GemmaRMSNorm` and stays byte-identical.

## Test plan

- [x] `cargo check -p mlxcel-core --lib --tests` and `cargo check -p mlxcel --lib`
- [x] `cargo clippy -p mlxcel-core --lib --tests -- -D warnings` and `cargo clippy -p mlxcel --lib -- -D warnings` (clean)
- [x] `cargo test --release -p mlxcel-core layers::tests::fused_qkv` (4 passed: standard-RMSNorm match RMS < 5e-3, Gemma unchanged, variant selects `(1 + weight)`, plus the existing fused-QKV bias test). Each comparison test asserts the projection is non-degenerate so it cannot pass on a zero projection.
- [x] `cargo fmt --check`
- [ ] Real-model validation (orchestrator): RMS < 5e-3 vs graph + greedy temp-0 parity + `mlxcel-bench-decode` on qwen3-0.6b/8b/30b-a3b-4bit and Apertus-8B-2509-4bit, with a non-QK-norm dense (Llama) regression guard.

Closes #326


inureyes:
## PR finalization

### Kill-switch: `MLXCEL_FUSED_QK_NORM`

Added `fused_qk_norm_enabled()` in `mlxcel_core::layers`, following the exact pattern of `fused_moe_enabled()` in `switch_layers.rs`:

- Backed by a `static OnceLock<bool>`, read from `MLXCEL_FUSED_QK_NORM` at first call and cached for the process lifetime. Per-step env lookups are not possible.
- `fused_qk_norm_enabled_from(Option<&str>)` is a pure, testable helper (same split as `fused_moe_enabled_from`).
- Accepted disable values: `0`, `false`, `off`, `no` (case-insensitive, trimmed). Unset or any other value keeps the kernel on.
- Gate is applied in both `qwen3.rs` and `qwen3_moe.rs`: fused path requires `l == 1` AND quantized weights AND kill-switch not disabled. `MLXCEL_FUSED_QK_NORM=0` forces the graph fallback for both.

### Defensive eps assertion

`FusedQKVLinear::forward_split_norm_rope_quantized` now has `debug_assert_eq!(q_norm.rms_eps(), k_norm.rms_eps(), ...)` immediately before the FFI call. Compiled out in release; zero hot-path cost.

### Test

New `fused_qk_norm_enabled_defaults_on_and_respects_disable_values` test covers the pure helper. All 25 `layers::` tests pass.

### Docs

`MLXCEL_FUSED_QK_NORM` added to `docs/environment-variables.md` next to the `MLXCEL_FUSED_MOE` family. No Korean mirror of that file exists.

### Commits

- `75947da` feat(layers): add MLXCEL_FUSED_QK_NORM kill-switch and eps assertion
- `90f2e45` docs: document MLXCEL_FUSED_QK_NORM in environment-variables.md

Clippy and fmt clean on both crates. Ready for real-model validation with `MLXCEL_FUSED_QK_NORM=0` for fused-vs-graph A/B.

inureyes:
## Real-model validation + perf decision (orchestrator)

Validated on M1 Ultra (Metal). Branch built clean (`metal,accelerate`).

### Correctness: PASS (byte-identical)
Within-binary A/B (fused default vs `MLXCEL_FUSED_QK_NORM=0` graph path), greedy `temp=0`, `--no-chat-template`, n=64:

| Model | Attention path | fused vs graph generated text |
|-------|----------------|-------------------------------|
| qwen3-0.6b-4bit | dense (`qwen3.rs`) | byte-identical |
| qwen3-8b-4bit | dense (`qwen3.rs`) | byte-identical |
| qwen3-30b-a3b-4bit | MoE (`qwen3_moe.rs`) | byte-identical |

Both wired attention structs produce output identical to the graph path (the RMSNorm reduces over the transpose-invariant head_dim axis, so norm-before-transpose and norm-after-transpose commute exactly). The unit test already shows RMS < 5e-3 vs graph on synthetic quantized weights.

### Performance: fused path is a small regression on M1 Ultra
Warmup (64) + measured (256) decode tok/s, 2 runs each, fused vs graph:

| Model | fused tok/s | graph tok/s | delta |
|-------|-------------|-------------|-------|
| qwen3-0.6b-4bit | 275.5 / 273.8 | 284.1 / 284.3 | about -3.4% |
| qwen3-8b-4bit | 82.3 / 82.4 | 83.2 / 83.2 | about -1.1% |

Consistent across runs, so not noise. As the security review noted, the fused kernel cuts Rust/C++ FFI crossings rather than MLX op count; on Apple Silicon's fast FFI that saving is net-negative.

### Decision: default the kill-switch OFF (opt-in)
Commit `5a5cb069d` flips `MLXCEL_FUSED_QK_NORM` to default off (opt-in via `1`/`true`/`on`/`yes`). The default decode path is now the unchanged graph path, so no regression ships on by default. The generalized `FusedQkNorm` primitive still ships as the reusable foundation for the deferred QK-norm families, and the fused dispatch can be enabled where it wins. This mirrors the opt-in treatment of the performance-neutral `MLXCEL_FUSED_MOE_RELU2` path. A per-backend re-bench (notably CUDA, where op-dispatch and FFI cost differ) is the follow-up to decide where, if anywhere, to default it on.

Reviewer, security-checker, and finalizer: 0 CRITICAL/HIGH/MEDIUM. Acceptance criteria met: generalized primitive accepts standard RMSNorm, Qwen3 + Qwen3-MoE dispatch the fused path (now opt-in), RMS < 5e-3 + byte-identical greedy parity, `// Used by:` updated, kill-switch + docs added.

Issue #289, opened 2026-06-15
perf: bf16->f16 quant-scale promotion (#260) regresses bf16-scale quantized model decode ~33-41% on M1 Ultra


inureyes:
## Summary
Commit `d80f78721` ("feat(models): add Apertus (xIELU, QK-norm, llama3 RoPE scaling) (#260)") added `convert_quant_scales_bf16_to_f16()` in `src/models/sanitize.rs::load_text_weights`, which promotes bf16 `.scales`/`.biases` to f16 for **all** quantized models. This regresses decode throughput by ~33-41% on M1 Ultra for every quantized checkpoint that ships **bf16** quant scales, while checkpoints shipping **f16** scales are unaffected.

## Evidence (M1 Ultra, 128GB, MLX pin a6ec712, decode tok/s)
Confirmed by a control-gated, median-of-3 git bisect (mixtral as a thermal control, since the M1 Ultra throttles under sustained load and naive measurement is noisy). The clean step is at #260:

| commit | qwen3-30b-a3b decode | verdict |
|---|---|---|
| `0b1606c31` (#257 plamo2) | 71.1 | good |
| `926f9a45a` (#259 granitemoehybrid) | 69.1 | good |
| `d80f78721` (#260 Apertus) | 44.4 | **bad (first)** |

Run-over-run (06-12 build `efa2c37bd` v0.1.4 -> v0.2.1), decode tok/s, all bf16-scale:
- qwen3-30b-a3b 70 -> 47, qwen3-moe 70 -> 47, qwen3-0.6b 290 -> 247, nemotron-h-30b 91 -> 51, nemotron-nas-30b 90 -> 51, plus gpt-oss-120b, solar-open-100b, minimax-m2, hunyuan-large, jamba, dots.llm1.

Controls (f16 scales) unchanged: mixtral-8x7b 54 -> 54, llama-3.1-8b 107 -> 109, qwen2.5-7b 110 -> 112, phi-4, starcoder2.

The split is exactly by scales dtype (safetensors headers): regressed models = BF16 scales; clean models = F16 scales. The one BF16 exception that did NOT regress is gemma-4-26b-a4b, because it also gained the GeGLU fused-MoE kernel (#281) which offsets the regression for that model.

It is **not** the fused-MoE kernel (`MLXCEL_FUSED_MOE` on vs off is identical for the affected models) and **not** an MLX change (pin a6ec712 across the whole range).

## Mechanism
[...]

PR #342, opened 2026-06-17
fix(sampling): correct f16/bf16 logprobs crash and corruption (#340)

### LINKED ISSUE #340: fix(sampling): top-k logprobs extraction crashes the server on f16/bf16 models (reads 2-byte logprobs as f32)
## Summary

`mlxcel-server` panics and aborts (`Abort trap: 6`) when a request to `/v1/completions` or `/v1/chat/completions` asks for top-k `logprobs` (for example `"logprobs": 5`) against a model whose logits / log-probs are f16 or bf16. After the #289 change ("keep quantized models fully bf16"), that is most quantized checkpoints, so this hits the common case.

Observed panic:

```
range end index 12 out of range for slice of length 10
src/lib/mlxcel-core/src/sampling.rs:722
```

This is pre-existing and unrelated to PR #339 (#325); see Provenance below.

## Reproduce

1. `./target/release/mlxcel-server -m models/qwen3-0.6b-4bit --parallel 8 --port 8131`
2. POST `/v1/completions` with `{"prompt":"...","max_tokens":32,"temperature":0,"logprobs":5}`

The worker thread panics and the process aborts. `qwen3-0.6b-4bit` carries bf16 logits under the current #289 policy, so any `top_k >= 1` triggers it.

## Root cause

`compute_logprobs()` in `src/lib/mlxcel-core/src/sampling.rs` (top-k branch, lines 714-725) extracts the top-k log-probabilities with `array_to_raw_bytes` plus a hardcoded 4-byte stride:

```rust
let idx_bytes = ffi::array_to_raw_bytes(&top_idx);   // int32 indices: 4 bytes/elem
let lp_bytes  = ffi::array_to_raw_bytes(&top_lp);    // f16/bf16 logprobs: 2 bytes/elem

let mut pairs: Vec<(i32, f32)> = (0..k_usize.min(idx_bytes.len() / 4))
    .filter_map(|i| {
        let tok_bytes:  [u8; 4] = idx_bytes[i * 4..(i + 1) * 4].try_into().ok()?;
        let lp_bytes4:  [u8; 4] = lp_bytes[i * 4..(i + 1) * 4].try_into().ok()?;  // line 722: panics
        Some((i32::from_ne_bytes(tok_bytes), f32::from_ne_bytes(lp_bytes4)))
    })
    .collect();
```

`array_to_raw_bytes` (`cpp/mlx_cxx_bridge.cpp:301`) dumps the array verbatim, `a.nbytes()` bytes, with no dtype conversion. `top_lp = take_along_axis(log_probs, top_idx)` inherits the log-softmax dtype, which inherits the model logit dtype. For f16/bf16 that is 2 bytes per element, so `lp_bytes.len() == 2 * k`. But:

- the loop bound is `idx_bytes.len() / 4`, driven by the int32 index array (genuinely 4 bytes per element), so the bound equals `k`;
- the `lp_bytes` read uses the same 4-byte stride against a 2-byte-per-element buffer.

For `logprobs: 5` and f16/bf16 logits: `idx_bytes` is 20 bytes (bound 5), `lp_bytes` is 10 bytes. At `i = 2` the slice `lp_bytes[8..12]` runs past length 10, which is exactly the reported panic. The no-overrun condition `4*(k-1) + 4 <= 2*k` reduces to `k <= 0`, so the slice overruns for any `top_k >= 1`: on a bf16 model, top-k logprobs always crashes. Even if the read did not overrun, reinterpreting two f16 elements as one f32 would yield garbage logprob values.

Note: the `tok_bytes` read is correct (indices really are 4 bytes). Only the logprob read is wrong.

## Selected-token logprob path: verified correct (does NOT share the bug)

Checked end-to-end per the request. The selected-token logprob (lines 680-684) does NOT use raw bytes:

```rust
let selected_lp_arr = ffi::take_along_axis(&log_probs, &idx, -1);
ffi::eval(&selected_lp_arr);
let selected_logprob = ffi::item_f32(&selected_lp_arr);
```

`item_f32` (`cpp/mlx_cxx_bridge.cpp:285`) calls MLX `array::item<float>()`, which is dtype-aware and converts the stored value to f32 regardless of f16/bf16/f32 storage. So the always-on selected-token logprob is correct on f16/bf16 models, and `logprobs` requests without top-k (`top_k == 0`) neither crash nor corrupt. The crash and the corruption are confined to the top-k alternatives path.

## Why existing tests miss it

The unit tests for this function (`compute_logprobs_returns_top_k_alternatives_sorted_descending` at `sampling.rs:1100`, `compute_logprobs_top_k_capped_at_vocab_size` at `sampling.rs:1116`) build logits with `ffi::from_slice_f32`, i.e. f32 (4 bytes), where the 4-byte stride is correct by accident. No test exercises an f16/bf16 logit array, so the dtype mismatch is invisible to CI.

## Provenance

`git blame -L 718,726 src/lib/mlxcel-core/src/sampling.rs` attributes the block to commit `892b24288a` (2026-04-02, "feat: implement logprobs support for chat completions and completions endpoints"). It predates and is unrelated to PR #339 (#325); it surfaced only during real-model validation of that PR.

## Suggested fix direction (not to implement in this issue)

At the logprobs boundary, either:

- cast `top_lp` to f32 with an explicit `astype` before `array_to_raw_bytes`, so the 4-byte stride is valid; or
- read the element size from the array dtype instead of hardcoding 4, and decode f16/bf16 accordingly.

Per project policy, quantized models keep bf16 scales/biases and (post-#289) bf16 logits, so the fix must handle bf16 logits at the logprobs boundary rather than promoting model weights. The selected-token path already does the right thing via `item<float>()`; the top-k path should reach the same f32 values.

## Acceptance Criteria

- [ ] Top-k logprobs return correct f32 values for f16, bf16, and f32 logit models.
- [ ] Selected-token logprob remains correct across f16/bf16/f32 (guard the current behavior so a refactor cannot regress it).
- [ ] `/v1/completions` and `/v1/chat/completions` with `logprobs: N` no longer crash on a bf16 model (for example `qwen3-0.6b-4bit`).
- [ ] A regression test covers the f16/bf16 logprobs path (both top-k and selected token), not only f32.

inureyes:
## Summary

Fixes a server crash (`Abort trap: 6`) and a silent garbage-value bug when a `/v1/completions` or `/v1/chat/completions` request asks for `logprobs` against a model whose logits are f16 or bf16. After #289 keeps quantized models fully bf16, that is the common case: any `top_k >= 1` crashed the worker, and even `logprobs` without top-k returned a corrupt selected-token logprob.

## Root cause

`compute_logprobs()` in `src/lib/mlxcel-core/src/sampling.rs` reads log-probabilities out of MLX arrays on the host through two byte-level accessors that do not convert dtype, while the arrays inherit the model logit dtype (f16/bf16, 2 bytes per element, for quantized checkpoints post-#289):

1. Top-k path (crash): the alternatives were extracted with `ffi::array_to_raw_bytes(&top_lp)` plus a hardcoded 4-byte stride. `array_to_raw_bytes` dumps the buffer verbatim, so for f16/bf16 the buffer is `2*k` bytes while the loop reads it with a 4-byte stride bounded by the int32 index array (`k` elements). For `logprobs: 5` on bf16, `lp_bytes` is 10 bytes and the read `lp_bytes[8..12]` overruns, producing the reported panic `range end index 12 out of range for slice of length 10`. Even without an overrun, reinterpreting two f16 elements as one f32 yields garbage.

2. Selected-token path (silent corruption): the value was read with `ffi::item_f32`, which calls MLX `array::item<float>()`. Contrary to the original issue's assumption, `item<float>()` is not dtype-aware: it returns `*data<float>()`, reinterpreting the element's raw bytes as f32 with no conversion. On bf16 it read a 2-byte element as a 4-byte f32 and returned garbage (for example `6.9e-41` instead of `-4.5028`). This path is always on whenever `logprobs` is enabled, so it was wrong on every bf16 model, not just the top-k case. The f16/bf16 regression tests added here caught it; reasoning alone (as in the issue) missed it.

## Fix

Promote to f32 at the host-extraction boundary in both places, casting only the small arrays so the full-vocab `log_probs` stays in its native dtype and the decode hot path is unaffected:

- Top-k: `ffi::astype(&top_lp, dtype::FLOAT32)` (then `eval`) before `array_to_raw_bytes`, so the 4-byte stride is valid (`k` elements of f32).
- Selected token: `ffi::astype(&selected_lp_arr, dtype::FLOAT32)` (then `eval`) before `item_f32`, so the single value converts correctly.

This mirrors the established in-repo pattern (`generate.rs` and `rope_proportional.rs` both `astype` to f32 before raw-byte extraction). The int32 token-id read is left unchanged because indices really are 4 bytes, and no model weights or dtype policy elsewhere are changed. `ffi::item_f32` itself is left as-is to keep the change at the logprobs boundary; its byte-reinterpretation behavior on non-f32 arrays is a latent footgun worth a separate follow-up.

## Tests

The existing tests only built f32 logits via `ffi::from_slice_f32`, where the 4-byte stride was correct by accident, so they never exercised either bug. The new tests build f16 and bf16 logit arrays from the same underlying f32 values and assert the results match an f32 reference run within a dtype-appropriate tolerance (bf16 ~0.1, f16 ~0.03, f32 exact), covering both paths:

- `compute_logprobs_top_k_bf16_no_panic_matches_f32` — unit-level reproduction of the server crash: `top_k = 5` on bf16 drives the identical top-k path the server hits. Asserts no panic, correct alternative count, descending sort, matching top-token id, and per-value agreement with the f32 reference.
- `compute_logprobs_top_k_f16_matches_f32` — same for f16 with a tighter tolerance.
- `compute_logprobs_top_k_f32_values_correct` — f32 reference path, exact values, guards the shared helper and sort.
- `compute_logprobs_selected_token_bf16_matches_f32` and `compute_logprobs_selected_token_f16_matches_f32` — `top_k = 0` on bf16/f16; these reproduced the selected-token corruption (pre-fix they returned `~6.9e-41` / `~7.0e-41`) and now assert the logprob matches the f32 reference.

## Test plan

- [x] `cargo test -p mlxcel-core --release --lib sampling` — 37 passed, 0 failed (run by the orchestrator with warm release artifacts).

Closes #340

### DIFF
--- a/src/lib/mlxcel-core/src/sampling.rs
+++ b/src/lib/mlxcel-core/src/sampling.rs
@@ -713,7 +713,15 @@ pub fn compute_logprobs(
 
         // Use raw bytes to extract i32 token IDs from top_idx.
         let idx_bytes = ffi::array_to_raw_bytes(&top_idx);
-        let lp_bytes = ffi::array_to_raw_bytes(&top_lp);
+        // `top_lp` inherits the model logit dtype, which is f16/bf16 for
+        // quantized models (post-#289). `array_to_raw_bytes` dumps the buffer
+        // verbatim with no dtype conversion, so a hardcoded 4-byte stride
+        // overruns a 2-byte-per-element buffer and reinterprets the bytes as
+        // garbage. Cast to f32 first so the stride is valid and the values are
+        // correct, mirroring the dtype-aware selected-token path (`item_f32`).
+        let top_lp_f32 = ffi::astype(&top_lp, dtype::FLOAT32);
+        ffi::eval(&top_lp_f32);
+        let lp_bytes = ffi::array_to_raw_bytes(&top_lp_f32);
 
         // Build (token_id, logprob) pairs for only the top-k partition.
         let mut pairs: Vec<(i32, f32)> = (0..k_usize.min(idx_bytes.len() / 4))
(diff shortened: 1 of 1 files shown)


inureyes:

### Intent
> Fix the f16/bf16 logprobs crash (`Abort trap: 6`) and silent garbage-value bug in `compute_logprobs()` for issue #340: read 2-byte f16/bf16 logprobs correctly instead of with a hardcoded 4-byte stride.

### Findings Addressed
- [x] New f16/bf16 regression tests failed the `cargo fmt --all -- --check` CI gate (ci.yml:86): chained `compute_logprobs(...).expect(...)` and `let x = logprobs_for_dtype(...)` calls were split across lines that rustfmt collapses (MEDIUM). Fixed by running `cargo fmt` on the test block (10 insertions, 20 deletions, test code only, no logic change).

### Remaining Items
- [ ] `ffi::item_f32` still byte-reinterprets `array::item<float>()` on non-f32 arrays (LOW) — the PR scopes this out as a documented follow-up footgun. Acceptable: every current caller now casts to f32 first, so it is latent, not live.
- [ ] No standalone test for size-1 vocab or `top_k`-clamped-on-bf16 combination (LOW) — pre-existing coverage gap, dtype-independent code path, not introduced by this PR.

### Verification
- [x] All stated requirements implemented (both top-k crash and selected-token corruption fixed; the second bug, which issue #340 assumed was safe, was found by the regression tests and fixed)
- [x] Integrated into project code flow (callers in scheduler.rs, speculative_burst.rs, gemma4_mtp_target.rs, dflash round_loop.rs consume corrected values unchanged; return shape stable)
- [x] Project conventions followed (mirrors the established `astype` to f32 before raw-byte extraction pattern in generate.rs and rope_proportional.rs)
- [x] Existing modules reused where applicable (`ffi::astype`, `dtype::FLOAT32`)
- [x] No unintended structural changes (fix confined to the logprobs host-extraction boundary; int32 token-id read, model-weight dtype policy, and the full-vocab `log_probs` dtype all unchanged)
- [x] Tests pass (orchestrator: 37 passed, 0 failed in release; the 2 selected-token tests failed pre-second-fix and pass after)

### Notes
- Dtype cast ordering verified correct in both paths: cast -> `eval` -> host read.
- Tolerances verified sound against an independent f32 reference: garbage gap (~4.5) is 45x the bf16 tolerance (0.1) and 150x the f16 tolerance (0.03), so they cannot pass on the known pre-fix garbage; worst-case legitimate bf16 rounding (~0.018) and f16 (~0.002) sit well inside, so they will not flake.
- Both bug claims confirmed against the C++ bridge: `item_f32` is `array::item<float>()` (line 285, no dtype conversion) and `array_to_raw_bytes` dumps `nbytes()` verbatim (line 301).

No CRITICAL or HIGH findings. The implementation is correct, complete, and properly integrated.

inureyes:

Reviewed the f16/bf16 logprobs dtype fix against the decode hot path, the FFI raw-bytes boundary, and edge-value handling. No CRITICAL or HIGH findings. No blockers.

### Security
The changed surface is internal sampling code: no untrusted input, no auth, no injection or serialization surface. The fix removes a denial-of-service vector, a normal `logprobs` request against a bf16 model aborted the worker (`range end index 12 out of range for slice of length 10`). Confirmed against the C++ bridge that the two accessors behave exactly as the PR claims: `item_f32` is `array::item<float>()` (raw reinterpret, no dtype conversion) and `array_to_raw_bytes` dumps `nbytes()` verbatim. Casting the extracted arrays to f32 before the host read is correct.

### Performance (decode hot path)
`compute_logprobs` runs per generated token when `logprobs` is enabled. Both casts operate only on the small extracted arrays, not the full vocab:
- Selected token: `astype` runs on `selected_lp_arr`, shape `[1, 1]` from `take_along_axis(log_probs, idx[1,1])`. 1 element.
- Top-k: `astype` runs on `top_lp`, shape `[1, k]` (`top_idx` is sliced to `k` at line 712 before the gather). k elements.

The full-vocab `log_softmax` + `eval` at lines 677-678 is pre-existing and unchanged. Neither cast widens the full vocab, so no full-vocabulary allocation or eval is added per token.

### FFI / memory safety at the raw-bytes boundary
After the top-k cast, `lp_bytes` is `4*k` bytes (k f32 elements). The loop is bounded by `idx_bytes.len() / 4 == k` and reads `lp_bytes[i*4..(i+1)*4]` for `i in 0..k`, so the last read is `[(k-1)*4 .. k*4]`, exactly in bounds. No remaining overrun. `eval` precedes the host read in both paths (line 690 before `item_f32`, line 730 before `array_to_raw_bytes`); the `array_to_raw_bytes` C++ also re-evals defensively.

### Edge values
- `top_k == 0`: selected-token-only path, top-k block skipped, `top_alternatives` empty. Covered by the two selected-token tests.
- `top_k` clamped to vocab: `k = min(top_k, vocab_size)` (line 697) feeds both `argpartition(k-1)` and the slice `stops`, so the cast array and the loop bound stay consistent at the clamp. The pre-existing `capped_at_vocab` test still guards this.
- Ties: alternatives are matched by token id, not position, and sorted descending with a total-order-safe comparator.
- `from_ne_bytes` / `item_f32` now receive genuine 4-byte f32 data after the cast.

### No new panics on the request path
The `filter_map` + `try_into` on fixed `[u8; 4]` slices cannot panic (out-of-range yields `None`). The fix is the established in-repo pattern (`generate.rs`, `rope_proportional.rs` both `astype` to f32 before raw-byte extraction). `cargo check -p mlxcel-core --release --lib` is clean.

### Note (non-blocking, LOW)
`ffi::item_f32` reinterpreting raw bytes on non-f32 arrays remains a latent footgun for any future caller that forgets the cast. The PR already calls this out as a separate follow-up; leaving the FFI as-is and fixing at the logprobs boundary is the right scope for this change.

inureyes:
## PR Finalization Complete

### Summary

- **Tests**: No new tests added. Coverage is already comprehensive: 5 regression tests covering f16/bf16/f32 on both the top-k path and the selected-token path, with sound tolerances (0.1 for bf16, 0.03 for f16). The two known LOW gaps (size-1 vocab, top_k clamped exactly on bf16) are pre-existing edge cases not worth adding here. All 37 tests pass.
- **Documentation**: No changes needed. This fix restores correct behavior with no API or behavior change. The project has no per-PR changelog convention (changelogs are written at release time). The one `logprobs` mention in docs is a feature-support table row that remains accurate.
- **Lint/Format**: `cargo fmt --all -- --check` is clean. One pre-existing clippy warning exists in `build.rs` (needless borrow on line 90), but `build.rs` was not touched by this PR and the warning is present on `main` as well. No warnings in the touched code (`sampling.rs`).

The branch is up to date with origin and the working tree is clean. Ready for merge.

Issue #325, opened 2026-06-17
perf(server): batched decode sampler fast path ([B, vocab] -> [B] fused sampling)


inureyes:
## Summary

Add a `[B, vocab] -> [B]` fused batched sampler fast path to the batched decode loop so that B active sequences are sampled with a single eval/sync point, instead of B per-row slice + sample + host scalar-extract round trips.

Source: idea.md "1. Batched Decode Sampler Fast Path" (P0, expected payoff medium to high for small/fast models and B>1 serving, risk medium).

## Current state

- `BatchScheduler::execute_batched_decode()` runs a single batched model forward for all active sequences, then slices one row at a time, calls `sample_token_optimized()`, evaluates the token array, extracts `item_i32`, and optionally computes per-row logprobs (`src/server/batch/scheduler.rs`, in the `execute_batched_decode` body near lines 4137 / 4213 / 4254).
- The standalone `batched_sample()` helper also loops by row and slices logits before calling the single-row sampler (`src/lib/mlxcel-core/src/sampling.rs:309`).
- `sample_token_optimized()` is already fused for a single row (`src/lib/mlxcel-core/src/sampling.rs:223`), so the missing piece is a true `[B, vocab] -> [B]` fused sampler.

## Proposal

- Add a scheduler fast path that triggers when ALL active rows share a compatible sampling config AND none of them needs a structured-output mask, a thinking-budget override, or a per-token logprobs payload.
- Dispatch a new batched fused sampler over the full `[B, vocab]` logits and return B token ids after one eval/sync point.
- Keep the existing per-row path as the correctness fallback for structured output, row-specific logprobs, token-bias observability, and mixed sampling configs.

## Scope

- New batched fused sampler in `src/lib/mlxcel-core/src/sampling.rs` (greedy first, then temperature/top-k/top-p in batched form where feasible).
- Fast-path gate plus dispatch in `execute_batched_decode()`.
- Preserve the per-row fallback exactly.

## Validation

- Server benchmark with B=2/4/8 on `qwen3-0.6b-4bit`, `qwen3-4b-4bit`, and a small Llama/Gemma dense model.
- Compare total decode tok/s, per-token latency, and greedy output parity at `temperature=0`.
- Include one fallback test with structured output or logprobs enabled to confirm behavior is unchanged.
[...]

Issue #203, opened 2026-06-10
fix: per-row position holes break B>1 batched MTP greedy parity after divergent accepts


inureyes:
## Problem / Background

Found while verifying issue #163 (PR #202) on an M1 Ultra with the Gemma 4 31B + MTP assistant pairing (REACHABLE_PAIRINGS[1], block_size 4, tests/speculative_parity.rs). Issue #163 hardened the ragged/batched MTP verify masks (NaN-safe padding rows, per-row stale-tail exclusion). Those fixes are correct and necessary, but real-model gates revealed a deeper, pre-existing structural gap that masking cannot fix.

In a B>1 batched MTP burst, every verify window is written at the shared physical cache offset and rotated (RoPE) at those physical positions for all rows. After a divergent (mixed-accept) round, `rollback_speculative_cache_explicit` trims all rows to the global max accept, so a row whose accept count was below the round max keeps a hole: its logical token sequence occupies positions [0, valid_end_r) plus the new window at [physical_offset, ...), with a gap in between that B=1 never has (B=1 trims its cache exactly, so its positions are contiguous).

Because both keys and queries rotate at physical positions, every subsequent attention score between a new token and the pre-hole history has its relative RoPE distance inflated by the cumulative hole size relative to the row's standalone B=1 run. The #163 stale-tail mask correctly removes the stale keys inside the hole from the key set, but no mask can repair the inflated relative distances. Greedy parity for a row is therefore only guaranteed while the row has accepted the round max in every round so far (the lockstep prefix); after the first sub-max round the row's logits diverge from B=1 structurally and near-tie argmaxes flip.

## Evidence

M1 Ultra, gemma-4-31b-it-4bit + gemma-4-31B-it-assistant-bf16, greedy, max_tokens 24, the four equal-length prompts from `greedy_parity_mtp_gemma4_batched_b4_matches_b1` and the four variable-length prompts from the new ragged gate:

1. The strongest single confirmation: in the ragged run, row 3 was the round-max row for many consecutive rounds (accept_lens [3, 3, 1, 1, 3, ...]) and stayed byte-identical to its B=1 reference for 17 tokens, then flipped within a round or two of its first sub-max round. Rows whose accepts dropped below the max early (accept_lens all zeros for row 0) flipped within one or two rounds of the first divergence.
[...]

PR #346, opened 2026-06-18
perf(server): split batched prefill into compatible cohorts

### LINKED ISSUE #332: perf(server): split batched prefill into compatible cohorts instead of all-or-nothing fallback
## Summary

When a collected prefill batch contains one incompatible request, the scheduler currently falls the WHOLE window back to sequential prefill. Instead, split the window into compatible cohorts so eligible concurrent work still runs batched.

Source: idea.md "4. Batched Prefill Cohort Splitting" (P1, expected payoff medium, risk medium).

## Current state

- Batched prefill is available only when the model opts in via `supports_batched_prefill()`; the default is false (`src/lib/mlxcel-core/src/generate.rs:467`).
- Qwen3 opts into batched prefill, maskless padded prefill, and paged decode (`src/models/qwen3.rs`, around line 821).
- In the scheduler, one incompatible request forces the whole collected batch back to sequential prefill: unsupported model, any VLM embeddings, any adopted prompt-cache prefix, or incompatible prompt lengths (`src/server/batch/scheduler.rs`, around lines 3092 / 3110 / 3129 / 3146).

## Proposal

- Instead of falling back the whole prefill window, split it into compatible cohorts:
  - cold text requests eligible for padded batched prefill,
  - adopted-prefix requests that need sequential offset-aware prefill,
  - VLM or custom-embedding requests,
  - equal-length-only models.
- Preserve request priority order at cohort boundaries so latency fairness does not regress.
- Expand batched-prefill support to additional dense families ONLY after cohort splitting is measured; otherwise support expansion is hidden by fallback behavior.

## Validation

- Synthetic server queue with mixed cold prompts and prompt-cache hits.
- Measure prefill tok/s, time-to-first-token, and head-of-line blocking.
- Correctness tests for cache-offset handling when adopted-prefix rows are split away from cold rows.
- Suggested first experiment (idea.md): add queue-shape tracing for batched-prefill fallback reasons first. If most fallback windows mix cold and adopted-prefix requests, cohort splitting is worth implementing.

## Acceptance criteria

- [x] Mixed prefill windows are partitioned into cohorts; the cold-text cohort runs padded batched prefill even when other requests in the window are incompatible.
- [x] Request priority order is preserved at cohort boundaries (no fairness regression).
- [x] Cache-offset correctness verified when adopted-prefix rows are separated from cold rows.
- [ ] Prefill tok/s, TTFT, and head-of-line blocking measured on a mixed synthetic queue vs the current all-or-nothing fallback.
- [x] INTEGRATION into the live scheduler prefill path.

## Risk and priority

idea.md priority P1; risk medium. Batched prefill is a high-leverage server feature; the current all-or-nothing fallback can waste eligible concurrent work.

## References

- idea.md "4. Batched Prefill Cohort Splitting" and "Suggested First Experiments" item 4.
- Code: `src/lib/mlxcel-core/src/generate.rs` (`supports_batched_prefill`), `src/models/qwen3.rs`, `src/server/batch/scheduler.rs` (prefill fallback gates).

inureyes:
## Summary

Closes #332. When a collected prefill window held one incompatible request (unsupported model, any VLM embeddings, any adopted prompt-cache prefix, or length-incompatible rows on equal-length-only models), the batch scheduler fell the whole window back to sequential prefill, wasting eligible concurrent work. This splits the window into compatible cohorts so cold text rows still run batched while incompatible rows take their existing offset-aware handling.

## What changed

- New `src/server/batch/prefill_cohort.rs`: pure `plan_prefill_cohorts` planner that partitions a priority-ordered window into `BatchedCold` and `Sequential` cohorts, plus its types and unit tests. Order-preserving, offset-isolating, and contiguity-respecting by construction.
- `src/server/batch/scheduler.rs`: `execute_batched_prefill` now drains the window, classifies each row (cold = no VLM/custom embeddings AND zero adopted-prefix offset), plans cohorts, and dispatches them in window order. The padded batched forward pass moved verbatim into a new `run_padded_batched_prefill` helper. Cold cohorts run batched even when the window contains incompatible rows; adopted-prefix and embedding-bearing rows take `execute_full_prefill`.
- `src/server/batch/mod.rs`: declares the `prefill_cohort` module.
- `src/server/batch/scheduler_cohort_parity_tests.rs`: new `#[ignore]` real-model qwen3 parity test.
- `src/server/batch/scheduler_prompt_cache_tests.rs`: refreshed a now-stale comment about the old whole-window fallback.

## Correctness (cache offsets)

A row is classified cold only when it has no embeddings AND `prefill_start_offset == 0`, which is exactly the zero-offset precondition the padded batched path assumes. The planner guarantees a `BatchedCold` cohort contains only cold rows, so an adopted prefix can never be folded into a batch and have its KV resumed at the wrong offset; it always takes the offset-aware `execute_full_prefill`. The refactor also routes the single-request and equal-length-only paths through `begin_prefill`, tightening two latent paths that previously reached `execute_full_prefill` in the `Queued` state and would have failed the `Queued` to `Decoding` transition.

## Fairness

Cohorts dispatch in window order and the queue dequeues in priority order, so concatenating cohort members reproduces window order exactly (asserted by a unit test). Cold rows batch only where contiguous, so the split never hoists a lower-priority request ahead of a higher-priority one.

## Scope

Batched-prefill support is unchanged (qwen3 / llama3 / qwen3.5). No new dense families were added; that expansion is deferred until cohort splitting is measured, per the issue.

## Test plan

- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate prefill_cohort::` (11 pure planner tests pass: order preservation, offset isolation, all-cold no-regression, all-incompatible, equal-length-only, contiguity).
- [x] `cargo check -p mlxcel --lib --tests` and `cargo clippy -p mlxcel --lib --tests -- -D warnings` clean.
- [ ] Orchestrator: full release build + `scheduler_cohort_parity_tests` against real qwen3 (`--ignored`), plus a mixed cold + prompt-cache-hit synthetic queue measuring prefill tok/s, TTFT, and head-of-line blocking vs the all-or-nothing fallback.

### DIFF
--- a/src/server/batch/mod.rs
+++ b/src/server/batch/mod.rs
@@ -28,6 +28,7 @@
 
 mod active;
 pub mod observability;
+mod prefill_cohort;
 mod queue;
 pub(crate) mod scheduler;
 #[cfg(test)]
--- a/src/server/batch/prefill_cohort.rs
+++ b/src/server/batch/prefill_cohort.rs
@@ -1,0 +1,426 @@
+// Copyright 2025-2026 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Cohort planning for batched prefill (#332).
+//!
+//! A collected prefill window can mix requests that the padded batched-prefill
+//! path supports with requests that it does not. Historically one incompatible
+//! request forced the *whole* window back to sequential prefill, wasting the
+//! eligible concurrent work. This module turns that all-or-nothing decision
+//! into a cohort split: the cold text requests that can share a single padded
+//! forward pass are grouped into a batched cohort, while incompatible requests
+//! (adopted prompt-cache prefixes, VLM / custom embeddings, length-incompatible
+//! rows on equal-length-only models) are routed to the offset-aware
+//! single-sequence path.
+//!
+//! The planner is a pure function over per-row classifications so its
+//! correctness invariants can be pinned by unit tests without a real model:
+//!
+//! - **Order preserving.** Concatenating the members of the returned cohorts in
+//!   order reproduces the input window order exactly. The window is drained in
+//!   priority order (high lane, then normal, then low; FIFO within a lane), so
+//!   preserving window order is exactly what keeps request priority / FIFO
+//!   fairness intact across cohort boundaries.
+//! - **Offset isolation.** A [`PrefillCohortKind::BatchedCold`] cohort contains
+//!   only rows flagged `is_cold`, i.e. rows with a zero KV-history offset and no
+//!   custom embeddings. The padded batched path assumes a zero cache offset for
+//!   every row, so this guarantees it never receives an adopted-prefix row whose
+//!   KV must resume at a non-zero offset.
+//! - **Contiguity.** Cold rows batch only where they are *contiguous* in the
+//!   window. A cold row separated from other cold rows by an incompatible row is
+//!   prefilled sequentially rather than reordered into a distant batch, so the
+//!   split never hoists a lower-priority request ahead of a higher-priority one.
+
+/// How a contiguous group of prefill-window rows is executed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum PrefillCohortKind {
+    /// Two or more cold text rows (zero KV-history offset, no custom
+    /// embeddings) that share a single padded batched-prefill forward pass.
+    BatchedCold,
+    /// Rows prefilled one at a time on the offset-aware single-sequence path:
+    /// adopted prompt-cache prefixes, VLM / custom-embedding rows, isolated
+    /// cold rows, and (on equal-length-only models) length-incompatible cold
+    /// rows.
+    Sequential,
+}
+
+/// A planned cohort: its execution kind and the window indices of its member
+/// rows, in priority (window) order.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct PrefillCohort {
+    pub kind: PrefillCohortKind,
+    pub members: Vec<usize>,
+}
+
+/// Per-row inputs to [`plan_prefill_cohorts`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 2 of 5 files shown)


inureyes:
## PR Finalization

**Tests:** 11 unit tests in `prefill_cohort::tests` all pass (empty window, all-cold, all-incompatible, mixed split, adopted-before-cold priority, isolated-cold fallback, two separate cold runs, single-row sequential, no-batched-support model, equal-length-only model, padded mixed lengths). The `#[ignore]` real-model parity test in `scheduler_cohort_parity_tests.rs` is wired and ready for the orchestrator. No coverage gap found.

**Docs:** `docs/CONTINUOUS_BATCHING.md` does not describe the batched-prefill all-or-nothing fallback behavior this PR replaces, so no doc update is needed and no new section was invented. No Korean mirror exists.

**Lint/format:** `cargo fmt --check` clean; `cargo clippy -p mlxcel --lib --tests -- -D warnings` clean. No fixes required.

No changes committed. Ready for merge.

inureyes:
## Real-model validation (orchestrator)

Cohort parity test on qwen3-0.6b-4bit. The originally-authored assertion (cohort-batched cold row == single-prefill reference) failed, but that was the wrong invariant: a cohort-batched row runs at B=2, and qwen3 padded batched prefill is not bitwise-identical to single-sequence (B=1) prefill on Metal (the documented #203 / #325 / #326 jitter class, where an early near-tie greedy token flips and cascades). The correct #332 invariant is that a cohort-split cold row decodes identically to the same row in an all-cold batched window of the same composition. Empirically confirmed:

- mixed-window cohort B (B=2, split from [A, B, adopted-C]) == all-cold batched B (B=2, window [A, B]): byte-identical. The cohort split is correct, cache offsets intact.
- single-prefill B (B=1) differs by one early near-tie token, exactly the expected jitter.

Fixed in `095f28489`: the test now asserts the batched-vs-batched invariant (`mixed_window_cold_cohort_matches_all_cold_batched_qwen3`), keeps the single-prefill values as a printed diagnostic only (never asserted), and documents at the cohort dispatch that a cold row which previously fell back to sequential prefill now runs batched (the intended effect of cohort splitting, not a regression). 11 planner unit tests pass; pr-reviewer and pr-security-checker approved (0 CRITICAL/HIGH). Cache-offset correctness holds by construction (a BatchedCold cohort contains only zero-offset, no-embedding rows).

Throughput acceptance item: the cold cohort now runs one padded batched forward instead of N sequential forwards, which is the established batched-prefill benefit now applying to mixed windows; a rigorous mixed-queue A/B against the all-or-nothing baseline is deferred (it needs a separate baseline binary), but the mechanism is proven correct and proven to batch the cold cohort in a mixed window.

Follow-up (separate issue, out of #332 scope): batched-prefill seed determinism. pr-reviewer noted that a seeded, non-greedy cold row in a batched cohort is governed by the last row's seed (last-seed-wins), which is a pre-existing property of the batched path that cohort splitting slightly widens. Output stays valid; only seed-reproducibility for seed+temperature>0 requests is affected.

PR #348, opened 2026-06-18
perf(server): adaptive B=1 MTP enable/decline policy from profiling

inureyes:
## Summary

Replace the static per-hardware B=1 MTP speculative-decode gate with an adaptive policy that profiles the first few B=1 bursts of each (target, drafter, hardware) pairing and settles to a data-driven enable/decline verdict, overriding the static gate where the measured profile is clearly favorable or unfavorable and deferring to it otherwise. MTP stays mathematically exact: the policy only decides when to run MTP, never what it emits, so temperature-0 output is byte-identical to classic decode.

## What changed

- `src/server/batch/mtp_policy.rs` (new): the policy. Pure decision core (`estimate_speedup`, `classify_speedup`, `resolve_verdict`), a round-weighted `ProfileAccumulator`, the `MtpPolicy` state machine (Forced / Profiling / Settled), the `PolicyKey` + coarse `PolicyHint` + per-pairing `PolicyStore` persistence, and the env parsing.
- `src/server/batch/mtp_policy_tests.rs` (new): 25 unit tests covering the decision logic, env overrides in both directions, the coarse-hint-only persistence shape, and the decision-only exactness contract.
- `src/lib/mlxcel-core/src/speculative/mtp/generator.rs`: additive `MtpGenerator::last_acceptance()` returning a new public `MtpAcceptanceSummary` (rounds, proposed/accepted draft tokens, draft/verify ms). Purely additive: it does not change the emitted tokens, logprobs, or stats, so MTP exactness is untouched.
- `src/server/batch/speculative_burst.rs`: thread the coarse profile out of the MTP B=1 burst (`DriveMtpOutput.acceptance` -> `BurstSuccess.profile` -> `BurstFinalized.mtp_profile`); DFlash and zero-round runs carry `None`.
- `src/server/batch/scheduler.rs`: hold an optional `MtpPolicy`, build it in `with_mtp_policy`, route the B=1 gate through `mtp_b1_should_run`, and feed the policy each completed B=1 burst's profile.
- `src/server/model_worker.rs`: chain `with_mtp_policy` after `with_speculative_dispatch`, keyed on the served model basename.
- `docs/environment-variables.md`, `docs/benchmarks.md`: document `MLXCEL_MTP_ADAPTIVE`, the persisted-hint location/format, and the manual-override semantics.

## How it decides

[...]

Issue #333, opened 2026-06-17
perf(inference): adaptive MTP enable/decline policy from per-pair profiling


inureyes:
## Summary

Replace the static environment gates for MTP speculative decode with an adaptive policy that profiles the first few requests for a target/drafter pair and auto-enables or declines MTP per pair and per hardware class.

Source: idea.md "7. Adaptive MTP Policy" (P2, expected payoff pair-dependent, risk medium).

## Current state

- The benchmark doc shows Gemma 4 12B + 4-bit assistant around 1.87x faster with MTP, while the 31B + bf16 assistant is pairing-dependent (`docs/benchmarks.md`, around lines 82 / 99).
- The scheduler already gates B=1 and B>1 MTP because some cases regress: B=1 can regress on pre-M5 hardware, and B>1 31B MTP is not consistently faster than classic batched decode (`src/server/batch/scheduler.rs`, around lines 2850 / 2872).

## Proposal

- Add an adaptive policy that profiles the first few requests for a target/drafter pair and records acceptance length, verify latency, drafter latency, batch size, and prompt-shape compatibility.
- Auto-enable or decline MTP per pair and per hardware class instead of relying only on static environment gates.
- Persist only coarse local hints, NOT prompt data.

## Validation

- `speculative_bench` baseline vs MTP for Gemma 4 12B and 31B pairings.
- Server benchmark with same-length and ragged prompt windows.
- Confirm `temperature=0` byte-identical output against classic decode.

## Acceptance criteria

- [x] Per-pair and per-hardware profiling records acceptance length, verify latency, drafter latency, batch size, and prompt-shape compatibility over the first few requests. (`MtpBurstProfile` + `ProfileAccumulator` in `src/server/batch/mtp_policy.rs`; profiling window `PROFILE_SAMPLE_TARGET`.)
- [x] MTP is auto-enabled or declined from the measured profile, overriding the static gates where the profile is favorable or unfavorable. (`estimate_speedup` + `classify_speedup` + `resolve_verdict`; ambiguous defers to the static per-hardware default.)
- [x] Only coarse local hints persisted; no prompt data stored. (`PolicyHint` carries verdict + coarse acceptance rate + sample count only; `${MLXCEL_CACHE_DIR:-$HOME/.cache/mlxcel}/mtp-policy/<key-hash>.json`; unit test asserts the field set.)
[...]

PR #351, opened 2026-06-18
fix(models): suppress gemma4_unified multimodal placeholder tokens

### LINKED ISSUE #350: fix(models): gemma4_unified leaks multimodal placeholder tokens (audio/image/video) into text generation
## Problem / Background

During text-only generation, `gemma-4-12b-it-4bit` (model_type `gemma4_unified`) occasionally emits a multimodal placeholder token in the middle of otherwise-coherent text.

Reproduction with `mlxcel generate` (chat template applied):

```
mlxcel generate -m models/gemma-4-12b-it-4bit -p "What is a hash table? Answer in one sentence." -n 60 -t 0
```

produced output containing a stray audio placeholder token:

```
...compute an<audio|>ex index, allowing for efficient data retrieval...
```

The leak is occasional and context-dependent. The "hash table" prompt leaks the audio placeholder, while "list three primary colors" and "write one sentence about the moon" generate cleanly.

Via `/v1/chat/completions`, the server strips special tokens from the response `content`, so a leak-heavy generation (the hash-table prompt) returns an EMPTY `content` with `finish_reason=length` despite ~60 generated tokens (per `usage.completion_tokens`).

## Root Cause

`gemma4_unified` is a multimodal model (audio + image + video). The placeholder token ids in `models/gemma-4-12b-it-4bit/config.json` are input-alignment placeholders that must never appear in generated text output. The leaked one is `audio_token_id` (258881):

| config field | token id |
|--------------|----------|
| `audio_token_id` | 258881 |
| `image_token_id` | 258880 |
| `boa_token_id`   | 256000 |
| `boi_token_id`   | 255999 |
| `eoa_token_index`| 258883 |
| `eoi_token_id`   | 258882 |
| `video_token_id` | 258884 |

mlxcel does not suppress these in the output logits during generation, so at a near-tie greedy step a placeholder can become the argmax and leak into the text stream.

## Proposed Solution

`src/lib/mlxcel-core/src/sampling.rs` already supports permanently suppressing a token by inserting `f32::NEG_INFINITY` into the per-step bias map (doc comment near line 93: "Use `f32::NEG_INFINITY` to permanently suppress a token"; an existing insert at line 1625 backed by a test asserting the suppressed-token probability is 0.0; language-bias suppression counters near lines 47-100).

Currently no `gemma4_unified` placeholder token is registered for OUTPUT suppression (model implementation in `src/models/gemma4.rs`, detected via `src/models/detection.rs:125`). The only model that references an image token at all today is `qwen3_vl` (`src/models/qwen3_vl.rs`), and that reference is for input embedding alignment, not output masking.

Fix: register the gemma4_unified multimodal placeholder token ids so they are masked to `-inf` at every decode step. Prefer deriving the suppressed set generically from the model config `*_token_id` / `*_token_index` fields rather than hardcoding only gemma4.

## Scope note (companion audit)

A broader audit of whether other multimodal/VLM families (gemma3, qwen2.5-vl, and similar) suppress their own placeholder tokens in the OUTPUT logits is a recommended companion. The fix should be generic where practical (derive the suppressed set from the model config `*_token_id` fields) rather than family-specific.

## Acceptance Criteria

- [ ] `gemma4_unified` never emits audio/image/video/boi/eoi/boa/eoa placeholder tokens as generated text.
- [ ] The hash-table prompt yields coherent, non-empty output via BOTH `mlxcel generate` and `/v1/chat/completions`.
- [ ] No regression on already-coherent prompts (for example "list three primary colors", "write one sentence about the moon").
- [x] A regression test pins that the placeholder token ids are suppressed (`NEG_INFINITY`) in the generation logits for `gemma4_unified`.

## Provenance

Pre-existing behavior, not introduced by the #325-#333 performance chain. Surfaced during #333 (PR #348) adaptive MTP validation using the gemma-4-12b-it-4bit plus assistant pairing.

inureyes:
## Summary

`gemma4_unified` (e.g. `gemma-4-12b-it-4bit`) reserves multimodal placeholder ids (audio/image/video span markers) for INPUT alignment only, but mlxcel did not mask them in the output logits. At a near-tie greedy decode step one could become the argmax and leak into generated text (a stray `<audio|>` mid-sentence on the "hash table" prompt). Through `/v1/chat/completions` the server strips special tokens, so a leak-heavy generation returned EMPTY `content` with `finish_reason=length`. This masks those ids to `-inf` at every decode step via the existing per-step token-bias mechanism, with a generic helper so other multimodal families can opt in later.

## What changed

- `LanguageModel::output_suppressed_token_ids()` (new trait method, default empty): the per-model "never generate these" set, mirroring `eos_token_ids()`. Non-multimodal models inherit the empty default and pay zero cost.
- `TokenBiasMap::suppress_tokens(ids)`: force-masks each id to `f32::NEG_INFINITY` (suppression always wins over any prior finite bias); empty slice is a no-op that preserves the bit-exact baseline.
- `models::MultimodalPlaceholderTokens`: a family-agnostic collector that, given the config `*_token_id` / `*_token_index` fields, returns the deduplicated, non-negative id set. `Gemma4UnifiedConfig::placeholder_tokens()` builds it from the parsed config, and the `Gemma4UnifiedModel` `LanguageModel` impl returns `suppressed_ids()` from its stored ids. `LoadedModel` delegates the new trait method.
- Wiring (scoped to gemma4_unified via the trait override): CLI `run_generation_mode` merges the set into the token-bias map before any generator is built; server `BatchScheduler` resolves it once in `with_config` and force-applies it on every sequence's sampling config in `enqueue_request`.

## Where suppression is applied (both decode paths)

- CLI `generate`: `run_generation_mode` merges `model.output_suppressed_token_ids()` into the token-bias map at the top, so it reaches the standard, VLM-embedding, and speculative sub-paths (all consume that map via `with_token_bias`).
- Server decode: `enqueue_request` applies the resolved set to each sequence's `SamplingConfig.token_bias` after the lang-bias merge and unconditionally, so it covers classic decode and the MTP burst (both sample through `sample_token_optimized`, which applies `config.token_bias`). Suppression is added last, so a per-request bias can never displace it.

## Derivation and safety

The suppressed set is derived from config (`audio_token_id`, `image_token_id`, `video_token_id`, `boa_token_id`, `boi_token_id`, `eoa_token_id`/`eoa_token_index`, `eoi_token_id`), never hardcoded as raw integers in the wiring. The real EOS ids (`eos_token_id` `[1, 106, 50]`) are not part of the placeholder struct, so end-of-sequence detection and normal text generation are unaffected. Non-multimodal models return an empty set, so `apply_token_bias` keeps its zero-overhead short-circuit.

## Test plan

- [x] `cargo test --release -p mlxcel-core sampling::tests::suppress` (2 tests: `-inf` + zero softmax probability, empty no-op)
- [x] `cargo test --release -p mlxcel --lib -- multimodal_placeholders gemma4_unified_config` (14 tests incl. `output_suppression_masks_placeholders_but_not_eos`: placeholder ids masked to `-inf`, EOS `[1,106,50]` untouched)
- [x] `cargo clippy -p mlxcel-core --lib --tests -- -D warnings` and `cargo clippy -p mlxcel --lib --tests -- -D warnings` (clean)
- [ ] Real-model validation (orchestrator): "hash table" prompt coherent and non-empty via `mlxcel generate -t 0` AND `/v1/chat/completions`; no regression on clean prompts ("list three primary colors", "write one sentence about the moon").

Closes #350


inureyes:

### Intent
> Suppress `gemma4_unified` multimodal placeholder ids (audio/image/video/boa/boi/eoa/eoi) in the output logits so they can never leak into generated text (#350).

### Findings Addressed
- No code changes were required. The implementation is correct and complete for the issue's stated scope; no CRITICAL or HIGH findings.

### Remaining Items (non-blocking, out of #350 scope)
- [ ] (MEDIUM) CLI pipeline-parallel path bypasses suppression. `pipeline_requested` (`src/commands/generate.rs:1280`) routes to `generate_pipeline_text` instead of `run_generation_mode`, so the placeholder set is never merged there. The path operates on raw layers with no `LoadedModel` handle, so wiring it needs a separate config read. Single-node `generate` and the server are the issue's scope and are covered.
- [ ] (MEDIUM) Distributed-handoff sequence builders do not apply the suppressed set: `prefill_text_request_for_handoff` (`src/server/batch/scheduler.rs:604`) and `ingest_handoff_as_active` (`:699`) construct a `SequenceInfo` without `suppress_tokens(&self.model_output_suppressed)`. Both are `#[allow(dead_code)]` today (not in the live request flow), so there is no current leak, but they should mirror the `enqueue_request` block when the disaggregated path (#126) goes live.
- [ ] (LOW) `Gemma4UnifiedConfig::placeholder_tokens()` is only exercised by tests; the runtime path rebuilds the set from the model's stored ids in `Gemma4UnifiedModel::output_suppressed_token_ids()`, so the two 7-field lists can drift over time. Consider having the model reuse the config builder (or store a `MultimodalPlaceholderTokens`).

### Verification
- [x] All stated requirements implemented: suppressed set is exactly the 7 placeholder ids; real EOS `[1,106,50]` is excluded by construction (no EOS field in the struct); negative ids dropped; result deduped/sorted.
- [x] No placeholder/mock code remaining.
- [x] Integrated into project code flow: CLI standard + VLM-embedding + speculative sub-paths (via `with_token_bias`, with `sampling_config.token_bias` empty so the cached map wins); server classic decode + MTP burst (both sample `seq.sampling` through `sample_token_optimized`). Both production `BatchScheduler` construction sites route through `with_config`.
- [x] Suppression always wins: `suppress_tokens` force-inserts `NEG_INFINITY`; applied last and unconditionally in `enqueue_request` (after the lang-bias merge) and overwrites any prior finite bias in the shared map.
- [x] Zero cost when empty: non-multimodal models inherit the empty default; `apply_token_bias` and the enqueue/with_config paths short-circuit on an empty set. EOS still terminates generation for `gemma4_unified`.
- [x] Generic but scoped: `MultimodalPlaceholderTokens` is config-driven and family-agnostic; only `gemma4_unified` overrides the trait.
- [x] Conventions: `_tests.rs` split, trait method mirrors `eos_token_ids()`, no AI attribution, no em dashes, all refs are same-repo.
- [x] Tests pass: 16 targeted tests green (2 in `mlxcel-core sampling::`, 14 in `mlxcel --lib`, incl. `output_suppression_masks_placeholders_but_not_eos`).

Real-model validation (hash-table prompt non-empty/coherent via CLI and `/v1/chat/completions`, plus clean-prompt no-regression) is left to the orchestrator's full release build.

inureyes:
## Implementation Review Update (supersedes the Verification section of my earlier comment)

A deeper pass plus the Codex reviewer found two CRITICAL leak paths that my first comment incorrectly marked as covered. Both are now fixed on this branch.

### CRITICAL findings (fixed)

1. **MTP speculative-burst verify bypassed suppression.** `Gemma4MtpTargetAdapter::verify_forward` (and the deferred `argmax_from_hidden_positions` path and the batched `verify_forward_batched`) derived target tokens with a RAW argmax that never applied the suppression bias. Only the first bonus token (sampled via `sample_token_optimized`) was masked, so every later token on the server MTP path could still emit a placeholder. This is the exact scenario #350 was surfaced in (#333 / PR #348 MTP validation with gemma-4-12b-it). Fix: apply the suppression bias (already carried in the `sampler` forwarded from `enqueue_request` through `MtpGenerator`) to the verify logits before the per-position argmax and before logprob extraction, so the emitted token and its logprob stay consistent. An empty bias short-circuits, so non-multimodal models stay bit-exact.

2. **Interactive chat bypassed suppression.** `mlxcel generate` with no `-p` (and `mlxcel run`) routes through `run_generate` to `run_chat`, which built its `CxxGenerator` without merging the model's suppressed set, so chat replies could leak. Fix: `run_chat` now attaches the bias via `with_token_bias`, mirroring `run_generation_mode`.

`apply_token_bias` was promoted from `pub(crate)` to `pub` so the MTP target adapter (binary crate) can reuse the existing masking helper. Three unit tests added: B=1 verify leak re-route, empty-bias no-op, batched per-row re-route.

### Verification

- `cargo check` lib + bins (release), `cargo test` MTP (`gemma4_mtp`, 35 passed) and core sampling (46 passed), clippy on both crates: all pass.
- Committed on this branch as `ea0a4aa7b` (held local, pending push by the orchestrator after the gate below).
- STILL OWED (orchestrator): real-model validation, specifically an MTP-enabled run of gemma-4-12b-it plus the hash-table prompt via CLI and `/v1/chat/completions` (no leak, non-empty, coherent) and no regression on clean prompts. The MTP fix only changes the argmax at a step where a placeholder would otherwise have won, so legitimate tokens are unaffected by construction, but a greedy-parity spot check is still worthwhile.

### Non-blocking (unchanged)

- MEDIUM: the CLI pipeline-parallel path and the `#[allow(dead_code)]` distributed-handoff sequence builders (`prefill_text_request_for_handoff`, `ingest_handoff_as_active`) still do not apply suppression. Both are outside #350's single-node scope; the handoff builders should mirror `enqueue_request` when the disaggregated path (#126) goes live.
- LOW: `Gemma4UnifiedConfig::placeholder_tokens()` is exercised only by tests; the runtime path rebuilds the set from the model's stored ids, so the two field lists can drift.

inureyes:
## PR Finalization

### Tests

No gaps found. The existing 16 suppression tests cover the suppressed-set derivation, EOS exclusion, NEG_INFINITY masking, and the 3 MTP verify-path cases. No new tests were needed.

### Documentation

No doc updates warranted. This is a correctness fix with no user-facing surface; there is no existing known-issues or model-support section that applies.

### Lint / Format

- `cargo fmt --check`: one long `assert_eq!` in `gemma4_mtp_target_tests.rs` exceeded the line limit. Reformatted to multi-line. Now clean.
- `cargo clippy -p mlxcel-core --lib --tests -- -D warnings`: clean.
- `cargo clippy -p mlxcel --lib --tests -- -D warnings`: clean.

### Latent-gap comment

Added a comment to `argmax_logits_to_array` in `src/lib/mlxcel-core/src/drafter/dflash/round_loop.rs` (the DFlash speculative verify round-loop). The function performs a raw argmax with no token bias. That is safe today because only `gemma4_unified` has a non-empty `output_suppressed_token_ids` set, and it runs on the MTP path, not DFlash. The comment notes that if a future model adopts DFlash verify with a non-empty suppressed-id set, this argmax must be preceded by `apply_token_bias` to prevent reintroducing the #350 leak.

Commit: `67336e0bc`

inureyes:
## Real-model validation (orchestrator)

Placeholder suppression validated on M1 Ultra:

- CLI `mlxcel generate` (run_generation_mode path), 4 prompts including the "What is a hash table?" prompt that previously leaked `<audio|>` mid-sentence: all coherent, NO placeholder leak.
- Server `/v1/chat/completions` with MTP forced on (the verify path the reviewer fixed), one-sentence hash-table prompt, adequate token budget: non-empty, coherent, no leak.

Root-cause note on the earlier "empty content" symptom (it is NOT a placeholder issue): gemma-4-12b-it is a thinking model by design. The server detects its `<|channel>thought`/`<channel|>` markers and defaults `enable_thinking=true` (intentional, upstream PR #1114). With thinking on, the model spends most of a small `max_tokens` budget inside the thinking block, so `max_tokens=50-80` truncates mid-thinking and the answer `content` is empty (`finish_reason=length`). With an adequate budget the answer is coherent: the one-sentence prompt finished in 275 tokens (`finish=stop`). The earlier #333 / initial #350 "empty" results were under-budgeted tests of a thinking model, not a leak.

pr-reviewer fixed two CRITICAL leak paths it found (MTP speculative verify and interactive chat), pr-security-checker confirmed MTP exactness consistency (verify applies the same suppression as classic), and 19 tests pass (16 suppression + 3 MTP verify). The #350 placeholder leak is fixed and validated.

Minor observations (out of #350 scope, not regressions): gemma-4-12b-it thinking is verbose (a three-sentence request did not finish within 800 tokens), and `reasoning_content` is not populated when thinking occurs. Neither affects answer correctness; worth a separate look if desired.

PR #356, opened 2026-06-18
fix(server): reseed RNG per row at batched-prefill first-token sample

### LINKED ISSUE #347: fix(server): batched-prefill seed determinism (last-seed-wins for seeded sampling in a batched cohort)
## Problem / Background

In the batched-prefill path, a per-request RNG seed is applied to the **global** MLX RNG, and the first-token sampler then consumes that global RNG with no per-call seed. Because a batched cohort runs every row's `begin_prefill` (seeding) before any row's `finish_prefill` (sampling), the **last** seeded row's seed governs first-token sampling for the entire cohort ("last-seed-wins").

Verified call chain (this repo):

- `seed_rng_if_needed` (`src/lib/mlxcel-core/src/generation_policy.rs:26-30`) seeds the global MLX RNG via `ffi::random_seed(seed)`, and only when `sampling.seed` is `Some`.
- `begin_prefill` (`src/server/batch/scheduler.rs:385-390`) calls `seed_rng_if_needed(&seq.sampling)` at line 388.
- `finish_prefill` (`src/server/batch/scheduler.rs:3765`) samples the first token via `sample_token_optimized(...)` at line 3802.
- `sample_token_optimized` -> `ffi::fused_sample` (`src/lib/mlxcel-core/src/sampling.rs:359-365`) takes only `(logits, temperature, top_k, top_p, min_p)`. There is no per-call seed argument; it draws from the global RNG.
- Ordering: `run_padded_batched_prefill` (`src/server/batch/scheduler.rs:3163`) calls `begin_prefill` for **all** rows up front (lines 3182-3186), runs one batched forward pass, then loops per row calling `finish_prefill` (around line 3309). So every row's seed is applied before the first sample is drawn, and the last row's seed is the live global RNG state when row 0 is sampled.

## Pre-existing, widened by #332

This is a pre-existing property of the all-cold batched-prefill path, which already ordered all `begin_prefill` before all `finish_prefill`. PR #346 (#332 cohort splitting) widens exposure: a cold row that previously fell back to **sequential** prefill (because its collected window held an incompatible sibling) now runs batched inside a `BatchedCold` cohort, so it newly inherits last-seed-wins. The in-code comment at `src/server/batch/scheduler.rs:3117-3128` documents exactly this "cold row now runs batched" behavior change.

## Impact (scoped honestly)

- Output tokens remain valid; there is **no corruption**.
- Greedy / `temperature == 0` requests are **unaffected**: the argmax path consumes no RNG (the `FusedSampleParams` field docs in `sampling.rs` note that `temperature == 0.0` and `top_k == 1` select the greedy argmax path).
- Affected only when a request supplies an explicit `seed` **and** `temperature > 0` **and** is batched with other seeded rows.
- Net: a per-request seed-reproducibility gap, not a crash or corruption. Hence priority:low.

## Suggested fix direction (not to implement here)

- Per-row reseed immediately before that row's sampling in `finish_prefill`; or
- A per-sequence RNG stream (instead of the shared global RNG) for seeded rows; or
- (cheap stopgap) exclude seeded non-greedy rows from batched cohorts in the prefill classifier (the `PrefillRow` classification in `execute_batched_prefill` / `plan_prefill_cohorts`).

The fix must be holistic: it has to cover the pre-existing all-cold case, not only the #332 mixed-window case.

## Acceptance Criteria

- [ ] A seeded request (`seed` set, `temperature > 0`) produces identical output whether run alone, in an all-cold batched cohort, or in a cohort-split mixed window.
- [x] Greedy / `temperature == 0` requests are unchanged.
- [x] A regression test pins per-request seed reproducibility under batched prefill, covering both the all-cold and the cohort-split cases.

## Technical Considerations

- The seed is applied to a process-global RNG (`ffi::random_seed`), so any solution that keeps the global RNG must guarantee the seed is live at the exact point each row is sampled (reseed inside `finish_prefill`, not in `begin_prefill`).
- Watch the interaction with the batched fused decode fast path (`config_supports_fused_batch` / `batched_fused_sample`): seeded non-greedy rows already need care there, and a per-sequence RNG stream would have to extend to decode steps for full end-to-end reproducibility, not just the first token.

Related: PR #346, issue #332.

inureyes:
## Summary

Fixes the "last-seed-wins" determinism bug in batched prefill so a seeded request's first token no longer depends on the seeds of the other rows it happens to be batched with. The change is a one-line reseed of the global MLX RNG at the exact first-token sample point, plus a real-model regression test.

## Root cause

`seed_rng_if_needed` seeds the process-global MLX RNG (`ffi::random_seed`) and is called from `begin_prefill`. In `run_padded_batched_prefill`, every row in a cohort runs `begin_prefill` up front (one loop) before any row runs `finish_prefill` (a later per-row loop that samples the first token). The first-token sampler is `fused_sample`, which takes only `(logits, temperature, top_k, top_p, min_p)` and draws from the global RNG with no per-call key. So by the time row 0 samples, the global RNG holds the LAST cohort row's seed, and every seeded row's first token is governed by whichever seed was applied last ("last-seed-wins"). PR #346 (#332 cohort splitting) widened the exposure: a cold row that previously fell back to sequential prefill now runs batched inside a `BatchedCold` cohort and newly inherits this behavior.

## Fix

Insert `seed_rng_if_needed(&seq.sampling)` in `finish_prefill`, immediately before the first-token `sample_token_optimized` call, so the global RNG is reseeded to THIS row's own seed at the exact point it samples. `begin_prefill`'s existing seeding is intentionally left in place (it is harmless once overwritten here, and is still reached from the disaggregated handoff path). `finish_prefill` is the single per-row first-token sampling chokepoint for every prefill path (batched cohort, sequential, full, chunked-final), so this one insertion fixes both the pre-existing all-cold case and the #332 cohort-split case holistically rather than patching each path.

## Greedy is a no-op

Greedy / `temperature == 0` / `top_k == 1` rows take the `fused_sample` argmax path, which consumes no RNG, so the reseed has no effect on them. The regression test pins this.

## Scope: decode is out of scope (honest AC #1 re-framing)

This fix makes a seeded row's FIRST token independent of its siblings' seeds, and greedy output is fully identical regardless of batch composition. It does NOT make a seeded request's full multi-token output byte-identical between "run alone" (B=1) and "batched" (B>1). Two pre-existing, already-documented effects remain and are not this seed bug: (1) padded batched prefill is not bitwise-identical to single-sequence prefill on Metal (the #203 / #325 / #326 forward-pass jitter class), and (2) the batched fused DECODE path (`execute_batched_decode` -> `batched_fused_sample`) samples all rows in ONE `fused_sample` draw over `[B, vocab]` from the shared global RNG, so a seeded row's decode trajectory depends on batch composition (the documented batched-vs-B=1 jitter class noted in `sampling.rs`). Full per-sequence multi-token determinism across batch sizes would require threading a per-row RNG key into `fused_sample` (an FFI / MLX-side change), which is disproportionate for this priority:low bug and is left as a follow-up. AC #1 is therefore satisfied at the first-token / batched-prefill scope of this bug; the residual is the existing jitter class, not last-seed-wins.

## Regression test (invariant choice)

`src/server/batch/scheduler_seed_determinism_tests.rs` (real qwen3-0.6b-4bit, `#[ignore]`, soft-skips when the model dir is absent), mirroring the `scheduler_cohort_parity_tests.rs` harness. It deliberately does NOT compare a batched (B>1) row against a single-prefill (B=1) row, which would conflate the forward-pass jitter class. Instead it holds the cohort COMPOSITION constant (same prompts in the same positions, so the measured row's logits are byte-identical) and varies ONLY seed values, isolating the RNG variable. The measured row is always at index 0 (sampled first) and the seeded sibling at index 1 (seeded last), which is exactly where "last-seed-wins" would leak. Checks:

- Sibling-seed invariance (load-bearing, bug-discriminating): the seeded stochastic row's first token is identical across distinct sibling seeds. This fails under "last-seed-wins" and holds exactly under the fix.
- Own-seed dependence (positive control): the same row's first token takes more than one distinct value across its own seeds, proving the sampler genuinely consumes RNG here (so the invariance check is not vacuous) and that the reseed reaches the row.
- Greedy invariance (AC #2): a `temperature == 0` row's first token is independent of any sibling's seed.
- Cohort-split coverage (AC #1 cohort-split clause / AC #3): a seeded row's first token in a #332 cohort-split mixed window (`[cold A, cold B, adopted C]`, which the planner splits into BatchedCold `{A, B}` + Sequential `{C}`) matches the same row in an all-cold batched window of the same cold composition (`[cold A, cold B]`).

## Test plan

- [x] `cargo test -p mlxcel --lib --release --features metal,accelerate batched_prefill_seed_determinism -- --ignored --nocapture` (observed: sibling seeds [1,2,3] -> first token [4792, 4792, 4792]; own seeds [10..60] -> 4 distinct first tokens; greedy [334, 334, 334]; cohort-split all-cold == mixed == 334)
- [x] `cargo fmt --check`

Closes #347

### DIFF
--- a/src/server/batch/scheduler.rs
+++ b/src/server/batch/scheduler.rs
@@ -3907,6 +3907,21 @@ impl BatchScheduler {
         } else {
             mlxcel_core::copy(&logits)
         };
+        // #347: reseed the global MLX RNG to THIS row's own seed at the exact
+        // point it samples its first token. `begin_prefill` already seeded once,
+        // but a batched cohort runs every row's `begin_prefill` up front before
+        // any row reaches `finish_prefill`, so by the time row 0 samples here the
+        // global RNG holds the LAST cohort row's seed ("last-seed-wins"). The
+        // fused sampler draws from that process-global RNG with no per-call key
+        // (`fused_sample` takes only the scalar params), so without this reseed a
+        // seeded row's first token would depend on its siblings' seeds. Reseeding
+        // here, rather than only in `begin_prefill`, guarantees the seed is live
+        // at the exact sample point and makes each row's first token depend only
+        // on its own seed. Greedy / `temperature == 0` / `top_k == 1` rows take
+        // the argmax path and consume no RNG, so this is a no-op for them. The
+        // batched fused DECODE path shares one global-RNG draw across the whole
+        // `[B, vocab]` batch and is out of scope here (see issue #347).
+        seed_rng_if_needed(&seq.sampling);
         let (first_token_arr, adjusted_logits) =
             sample_token_optimized(&logits_for_sampling, &seq.sampling, &token_history);
         mlxcel_core::eval(&first_token_arr);
@@ -4963,3 +4978,7 @@ mod serving_handoff_parity_tests;
 #[cfg(test)]
 #[path = "scheduler_cohort_parity_tests.rs"]
 mod scheduler_cohort_parity_tests;
+
+#[cfg(test)]
+#[path = "scheduler_seed_determinism_tests.rs"]
+mod scheduler_seed_determinism_tests;
--- a/src/server/batch/scheduler_seed_determinism_tests.rs
+++ b/src/server/batch/scheduler_seed_determinism_tests.rs
@@ -1,0 +1,443 @@
+// Copyright 2025-2026 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Real-model regression test for #347 batched-prefill seed determinism.
+//!
+//! The first-token sampler draws from the process-global MLX RNG with no
+//! per-call key (`fused_sample` takes only the scalar sampling params). A
+//! batched cohort runs every row's `begin_prefill` (which seeds the global RNG)
+//! up front, before any row reaches `finish_prefill` (which samples), so before
+//! the fix the LAST cohort row's seed governed every row's first-token sample
+//! ("last-seed-wins"): a seeded row's first token depended on its siblings'
+//! seeds. The fix reseeds the global RNG to each row's own seed inside
+//! `finish_prefill`, immediately before that row samples, so a row's first token
+//! depends only on its own seed.
+//!
+//! Invariant choice (the careful part). This test does NOT assert byte-identity
+//! of a batched (B > 1) row against a single-prefill (B = 1) row. Padded batched
+//! prefill is not bitwise-identical to single-sequence prefill on Metal (the row
+//! is padded up to the cohort's longest prompt and run through a wider matmul,
+//! so f16 rounding differs), which is the documented #203 / #325 / #326 jitter
+//! class, not a seed bug. To isolate the RNG variable, every assertion holds the
+//! cohort COMPOSITION constant (same prompts in the same positions, so the
+//! forward pass and therefore the measured row's logits are byte-identical) and
+//! varies ONLY seed values. Seeds never touch the forward pass, so any change in
+//! the measured row's first token under a constant composition is attributable
+//! purely to RNG sequencing.
+//!
+//! The measured row is always enqueued at index 0, so it is begin- and
+//! finish-prefilled first; the sibling is enqueued last, so under the old
+//! "last-seed-wins" behavior the sibling's seed was the live global RNG state
+//! when the measured row sampled. That is exactly the dependency the fix
+//! removes.
+//!
+//! Checks (all under the fixed code):
+//! - Sibling-seed invariance (the load-bearing, bug-discriminating check): a
+//!   seeded stochastic row at index 0 produces the SAME first token across
+//!   several distinct sibling seeds. This fails under "last-seed-wins" and holds
+//!   exactly under the fix.
+//! - Own-seed dependence (positive control, so the invariance check is not
+//!   vacuous): the same row produces MORE THAN ONE distinct first token across
+//!   several of its own seeds, i.e. the sampler genuinely consumes RNG at this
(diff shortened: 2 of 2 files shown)


inureyes:

### Intent
Reseed the global MLX RNG to each row's own seed at the batched-prefill first-token sample point, eliminating "last-seed-wins" for seeded sampling in a batched cohort (#347).

### Findings
No CRITICAL, HIGH, or MEDIUM correctness findings. The one-line reseed and its regression test are correct as written; no changes were made by this review.

### Verification
- [x] **Reseed placement correct.** `seed_rng_if_needed(&seq.sampling)` (scheduler.rs:3924) sits after the structured-output mask (3885-3909) and immediately before `sample_token_optimized` (3925). Nothing between consumes RNG: `sample_token_optimized_core` applies only deterministic token-bias and penalty graph ops before `fused_sample`, and the lone `eval` in the token-bias branch is an argmax, not a draw.
- [x] **`finish_prefill` is the single first-token chokepoint for every classic prefill path:** batched cohort (3418), full (3577), chunked first-chunk-terminal (3731), chunked continuation-final (3868). Sequential-cohort rows reuse the full/chunked finish sites. The speculative MTP/DFlash burst is a separate, pre-existing pipeline that bypasses `finish_prefill`, but it is structurally immune to last-seed-wins: a batched burst window only admits seed-identical rows (`sampling_config_eq` compares `seed`) and a B=1 burst is single-seed, so there is no distinct "last seed" to leak.
- [x] **Regression test discriminates the bug without the f16 batched-vs-B=1 jitter confound.** It holds cohort composition constant and varies only sibling seeds, measuring row 0 (sampled first) with the seeded sibling at row 1 (seeded last), which is exactly where last-seed-wins leaks. The positive control (own-seed variation yielding two or more distinct tokens) keeps the invariant non-vacuous. Two cold rows of different lengths do batch into one BatchedCold cohort on qwen3 (`can_pad == true`), so the test genuinely exercises the begin-all-then-finish-all ordering.
- [x] **Greedy unaffected.** `seed_rng_if_needed` is gated on `seed.is_some()` and the argmax path consumes no RNG; pinned by the greedy-invariance assertion.
- [x] **AC #1 reframing is honest.** First-token / batched-prefill determinism is achieved. The residual run-alone (B=1) vs batched (B>1) multi-token difference is the pre-existing forward-pass jitter class plus the batched-decode shared-RNG draw, both correctly identified and deferred to the FFI key change.
- [x] `cargo check --lib --tests --features metal,accelerate` passes. No placeholder code, conventions followed (test mirrors `scheduler_cohort_parity_tests.rs`), `Closes #347` present.

Verdict: correct and complete for its stated scope.

PR #359, opened 2026-06-18
fix(server): surface reasoning_content on non-streaming chat

### LINKED ISSUE #352: fix(server): gemma-4-12b-it defaults to enable_thinking, yielding verbose thinking and truncated/empty answers at normal max_tokens
## Problem / Background

`gemma-4-12b-it-4bit` is `model_type: gemma4_unified` (`ModelType::Gemma4Unified`). Its tokenizer carries the `<|channel>` / `<channel|>` added tokens, so on server startup `tokenizer.infer_thinking_markers()` resolves a think-marker pair through the multi-token (Gemma 4) branch and `has_thinking()` returns true. As a result the server flips the `chat_template` Jinja kwarg `enable_thinking` to `true` by default.

Verified code path:

- `src/tokenizer/mod.rs` (`infer_thinking_markers`, around line 214). The multi-token branch (around line 242) returns a populated `ThinkingMarkers` when both `<|channel>` and `<channel|>` are present in the vocab, with `think_start = "<|channel>thought"` and `think_end = "<channel|>"`.
- `src/server/startup.rs` (around lines 1591 to 1622). When `thinking_markers.has_thinking()` is true the server calls `chat_template.set_default_enable_thinking(true)`. The in-tree comment cites `ml-explore/mlx-lm#1114` and the `<|channel>thought` / `<channel|>` markers "for Gemma 4 and friends".
- Per-request `chat_template_kwargs` still win on conflict via `merge_server_and_request` (`src/server/chat_template_kwargs.rs`, around line 306), and `OPEN_THINKING_SUFFIXES = ["<|channel>thought\n", "<think>\n"]` (`src/server/routes/chat.rs`, around line 918) confirms the primed open-thinking block for both Gemma 4 and Qwen families.

This is the intended behavior introduced to mirror upstream `ml-explore/mlx-lm#1114` (`TokenizerWrapper.apply_chat_template` defaulting `enable_thinking=self.has_thinking`). It is not a regression. It was observed while validating #350 / #351 (and #333), and is orthogonal to #350 (the placeholder-leak fix, now merged).

## Symptom

With `enable_thinking=true` (the startup default for this model), gemma-4-12b-it produces very verbose thinking, so the answer `content` is truncated or empty at normal token budgets. Observed against `gemma-4-12b-it-4bit` on `/v1/chat/completions`, temperature 0:

- Prompt "What is a hash table? Answer in one sentence." at `max_tokens=80`: empty `content`, `finish_reason=length` (the whole budget is consumed inside the thinking block).
- Same prompt at `max_tokens=400`: a coherent one-sentence answer, but `finish_reason=stop` only after roughly 275 completion tokens, meaning about 245 tokens were spent thinking for a one-sentence answer.
- Prompt "Explain in three sentences how a hash table achieves average O(1) lookup." at `max_tokens=800`: `finish_reason=length`, `content` truncated to roughly 33 characters (still not finished after 800 tokens).
- With `chat_template_kwargs={"enable_thinking": false}`, the same prompts answer concisely and correctly with `finish_reason=stop` in well under 80 tokens.

So the verbosity and truncation are driven specifically by the thinking default, not by the model being broken.

## Secondary observation (verified): reasoning_content is dropped on the non-streaming path

In the repro the non-streaming response carried no reasoning content even though about 245 thinking tokens were generated for the one-sentence case. This is confirmed at the code level, and it is not gemma-4-specific:

- The non-streaming response message type `ChatMessage` (`src/server/types/response.rs`, around line 113) has only `role`, `content`, and `tool_calls`. There is no `reasoning_content` field. The non-streaming handler `non_stream_chat_completion` (`src/server/routes/chat.rs`, around line 341) strips the thinking scratchpad via `strip_unclosed_primed_thinking` and `clean_structural_tokens` and never surfaces it. So for any thinking model (qwen3 with `<think>`/`</think>` included), the non-streaming `/v1/chat/completions` response discards the reasoning entirely.
- The streaming path does surface it: `src/server/routes/chat.rs` (around lines 729 and 793) emits `delta.reasoning_content` chunks via the `StreamFilter` for both the `<think>` (Qwen) and `<|channel>` (Gemma 4) families.

Net: the discarded thinking is a general gap in the non-streaming response shape, not a gemma-4-only extraction bug. It is worth tracking here because it compounds the user-visible impact for this model.

## Impact

A default `/v1/chat/completions` request to gemma-4-12b-it with a typical `max_tokens` returns an empty or truncated answer, which looks like the model is broken when it is not. This affects the default serving experience for this model and is user-visible.

## Directions to investigate (not prescriptive)

1. Reconsider whether `<|channel>` detection should default `enable_thinking=true` for `gemma4_unified`, given the model over-thinks simple prompts, or whether the default should be off, opt-in, or at least documented.
2. If thinking stays enabled, surface `reasoning_content` on the non-streaming response (the field is absent today) rather than silently stripping it, and document a recommended higher `max_tokens` for this model.
3. Confirm whether the model actually emits the `<|channel>thought` ... `<channel|>` markers in its output (which bears on extraction), or whether the over-thinking is unmarked verbose generation that the marker-based stripper never matches.

## Acceptance Criteria

- [x] A default `/v1/chat/completions` request to gemma-4-12b-it with a typical `max_tokens` (for example 256) returns a non-empty, coherent answer, achieved either by adjusting the thinking default for this model or by ensuring the answer fits within a normal budget.
- [x] The thinking content is either surfaced via `reasoning_content` (non-streaming included) or intentionally documented as stripped.
- [x] The chosen behavior (thinking default and recommended `max_tokens`) is documented for `gemma4_unified`.

## Technical Considerations

- The thinking default originates in `set_default_enable_thinking(true)` driven by `infer_thinking_markers().has_thinking()`. Any change should preserve the existing precedence: explicit per-request `chat_template_kwargs.enable_thinking` and the CLI/env defaults (`--chat-template-kwargs`, `LLAMA_ARG_CHAT_TEMPLATE_KWARGS`) must continue to win via `merge_server_and_request`.
- Adding `reasoning_content` to the non-streaming `ChatMessage` would close the gap for every thinking family at once (qwen3 and gemma4), not just Gemma 4.

## Provenance

Observed during validation of #350 / #351 and #333. Not introduced by those changes: the thinking default is intentional and mirrors `ml-explore/mlx-lm#1114`. Bare #350, #351, #333, and #348 are references to this repository.

inureyes:
## Summary

`gemma-4-12b-it` (`model_type: gemma4_unified`) carries the `<|channel>` / `<channel|>` thinking markers, so on startup the server resolves a think-marker pair and defaults the `enable_thinking` chat-template kwarg to `true`, mirroring ml-explore/mlx-lm#1114. Per the maintainer decision this default stays as-is: the marker detection in `src/tokenizer/mod.rs` and `set_default_enable_thinking` in `src/server/startup.rs` are untouched. The real bug is that the non-streaming `/v1/chat/completions` response silently discarded the thinking scratchpad, so a verbose thinking model looked like it returned an empty or truncated answer at a normal `max_tokens`.

This surfaces the reasoning as `reasoning_content` on the non-streaming response, which closes the gap for every thinking family at once (qwen3 `<think>` and gemma4 `<|channel>`), and documents the thinking default plus a recommended higher `max_tokens` for gemma4_unified.

## What changed

- `src/server/types/response.rs`: add an optional `reasoning_content: Option<String>` to the response `ChatMessage` (`#[serde(skip_serializing_if = "Option::is_none")]`, so non-thinking responses keep their exact wire shape), matching the OpenAI/vLLM convention and the streaming `delta.reasoning_content`. Add a `with_reasoning_content` chained setter that mirrors `with_cached_tokens`.
- `src/server/routes/chat.rs`: add `extract_reasoning_content`, which replays the raw generation through the same `StreamFilter` the streaming path constructs, then wire it onto all three non-streaming return points (tool-calls, tools-enabled-no-calls, plain). The existing `content` computation (`strip_unclosed_primed_thinking` / `clean_structural_tokens` / tool-call parsing) is unchanged; reasoning is purely additive.
- `docs/supported-models.md`: document the gemma4_unified thinking default, the recommended `max_tokens` (at least 512), the `reasoning_content` surfacing on both streaming and non-streaming paths, and that per-request/CLI/env `enable_thinking=false` disables it.
- `docs/block-diffusion.md`: correct the stale claim that non-streaming responses strip `reasoning_content`.

## StreamFilter reuse (streaming / non-streaming parity)

Rather than write a new marker parser, `extract_reasoning_content` builds the same `StreamFilter` the streaming SSE path builds (primed-open vs fresh, keyed on `is_prompt_primed_open_thinking`), feeds the whole generated string once, calls `flush()`, and accumulates the `reasoning` pieces. The filter state machine is deterministic over the concatenated input, so feeding the whole string yields the same reasoning/content split as feeding it token-by-token. A unit test (`reasoning_split_identical_whole_vs_chunked`) locks this equivalence, which is what guarantees streaming and non-streaming surface identical reasoning. Per-request `chat_template_kwargs.enable_thinking`, `--chat-template-kwargs`, and `LLAMA_ARG_CHAT_TEMPLATE_KWARGS` precedence is untouched (`merge_server_and_request` unchanged).

## Real-model validation (gemma-4-12b-it-4bit, temperature 0)

Startup log confirms the intended default is preserved: `Tokenizer recognizes a think marker pair; defaulting chat_template kwarg enable_thinking=true (upstream PR #1114) think_start=Some("<|channel>thought") think_end=Some("<channel|>")`.

Prompt "What is a hash table? Answer in one sentence." at `max_tokens=400`, default thinking:

```
finish_reason: stop
completion_tokens: 275
has reasoning_content key: True
reasoning_content len: 980
reasoning_content (head): "*   Topic: Hash table. ... *Draft 1:* A hash table is a data structure that maps keys to values by using a hash function ..."
content: "A hash table is a data structure that maps keys to values by using a hash function to compute an index into an array, allowing for efficient data storage and retrieval."
```

Same prompt with `chat_template_kwargs={"enable_thinking": false}` (per-request precedence preserved):

```
finish_reason: stop
completion_tokens: 33
has reasoning_content key: False
content: "A hash table is a data structure that maps keys to values by using a hash function to compute an index into an array, allowing for efficient data retrieval and storage."
```

This also answers the issue open question 3: the over-thinking is marker-delimited (`<|channel>thought` ... `<channel|>`), since the filter captured the full 980-char block into `reasoning_content` while `content` holds only the answer (an empty `content` plus `finish_reason=length` at a tiny budget was the tokens being stripped as a marked thinking block, not leaked unmarked text).

## This fixes qwen3 too

qwen3-4b-4bit thinking request "What is 17 times 23? Answer in one sentence." at `max_tokens=512`, temperature 0:

```
finish_reason: stop
completion_tokens: 375
has reasoning_content key: True
reasoning_content len: 932
reasoning_content (head): " Okay, the user is asking for 17 multiplied by 23 ... 340 + 51 equals 391 ..."
content: "17 times 23 equals 391."
```

## Test plan

- [x] `cargo test -p mlxcel --lib --features metal,accelerate` for the new tests: `server::routes::chat::tests::extract_reasoning_*` (5), `server::routes::chat::tests::reasoning_split_identical_whole_vs_chunked`, `server::types::response::tests::reasoning_content_*` (2). All 8 pass.
- [x] `cargo fmt --check`
- [x] Real-model smoke on gemma-4-12b-it-4bit and qwen3-4b-4bit (above)

Closes #352

### DIFF
--- a/docs/block-diffusion.md
+++ b/docs/block-diffusion.md
@@ -123,7 +123,7 @@ Image input follows the standard OpenAI `image_url` content part format. Request
 The checkpoint is a thinking-channel model. The assistant fills a `reasoning_content` channel before writing the visible reply. A few practical consequences:
 
 - **Give requests enough `max_tokens`.** With a budget of 64 tokens a simple question may produce only reasoning, leaving `content` empty with `finish_reason: "length"`. A budget of 256 produces the full answer for most short prompts.
-- **Non-streaming responses omit reasoning.** The server's existing convention is to strip `reasoning_content` from non-streaming chat completions. Streaming responses carry it as `delta.reasoning_content`.
+- **Reasoning is surfaced on both paths.** Streaming responses carry it as `delta.reasoning_content`; non-streaming `/v1/chat/completions` responses include a `reasoning_content` field on the assistant message (present only when the model produced reasoning, omitted otherwise).
 - **`/v1/completions` works but raw prompts degrade.** The endpoint accepts un-templated text mechanically, but the model depends on its chat structure, so output quality degrades significantly. Prefer `/v1/chat/completions`.
 - **Cancellation resolution.** A cancelled request is aborted within at most one denoising step, not instantaneously.
 
--- a/docs/supported-models.md
+++ b/docs/supported-models.md
@@ -85,6 +85,28 @@ Audio/video capability is model-specific. The server request types include
 must advertise support for the corresponding modality. Video frame extraction
 uses the system `ffmpeg`/`ffprobe` binaries at runtime.
 
+### Thinking default for `gemma4_unified`
+
+Gemma 4 Unified (`gemma4_unified`) ships `<|channel>` / `<channel|>` thinking
+markers in its tokenizer, so the server defaults `enable_thinking=true` for this
+family on startup, mirroring [ml-explore/mlx-lm#1114](https://github.com/ml-explore/mlx-lm).
+With thinking on, the model writes an internal scratchpad before the visible
+reply, so simple prompts spend more of the budget on reasoning than on the
+answer. A one-sentence answer can take roughly 275 completion tokens, and a
+default `max_tokens` of 64 to 80 may return an empty `content` with
+`finish_reason` of `length` because the whole budget went to thinking. Set
+`max_tokens` to at least 512 for this family, and higher for multi-sentence
+answers.
+
+The scratchpad is no longer dropped: it is surfaced as `reasoning_content` on
+both streaming responses (`delta.reasoning_content`) and non-streaming responses
+(a `reasoning_content` field on the assistant message, present only when the
+model produced reasoning). This applies to every thinking family, including
+Qwen-style `<think>` models. To turn thinking off, pass
+`chat_template_kwargs={"enable_thinking": false}` per request, or set the server
+default via `--chat-template-kwargs` or `LLAMA_ARG_CHAT_TEMPLATE_KWARGS`. A
+per-request value always wins over the server default.
+
 ## Quantization formats
 
 | Format | Status | Notes |
--- a/src/server/routes/chat.rs
+++ b/src/server/routes/chat.rs
@@ -440,6 +440,15 @@ async fn non_stream_chat_completion(
 
     let cached_tokens = result.cached_tokens;
 
+    // Surface the thinking scratchpad as `reasoning_content`. This is additive:
+    // the `content` computation below (strip_unclosed_primed_thinking /
+    // clean_structural_tokens / tool-call parsing) is unchanged. Reusing the
+    // streaming `StreamFilter` here means streaming and non-streaming responses
+    // split reasoning from content identically. `None` for non-thinking models
+    // leaves the field absent, closing the dropped-reasoning gap for every
+    // thinking family at once (Qwen `<think>`, Gemma 4 `<|channel>`).
+    let reasoning = extract_reasoning_content(&result.text, primed_open_thinking);
+
     // Try to parse tool calls from the output
     if tool_calls::should_parse_tool_calls(&request) {
         let tools = request.tools.as_deref();
@@ -463,7 +472,8 @@ async fn non_stream_chat_completion(
                         result.completion_tokens,
                         logprobs,
                     )
-                    .with_cached_tokens(cached_tokens, prompt_cache_enabled),
+                    .with_cached_tokens(cached_tokens, prompt_cache_enabled)
+                    .with_reasoning_content(reasoning.clone()),
                 ));
             }
         }
@@ -483,7 +493,8 @@ async fn non_stream_chat_completion(
                 Some(result.finish_reason),
                 logprobs,
             )
-            .with_cached_tokens(cached_tokens, prompt_cache_enabled),
+            .with_cached_tokens(cached_tokens, prompt_cache_enabled)
+            .with_reasoning_content(reasoning.clone()),
         ));
     }
 
@@ -506,7 +517,8 @@ async fn non_stream_chat_completion(
             Some(result.finish_reason),
             logprobs,
         )
-        .with_cached_tokens(cached_tokens, prompt_cache_enabled),
+        .with_cached_tokens(cached_tokens, prompt_cache_enabled)
+        .with_reasoning_content(reasoning),
     ))
 }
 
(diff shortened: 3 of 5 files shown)


inureyes:
Follow-up commit `8402a75e1` addresses the HIGH finding from review: the non-streaming chat `content` leaked the reasoning text and a literal `</think>` for a primed-open `<think>` checkpoint that emits a close-only `</think>` (no opening tag in the output). The bare-leading-close pre-pass in `strip_thinking` handled only the Gemma 4 `<channel|>` primed case; it is now generalized to both `(open, close)` pairs, so chat non-streaming matches the split already done by `/v1/responses`, `/v1/anthropic`, and the chat streaming filter.

Added parser tests `strip_thinking_primed_qwen_close_only` and `clean_structural_tokens_primed_qwen_close_only`. Verified locally under the pinned toolchain (rust 1.93.1, matching CI): clippy `-D warnings` clean, 47 parser tests pass, the 5 `extract_reasoning` tests and the whole-vs-chunked parity test pass, fmt clean.

Issue #365, opened 2026-06-19
feat(models): Whisper ASR port for speech-to-text (/v1/audio/transcriptions)


inureyes:
Part of #363

Blocked by #364 (Phase 1 server audio plumbing). Start once the audio routes, multipart parsing, and audio-model interface from #364 are in place.

## Problem / Background

The `/v1/audio/transcriptions` and `/v1/audio/translations` endpoints (epic #363) need a real speech-to-text model. mlxcel has the log-mel front-end and audio-input encoders but no ASR decoder. This sub-issue ports an ASR model and wires it to the transcription/translation endpoints built in #364.

## Proposed Solution

- Port a Whisper-style encoder-decoder ASR model to mlxcel/MLX (mel encoder plus an autoregressive text decoder with the multilingual transcribe/translate task tokens). Reuse the log-mel front-end in `src/audio/feature_extractor.rs`. Register the `model_type` in `src/models/detection.rs`. Suggested checkpoint family: mlx-community Whisper (implementer confirms the exact MLX checkpoint and size).
- Alternative: a Conformer CTC/RNNT ASR (Parakeet-style) is also viable and partially aligns with the existing Nemotron Omni Conformer encoder infrastructure. Whisper is the recommended first target for broad language coverage and the built-in translate task.
- Wire to `/v1/audio/transcriptions` and `/v1/audio/translations` (translate task) on the #364 plumbing: uploaded audio (WAV via the existing reader) to mel features to ASR to text to OpenAI transcription JSON (`{ "text": ... }`, optional segments / verbose_json).

## Acceptance Criteria (integrated, not build-only)

- [ ] A real Whisper MLX checkpoint transcribes an uploaded WAV via `POST /v1/audio/transcriptions` end-to-end.
- [ ] `POST /v1/audio/translations` returns translated (to-English) text for a non-English clip.
- [ ] `mlxcel list` shows the ASR architecture.
- [ ] Unit tests added; `cargo test`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --check` clean.
- [ ] `docs/supported-models.md` updated.

## Technical Considerations

- Reuse the existing mel front-end rather than adding a second one.
- Depends on the audio-model interface, multipart upload, and route wiring from #364.

Issue #364, opened 2026-06-19
feat(server): audio endpoint plumbing (multipart upload, binary response, WAV output)


inureyes:
Part of #363

## Problem / Background

The OpenAI audio endpoints (epic #363) need a server foundation that does not exist yet in mlxcel: no `/v1/audio/*` routes, no multipart/form-data upload parsing, no binary audio responses, and no audio-OUTPUT encoder (only a WAV reader exists). This sub-issue builds that foundation so the STT and TTS model work plug in without reworking the server surface. No model is added here.

## Proposed Solution

Routes: register in `src/server/app.rs` (`create_app`), both the unversioned and `/v1` forms:
- `/audio/speech` + `/v1/audio/speech`
- `/audio/transcriptions` + `/v1/audio/transcriptions`
- `/audio/translations` + `/v1/audio/translations`

New handler module `src/server/routes/audio.rs`, exported via `src/server/routes/mod.rs`.

Request/response types in `src/server/types/` (mirror the OpenAI audio API):
- `AudioSpeechRequest`: `model`, `input`, `voice`, `response_format`, `speed`.
- `AudioTranscriptionRequest`: `model`, `file`, `language`, `response_format`, `temperature`.

Multipart upload (new; no existing pattern in the codebase): parse `multipart/form-data` for the STT upload via `axum::extract::Multipart`, extracting the audio file part plus the form fields.

Binary audio response (new) for TTS: `Content-Type: audio/{mpeg,wav}`, `Content-Disposition: attachment; filename=speech.<fmt>`, body via `axum::body::Body`.

Audio OUTPUT encoding (new) in `src/audio/`: at minimum a WAV writer (the counterpart to the existing `load_wav_*` reader). mp3/opus are an optional follow-up.

Audio-model interface: define a model-provider-facing trait/enum for audio model kinds so the STT (#365) and TTS (#366) work plugs in, with an in-process request/admission path. Reuse `state.can_accept_request()`, the existing metrics, and `ErrorResponse`.

Until the models land, the endpoints return a clear structured "audio model kind not loaded" error (e.g. HTTP 501).

## Acceptance Criteria (integrated, not build-only)

- [ ] All six routes registered in `create_app` and reachable.
- [ ] Multipart (STT) and JSON (TTS) requests parse and return structured responses/errors through the real server flow.
- [ ] WAV encode round-trips against the existing reader (unit test).
[...]

PR #374, opened 2026-06-19
feat(models): Kokoro-82M TTS with iSTFTNet vocoder for /v1/audio/speech

### LINKED ISSUE #366: feat(models): TTS model + vocoder for text-to-speech (/v1/audio/speech)
Part of #363

Blocked by #364 (Phase 1 server audio plumbing). Recommended to land after the STT sub-issue (#365), but the only hard dependency is #364.

## Problem / Background

The `/v1/audio/speech` endpoint (epic #363) needs a real text-to-speech model plus a vocoder/codec to turn model output into a waveform. mlxcel has no TTS path. The server plumbing (#364) and the shared dedicated-thread MLX worker (`AudioWorker`, landed with #365) already exist, so this sub-issue ports a TTS model and vocoder and wires synthesis to `/v1/audio/speech`.

## Proposed Solution

- Port a lightweight TTS model available as MLX. Recommended first target: Kokoro (phoneme-to-mel acoustic model with a built-in iSTFTNet vocoder, so it needs no separate neural codec, which makes it the simplest self-contained first TTS).
- For TTS families that do not bundle a vocoder, a separate neural vocoder/codec decode to waveform is required (e.g. Vocos, BigVGAN, EnCodec, or SNAC). Scope this sub-issue to one acoustic model plus whatever vocoder it requires; Kokoro keeps that to a single self-contained port.
- Implement the synthesis direction of the existing `AudioModelProvider` seam (`synthesize`) and run it on the shared `AudioWorker` (`src/server/audio_worker.rs`), so model load and MLX evaluation happen on one stream-initialized thread, exactly as the STT provider does. The route handler (`audio_speech` in `src/server/routes/audio.rs`) and the binary-response / WAV-writer plumbing are already in place from #364.
- Flow: text (plus `voice`, `speed`) to grapheme-to-phoneme to acoustic model to vocoder to waveform to encode (WAV via the #364 writer; mp3 optional) to binary response.

## Checkpoint (local validation asset)

Already downloaded for development and e2e validation:

- Path: `models/kokoro-82m/` (repo `prince-canuma/Kokoro-82M`; the identical `mlx-community/Kokoro-82M-bf16` works too).
- Model: `kokoro-v1_0.safetensors` (~312 MB, 548 tensors), non-quantized.
- Vocoder: iSTFTNet, built into the model (no separate codec checkpoint).
- Voices: `voices/<name>.safetensors` (54 voices; default `af_heart`). The redundant `.pt` copies are unused (read the `.safetensors` voice packs).
- `config.json` carries the architecture under an `istftnet` block plus `dim_in` / `hidden_dim` / `max_conv_dim` / `max_dur` and related fields.

## Key design notes

- Detection: Kokoro's `config.json` has NO top-level `model_type` field, so the `model_type`-based dispatch in `src/models/detection.rs` will not auto-detect it. Detection must key off another signal, for example the presence of the `istftnet` config block or the `kokoro-v1_0.safetensors` weight filename. Decide this so `-m models/kokoro-82m` resolves to the TTS provider.
- Grapheme-to-phoneme (g2p): Kokoro consumes phonemes, not raw text, so a g2p front-end is required (the analog of the STT mel front-end). This is the main new host-side component beyond the acoustic model and vocoder; decide the approach (a Rust phonemizer, a bundled lexicon plus rules, or an espeak-style backend) and the language scope. The voice embedding is selected from the `voices/` pack by the `voice` request field.
- Reuse: the `AudioWorker` from #365 (dedicated MLX thread + thread-local stream) handles the MLX thread/stream affinity, so the TTS provider must not evaluate MLX on `spawn_blocking` threads directly. The WAV output encoder and binary-response path come from #364.

## Acceptance Criteria (integrated, not build-only)

- [x] A real Kokoro MLX checkpoint synthesizes audio via `POST /v1/audio/speech` end-to-end, returning playable audio (verify by decoding the response and checking sample rate / duration).
- [x] `voice` (selects a `voices/` pack), `speed`, and `response_format` are honored.
- [x] The server resolves a Kokoro checkpoint to the TTS provider (detection works without a `model_type` field).
- [x] `mlxcel list` shows the TTS architecture.
- [x] Unit tests added; `cargo test`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --check` clean.
- [x] `docs/supported-models.md` updated.

## Technical Considerations

- Needs the audio-output encoder (WAV writer) and binary-response plumbing from #364, and the `AudioWorker` from #365.
- The vocoder decode is part of the acoustic model here (Kokoro's iSTFTNet is built in); the g2p front-end is the main new host-side piece.

inureyes:
## Summary

Ports the Kokoro-82M text-to-speech model (a StyleTTS2 phoneme-to-mel acoustic model with a built-in iSTFTNet vocoder) to MLX/Rust and wires it to `POST /v1/audio/speech`, plus a self-contained English grapheme-to-phoneme front-end since Kokoro consumes IPA phonemes rather than raw text. The audio scaffolding from #364 (route, WAV writer, binary response) and #365 (the dedicated-thread `AudioWorker` and `AudioModelProvider` seam) is reused unchanged; this PR adds the `synthesize` direction.

A real synthesis through the full pipeline produces playable audio: for the phonemes of "hello" the model emits a 24 kHz mono waveform of 15900 samples (0.663 s), finite and bounded (rms 0.012, peak 0.083), written through the existing WAV writer with a valid RIFF header.

## What changed

g2p front-end (`src/models/g2p/`): an American-English phonemizer. `normalize` tokenizes text and speaks integers; `lexicon` is a compiled-in word-to-IPA table; `rules` is a deterministic letter-to-sound fallback for out-of-vocabulary words; `mod` orchestrates `text_to_phonemes`. Output uses only symbols in Kokoro's vocab (audited by a test). No external binary or download.

Kokoro model (`src/models/kokoro/`): the StyleTTS2 + iSTFTNet forward pass split into focused modules: `bert` (PLBert / weight-shared ALBERT), `predictor` (duration + F0/N prosody with AdaLayerNorm and AdaIN), `text_encoder` (acoustic), `decoder` (iSTFTNet generator with NSF source and Snake MRF blocks), `stft` (STFT/iSTFT + harmonic source), `lstm` (hand-rolled bi-LSTM, since MLX has no recurrent op), `blocks` (shared AdaIN residual blocks), `ops`/`weights` (MLX helpers, weight-norm reconstruction, channels-last conv). `mod` is the `KokoroModel::load` / `synthesize` entry plus the `is_kokoro_checkpoint` detector.

Provider (`src/server/kokoro_tts.rs`): `KokoroTtsProvider` mirrors the Whisper STT provider, loading and running all MLX on one `AudioWorker` thread and forwarding requests over its channel.

Registry and wiring: `ModelType::Kokoro` added to the enum, `ALL_MODEL_TYPES`, metadata, the static-descriptor and tensor-parallel fallback tables, and the `mlxcel arch` family ordering; detection in `get_model_type` keys off the `istftnet` config block or the `kokoro-v1_0.safetensors` filename (Kokoro has no `model_type`) before the `model_type` dispatch; startup loads the provider for a Kokoro checkpoint. `docs/supported-models.md` gains a Text-to-speech section.

## Design decisions and scope

g2p approach: a bundled lexicon plus letter-to-sound rules, scoped to American English (the default voice `af_heart` and the other `a*` voices). This is the lowest-risk self-contained choice: no mandatory external binary (the repo has no espeak/phonemizer dependency) and it yields vocab-valid IPA for the default voice. Limit: non-English voices still load and synthesize, but their phonemes come from the English front-end, so pronunciation quality is reduced; per-language g2p (the analogue of upstream `misaki[xx]`) is future work.

iSTFT: the inverse follows the PyTorch reference (`magnitude * exp(j*phase)`, periodic Hann, no phase unwrap). It is computed with an explicit real inverse-DFT (Hermitian bin doubling, `1/N` normalization) rather than MLX complex64 array construction, which a round-trip test pinned as the correct unit-scale path. The NSF harmonic source is deterministic (zero initial phase, no additive noise) so synthesis is reproducible.

`response_format`: WAV is implemented (via the #364 writer). mp3 is not added; no audio codec is currently a dependency, so it stays a documented follow-up and the route returns 400 for non-wav as before.

## Validation

`cargo check --lib --tests`: clean. `cargo fmt --check`: clean. `cargo clippy --lib --tests -- -D warnings`: clean (also rewrote a pre-existing `mean_prompt_len` division that the gate flagged).

Unit tests (narrow selectors): 17 g2p tests pass (tokenizer, number expansion, lexicon lookup, symbol audit, rules, OOV, punctuation). 7 Kokoro tests pass including the STFT round-trip, alignment-matrix construction, voice-pack loading, detection, a real-checkpoint end-to-end synthesis (24 kHz, non-trivial duration, finite bounded samples), and a speed-scaling check (speed 1.5 yields fewer samples than 0.7). The end-to-end tests load `prince-canuma/Kokoro-82M` when present and skip otherwise.

## Acceptance criteria

All six issue checkboxes are satisfied: end-to-end synthesis through `/v1/audio/speech` returns playable 24 kHz audio; `voice`, `speed`, and `response_format` (wav) are honored; detection resolves a Kokoro checkpoint without a `model_type`; `mlxcel arch` lists the TTS architecture; unit tests are added and fmt/clippy/check are clean; `docs/supported-models.md` is updated.

## Known gaps and risks

mp3 output is deferred. g2p is English-only. The bundled lexicon is curated for high-frequency words; rarer words use the rule-based fallback, which is approximate. The deterministic NSF makes the waveform slightly quieter than a randomized run but perceptually equivalent; exact perceptual parity with the upstream MLX/PyTorch implementations was not bit-compared (a NumPy reference oracle was used during development to validate the forward-pass structure and the unit-scale iSTFT).

Closes #366


inureyes:

### Intent
> Port Kokoro-82M (StyleTTS2 acoustic + built-in iSTFTNet vocoder) plus a from-scratch English g2p front-end, wired to POST /v1/audio/speech, with detection that works without a top-level model_type.

### Findings Addressed
- [x] Generator fed the wrong F0 curve, producing half-duration audio (HIGH). The iSTFTNet generator received the conv-downsampled F0 (f0_a, length T) instead of the raw predictor F0 (f0_pred, length 2T). Since the generator upsamples its F0 input by 300, the harmonic source came out at T*300 (half of the intended 2T*300), and the align_time trim left the waveform at roughly half its correct duration, with pitch driven off the smoothed/strided curve rather than the raw pitch. Fixed in 2a5123d to pass the raw f0_pred, matching both the upstream MLX reference (istftnet.py Decoder.__call__) and the NumPy oracle. Verified: the real-checkpoint e2e test for the test phonemes now emits exactly 31800 samples (1.325 s = 2T*300), matching ref_audio.npy and doubling the prior 15900. The conv-downsampled f0_a stays correct for the encode/decode feature concat.

### Remaining Items
- None blocking. The known gaps the PR already documents (mp3 output deferred; English-only g2p with rule-based fallback for rare words; deterministic NSF making output slightly quieter than a randomized run) are acceptable, honestly stated, and in scope as follow-ups.

### Verification
- [x] All stated requirements implemented (e2e synthesis via /v1/audio/speech; voice, speed, response_format=wav honored; detection without model_type via istftnet block / weight filename; mlxcel arch and list show the TTS architecture; docs updated)
- [x] No placeholder/mock code remaining (real forward pass; numeric path validated against a NumPy oracle and an iSTFT round-trip test)
- [x] Integrated into project code flow (provider loaded in startup.rs for ModelType::Kokoro, stored in AppState via with_audio_model, reached through the route which gates on supports(Tts); MLX runs strictly on the AudioWorker thread, spawn_blocking only forwards over the channel)
- [x] Project conventions followed (Apache license headers, attribution-free git output, no em dashes, thiserror/anyhow-style error handling, Rust import/format conventions)
- [x] Existing modules reused (AudioWorker / AudioModelProvider seam and WAV writer from prior PRs; weight-map and stream helpers reused; no duplication)
- [x] No unintended structural changes (registry wiring is additive across the enum, ALL_MODEL_TYPES, metadata, tensor-parallel fallback, arch ordering; the mtp_policy.rs change is a benign checked_div refactor flagged by lint, semantics-preserving)
- [x] Tests pass (cargo check --lib --tests, clippy --lib --tests -D warnings, fmt, and test --lib models::kokoro 7/7 including the real-checkpoint e2e and STFT round-trip; g2p 17/17 per the orchestrator)

### Robustness (spot-checked, all sound)
- speed is sanitized in KokoroModel::synthesize (finite and > 0, else 1.0) and per-token frame counts are clamped to [1, 100] with NaN mapped to 1, so a zero/negative/huge/NaN speed cannot panic, hang, or allocate unboundedly.
- voice name handling is whitelisted to [A-Za-z0-9_-] at both resolve_voice and VoicePack::load, must also match an existing pack, and falls back to af_heart; path traversal is fully guarded (safe_name_rejects_traversal covers ../secret, a/b, a.b, empty).
- empty / whitespace / no-phoneme input returns a structured error; phoneme ids are truncated to MAX_TOKENS (510); g2p output is audited to stay within Kokoro's vocab.

inureyes:
## Security & performance review (PR #374, Kokoro TTS)

Reviewed the new code on the trust boundary `POST /v1/audio/speech` -> g2p -> Kokoro model, plus safetensors/voice loading. The model-side guards from the prior numeric-fidelity review hold up. Two HIGH findings on the attacker-reachable path are fixed in 9204867; one residual item needs a maintainer decision.

### Fixed

**HIGH, DoS: unbounded g2p work before the 510-token cap.** The TTS route inherits the 25 MiB audio body limit (`AUDIO_MAX_UPLOAD_BYTES`, sized for STT audio uploads), so `request.input` can be ~25 MB of free text. `KokoroEngine::synthesize` ran the full g2p front-end (`normalize` over the whole string, then a per-word lexicon/rules lookup) before `KokoroModel::synthesize` truncates to `MAX_TOKENS=510`. So the 510-token cap protected only the MLX/BERT stage, not g2p. With all synthesis on the single `AudioWorker` thread, one oversized request monopolizes it and stalls every audio request; repeated requests sustain an outage. Fix: cap `input` to `MAX_INPUT_CHARS=4096` on a char boundary before any g2p work (truncate, not reject; 4096 is already ~8x the 510-token ceiling). Unit-tested independently of the model.

**HIGH, availability: a synthesis panic kills the audio worker.** The worker loop called `engine.synthesize/transcribe` with no panic boundary. The kokoro forward pass carries ~20 `expect`/`unwrap` calls reachable from request input (MLX FFI helpers, shape math). A panic there unwound out of the loop, dropped the command channel, and made every later `/v1/audio/speech` and `/v1/audio/transcriptions` request fail with `WORKER_GONE` for the rest of the process (the provider never respawns the thread). The route-layer `spawn_blocking` `JoinError` branch does not catch this (the panic is on the worker thread, not the pool thread). Fix: wrap each engine call in `catch_unwind`, still synchronize the stream on both paths, and convert a caught panic into a per-request `Inference` error so the worker keeps serving. Regression test proves the worker survives a panicking request and serves the next one. I could not construct a phoneme string that provably panics under the current shape math, so triggerability is plausible-but-unproven; the backstop is cheap and the `expect` surface is large enough to warrant it.

### Needs a maintainer decision (not changed here)

**`panic = "abort"` defeats the panic backstop in release builds.** `[profile.release]` in `Cargo.toml` sets `panic = "abort"`, and the release workflow / README build with `cargo build --release`. In a production binary, any attacker-reachable panic in the synthesis path aborts the entire inference server (not just audio), and `catch_unwind` cannot intercept an abort. The backstop above still works in dev/test and any future unwind build, and documents intent, but it does not close the production hole. Options: (a) set `panic = "unwind"` for the release profile (reverses a deliberate size/codegen choice, affects the whole binary), or (b) de-panic the reachable kokoro forward-path `expect`/`unwrap` sites (return `Result` instead). Both are broader than this TTS PR's scope, hence flagged rather than applied. Recommendation: track as a follow-up; option (b) is the more robust long-term fix.

### Verified as already-safe (no change needed)

- Voice names whitelisted `[A-Za-z0-9_-]` and path-traversal-guarded at both `resolve_voice` and `VoicePack::load`; no user value reaches the filesystem except the whitelisted voice name; the voice name cannot escape `voices/`.
- `speed` sanitized (finite, > 0); per-token frame counts clamped to `[1,100]`; `total_frames` bounded by `510*100`; input ids truncate to 510 before the O(L^2) BERT attention and the alignment-matrix allocation; PLBert position-embedding indices stay in-bounds at the padded length of 512.
- No request input text logged at info level. Safetensors/voice loading is operator-controlled and returns `Result` (not panics) at the `load_safetensors` boundary.
- g2p lexicon is a `HashMap` (O(1) lookup), not a linear scan; `expand_integer` is O(n) with a digit-by-digit fallback past `u64`. No O(n^2) in normalize/lexicon/rules.

### Verification

Toolchain 1.93.1, narrow selectors: `cargo check --lib --tests` clean; `cargo clippy --lib --tests -- -D warnings` clean; `models::g2p` 17/17, `server::kokoro_tts` 4/4, `server::audio_worker` 4/4, `models::kokoro` 7/7 (the real-checkpoint e2e synthesis ran and passed).

inureyes:
## PR Finalization Complete

### Tests added

Three new test groups covering previously untested code paths (all deterministic, no checkpoint required):

- **g2p/mod.rs**: possessive clitic strip-and-retry (`cat's` -> `kætz`) and whitespace-only input (tab/newline)
- **kokoro/mod.rs**: weight-filename detection branch (file present, no `istftnet` block in config)
- **kokoro/voices.rs**: `resolve_voice` fallback behavior (empty voices dir, None, unknown name, unsafe traversal name) and known-voice resolution via a tmpdir with a stub safetensors file

Running total: 19 g2p tests, 7 kokoro tests (non-checkpoint), 3 voice tests, 4 kokoro_tts tests, 4 audio_worker tests.

### Documentation updated

**docs/audio-api.md**:
- Removed the stale "TTS always returns 501" statement from the Current Status paragraph
- Added `src/server/kokoro_tts.rs` to the implementation source map table
- Added a Kokoro TTS setup section covering checkpoint detection, voice selection, language scope, input length cap, and current limitations

**docs/supported-models.md**: already accurate from prior commits; no changes needed.

### Lint / format

`cargo fmt` reformatted two `run_guarded` call sites in `audio_worker.rs` (style only, no logic change). Clippy passes clean with `-D warnings`.

### Verification

`cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --check`: all green. All 33 targeted tests pass.

PR #380, opened 2026-06-21
fix: accept system-role turns inside /v1/messages array

### LINKED ISSUE #349: fix(server): /v1/messages returns HTTP 422 for Claude Code >= 2.1.156 (system-role turns inside the messages array)
## Summary

The Anthropic Messages endpoint (`/v1/messages`) rejects every request from Claude Code 2.1.156 and later with HTTP 422, before any generation runs. Starting with that build, Claude Code interleaves `{"role":"system", ...}` turns inside the `messages` array. mlxcel models only `user` and `assistant` roles for those turns, so JSON deserialization fails and the whole request is rejected. The endpoint is effectively unusable as a Claude Code backend for any current build.

## Background

The Anthropic spec places system text in the top-level `system` field and allows only `user`/`assistant` in `messages`. Claude Code 2.1.156+ no longer follows that for mid-conversation system reminders: it emits `role:"system"` entries inside `messages`. The reported version bisect:

| Claude Code | messages roles | result |
|-------------|----------------|--------|
| 2.1.145 | `[user]` | 200 |
| 2.1.153 | `[user]` | 200 |
| 2.1.156 | `[user, system, ...]` | 422 |
| 2.1.159 / 162 / 165 / 169 | `[user, system, ...]` | 422 |

The behavior changed between 2.1.153 and 2.1.156. Everything else Claude Code sends (top-level `system`, `tools`, `tool_use`/`tool_result`, `metadata`, image blocks) already returns 200. The OpenAI-compatible `/v1/chat/completions` path is unaffected because it already accepts `system` in `messages`.

The failure is confirmed statically against the current source:

- `AnthropicRole` (`src/server/types/anthropic_request.rs:139`) is `#[serde(rename_all = "lowercase")]` with only `User` and `Assistant`, so `role:"system"` is an unknown variant.
- `AnthropicMessage.role: AnthropicRole` (`src/server/types/anthropic_request.rs:132`) is the field that fails to parse.
- `anthropic_messages` (`src/server/routes/anthropic.rs:65`) takes `Json(request): Json<AnthropicRequest>` (`anthropic.rs:68`). There is no explicit 422 in mlxcel; this is axum's default when the `Json` extractor fails to deserialize.
- `append_message` (`src/server/anthropic_translator.rs:158`) matches `AnthropicRole` exhaustively over `User`/`Assistant` only (`anthropic_translator.rs:159-161`).

## Proposed Solution

1. Add a `System` variant to `AnthropicRole` and to `AnthropicRole::as_str` in `src/server/types/anthropic_request.rs` so the request deserializes instead of returning 422.
2. Map it in `append_message` (`src/server/anthropic_translator.rs:159`) as `AnthropicRole::System => Role::System`. The internal `Role` enum (`src/server/types/request.rs:23`) already has a `System` variant, so this mapping is available with no new type work.
3. Verify the mid-array system turn actually renders through the chat template. Many templates emit `system` only at the head of the conversation, so a system turn at `messages[1]` may be dropped or fall through to the fallback renderer. If a template cannot place a mid-conversation system message, fold it into the adjacent user turn (or merge it into the leading system block) rather than relying on the template.

## Implementation Notes

- The top-level `system` field is already converted to a `Role::System` message in `anthropic_translator.rs` (see the system-block handling around `anthropic_translator.rs:84`), so a `Role::System` message in the stream is already a shape the downstream path expects.
- Adding the enum variant alone removes the 422, but item 3 (template rendering) is the part that determines whether the system text actually influences generation. Treat the deserialization fix and the rendering check as one change, not two.
- Existing translator tests live in `src/server/anthropic_translator.rs` (for example the system-role assertions around `anthropic_translator.rs:560`); add a case there for a `system`-role turn inside `messages`.

## Acceptance Criteria

- [ ] A request whose `messages` contains a `system`-role turn deserializes (no 422) and the system text influences generation.
- [ ] Claude Code 2.1.156 or later connects and completes a turn end-to-end against `/v1/messages`.
- [ ] Top-level `system`, `tools`, and `tool_use`/`tool_result` behavior is unchanged.
- [ ] A translator test covers a `system`-role turn inside `messages`.

---

## Problem / Background

Pointing **Claude Code ≥ 2.1.156** at the mlxcel Anthropic Messages endpoint (`/v1/messages`, e.g. via `ANTHROPIC_BASE_URL`) makes every request fail with **HTTP 422** before any generation:

```
Failed to deserialize the JSON body into the target type:
messages[1].role: unknown variant `system`, expected `user` or `assistant`
```

The Anthropic spec only allows `user`/`assistant` in `messages` (system text goes in the top-level `system` field). However, Claude Code ≥ 2.1.156 interleaves `{"role":"system", …}` turns **inside the `messages` array** (mid-conversation system reminders). mlxcel's `AnthropicRole` only models `User`/`Assistant`, so axum's `Json` extractor fails to deserialize and rejects the whole request with 422.

Call chain (this repo):

- `AnthropicRole` (`src/server/types/anthropic_request.rs`) defines only `User`/`Assistant` under `#[serde(rename_all = "lowercase")]`, so `role:"system"` is an unknown variant.
- `AnthropicMessage.role: AnthropicRole` (`src/server/types/anthropic_request.rs`) is the field that fails to parse.
- `anthropic_messages` handler (`src/server/routes/anthropic.rs`) takes `Json(request): Json`; the extractor returns 422 on the deserialize error (there is no explicit 422 anywhere in mlxcel, it is axum's default for a failed `Json`).
- `append_message` (`src/server/anthropic_translator.rs:159`) matches `AnthropicRole` exhaustively (`User`/`Assistant` only).

Everything else Claude Code sends already returns 200: top-level `system` (string or blocks), `tools`, `tool_use`/`tool_result`, `metadata`, image blocks. Only a `system` role **inside `messages`** is rejected.

## Version bisect (measured)

Captured the actual `POST /v1/messages` body each Claude Code build sends, via a local mock Anthropic endpoint driven by `claude -p "hi"`:

| Claude Code | `messages` roles | result |
|-------------|------------------|--------|
| 2.1.145     | `[user]`               | 200 ✅ |
| 2.1.153     | `[user]`               | 200 ✅ |
| **2.1.156** | `[user, system, …]`    | **422** ❌ |
| 2.1.159 / 162 / 165 / 169 | `[user, system, …]` | **422** ❌ |

The behavior changed between **2.1.153 and 2.1.156** (2.1.153 is the last build that keeps `system` top-level; 2.1.154/2.1.155 were not available to test).

## Reproduction

```bash
# 200 — top-level system (spec-compliant)
curl -s -w '\n%{http_code}\n' http://127.0.0.1:8080/v1/messages -H 'content-type: application/json' \
  -d '{"model":"","max_tokens":32,"system":"be terse","messages":[{"role":"user","content":"hi"}]}'

# 422 — system role inside messages (what Claude Code >= 2.1.156 sends)
curl -s -w '\n%{http_code}\n' http://127.0.0.1:8080/v1/messages -H 'content-type: application/json' \
  -d '{"model":"","max_tokens":32,"messages":[{"role":"user","content":"hi"},{"role":"system","content":"be terse"},{"role":"user","content":"greet me"}]}'
```

## Impact

- mlxcel's `/v1/messages` is unusable as a Claude Code backend for any build **≥ 2.1.156** — it fails on the very first turn.
- Workaround: pin Claude Code to **≤ 2.1.153**.
- Scope is deserialization-level rejection, not corruption. Only the Anthropic endpoint is affected; the OpenAI `/v1/chat/completions` path already accepts `system` in `messages`.

## Suggested fix direction

1. Add a `System` variant to `AnthropicRole` (and `as_str`) so the request deserializes instead of 422-ing.
2. Map it in `anthropic_translator::append_message` (`AnthropicRole::System => Role::System`).
3. ⚠️ Then verify the **mid-array system turn actually renders** through the chat template. Most templates (Qwen, etc.) only emit `system` at the head, so a system turn at `messages[1]` may be dropped or fall through to the simple fallback renderer. If so, fold it into the adjacent user turn (or merge into the leading system block) rather than relying on the template to place a mid-conversation system message.

## Acceptance Criteria

- [ ] A request whose `messages` contains a `system`-role turn deserializes (no 422) and the system text influences generation.
- [ ] Claude Code ≥ 2.1.156 connects and completes a turn end-to-end against `/v1/messages`.
- [ ] top-level `system`, `tools`, and `tool_use`/`tool_result` behavior unchanged.
- [ ] A translator test covers a `system`-role turn inside `messages`.

## Environment

mlxcel v0.3.1, macOS (Apple Silicon), model `mlx-community/Qwen3.6-27B-4bit`, Claude Code 2.1.169.

inureyes:
## Summary

`/v1/messages` rejected every request from Claude Code 2.1.156 and later with HTTP 422 before any generation. Starting with that build, the client interleaves `{"role":"system", ...}` turns inside the `messages` array for mid-conversation reminders, and `AnthropicRole` only modeled `user`/`assistant`, so axum's `Json` extractor failed to deserialize the body. This restores the endpoint for current client builds and guarantees the interleaved system text actually reaches the model.

## What changed

- `src/server/types/anthropic_request.rs`: add a `System` variant to `AnthropicRole` (and to `as_str`, returning `"system"`) so the request deserializes instead of 422-ing. Updated the role doc comments to explain the non-spec interleaving.
- `src/server/anthropic_translator.rs`: map `AnthropicRole::System` to the internal `Role::System` in `append_message`, then add a `fold_system_messages` pass that relocates any non-head system turn so its text renders under any chat template.

## Why folding is required (not just the enum variant)

Adding the variant alone stops the 422 but does not guarantee the system text influences generation. Many production chat templates (Qwen-family, Llama 3) special-case `messages[0]` for the system block and their `{% for message in messages %}` loop only handles `user`/`assistant`/`tool`. A bare `system` turn at `messages[1]` is silently dropped by such a template, so the reminder would never reach the model. The fold pass moves the text into a turn every template renders.

## Merge strategy

- Leading `system` turns (before the first user/assistant/tool turn) merge into the single head system block, together with the top-level `system` field when present, so neither is dropped.
- A mid-conversation `system` turn is prepended to the following user turn. Claude Code emits the reminder immediately before a user turn, so this preserves its positional/contextual meaning; folding into an existing turn means we never depend on a template rendering a non-head `system` role, and no new user turn is created so user/assistant alternation is unchanged.
- A trailing `system` turn with no following user turn is appended to the preceding user turn, or merged into the head block as a last resort.
- Multimodal user turns keep their image parts; the reminder is added as a leading text part.

## Test plan

- [x] `cargo test --lib server::anthropic_translator` (29 passed, including 7 new cases: deserialization without 422, fold into following user turn, top-level plus mid-array system both surviving, leading-array merge with top-level system, trailing reminder preserved, multimodal fold preserving the image, and the unchanged top-level-system path)
- [x] `cargo test --lib server::types::anthropic_request` (10 passed, including a new deserialization regression test for a system-role turn inside `messages`)

The new translator tests render the folded output through a head-only Qwen-style template (system only at `messages[0]`, per-turn loop handles user/assistant only) and assert the reminder text survives, proving it was folded rather than left as an orphan mid-array system turn that such a template would drop.

## Follow-up (manual)

Live end-to-end verification with a real Claude Code 2.1.156+ client against `/v1/messages` could not be run here and is left as a manual check.

Closes #349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Fixed
+- **HTTP 422 from `/v1/messages` for Claude Code >= 2.1.156** (#380). Claude Code interleaves `{"role":"system", ...}` turns inside the `messages` array as mid-conversation reminders. The missing `System` variant in `AnthropicRole` caused `serde_json` to reject those requests before any generation. A new `fold_system_messages` translator pass now relocates mid-conversation system turns into the adjacent user turn (or the head system block) so the text reaches the model under any chat template, including head-only templates (Qwen, Llama 3) that silently drop non-head system messages.
+
 ## [v0.3.2] - 2026-06-20
 
 ### Added
--- a/src/server/anthropic_translator.rs
+++ b/src/server/anthropic_translator.rs
@@ -94,6 +94,10 @@ pub fn anthropic_request_to_chat(request: &AnthropicRequest) -> AnthropicTransla
         append_message(&mut messages, message);
     }
 
+    // 2b. Relocate any `system`-role turn that landed mid-conversation so its
+    // text actually reaches the model regardless of the chat template.
+    fold_system_messages(&mut messages);
+
     // 3. Tools / tool_choice.
     let tools = convert_tools(request);
     let tool_choice = convert_tool_choice(request.tool_choice.as_ref());
@@ -159,6 +163,10 @@ fn append_message(out: &mut Vec<Message>, message: &AnthropicMessage) {
     let role = match message.role {
         AnthropicRole::User => Role::User,
         AnthropicRole::Assistant => Role::Assistant,
+        // Claude Code >= 2.1.156 interleaves `system` turns inside `messages`.
+        // Map them to the internal `System` role here; `fold_system_messages`
+        // then relocates any that are not at the head of the conversation.
+        AnthropicRole::System => Role::System,
     };
 
     match &message.content {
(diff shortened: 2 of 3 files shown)


inureyes:

### Intent
> Accept Claude Code >= 2.1.156 requests that interleave `{"role":"system"}` turns inside the `/v1/messages` array (no more HTTP 422) AND guarantee that interleaved system text actually reaches the model under head-only chat templates, without changing top-level `system`/`tools`/`tool_use`/`tool_result` behavior.

### Findings Addressed
- [x] Multimodal fold concatenated the reminder and the user's text with no separator (MEDIUM). `MessageContent::text()` joins parts with an empty string, so a mid-array reminder before a multimodal user turn rendered as `watch outsee this` instead of `watch out\n\nsee this`. Fixed in `combine_user_text` by merging the reminder into the existing text part with the same blank-line separator the text-only path uses (falling back to a fresh part only for image-only turns). Independently flagged by the Codex reviewer (P2).

### Remaining Items
- None blocking. Two informational notes (no change required):
  - The literal `[user, system, user]` shape folds to `[user, user]` (two consecutive user turns). This adjacency is inherent to the input (two real user turns separated by a system turn) and is not introduced by the fold; the alternative (a bare mid-array system turn) is strictly worse since head-only templates drop it. Qwen-family templates (the issue's target) tolerate consecutive user turns.
  - A mid-array reminder followed only by an assistant/tool turn (no later user) is relocated to the head system block as a documented last resort. Text is preserved; positioning shifts from mid-conversation to global.

### Verification
- [x] All stated requirements implemented (System variant deserializes; fold pass routes the text into a turn every template renders; head-only Qwen-style render tests prove the reminder reaches the prompt)
- [x] Integrated into project code flow (`anthropic_request_to_chat` calls `fold_system_messages`; `append_message` maps `AnthropicRole::System`)
- [x] Project conventions followed (no `unwrap`/`expect` in new paths, inline format args, clippy clean)
- [x] Existing modules reused where applicable (internal `Role::System`, `MessageContent`, image plumbing)
- [x] No unintended structural changes (fold pass early-returns for the common top-level-system / no-system / tools / tool_use / tool_result paths)
- [x] No dropped system reminder or lost user/assistant/tool turn; user/assistant ordering preserved (verified by trace over leading / mid / trailing / consecutive / only-system / assistant-follows / tool-interaction cases, corroborated by the Codex reviewer)
- [x] Tests pass (`cargo test --lib server::anthropic_translator`: 30 passed; `cargo clippy --lib --tests -- -D warnings` clean; `cargo fmt --check` clean)

Edge-case tests added: exact-separator assertion on the multimodal fold and a consecutive-mid-array-system-turns case.

inureyes:
## PR Finalization Complete

### Tests
30 tests in `server::anthropic_translator` pass (all pre-existing; no gaps found that warranted new cases given the existing coverage of happy path, consecutive system turns, multimodal fold, leading/trailing edge cases, and head-only template render proof).

### Documentation
Added an `[Unreleased]` section to `CHANGELOG.md` (commit `3d5c2be`) recording the HTTP 422 fix and the `fold_system_messages` strategy. No dedicated Anthropic `/v1/messages` endpoint doc page exists in `docs/`, so no new file was created.

### Lint / Format
`cargo fmt -- --check`: clean. `cargo clippy --lib --tests -- -D warnings`: clean (17.6s, 0 warnings).

All checks pass. Ready for merge.

PR #371, opened 2026-06-19
feat(models): Whisper ASR speech-to-text for /v1/audio endpoints

inureyes:
## Summary

Port a Whisper-style encoder-decoder ASR model (Phase 2 of the audio-endpoints epic #363) and wire it to the speech-to-text endpoints from #364. When the loaded checkpoint is detected as Whisper, the server populates its speech-to-text slot so POST /v1/audio/transcriptions transcribes in place and POST /v1/audio/translations translates to English.

## What changed

- src/audio/whisper_mel.rs (new): Whisper log-mel front-end (80/128 mel bins, n_fft 400, hop 160, 16 kHz, 30 s windows) with a Slaney-scale, Slaney-normalized filterbank and global-max log normalization; reuses the shared real_fft_magnitude primitive and adds 16 kHz linear resampling.
- src/models/whisper/ (new module: mod, layers, encoder, decoder, tokenizer, decoding): convolutional plus transformer audio encoder, autoregressive text decoder with cross-attention to the encoder output and tied logits, greedy decode with a KV cache, first-step language detection, and Whisper token suppression (suppress_blank, non-speech symbols, timestamp range). Accepts native MLX and HuggingFace key layouts; applies the bf16 to f16 conversion for non-quantized weights.
- src/models/detection.rs, src/models/mod.rs, src/model_metadata.rs: register model_type "whisper" as ModelType::Whisper in detection, the arch/list registry, metadata, and the load-policy table.
- src/loading/mod.rs, src/distributed/tensor_parallel/inference.rs: the text loader fails clearly for a Whisper checkpoint (ASR, not a generator) and the tensor-parallel dispatch table stays total.
- src/server/whisper_stt.rs (new): WhisperSttProvider implements AudioModelProvider for speech-to-text, serializing MLX access behind a mutex (matching the existing unsafe Send/Sync holder pattern).
- src/server/startup.rs: detect Whisper at startup and attach the provider via AppState::with_audio_model.
- docs/supported-models.md: add a Speech-to-text (ASR) section and update the variant count.

## Validation

Unit-tested here (cargo test --lib whisper, cargo clippy --lib --tests -D warnings, cargo fmt --check, all clean):

- mel front-end: frame counts (100 frames per 1 s, 3000 per 30 s), filterbank shape, periodic Hann window, normalization band, tone energy in low mel bins, resampling.
- tokenizer: special-token and language-tag resolution, sot-sequence ordering, suppression-set union, text decode dropping specials and timestamps.
[...]

PR #381, opened 2026-06-21
perf(server): bound the audio request queue and add a per-request timeout

### LINKED ISSUE #373: perf(server): bound the audio request queue and add a per-request timeout
## Problem / Background

The OpenAI audio endpoints (speech-to-text landed in #365; text-to-speech is upcoming in #366) dispatch work to a single dedicated `AudioWorker` thread over an **unbounded** `mpsc::channel` (`src/server/audio_worker.rs`). There is no admission bound and no per-request timeout on the audio path: `can_accept_request()` (`src/server/state.rs`, around line 513) gates only on the generation batch queue depth, which the audio path never increments, so it is effectively always true for audio requests.

The PR #371 security review classified the following as MEDIUM under load:

1. **Head-of-line blocking**: all audio requests serialize through one worker thread, so a single slow or large clip stalls the whole audio queue.
2. **Unbounded memory growth**: each queued command holds up to the 25 MiB per-request audio payload, and nothing caps the number of concurrent in-flight requests, so a burst can grow memory without bound.

Neither is a crash (the per-request resample/alloc bounds from #365 cap a single request), but both are availability/throughput risks on the network-facing path.

## Proposed Solution

- Switch the `AudioWorker` channel to a bounded `mpsc::sync_channel(N)` and use `try_send`, mapping a full queue to the existing 503 "all slots are busy" response path (reuse the server's admission/error envelope rather than inventing a new one).
- Add a per-request reply timeout so a stuck or pathologically slow request frees its blocking thread and returns a structured error instead of hanging.
- Make the queue bound and timeout configurable via the server config, with sensible defaults.
- This lives in the shared `AudioWorker`, so it covers both the STT path and the TTS path (#366).

## Acceptance Criteria

- [x] The audio queue is bounded; when full, new audio requests get a structured 503 through the real server flow (unit/integration test).
- [x] A per-request timeout returns a structured error and releases the blocking thread (test).
- [x] Bound and timeout are configurable with documented defaults.
- [x] Integrated into the live audio request path (not a standalone helper).
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` clean.

## Technical Considerations

Surfaced by the security review of PR #371 (merged). Anchors: `src/server/audio_worker.rs`, `src/server/state.rs` (`can_accept_request`). Related epic: #363.

inureyes:
## Summary

Bounds the OpenAI audio worker queue and adds a per-request reply timeout so the network-facing audio path cannot grow memory without bound or hang a caller. Both fixes live in the shared `AudioWorker`, so they cover the Whisper STT and Kokoro TTS paths together. Surfaced by the PR #371 security review (two MEDIUM availability risks: head-of-line blocking through one worker, and unbounded memory growth from queued 25 MiB payloads).

## What changed

- `src/server/audio_worker.rs`: command channel switched from an unbounded `mpsc::channel` to a bounded `mpsc::sync_channel(queue_depth)`; `dispatch` uses `try_send` (full queue maps to the new `QueueFull` error); `transcribe`/`synthesize` use `recv_timeout(request_timeout)` (timeout maps to the new `Timeout` error). `spawn` gains `queue_depth` and `request_timeout` params and clamps a zero capacity to at least one queued command. `Drop` keeps a blocking shutdown send, which is correct because the worker keeps draining.
- `src/server/audio_model.rs`: two new `AudioModelError` variants, `QueueFull` and `Timeout`.
- `src/server/routes/audio.rs`: `audio_model_error_response` maps `QueueFull` to the shared `503` `server_busy` envelope and `Timeout` to a structured `504` `server_timeout`.
- `src/server/types/response.rs`: new `ErrorResponse::gateway_timeout` helper (`504`, `server_timeout`).
- `src/server/whisper_stt.rs`, `src/server/kokoro_tts.rs`: provider `load` takes `queue_depth` and `request_timeout` and forwards them to `AudioWorker::spawn`.
- Config plumbing mirrored on `max_queue_depth`: new `ServerConfig` fields `audio_queue_depth` (default 8) and `audio_request_timeout_secs` (default 120), wired through `ServerStartupInput`/`ServerStartupConfig`/`build_server_config` and read in the startup audio block (a `0` timeout falls back to the default). New flags `--audio-queue-depth` / `--audio-request-timeout-secs` and env vars `MLXCEL_AUDIO_QUEUE_DEPTH` / `MLXCEL_AUDIO_REQUEST_TIMEOUT_SECS` on both `mlxcel serve` and `mlxcel-server`.
- Tests: `queue_full_returns_queuefull_error` and `request_timeout_returns_timeout_error` (worker level, no real model, deterministic), plus route-level assertions that `QueueFull` is `503` and `Timeout` is `504`.
- Docs: new "Queue bound and per-request timeout" section and `504` rows in `docs/audio-api.md`; new "Server audio admission variables" section in `docs/environment-variables.md`.

## Test plan

- [x] `cargo check --bins` (CLI arg-struct wiring compiles)
- [x] `cargo test --lib server::audio_worker` (6 passed, incl. the two new tests)
- [x] `cargo test --lib server::routes::audio` (16 passed)
- [x] `cargo test --lib server::config` and `cargo test --lib server::cli_input` and `cargo test --lib server::startup` (pass)
- [x] `cargo clippy --lib --tests -- -D warnings` and `cargo clippy --bins -- -D warnings` (clean)

Closes #373

### DIFF
--- a/docs/audio-api.md
+++ b/docs/audio-api.md
@@ -89,7 +89,8 @@ Content-Type: `audio/wav`. Content-Disposition: `attachment; filename=speech.wav
 |--------|-----------|
 | 400 | Unsupported `response_format` (any value other than `wav` or absent). |
 | 501 | No TTS model is registered. Body: `{"error":{"type":"not_implemented","message":"audio model kind not loaded: tts"}}` |
-| 503 | All inference slots are busy. |
+| 503 | All slots are busy: either the generation batch queue or the bounded audio worker queue (`--audio-queue-depth`) is full. |
+| 504 | The audio worker did not reply within the per-request timeout (`--audio-request-timeout-secs`). |
 
 **Example (curl):**
 
@@ -127,7 +128,8 @@ Unknown multipart parts are drained and ignored. The upload body limit is 25 MiB
 | 400 | Malformed multipart, non-numeric `temperature`, or unsupported `response_format`. |
 | 413 | Upload body exceeds 25 MiB. |
 | 501 | No STT model is registered. Body: `{"error":{"type":"not_implemented","message":"audio model kind not loaded: stt"}}` |
-| 503 | All inference slots are busy. |
+| 503 | All slots are busy: either the generation batch queue or the bounded audio worker queue (`--audio-queue-depth`) is full. |
+| 504 | The audio worker did not reply within the per-request timeout (`--audio-request-timeout-secs`). |
 
 **Example (curl):**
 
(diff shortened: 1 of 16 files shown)


inureyes:
## PR Finalization Complete

### Summary

**Tests:** 7 audio_worker tests pass (including the new \`zero_request_timeout_falls_back_to_default\`); 16 audio route tests pass. All existing coverage for the bounded-queue 503 and per-request timeout 504 paths is present and deterministic.

**Code change (commit 2cf255539):** Added a \`Duration::ZERO\` self-guard inside \`AudioWorker::spawn\`, symmetric with the existing \`queue_depth.max(1)\` clamp. A zero timeout would make \`recv_timeout\` return \`Err(Timeout)\` before the worker thread can reply, rejecting every request. The guard falls back to the documented default (120 s) so the worker is correct in isolation regardless of what the caller passes. Added a focused unit test (\`zero_request_timeout_falls_back_to_default\`) that passes \`Duration::ZERO\` and verifies a fast request completes normally.

**Documentation:** \`docs/audio-api.md\` and \`docs/environment-variables.md\` accurately reflect both knobs: \`--audio-queue-depth\` / \`MLXCEL_AUDIO_QUEUE_DEPTH\` (default 8) and \`--audio-request-timeout-secs\` / \`MLXCEL_AUDIO_REQUEST_TIMEOUT_SECS\` (default 120), with the head-of-line and memory rationale. No changes needed.

**Lint/Format:** \`cargo clippy --lib --tests -- -D warnings\` clean; \`cargo fmt\` applied (reformatted the new block).

Issue #382, opened 2026-06-21
fix(server): make audio synthesis robust to MLX FFI exceptions (cxx std::terminate bypasses the panic backstop)


inureyes:
## Problem / Background

The audio synthesis and transcription forward paths (`src/server/kokoro_tts.rs`, `src/models/kokoro/*`, `src/server/whisper_stt.rs`) build MLX graphs through `mlxcel-core` ops that are declared in the cxx bridge (`src/lib/mlxcel-core/src/lib.rs`) as `-> UniquePtr<MlxArray>`, i.e. NOT `Result`. Only a few boundaries (for example `try_eval`) return `Result`.

When the underlying MLX C++ throws (shape mismatch, allocation failure, etc.) inside a non-`Result` cxx function, cxx hard-aborts the process via `std::terminate`. This is NOT a Rust panic, so it is NOT intercepted by the `catch_unwind` backstop in `src/server/audio_worker.rs` (`run_guarded`), and it cannot be caught regardless of the release `panic` setting (`abort` or `unwind`).

Consequence: an MLX FFI exception on the audio forward path aborts the whole server process. This is the dominant residual fault vector for the audio path and is the one class that the Rust-panic-class remediations cannot cover. The existing input guards (4096-char input cap, Kokoro-vocab-restricted g2p output, 510-token truncation, finite/positive speed, per-token frame counts clamped to [1,100], whitelisted voice names) make it largely unreachable today, but it remains the gap.

## Proposed Solution

Make the MLX ops used on the audio forward path fallible at the FFI boundary so a C++ exception becomes a recoverable `Err` instead of a `std::terminate`. Options to evaluate:

- (a) Add `Result`-returning variants of the specific ops used on the audio forward path (cxx converts the thrown C++ exception into `Err`), and thread them through the Kokoro/Whisper forward code. Scoped to the audio path.
- (b) A broader `mlxcel-core` policy of `Result`-returning graph ops. Larger, cross-cutting, affects every model. Likely out of scope for a first pass.

Feasibility and the exact op surface to convert need assessment. This is a `mlxcel-core` FFI change, intentionally separated from the panic-safety issue #375 (which addresses only Rust-panic-class containment).

## Acceptance Criteria

- [x] An MLX C++ exception raised on the audio synthesis or transcription forward path results in a structured per-request error, not a process abort, in a release build.
- [x] A regression test (or documented manual verification) demonstrates the behavior.
[...]

PR #383, opened 2026-06-21
fix(server): make audio synthesis panic-safe in release via panic=unwind with explicit core-thread abort

### LINKED ISSUE #375: fix(server): make audio synthesis panic-safe in release builds (panic="abort" defeats the AudioWorker catch_unwind backstop)
## Problem / Background

PR #374 added the Kokoro text-to-speech provider, which runs the StyleTTS2 + iSTFTNet forward pass on the shared single `AudioWorker` thread (`src/server/audio_worker.rs`) behind `POST /v1/audio/speech`. To keep one bad request from bricking all audio requests, PR #374 added a `catch_unwind` boundary around each engine call so a synthesis panic becomes a per-request error and the worker thread survives.

The problem is that `Cargo.toml` sets `[profile.release] panic = "abort"`, and the release workflow and documented production build use `cargo build --release`. With `panic = "abort"`, a panic does not unwind, so `catch_unwind` cannot intercept it: in a release/production build a synthesis panic aborts the entire server process, not just the offending request. The audio forward path contains multiple `expect`/`unwrap` sites. The existing input guards (4096-char input cap, Kokoro-vocab-restricted g2p output, 510-token truncation, finite/positive `speed`, per-token frame counts clamped to [1,100], whitelisted voice names) make these largely unreachable today, but the backstop is ineffective in release builds, so any future regression that introduces a reachable panic would crash the whole server.

## Proposed Solution

Pick one of the following remediations, to be decided in this issue:

- Option A: reconsider `panic = "abort"` for the server release build (for example, use `panic = "unwind"` for the binary, weighing the binary-size and performance tradeoff that the setting was chosen for) so the per-request `catch_unwind` backstop works in production.
- Option B: keep `panic = "abort"` and instead make the audio synthesis forward path panic-free by converting the attacker-reachable `expect`/`unwrap` sites in `src/models/kokoro/*` and `src/server/kokoro_tts.rs` to recoverable `Result` errors, so no input can panic regardless of unwind policy.

## Acceptance Criteria

- [x] A synthesis-path panic in a release build results in a per-request error response and does NOT abort the server process.
- [x] A regression test (or documented manual verification) demonstrates this behavior in release mode.
- [x] The chosen approach does not regress the input guards or the debug-build behavior already covered by the `audio_worker` panic-survival test.

## Reference

Surfaced in PR #374; see the security-review comment on that PR for the two documented remediation options.

inureyes:
## Summary

Release builds set `[profile.release] panic = "abort"`, which defeats the `catch_unwind` request-isolation backstop PR #374 added to the audio worker (`run_guarded`). Under abort a synthesis panic does not unwind, so a single bad `/v1/audio/speech` request aborts the whole server instead of degrading to a per-request error. This flips the release profile to `panic = "unwind"` so that backstop (and the pipeline stage-isolation `catch_unwind`) work in the shipped binary, and re-imposes fail-fast explicitly on the core generation worker threads so a broken core invariant still aborts cleanly for a supervised restart.

## What changed

- `Cargo.toml`: `[profile.release]` `panic = "abort"` becomes `panic = "unwind"`, with a comment explaining the decision. This is the only profile that sets `panic`; no member crate overrides it.
- `src/server/model_worker.rs`: adds `run_core_thread_or_abort`, a narrow `catch_unwind` + `std::process::abort()` wrapper that logs without request content, and wraps the two core generation worker thread bodies (the batched worker and the legacy `--no-batch` sequential worker). The batch scheduler, the DiffusionGemma batch-1 loop, and the disaggregated serving-role loop all run inside those two thread bodies, so the two wrappers cover every core generation path. The wrapped-body diff is large because rustfmt re-indents the moved blocks; `git diff --ignore-all-space` shows the logical change is just the helper plus the two wrappers.
- `src/server/model_worker_tests.rs`: adds a happy-path unit test for `run_core_thread_or_abort` (a non-panicking body runs to completion and its side effect is observed). The abort path terminates the test runner and is documented as manually/subprocess-verified rather than unit-tested in-process.
- `src/server/audio_worker.rs`: adds a comment on `worker_survives_engine_panic_and_keeps_serving` noting it now represents release-build behavior, because release uses unwind to match the always-unwind test profile. The test is kept unchanged.
- `docs/adr/0003-release-panic-unwind-with-core-thread-abort.md` (new) and `docs/adr/README.md`: record the decision, the rejected options (keep abort and document only; full Kokoro-to-`Result` refactor as disproportionate), the no-global-abort-hook trap, that `cargo test` always unwinds, and the residual MLX-FFI `std::terminate` vector.
- `docs/audio-api.md` and `docs/architecture.md`: short notes that synthesis panics are now contained as per-request errors in release, core generation threads abort on panic by design, and MLX FFI exceptions still terminate.

## Design notes / traps respected

- No global `std::panic::set_hook` that aborts: the hook runs before unwinding and would kill the audio worker `run_guarded` and the pipeline stage containment this change exists to restore. Only targeted per-thread `catch_unwind` + `abort` wrappers are used. The default panic hook (prints panic + backtrace) is kept.
- The audio worker `run_guarded`, the pipeline runtime `catch_unwind`, and Tokio async request handlers are deliberately left contained and are not wrapped.
- `AssertUnwindSafe` in the wrapper is sound because the wrapper aborts (never continues) on a caught panic, so no torn state is observed.

## Residual abort vector

An MLX C++ FFI exception thrown through a non-`Result` `cxx` op becomes `std::terminate`, not a Rust panic, so neither the `catch_unwind` backstops nor the unwind policy intercept it. This change does nothing for that path; it was filed separately as #382 and is referenced in the ADR.

## Thread spawn audit

Production long-lived threads found (`thread::spawn` / `thread::Builder`), and the disposition of each:

- `src/server/model_worker.rs:188` (batched model worker, core generation): wrapped.
- `src/server/model_worker.rs:714` (legacy sequential model worker, core generation): wrapped.
- `src/server/audio_worker.rs:165` (audio worker): left contained; it has its own `run_guarded` `catch_unwind` request isolation, which is exactly the boundary this change restores.
- `src/distributed/pipeline/remote_service.rs:597` (remote pipeline stage service): left contained as a pipeline stage-isolation boundary (outside the `src/server` / `src/lib/mlxcel-core/src` audit scope; a stage fault is contained to its request/connection).

`src/server/diffusion_worker.rs` spawns no thread of its own; `run_diffusion_worker_loop` runs on the model-worker thread, so it is already covered by the two wrappers. The remaining `thread::spawn` hits in the tree are `#[cfg(test)]` stubs and test helpers.

## Test plan

- [x] `cargo check --bins`
- [x] `cargo test --lib server::audio_worker` (7 passed, including `worker_survives_engine_panic_and_keeps_serving`)
- [x] `cargo test --lib run_core_thread_or_abort` (1 passed)
- [x] `python3 scripts/ci/check_cross_repo_refs.py` (OK)

Closes #375


inureyes:
## PR Finalization

### Documentation

All four docs verified against the final design:

- : ADR 0003 is indexed with an accurate one-line summary.
- : context, options, decision, no-global-hook trap, residual vector (#382), and consequences are all consistent with the implementation.
-  (Fault isolation section): correctly states that synthesis/transcription panics are contained as per-request errors since #375, that core generation threads take the opposite posture, and that the MLX FFI std::terminate path is tracked as #382.
-  (Panic and threading posture section): correctly identifies the audio worker as the only contained boundary, names the two wrapped server threads and the remote pipeline stage, describes coordinator-level fault handling (Result propagation plus timeout/health probing) rather than a catch_unwind, and references #382 and ADR 0003.

No em dashes in any of the new content. No stale or drifted phrasing found.

### Tests

-  (existing): exercises the void-return happy path.
-  (added, commit c89d6015b): confirms a Result-returning body that returns Err is forwarded unchanged, not treated as a panic. Cheap, robust, no flakiness.
- The abort path is documented as manually verified rather than tested in-process; a subprocess harness would add flakiness for a one-line abort and was correctly left out.
- : kept unchanged with its comment noting it now represents release behavior.

### Lint / Format

- : clean.
- : clean.
- : clean.
- 
running 17 tests
test server::model_provider::model_worker::tests::parse_byte_fallback_token_rejects_non_hex_tokens ... ok
test server::model_provider::model_worker::tests::parse_byte_fallback_token_recognises_hex_tokens ... ok
test server::model_provider::model_worker::tests::parse_byte_fallback_token_rejects_leading_sign ... ok
test server::model_provider::model_worker::tests::run_core_thread_or_abort_forwards_err_without_aborting ... ok
test server::model_provider::model_worker::tests::run_core_thread_or_abort_runs_body_to_completion ... ok
test server::model_provider::model_worker::tests::safe_emit_boundary_stops_before_trailing_replacement_chars ... ok
test server::model_provider::model_worker::tests::build_generation_result_computes_finish_reason_and_generation_split ... ok
test server::model_provider::model_worker::tests::merge_config_stop_tokens_appends_only_missing_ids ... ok
test server::model_provider::model_worker::tests::decode_request_images_rejects_decompression_bomb_header ... ok
test server::model_provider::model_worker::tests::decode_request_images_keeps_valid_images_and_rejects_all_invalid_input ... ok
test server::model_provider::model_worker::tests::token_piece_identifies_byte_fallback_tokens ... ok
test server::model_provider::model_worker::tests::byte_fallback_incomplete_sequence_flushed_at_end_of_stream ... ok
test server::model_provider::model_worker::tests::byte_fallback_single_byte_ascii_emits_immediately ... ok
test server::model_provider::model_worker::tests::byte_fallback_incomplete_then_regular_token_does_not_lose_text ... ok
test server::model_provider::model_worker::tests::byte_fallback_emoji_streams_only_complete_sequences ... ok
test server::model_provider::model_worker::tests::byte_fallback_cjk_streams_only_complete_sequences ... ok
test server::model_provider::model_worker::tests::byte_fallback_mixed_regular_and_cjk_streams_correctly ... ok

test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 3532 filtered out; finished in 0.02s: 17 passed.
- 
running 7 tests
test server::audio_worker::tests::worker_surfaces_loader_failure ... ok
test server::audio_worker::tests::default_engine_reports_unsupported_direction ... ok
test server::audio_worker::tests::zero_request_timeout_falls_back_to_default ... ok
test server::audio_worker::tests::worker_round_trip_runs_mlx_op_and_replies ... ok
test server::audio_worker::tests::worker_survives_engine_panic_and_keeps_serving ... ok
test server::audio_worker::tests::queue_full_returns_queuefull_error ... ok
test server::audio_worker::tests::request_timeout_returns_timeout_error ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 3542 filtered out; finished in 0.92s: 7 passed.

PR #385, opened 2026-06-21
feat(generate): wire MTP speculative decode into offline generate

### LINKED ISSUE #166: feat: wire MTP speculative decoding into the offline `mlxcel generate` path
Wire the MTP speculative-decoding round loop into the **offline `mlxcel generate`** path. Today MTP speculative decode only runs in the server burst path; the offline CLI returns a deferred error for every target.

## Current state

`src/commands/generate.rs` (~lines 899-926): when `--draft-kind mtp` is passed to offline `mlxcel generate`, it errors:

```
"--draft-kind mtp is plumbed end-to-end but the offline `mlxcel generate` path
 does not yet construct the kind-specific MtpGenerator round loop on this target model..."
```

So a one-shot CLI user cannot get the MTP speedup (~1.2-1.87x, measured) without standing up a server. The server burst path already drives `MtpGenerator` via `Gemma4MtpTargetAdapter` / `Gemma4UnifiedMtpTargetAdapter`; the offline path just never builds that loop.

## Goal

Construct and drive the `MtpGenerator` round loop in the offline generate path for MTP-capable targets, reusing the same per-target `MtpTarget` adapters the server uses (`src/models/gemma4_mtp_target.rs`), so `mlxcel generate --draft-kind mtp` performs speculative decode.

## Touchpoints

- `src/commands/generate.rs` — the `--draft-kind mtp` branch (replace the error with the real loop).
- `src/models/gemma4_mtp_target.rs` — reuse `Gemma4MtpTargetAdapter` / `Gemma4UnifiedMtpTargetAdapter`.
- `src/lib/mlxcel-core/src/speculative/mtp/` — `MtpGenerator`.
- The same B=1 default-on / `MLXCEL_ENABLE_MTP_B1` semantics as the server should apply.

## Acceptance criteria

- `mlxcel generate -m <target> --draft-model <assistant> --draft-kind mtp -p "..." -n N` runs MTP speculative decode (no error).
- At `--temp 0` the output is byte-identical to the non-speculative `mlxcel generate` output.
- A measured decode speedup on the 12B Unified and/or 31B + assistant pair (mirroring the server numbers).
- DFlash and the classic `SpeculativeGenerator` offline paths are unaffected.

inureyes:
## Summary

Wire the MTP speculative-decoding round loop into the offline `mlxcel generate` path. Previously `--draft-kind mtp` returned a deferred error for every target in the offline CLI; speculative decode only ran in the server burst path. The offline path now constructs and drives the same `MtpGenerator` round loop the server uses, reusing the same per-target `MtpTarget` adapters (`src/models/gemma4_mtp_target.rs`), so a one-shot CLI user gets the MTP speedup without standing up a server.

## What changed

- `src/commands/generate.rs`: `run_generation_mode` now takes a concrete `&LoadedModel` instead of a generic `M: LanguageModel`, so the `--draft-kind mtp` branch can match the target family and select the matching adapter. The sole caller already passes `&LoadedModel`, and every inner call (`generate_standard`, `generate_with_embeddings`, `SpeculativeGenerator::generate`) stays generic over `LanguageModel` (which `LoadedModel` implements), so the non-MTP monomorphized code is unchanged.
- New `run_offline_mtp`: resolves the concrete Gemma 4 target (text / VLM / Unified) the same way the server's `run_mtp_burst` does, loads the assistant via `load_drafter` (an MTP assistant is a `Drafter`, not a full `LoadedModel`), runs the compat check, binds the drafter to the same concrete target the adapter wraps (load-bearing: `MtpGenerator::generate` does not bind internally), then selects `Gemma4MtpTargetAdapter` / `Gemma4VLMtpTargetAdapter` / `Gemma4UnifiedMtpTargetAdapter` and drives the loop through `drive_offline_mtp`.
- Byte-identical parity: the resolved token bias (CLI `--logit-bias` plus the issue #350 multimodal-placeholder suppression that `run_generation_mode` already merges) is injected into `SamplingConfig.token_bias` so the adapter applies the SAME bias the non-speculative `CxxGenerator` applies via `with_token_bias`. The first-bonus sample also seeds the same history-dependent-penalty `token_history` as the classic decode path.
- Clear errors, no silent fallback: a non-MTP-capable target returns an actionable error, and a multimodal (`--image` / `--audio` / `--video`) request under `--draft-kind mtp` is rejected with guidance, mirroring the server's multimodal decline.
- Pure routing gate `should_route_offline_mtp` keeps the loop-construction decision unit-testable without loading a model.
- `src/commands/generate_tests.rs`: three model-free unit tests pin the routing gate.

## Preserved paths

The explicit `--draft-kind dflash` / internal-mtp branches keep their existing deferred error, and the auto-detect classic `SpeculativeGenerator` path (no explicit `--draft-kind`) is byte-for-byte unchanged. Only the explicit `--draft-kind mtp` branch now builds the real loop.

## Test plan

- [x] `cargo check -p mlxcel --bin mlxcel --features metal,accelerate`
- [x] `cargo test -p mlxcel --bin mlxcel --features metal,accelerate should_route_offline_mtp` (3 passed)
- [x] `cargo clippy -p mlxcel --bin mlxcel --features metal,accelerate --tests -- -D warnings` (clean)
- [x] `cargo fmt -p mlxcel -- --check` (clean)
- [ ] Real-model validation (deferred to the orchestrator, needs the release binary + multi-GB checkpoints): byte-identical temp-0 output and decode speedup, see below.

## Orchestrator validation (real models)

MTP speculative run (12B Unified pair):

`./target/release/mlxcel generate -m models/gemma-4-12b-it-4bit --draft-model models/gemma-4-12b-it-assistant-4bit --draft-kind mtp -p "Explain Apple Silicon's unified memory architecture in one short paragraph." -n 128`

Non-speculative baseline (must be byte-identical at temp 0, which is the default):

`./target/release/mlxcel generate -m models/gemma-4-12b-it-4bit -p "Explain Apple Silicon's unified memory architecture in one short paragraph." -n 128`

Same pattern for the 31B pair (`models/gemma-4-31b-it-4bit` + `models/gemma-4-31b-it-assistant-bf16`). The speedup is `baseline decode tok/s` vs the MTP run. Note: per the #203 jitter class, temp-0 classic-vs-MTP can show occasional word-level near-tie flips on some hardware; that is the same jitter the server path exhibits, not a wiring defect.

Closes #166

### DIFF
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -81,6 +81,10 @@ Measure with the `speculative_bench` harness or the server:
 mlxcel serve -m <target> --draft-model <drafter> --draft-kind mtp
 ```
 
+The offline `mlxcel generate` command also supports the MTP round-loop path (issue #166) for Gemma 4 text, VLM, and Unified targets. Pass `--draft-model <drafter> --draft-kind mtp` explicitly; without `--draft-kind mtp` the command keeps the classic speculative path for backward compatibility even when the drafter auto-detects as MTP.
+
+Parity: at `temperature 0` with no repetition, frequency, presence, or DRY penalties, the offline MTP output matches the non-speculative path within the documented f16 / #203 batched-kernel jitter class (on some hardware, notably M1 Ultra, near-tie token choices can differ). When any sampling penalty is active, only the first bonus token samples from the penalized distribution; subsequent tokens in each verify window are accepted or rejected greedily, so non-greedy or penalized requests are not byte-identical to the non-speculative path. This matches the known limitation of the server burst path.
+
 ### Gemma 4 Unified (12B) + 4-bit assistant
 
 Measured on Apple M5 Max (128 GB) with `mlx-community/gemma-4-12b-it-4bit` as the
--- a/src/cli/speculative_args.rs
+++ b/src/cli/speculative_args.rs
@@ -86,6 +86,14 @@ pub struct SpeculativeArgs {
     /// classic non-MTP / non-DFlash `SpeculativeGenerator` path when a
     /// drafter is supplied without an explicit kind.
     ///
+    /// Parity note (offline `mlxcel generate` with `--draft-kind mtp`):
+    /// at temperature 0 with no sampling penalties the output matches the
+    /// non-speculative path (within the f16 / #203 jitter class). When any
+    /// repetition, frequency, presence, or DRY penalty is active, only the
+    /// first bonus token is penalized; subsequent tokens in each verify
+    /// window are greedy, so penalized requests are not byte-identical to
+    /// the non-speculative path.
+    ///
     /// Also read from `LLAMA_ARG_DRAFT_KIND` (and the mlxcel-native
     /// alias `MLXCEL_DRAFT_KIND`).
     #[arg(long = "draft-kind", env = "LLAMA_ARG_DRAFT_KIND", value_name = "KIND")]
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -1154,7 +1154,7 @@ fn run_offline_mtp(
         .map_err(|e| anyhow!("MTP drafter bind failed: {e}"))?;
     println!("MTP drafter loaded and bound (block_size = {block_size}).");
 
-    // Inject the resolved token bias (CLI `--logit-bias` plus the model's
+    // Inject the resolved token bias (CLI `--lang-bias` plus the model's
     // reserved multimodal placeholder suppression from issue #350) into the
     // sampling config so the adapter applies the SAME bias the non-speculative
     // `CxxGenerator` path applies via `with_token_bias`. This is what keeps the
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -1171,7 +1171,7 @@ fn run_offline_mtp(
     // Select the per-target adapter exactly as the server does, then drive the
     // round loop. `seq_id = None` selects the wrapper's internal single-sequence
     // fallback slot, the documented offline / single-row CLI usage.
-    let (tokens, stats) = match model {
+    let (mut tokens, mut stats) = match model {
         LoadedModel::Gemma4(wrapper) => drive_offline_mtp(
             Gemma4MtpTargetAdapter::new_with_block_size(wrapper, None, block_size),
             drafter,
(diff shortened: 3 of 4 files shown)


inureyes:
## Security & performance review (offline MTP wiring)

Reviewed the changed files (`src/commands/generate.rs`, `src/commands/generate_tests.rs`) at HEAD `6d6999b44`, weighted toward performance, resource handling, and panic-safety since this is a local one-shot CLI path with minimal network surface.

**Verdict: no CRITICAL or HIGH findings. No code changes required.**

### Confirmed

- **Hot path / no per-token overhead.** `run_offline_mtp` does only one-time setup (load + validate + bind + a single `sampling_config.clone()`), then hands off to the existing `MtpGenerator` loop. `strip_trailing_eos` runs once at end-of-generation (single `position()` scan, O(n)), and the stats realignment runs once. Nothing is added inside the per-token round loop.
- **Monomorphization change is codegen-neutral.** `run_generation_mode` was generic only ever instantiated with `LoadedModel` (the sole caller passes `&LoadedModel`); switching the signature to concrete `&LoadedModel` produces identical code and adds no clones or allocations on the non-MTP path. The MTP-only `sampling.clone()` is off that path.
- **Resource safety.** The drafter is loaded exactly once via `load_drafter`; the MTP branch returns before the classic `load_model(draft_model_path)`, so there is no double-load. A load / validate / bind failure returns `Err` and the partially-set-up drafter drops via RAII with no partially-initialized state and no panic. `bind` takes the target by shared reference, so the target model is never left in a partial state.
- **Guards precede heavy allocation.** `block_size < 2` and the non-MTP-capable-target gate both return `Err` before `load_drafter` runs.
- **Panic-safety of new code.** `strip_trailing_eos` is safe for empty token vec, empty eos set, and eos at index 0 (each yields a no-op or empty result, pinned by unit tests). The stats realignment guards `decode_time_ms > 0.0`, so no divide-by-zero.
- **DoS / unbounded work.** The round loop is strictly bounded by `max_tokens` (loop-head check, the per-round `remaining`/`bs` clamp, and the walk `budget`). A misbehaving drafter that errors or under-produces causes a clean `break`, never unbounded generation.
- **Conventions.** No new `unsafe`, no AI attribution, no em dashes in the added lines, no secret/PII in the new log lines (only the drafter path, consistent with the existing classic-path log).

### Non-blocking observations (MEDIUM / LOW, no action taken)

- **MEDIUM (pre-existing, shared code):** `MtpGenerator::generate` asserts `!prompt_tokens.is_empty()`. An empty tokenized prompt under `--draft-kind mtp` would abort instead of returning an error. In practice Gemma 4 prepends BOS, so this is not reachable on normal input; the classic path degrades rather than asserting. Worth a guard if offline empty-prompt robustness is desired, but it is shared-generator behavior, not introduced here.
- **LOW:** `run_offline_mtp` matches over `LoadedModel` twice (target resolution, then adapter selection) with an `unreachable!()` in the second match. The branch is provably dead given the first gate, but folding both into a single match would remove the parallel-match fragility.

inureyes:
## PR Finalization (959afba17)

### Changes

**Comment fix (`src/commands/generate.rs` line 1157):** The comment named `--logit-bias` as the source of the resolved token bias. The actual flag is `--lang-bias` (resolved via `resolve_cli_token_bias` in `src/lang_bias.rs`). Corrected to `--lang-bias`. Functional behavior was already correct; this was comment accuracy only.

**Documentation (`docs/benchmarks.md`):** Added a paragraph in the existing "Speculative decoding (MTP)" section covering the offline `mlxcel generate --draft-kind mtp` path (issue #166): supported targets (Gemma 4 text / VLM / Unified), the temp-0 parity condition (within the f16 / #203 batched-kernel jitter class), and the greedy-parity-only limitation when any sampling penalty is active (first bonus token penalized, subsequent tokens in each verify window are greedy, so penalized requests are not byte-identical to the non-speculative path). This mirrors the server burst path's known limitation.

**CLI help text (`src/cli/speculative_args.rs`):** Added the same greedy-parity caveat in brief to the `--draft-kind` doc comment for discoverability via `--help`.

### Checks

- `cargo fmt --check`: clean
- `cargo clippy -p mlxcel --bin mlxcel --features metal,accelerate --tests -- -D warnings`: clean (15s, no warnings)

### Tests

Existing routing + `strip_trailing_eos` unit tests in `src/commands/generate_tests.rs` cover the PR's logic. No obvious cheap gap identified that doesn't require model loading.

PR #386, opened 2026-06-21
feat(server): add /v1/completions to the disaggregated router

### LINKED ISSUE #200: Disaggregated router: add the /v1/completions endpoint
## Context
Follow-up from epic #116 (#126, router front). The disaggregated router only serves chat completions.

## Problem
`create_router_app` (`src/server/router_front.rs`) registers only `POST /v1/chat/completions` and `GET /health`. The text-completion endpoint `POST /v1/completions` (served single-node by `routes::completions`) is not available through the router.

## Implementation
- Add a `router_completions` handler that parses a completion request, tokenizes the raw `prompt` with the same `add_special` rule the worker uses (`!prompt.starts_with("<bos>") && !prompt.starts_with("<s>")`), and reuses the same route -> send `PrefillRequestFrame` -> demux -> SSE/JSON path as the chat handler.
- Factor the shared body (tokenize -> select_prefill -> send -> drive_handoff_result -> emit) so chat and completions share it; the difference is request parsing and the response chunk shape (completion chunk vs chat chunk).
- Register the route in `create_router_app`.

## Acceptance criteria
- `POST /v1/completions` through the router returns output byte-identical to single-node `/v1/completions` for the same prompt and sampling.

inureyes:
## Summary

Add `POST /v1/completions` to the disaggregated router (`src/server/router_front.rs`), which previously served only `POST /v1/chat/completions` and `GET /health`. The new path reuses the single-node completion serializers so its output is byte-identical to single-node `/v1/completions` for the same prompt and sampling. Follow-up from epic #116 / #126.

## What changed

- Factor the shared dispatch body (tokenize -> select_prefill -> send `PrefillRequestFrame`) out of `route_chat` into `start_handoff(state, prompt, request_id, response_id, opts) -> Result<(prompt_tokens, rx)>`. Both the chat and the completion handlers call it. The chat id scheme (`chatcmpl-<n>`), routing, and tokenization are unchanged, so the existing chat endpoint stays byte-identical to before this PR. `start_handoff` applies the same `add_special` rule the worker uses (`!prompt.starts_with("<bos>") && !prompt.starts_with("<s>")`) and removes the registered `pending` channel on any pre-send failure.
- Add `router_completions` / `route_completion`. They reuse the EXISTING single-node serializers (no hand-rolled completion shape): non-streaming returns `CompletionResponse::new_with_logprobs` (object `text_completion`, `choices[index, text, finish_reason, logprobs]`, `usage`); streaming emits `CompletionChunk::content` per token, then `CompletionChunk::finish`, an optional `CompletionChunk::usage` chunk when `stream_options.include_usage` is set, then the `[DONE]` sentinel, matching `routes::completions` chunk-for-chunk.
- `finish_reason` mirrors the worker exactly: `completion_tokens >= max_tokens` -> "length" else "stop" (same logic as `src/server/model_worker.rs`). `prompt_tokens` for the `usage` block comes from the router's own tokenization (equal to the worker's count under the shared `add_special` rule).
- Per-token logprobs are rejected with a `400 invalid_request_error` because the disaggregated `ResultFrame`s carry detokenized text only and cannot reproduce them; `n` and `echo` are not in `CompletionRequest`, so they are silently ignored exactly as the single-node path does.
- Register `POST /v1/completions` in `create_router_app` and update the module / factory docs.
- Add an integration test `disaggregated_router_completions_match_single_node` in `tests/disaggregated_router_e2e.rs` covering both the non-streaming JSON body and the streaming SSE chunk sequence.

## Parity notes

- Byte-identity is structural, not reimplemented: both endpoints serialize the same `CompletionResponse` / `CompletionChunk` types, and `serde_json` is built with `preserve_order`, so field order matches. The only inherently volatile fields are `id` (`cmpl-<uuid>`, same format as single-node) and `created` (timestamp); the test normalizes just these two before comparing.
- Pre-existing router limitation (unchanged by this PR, shared with the chat path): the `PrefillRequestFrame` does not carry stop sequences, so `stop` is not honored on the disaggregated path. The byte-identity claim holds for requests without stop sequences.

## Test plan

- [x] `cargo check -p mlxcel --lib --tests`
- [x] `cargo clippy -p mlxcel --lib --tests -- -D warnings`
- [x] `cargo fmt --check -p mlxcel`
- [ ] `cargo test --test disaggregated_router_e2e --release --features metal,accelerate disaggregated_router_completions_match_single_node -- --ignored --nocapture` (heavy; spawns four real `mlxcel-server` processes loading qwen3-0.6b-4bit, run by the orchestrator)

Closes #200

### DIFF
--- a/docs/CONTINUOUS_BATCHING.md
+++ b/docs/CONTINUOUS_BATCHING.md
@@ -124,12 +124,11 @@ curl http://127.0.0.1:8300/v1/chat/completions \
   excluded from the paged handoff; the exact-prefix snapshot cache described
   above is a single-node prompt-cache optimization and is not serialized across
   disaggregated prefill/decode roles.
-- Text-only. The router serves `/v1/chat/completions`; multimodal requests are
-  rejected by the router.
-- The router does not yet apply the chat stream filter that the single-node chat
-  route uses, so reasoning/tool-call markers pass through verbatim. Use a
-  non-reasoning prompt or `chat_template_kwargs.enable_thinking=false` for clean
-  content.
+- Text-only. The router serves `POST /v1/chat/completions` and `POST /v1/completions`; multimodal requests are rejected.
+- On `/v1/completions`, three option groups are rejected with HTTP 400 because the disaggregated wire protocol does not carry the data they require: `logprobs`, `response_format` (structured output), and explicit reasoning/thinking budgets. Requests without those fields work normally.
+- `stop` sequences are not forwarded to the worker nodes by the router. This is a pre-existing limitation shared with the chat path.
+- `completion_tokens` in the usage block is counted from emitted detokenized text pieces. For byte-level BPE tokenizers (Qwen, Llama) this equals the worker's token count exactly. For byte-fallback tokenizers (Gemma `<0xXX>` byte sequences) it can under-count, which may flip `finish_reason` between `"length"` and `"stop"`. A precise fix requires the disaggregated wire protocol to carry the worker's token count; a follow-up issue tracks the protocol extension.
+- The router stream filter suppresses model-specific structural markers (`<think>`, tool-call delimiters) and routes thinking content to `reasoning_content`. Tool-call parsing is not yet supported on the router path: only `content` and `reasoning_content` are emitted.
 - Co-locating the roles on one machine adds transport hops without the scaling
   benefit of separate prefill and decode hardware, so a single-machine setup is
   for validation. The throughput case for disaggregation is separate prefill and
--- a/src/server/router_front.rs
+++ b/src/server/router_front.rs
@@ -15,10 +15,16 @@
 //! Model-free HTTP router front-end for disaggregated serving (#126 B3b2b).
 //!
 //! The router loads a tokenizer and chat template but never loads model
-//! weights. Incoming `/v1/chat/completions` requests are tokenized, forwarded
-//! to a prefill node via [`PrefillRequestFrame`], and the two-part result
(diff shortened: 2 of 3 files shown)


inureyes:
## Finalization complete

**Documentation** (`docs/CONTINUOUS_BATCHING.md`): updated the "Scope and limitations" section in the disaggregated serving doc to reflect the new `POST /v1/completions` support on the router. Changes:

- Added `POST /v1/completions` next to `POST /v1/chat/completions` in the supported-endpoints sentence.
- Documented the three unsupported option groups that return HTTP 400: `logprobs`, `response_format`, and explicit reasoning/thinking budgets.
- Documented the two wire-protocol limitations callers should know: `stop` sequences are not forwarded (pre-existing, shared with the chat path), and `completion_tokens` can under-count for byte-fallback tokenizers (a follow-up issue tracks the protocol extension).
- Corrected the outdated bullet that said the stream filter was not yet applied; it has been active since PR #198.

**Tests**: no new unit tests added. The 400-guard conditions reduce to `Option::is_some()` checks and `pick_budget_alias()` (which has its own tests); extracting them into separate unit tests adds no coverage beyond what already exists. The byte-identity E2E test (`disaggregated_router_completions_match_single_node`) is the substantive coverage and has been confirmed passing.

**Lint/format**: `cargo fmt --check` passes clean; `cargo clippy -p mlxcel --lib --tests -- -D warnings` exits with no diagnostics.

Commit: `f9b0a459c` on `feature/issue-200-router-completions`.

PR #388, opened 2026-06-21
feat(router): drive multi-node disaggregated routing, health, and failover

### LINKED ISSUE #201: Disaggregated serving: multi-node routing, load balancing, and failover
## Context
Follow-up from epic #116 (#126, router front). The router wires `RequestRouter` but the single-prefill / single-decode setup is degenerate.

## Problem
`RouterState` (`src/server/router_front.rs`) registers the configured `--prefill-peers` and `--decode-peers` in a `NodeRegistry` and calls `RequestRouter::route_to_prefill` for selection, but with one prefill node it always returns that node, and the decode node is chosen by the prefill node's own `--decode-peers` config rather than by the router. So the load-balancing, health, and failover machinery (`RequestRouter::apply_backpressure` / `handle_node_failure`, `NodeRegistry` status) is unexercised. Scaling prefill and decode pools independently is the reason to run disaggregated, so this is the feature that makes it worthwhile.

## Implementation
- Router-driven decode selection: have the router pick the decode node (`route_to_decode`) and pass it to the prefill node in the request frame, instead of the prefill node choosing from static config, so the router balances both pools.
- Health and failover: probe peer liveness (or mark a node unreachable on a transport error) and update `NodeRegistry` status so `route_to_prefill` / `route_to_decode` skip dead nodes; on a node failure mid-request, re-route or fail the request cleanly via `handle_node_failure`.
- Backpressure: wire `apply_backpressure` into the admission path so the router queues or rejects when all nodes are at capacity, instead of unconditionally sending.
- Tests: a multi-prefill / multi-decode E2E (at least 2 of each) asserting requests spread across nodes and that killing one node reroutes subsequent requests.

## Acceptance criteria
- With multiple prefill and decode nodes, requests distribute across them per the routing strategy.
- A node failure does not wedge the router; subsequent requests route to healthy nodes.

inureyes:
## Summary

Make the disaggregated router actually balance independent prefill and decode pools, with health-driven failover and admission backpressure (issue #201). Before this change the router registered both pools but, with one prefill node, always returned it, and the decode node was chosen by the prefill node's own `--decode-peers` config rather than by the router, so the load-balancing, health, and failover machinery was never exercised.

## What changed

(1) Router-driven decode selection. PrefillRequestFrame gains an optional `decode_target: Option<String>` (`serde(default)`, in `src/distributed/disaggregated/serving_protocol.rs`). The router selects the prefill node with `route_to_prefill` and the decode node with `route_to_decode` (both round-robin, since the router has no live per-node load telemetry), and ships the chosen decode address in `decode_target`. The prefill role loop (`coordinator.rs::run_prefill_role_networked`) forwards the KV handoff to `decode_target` when present and falls back to its own `--decode-peers` config when absent, so a mixed-version pool keeps working: an older router omits the field and a new prefill falls back to config; a new router populates it and an older prefill ignores the unknown field.

(2) Health and failover. A transport error when sending a prefill frame marks the node `Unreachable` in the registry (so `route_to_prefill` / `route_to_decode` skip it), re-routes its in-flight requests via `handle_node_failure`, and retries the request on a healthy node, failing cleanly with an error when none remains. A background `spawn_health_monitor` task TCP-probes every peer's serving address on an interval so a dead decode node (which the router never sends to directly) is detected and skipped, and a recovered node is restored to `Online`. The prefill loop is now resilient: a failed first-token send or decode handoff fails only that one request (the client receives a terminal error frame instead of blocking) and the worker keeps serving the rest. Tracked requests reach a terminal phase on completion/failure and are purged by the monitor so request state does not leak.

(3) Backpressure. Every request passes `apply_backpressure` admission control before dispatch (`route_chat` / `route_completion`); on `Queue` or `Reject` the router returns HTTP 503 instead of attempting a send that cannot succeed.

(4) Observability. `GET /router/stats` reports per-node prefill/decode dispatch counts, node health, and the `RouterMetrics` snapshot.

(5) Docs. `docs/distributed.md` and the disaggregated section of `docs/CONTINUOUS_BATCHING.md` describe router-driven decode selection, health/failover, and backpressure.

Locking discipline is preserved: no registry, request-map, or pending-map lock is held across an await or send.

## Test plan

- [x] `cargo check --release -p mlxcel --lib --tests --features metal,accelerate` (clean)
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate distributed::disaggregated::request_router::tests` (25 passed)
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate distributed::disaggregated::serving_protocol::tests` (8 passed)
- [x] `cargo fmt --check -p mlxcel` (clean)
- [x] clippy clean for all changed files (the only clippy `-D warnings` failure is a pre-existing `mlxcel-core/build.rs:90` lint on `origin/main`, unrelated to this PR)
- [ ] heavy multi-node E2E (orchestrator): `cargo test --test disaggregated_router_e2e --release --features metal,accelerate -- --ignored --nocapture disaggregated_router_balances_and_survives_node_failure`
- [ ] existing parity E2E still green (orchestrator): `cargo test --test disaggregated_router_e2e --release --features metal,accelerate -- --ignored --nocapture disaggregated_router_completions_match_single_node`

Closes #201

### DIFF
--- a/docs/CONTINUOUS_BATCHING.md
+++ b/docs/CONTINUOUS_BATCHING.md
@@ -78,8 +78,39 @@ Networking flags:
 | Flag | Role | Purpose |
 |------|------|---------|
 | `--serving-bind <addr>` | prefill, decode, router | This node's own role-transport listener (`host:port`). |
-| `--decode-peers <addr,...>` | prefill, router | Decode node(s) a prefill node hands KV off to; routers also use these addresses for decode continuation routing. |
-| `--prefill-peers <addr,...>` | decode, router | Prefill node(s) accepted by a decode node and selected by a router. |
+| `--decode-peers <addr,...>` | prefill, router | On a router, the decode pool it balances; on a prefill node, the static fallback used only when the router does not pick a decode target for a request. |
+| `--prefill-peers <addr,...>` | decode, router | On a router, the prefill pool it balances; on a decode node, the prefill node(s) it accepts handoffs from. |
+
+### Multi-node routing, load balancing, and failover
+
+With more than one prefill or decode node, the router balances both pools
+(issue #201). For each request it picks a prefill node and, independently, a
+decode node, both round-robin across the nodes the registry currently considers
+online (the router has no live per-node load telemetry, so round-robin is the
+strategy that actually spreads load). The chosen decode node travels to the
+prefill node in the request frame's `decode_target` field, so the prefill node
+hands the KV cache to the router-balanced decode node rather than to its own
+`--decode-peers` config. The field is optional on the wire: a frame without it
+(an older router) leaves the prefill node on its config fallback, and an older
+prefill node ignores it and uses config, so mixed-version pools keep working.
+
+The router tracks node health and fails over without wedging:
+
+- A transport error when sending a request to a prefill node marks that node
+  unreachable, re-routes its in-flight requests, and retries the request on a
+  healthy node. When no prefill node is left, the request fails cleanly with an
+  error rather than hanging.
+- A background health monitor TCP-probes every peer's serving address on an
+  interval. A node that stops accepting connections is marked unreachable (so
+  routing skips it, including a dead decode node the router never sends to
+  directly), and a node that starts accepting again is restored to online.
+- Admission control runs before every dispatch: when the prefill queue is full
+  or no prefill node is available, the router returns HTTP 503 instead of
+  attempting a dispatch that cannot succeed.
+
+`GET /router/stats` reports the per-node dispatch counts for both pools, the
+registered nodes with their current health status, and the routing-metrics
+snapshot, so an operator can confirm the spread and see which nodes are down.
 
 The handoff transfers the paged block contents (not just metadata) over the
 transport, so the decode node reconstructs the exact KV the prefill node built;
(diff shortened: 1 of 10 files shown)


inureyes:
## PR Finalization Summary

**Commit:** `c1fbf6223` - docs only

**Documentation review**

`docs/CONTINUOUS_BATCHING.md` and `docs/distributed.md` already accurately describe router-driven decode selection, health/failover, and backpressure (the `apply_backpressure` admission control path returning HTTP 503 is documented as admission control in both files, which is accurate and sufficient).

Added a "Security and trust model" subsection to `docs/distributed.md` inside the Disaggregated inference section, recording the two MEDIUM review findings as operational guidance:

1. `GET /router/stats` on the client-facing port discloses internal cluster topology (peer addresses, health, request counts). Guidance: restrict the router's client port to a trusted segment, or proxy it.
2. Prefill nodes connect to the router-provided `decode_target` to ship the KV handoff. Safe under the trusted-segment assumption the transport already requires. Defense-in-depth allowlist tracked as follow-up.

**Tests**

No new unit tests added. `request_router_tests.rs` already covers every logical path (round-robin, least-loaded, memory-aware, prompt-length-aware, backpressure, phase transitions, node failure rerouting, unreachable-node skipping, failover to survivor, auto-purge). `serving_protocol_tests.rs` covers `decode_target` serde round-trips including the backward-compatibility `None` default for older routers. No obvious gap in pure logic.

**Lint / format**

- `cargo fmt -p mlxcel --check`: clean
- `cargo clippy -p mlxcel --lib --tests -- -D warnings`: clean (no warnings)

Ready for merge.

PR #392, opened 2026-06-21
fix(router): use worker's authoritative token count for usage

### LINKED ISSUE #387: fix(router): carry the worker's true generated-token count over the disaggregated wire protocol
## Problem

The disaggregated router (`src/server/router_front.rs`) derives `completion_tokens` (and the chat equivalent) by counting emitted text frames: one increment per `ResultFrame` text entry in the per-request `on_token` callback. This equals the worker's true generated-token count for byte-level-BPE tokenizers (e.g. Qwen), but can UNDER-COUNT for byte-fallback tokenizers (e.g. Gemma, which emit multi-frame `<0xXX>` byte sequences for a single token, or where one model token maps to multiple emitted text pieces).

Because `finish_reason` is computed as `completion_tokens >= max_tokens ? "length" : "stop"` (matching `src/server/model_worker.rs` ~line 1401), an under-count can also flip the reported `finish_reason` between "length" and "stop", and makes the router's `usage` diverge from the single-node `/v1/completions` and `/v1/chat/completions` responses for those tokenizers.

## Scope / why it is a separate issue

PR #386 intentionally kept byte-identity for byte-level-BPE tokenizers (validated via the `disaggregated_router_completions_match_single_node` E2E test on qwen3-0.6b-4bit) and documented this limitation rather than expanding scope. A correct fix requires extending the disaggregated serving wire protocol to carry the worker's authoritative generated-token count (and likely the worker-computed finish_reason) from the decode node to the router, which touches `src/distributed/disaggregated/serving_protocol.rs` plus both the prefill and decode emit paths and the router demux. That is a protocol change, separate from the byte-parity feature PR (one PR, one logical change).

## Proposed direction

Add the worker's true generated-token count (and optionally its finish_reason) to the result/serving protocol frames, and have the router use those authoritative values for `usage.completion_tokens` / `finish_reason` instead of frame-counting. Then extend the E2E parity test to a byte-fallback tokenizer (e.g. a small Gemma checkpoint) to assert usage + finish_reason parity with single-node.

## Acceptance Criteria

- [ ] The router's `usage.completion_tokens` and `finish_reason` for `/v1/completions` and `/v1/chat/completions` match single-node for a byte-fallback tokenizer (e.g. Gemma), not just byte-level-BPE.
- [ ] The disaggregated wire protocol carries the worker's authoritative token count.
- [ ] An E2E test asserts the parity on a byte-fallback tokenizer.

## References

- Surfaced in PR #386 review (issue #200).
- Counting site and "Parity notes" are in `src/server/router_front.rs::route_completion`.
- Protocol: `src/distributed/disaggregated/serving_protocol.rs`.
- finish_reason: `src/server/model_worker.rs`.

inureyes:
## Summary

The disaggregated router derived `usage.completion_tokens` by counting emitted detokenized text pieces in its per-request `on_token` callback. That equals the worker's generated-token count for byte-level-BPE tokenizers (Qwen) but under-counts for byte-fallback tokenizers (Gemma), where a multi-byte character arrives as several `<0xXX>` model tokens but surfaces as one detokenized text piece (or none, for a byte-fallback first token). Because `finish_reason` derives from `completion_tokens >= max_tokens`, the under-count could also flip it between "length" and "stop", diverging from single-node.

This carries the worker's authoritative generated-token count over the serving wire protocol and has the router report it instead of frame-counting.

## What changed

- `src/distributed/disaggregated/serving_protocol.rs`: add `generated_tokens: Option<u64>` to `ResultFrame`, `#[serde(default)]` for backward compatibility (a node predating the field decodes to `None`; an older router ignores the unknown field).
- `src/distributed/disaggregated/coordinator.rs`: `drain_generation_events` now returns the worker's authoritative count from the terminal `GenerateEvent::Done`. The prefill node sets `generated_tokens` on its `FirstToken` frame (its true first-token count: 1 normally, 0 on immediate EOS); the decode node sets it on its terminal `Continuation` frame (its post-handoff count, accumulated across decode ticks). Intermediate continuation frames leave it `None`.
- `src/server/router_front.rs`: `drive_handoff_result` now returns a `HandoffOutcome` summing the per-node counts. `resolve_completion_tokens` prefers the authoritative total, clamps it to `max_tokens` (the router bounds generation to its own budget, so a larger reported count can only come from a buggy or hostile node), and falls back to the emitted-piece count when no frame carried one. `finish_reason` is derived from the authoritative count with the same `count >= max_tokens` formula the worker uses, so it matches single-node. Applies to `/v1/completions` (streaming and non-streaming) and `/v1/chat/completions` (the finish chunk and non-streaming body now carry the derived `finish_reason` instead of a hard-coded "stop").

## Backward compatibility

The wire change is additive and optional. In a mixed-version cluster, a frame from an older prefill or decode node arrives with `generated_tokens = None` (serde default), and the router transparently falls back to the previous frame-counting behavior. A new node's extra field is ignored by an older router.

## Test plan

- [x] `cargo check --lib --tests --features metal,accelerate` (exit 0)
- [x] `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` (exit 0)
- [x] `cargo test --lib --features metal,accelerate -- router_front::tests serving_protocol::tests` (15 passed): protocol round-trip with the new field present and absent; router resolution covering authoritative-vs-fallback selection, the `max_tokens` clamp, and the finish_reason-flip case.
- [ ] `tests/disaggregated_router_e2e.rs::disaggregated_router_completions_match_single_node_byte_fallback`: a new byte-fallback (Gemma) `/v1/completions` parity test asserting `usage.completion_tokens` and `finish_reason` match single-node. Gated behind `#[ignore]` plus a checkpoint-presence guard like the other real-model tests; not run in this environment (no Gemma checkpoint available). It requires a byte-fallback checkpoint that also supports the pool-backed paged handoff and must be run explicitly with `--ignored`.

Closes #387

### DIFF
--- a/src/distributed/disaggregated/coordinator.rs
+++ b/src/distributed/disaggregated/coordinator.rs
@@ -294,7 +294,17 @@ impl ServingCoordinator {
                 token_tx,
                 Arc::new(AtomicBool::new(false)),
             )?;
-            let (tokens, done, error) = drain_generation_events(&token_rx);
+            let drained = drain_generation_events(&token_rx);
+
+            // The prefill node's authoritative count of model tokens it
+            // generated for this request (issue #387). When it finished here
+            // (immediate EOS or a single-token budget) the terminal `Done`
+            // reports the exact count (0 or 1); when it produced a handoff and
+            // continues on the decode node, it sampled exactly one first token
+            // and emitted no `Done`, so the count is 1. Carrying this lets the
+            // router count a byte-fallback first token (which surfaces as no
+            // text piece) correctly instead of under-counting it.
+            let prefill_generated = drained.completion_tokens.unwrap_or(1);
 
             // Return the prefill node's first token to the client. When prefill
             // produced no handoff frame (immediate EOS), this first-token result
@@ -304,10 +314,11 @@ impl ServingCoordinator {
             let first_result = ResultFrame {
                 request_id: request.request_id,
                 phase: ResultPhase::FirstToken,
-                tokens,
+                tokens: drained.tokens,
                 start_sequence: 0,
-                done: done || frame.is_none(),
-                error,
+                done: drained.done || frame.is_none(),
+                error: drained.error,
+                generated_tokens: Some(prefill_generated),
             };
             if let Err(e) = self
                 .transport
@@ -358,6 +369,7 @@ impl ServingCoordinator {
                         start_sequence: 0,
                         done: true,
                         error: Some(format!("decode handoff to {decode_peer} failed: {e}")),
+                        generated_tokens: None,
                     };
                     let _ = self
                         .transport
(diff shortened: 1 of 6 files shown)


inureyes:
## Finalization complete

### Changes applied (commit bdc74f2)

**`src/server/router_front.rs`**

- `drive_handoff_result`: changed `generated_tokens.unwrap_or(0) + n` to `.saturating_add(n)` in the `add_count` closure. A hostile or buggy remote node reporting a very large frame count can no longer cause a debug-build overflow panic.
- `resolve_completion_tokens`: changed `.unwrap_or(frame_counted)` to `.unwrap_or(frame_counted.min(max_tokens))`. The fallback (frame-counted) branch now clamps to `max_tokens` the same way the authoritative path already does, making both branches uniformly defensive.

**`src/server/router_front_tests.rs`**

- Added `fallback_frame_count_clamped_to_max_tokens`: asserts that when no authoritative count is present and the emitted-piece count exceeds the budget, `resolve_completion_tokens` returns `max_tokens`, not the raw frame count.

**`docs/distributed.md`**: no change. The file's disaggregated-inference section describes topology and the security trust model; it does not document token-count or usage derivation, so there was nothing to update.

### Quality gate results

- `cargo clippy --lib --tests --features metal,accelerate -- -D warnings`: clean (Finished in 15.56s, no warnings)
- `cargo test --lib --features metal,accelerate -- router_front::tests`: 6/6 passed (5 pre-existing + 1 new)
- `cargo test --lib --features metal,accelerate -- serving_protocol::tests`: 10/10 passed
- `cargo fmt`: pre-existing failure on `router_front.rs` due to let-chain syntax (lines 368, 370, 604, 613) that rustfmt 1.8.0 cannot parse; this is not introduced by this commit. The test file (`router_front_tests.rs`) is format-clean per `rustfmt --check`.

PR #394, opened 2026-06-22
fix(server): preserve assistant reasoning field across turns

### LINKED ISSUE #362: fix(server): preserve assistant reasoning field across turns (interleaved reasoning)
## Problem / Background

mlxcel-server drops the assistant `reasoning` field on input, so prior-turn reasoning is lost in multi-turn "interleaved reasoning" conversations. Chat templates that render `message.get('reasoning')` (e.g. Gemma 4) never receive it, so the model cannot see its own prior thinking across turns.

### Current gap (verified by mlxcel code analysis)

- `src/server/types/request.rs:299-319`: the `Message` struct has no `reasoning` / `reasoning_content` field; serde silently drops it from incoming assistant turns.
- `src/server/chat_request.rs:325-341`: `build_raw_json_messages_with_thinking()` builds raw JSON messages (role / content / name / tool_call_id / tool_calls) but never forwards reasoning.
- `src/server/chat_template.rs:1268`: the Gemma 4 chat template already consumes `message.get('reasoning') or message.get('reasoning_content')`. The field is expected by the template but never supplied, so multi-turn reasoning context is lost.
- Output side is already complete: `src/server/types/stream.rs:23-38` (`Delta.reasoning_content`) and `ChatCompletionChunk::reasoning_content()`; Responses API `ReasoningTextDelta` / `ReasoningTextDone`. This is strictly an input-side fix.

## Proposed Solution

- Add `reasoning: Option<String>` to `Message`, accepting both `reasoning` and `reasoning_content` via serde alias for OpenAI compatibility.
- In `build_raw_json_messages_with_thinking()`, when `m.reasoning` is present, set it on the raw JSON message (`msg["reasoning"]`) so consuming templates render it.
- Define and document the interaction with `preserve_thinking` / rolling-checkpoint stripping (`src/server/chat_template_kwargs.rs:360-384`, `strip_rolling_checkpoint` / `strip_think_block`): when `preserve_thinking=false` (default), strip prior reasoning consistently (both inline `<think>` blocks and the parallel `reasoning` field) to avoid leaking or duplicating; when `preserve_thinking=true`, forward it. Avoid double-injection when content already contains a `<think>` block.
- Apply the same forwarding in the Anthropic Messages translator and the Responses API input mapping if they reconstruct assistant turns, for cross-API consistency.

## Acceptance Criteria

Must be integrated into the actual chat flow, not a standalone helper.

- [ ] A multi-turn request with an assistant message carrying `reasoning` renders that reasoning into the prompt via the template path, verified by a regression test asserting the reasoning field is preserved into the rendered prompt.
- [ ] Existing `preserve_thinking` behavior unchanged (regression tests pass).
- [ ] Wired into the real `/v1/chat/completions` request-building flow.
- [ ] `cargo test --release`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` all clean.

## References

- mlxcel anchors: `src/server/types/request.rs:299`, `src/server/chat_request.rs:325-341`, `src/server/chat_template.rs:1268`, `src/server/chat_template_kwargs.rs:360-384`, `src/server/types/stream.rs:23-38`.

inureyes:
## Summary

mlxcel-server dropped the assistant `reasoning` field on input, so prior-turn reasoning was lost in multi-turn interleaved-reasoning conversations. Chat templates that render `message.get('reasoning')` (e.g. Gemma 4) never received it. This is the input-side fix; the output side (streaming `reasoning_content`) was already complete.

## What changed

- `src/server/types/request.rs`: add `reasoning: Option<String>` to `Message`, accepting both `reasoning` and the OpenAI-compatible `reasoning_content` via `#[serde(alias = "reasoning_content")]`, omitted from output when absent.
- `src/server/chat_request.rs`: route requests carrying a non-empty reasoning field through the raw-JSON render path (the typed `ChatMessage` path only carries role + content). In `build_raw_json_messages_with_thinking`, forward `msg["reasoning"]` using the same decision applied to inline `<think>` blocks.
- `src/server/anthropic_translator.rs`: assistant `thinking` blocks are forwarded onto the parallel `reasoning` field instead of being dropped.
- `src/server/responses_translator.rs`: a Responses `reasoning` input item is buffered and attached to the following assistant turn's `reasoning` field; it does not leak across a turn boundary.
- Existing `Message` literals updated for the new field; tests added.

## preserve_thinking interaction and double-injection

- `preserve_thinking=false` (default): a prior assistant turn (before the rolling checkpoint) has both its inline `<think>` block and its parallel `reasoning` field dropped, so prior thinking is removed consistently and cannot leak into the next prompt.
- `preserve_thinking=true`: the reasoning field is forwarded for retained turns.
- Double-injection guard: when a retained turn's content already contains an inline `<think>` block, the parallel `reasoning` field is not also forwarded, so templates that render both channels never emit the same reasoning twice.

## Cross-API wiring

- Anthropic Messages and Responses translators both reconstruct assistant turns that reach the chat-template path, so both forward reasoning. The Anthropic `Text`-only message shape and tool-result messages have no reasoning to carry, so they are unaffected.

## Test plan

- [x] `cargo test --lib --features metal,accelerate -- server::chat_request:: server::types::request:: server::anthropic_translator:: server::responses_translator::` (109 passed)

New tests: multi-turn render asserts reasoning reaches the prompt when `preserve_thinking=true` and is stripped by default; double-injection guard; raw-path routing; serde `reasoning`/`reasoning_content` round-trip; Anthropic and Responses forwarding.

Closes #362

### DIFF
--- a/docs/responses-api.md
+++ b/docs/responses-api.md
@@ -101,8 +101,7 @@ Phase 1 supports these typed items:
 ]
 ```
 
-`developer` role is treated like `system`. Reasoning input items are accepted
-but are not fed back into the next prompt.
+`developer` role is treated like `system`. Reasoning input items are accepted and forwarded: the text content is buffered and attached to the parallel `reasoning` field of the following assistant turn. Chat templates that render `message.get('reasoning')` (such as Gemma 4) receive it there. The `preserve_thinking` kwarg controls whether the field survives the rolling-checkpoint strip: `false` (the default, unless the prompt cache is on) drops prior-turn reasoning along with any inline `<think>` blocks; `true` retains it.
 
 Message content parts reuse mlxcel's chat-completions content part types. This
 means `text`, `image_url`, `video_url`, and `input_audio` can deserialize, but
--- a/src/server/anthropic_translator.rs
+++ b/src/server/anthropic_translator.rs
@@ -85,6 +85,7 @@ pub fn anthropic_request_to_chat(request: &AnthropicRequest) -> AnthropicTransla
             content: MessageContent::Text(text),
             name: None,
             tool_call_id: None,
+            reasoning: None,
             tool_calls: None,
         });
     }
@@ -176,6 +177,7 @@ fn append_message(out: &mut Vec<Message>, message: &AnthropicMessage) {
                 content: MessageContent::Text(text.clone()),
                 name: None,
                 tool_call_id: None,
+                reasoning: None,
                 tool_calls: None,
             });
         }
@@ -184,6 +186,7 @@ fn append_message(out: &mut Vec<Message>, message: &AnthropicMessage) {
             let mut image_parts: Vec<ContentPart> = Vec::new();
             let mut tool_calls: Vec<ToolCallInMessage> = Vec::new();
             let mut tool_results: Vec<Message> = Vec::new();
+            let mut reasoning_parts: Vec<String> = Vec::new();
 
             for block in blocks {
                 match block {
@@ -228,9 +231,22 @@ fn append_message(out: &mut Vec<Message>, message: &AnthropicMessage) {
                             ));
                         }
                     }
-                    AnthropicContentBlock::Thinking { .. } | AnthropicContentBlock::Unknown => {
-                        // Dropped: thinking traces and unknown blocks do not
-                        // feed back into the next prompt.
+                    AnthropicContentBlock::Thinking { thinking } => {
+                        // Forward extended-thinking traces from assistant turns
+                        // into the parallel `reasoning` field (issue #362) so
+                        // templates that render `message.get('reasoning')` see
+                        // prior thinking across turns. Thinking on non-assistant
+                        // turns is meaningless and ignored.
+                        if message.role == AnthropicRole::Assistant
+                            && let Some(text) = thinking.as_ref()
+                            && !text.is_empty()
+                        {
+                            reasoning_parts.push(text.clone());
+                        }
+                    }
+                    AnthropicContentBlock::Unknown => {
+                        // Dropped: unknown blocks do not feed back into the
+                        // next prompt.
                     }
                 }
             }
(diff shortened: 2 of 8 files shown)


inureyes:
## PR Finalization

### MEDIUM fix applied

`src/server/responses_translator.rs` - `FunctionCallOutput` arm: the arm was already calling `pending_reasoning.take()` inside the tool-call flush block, but that block is skipped when `pending_tool_calls` is empty (malformed input: Reasoning item immediately followed by FunctionCallOutput with no preceding FunctionCall). The buffered reasoning then leaked onto the next assistant turn. Fixed by adding `pending_reasoning = None;` after the Tool message push, matching the pattern already used in the `Message` arm. In the normal Reasoning->FunctionCall->FunctionCallOutput flow the flush `take()` already consumes the reasoning, so this assignment is a no-op.

### Tests

Two new tests in `server::responses_translator::tests`:
- `reasoning_does_not_leak_when_function_call_output_has_no_preceding_function_call`: asserts `reasoning == None` on the assistant turn after the malformed sequence `[Reasoning, FunctionCallOutput(no preceding FunctionCall), Message(assistant)]`.
- `reasoning_attaches_to_function_call_turn_in_normal_tool_flow`: confirms the normal `[Reasoning, FunctionCall, FunctionCallOutput, Message(assistant)]` flow still attaches reasoning to the function-call assistant turn and not to the later text assistant turn.

73 tests passed in `server::responses_translator::` and `server::chat_request::`.

### Docs

`docs/responses-api.md`: corrected the stale claim that reasoning input items "are not fed back into the next prompt" to accurately describe the preservation behavior added by this PR.

### Lint / format

`cargo clippy --lib --tests --features metal,accelerate -- -D warnings`: clean. The fmt tool cannot parse pre-existing let-chains in this file (confirmed pre-existing before this PR); the new code is formatted consistently with the surrounding file.

Commit: `bd3bef703`

PR #400, opened 2026-06-22
feat: add video input support for Gemma 4 Unified (gemma4_unified)

inureyes:
## Summary

Add video input support to the encoder-free Gemma 4 Unified (`gemma4_unified`) runtime, replacing the previous "video input is not yet supported" gate. Video is handled as images-per-frame: frames are extracted with `ffmpeg` (uniform sampling, default 2.0 fps via the shared `multimodal::video` module), patchified through the existing patch-projection vision embedder, and the per-frame soft tokens scatter into `video_token_id` placeholder spans, mirroring the image scatter. Works on both the CLI (`--video`) and the server (`video_url` content blocks).

## What changed

- **Processor (`src/vision/processors/gemma4_unified.rs`)**: factored the image patch loop into a shared `patchify(image, soft_token_cap)` and added `preprocess_video_frames`, which caps each frame at `vision_soft_tokens_per_video_frame` (70 by default) instead of the larger per-image budget (280). `resize_dims` now takes the cap as a parameter so image and video share one loop. New `DEFAULT_VIDEO_SOFT_TOKENS_PER_FRAME` constant and `set_video_soft_tokens_per_frame` setter.
- **Config (`src/vision/gemma4_unified_config.rs`)**: new `vision_soft_tokens_per_video_frame` field (default 70), wired into the processor by the loader (`src/loading/vlm_gemma_unified.rs`).
- **Model (`src/vision/gemma4_unified.rs`)**: factored the per-input project/trim/concat into `project_vision_features`; added `get_video_features` and `get_input_embeddings_with_video`; the image/video/audio scatters now run on the running embeddings inside a shared `merge_multimodal` core. The image/audio/text paths run the identical op sequence as before, so their output is unchanged.
- **Token expansion (`src/multimodal/vlm_runtime.rs`)**: new `expand_gemma4_unified_video_tokens` frames each frame as `<boi> video_token*N <eoi>` and either replaces a `<|video|>` placeholder in prompt order or splices after BOS when absent. Per-frame soft tokens use `video_token_id` (not `image_token_id`), matching the encoder-free scatter target.
[...]

Issue #164, opened 2026-06-05
feat: video input support for Gemma 4 Unified (gemma4_unified)


inureyes:
Add video input support to the Gemma 4 Unified (`gemma4_unified`) runtime. The architecture (added in #151 / #153) supports text + image + audio, but **video input is gated behind an explicit "not yet supported" error** and was an intentional deferral.

## Current state

`src/server/model_worker.rs:947` rejects video for `gemma4_unified`:

```
"video input is not yet supported for gemma4_unified models; ..."
```

The model config already carries the video plumbing: `video_token_id` (258884), `vision_soft_tokens_per_video_frame` (70). The encoder-free vision embedder and the image path already exist; video is "images per frame" with `video_token_id` placeholders.

## Goal

Accept `--video` (and the server's video content parts) for `gemma4_unified`: extract frames (uniform FPS resample via ffmpeg, as the existing VLM video path does), patchify each frame through the existing encoder-free vision embedder, and scatter the per-frame soft tokens into `video_token_id` placeholder spans.

## Touchpoints

- `src/vision/processors/gemma4_unified.rs` — frame extraction + per-frame patchify (reuse the image patchify; `vision_soft_tokens_per_video_frame`).
- `src/vision/gemma4_unified.rs` — `get_video_features` / merge into `video_token_id` placeholders (mirror the image scatter).
- `src/commands/generate_vlm.rs` / `src/server/model_worker.rs` — remove the gate, wire the video input path.
- Reuse the bidirectional blockwise mask over image/video spans (video tokens are type 2).

## Acceptance criteria

- `mlxcel generate -m mlx-community/gemma-4-12b-it-4bit --video clip.mp4 -p "Describe this video."` produces a coherent, grounded description.
- Text / image / audio / image+audio paths are unaffected.
- Server video content parts work through `/v1/chat/completions`.
- Frame-extraction FPS matches the existing VLM video convention (default 2.0 fps).

PR #405, opened 2026-06-23
fix(models): correct gemma sliding-window prefill beyond the window

### LINKED ISSUE #401: fix(models): gemma4 sliding-window single-pass prefill degenerates to <pad> beyond the window
## Problem / Background

For Gemma 4 sliding-window models (`gemma4_unified` and `gemma4_vl`), a single-pass prefill whose length exceeds the model's sliding window (1024 tokens for `gemma-4-12b-it-4bit`) degenerates. The rotating KV cache evicts keys that the earliest query rows still need, leaving an all-masked attention row. That row propagates NaN and the model emits `<pad>` output.

The failure is modality-independent. It reproduces with:

- a coherent text-only prompt of ~1700 tokens,
- the existing gemma4 VLM video path (`gemma4_vl`), and
- the new `gemma4_unified` video path added in #400.

This was surfaced while implementing #164 (video input for `gemma4_unified`, merged as PR #400).

### Concrete trigger from #164

The literal acceptance command from #164:

```bash
mlxcel generate -m mlx-community/gemma-4-12b-it-4bit --video clip.mp4 -p "Describe this video."
```

At the default `2.0` fps, a multi-second clip expands into ~20 frames (~1377 tokens), which exceeds the 1024-token sliding window, so the prefill degenerates and the output is `<pad>`.

The current user-side mitigation is to lower `--fps` (for example `--fps 0.6`) so the expanded prompt fits inside the window. This is documented in `docs/supported-models.md` as a known limitation.

### Relationship to PR #400 (distinct bug, already fixed)

PR #400 already fixed a related but distinct bug. `overlay_block_bidirectional` in `src/models/gemma4.rs` used to **panic** (\"Cannot reshape array of size N into shape (1, window)\") when the prefill exceeded the window. It now aligns block ids to the trailing windowed keys, so the panic is gone.

This issue is about the **remaining numerical degeneration** (NaN -> `<pad>`) caused by rotating-cache eviction during a single-pass prefill. The panic fix did not address the eviction-of-needed-keys problem.

## Proposed Solution

Make long-prompt prefill numerically valid for sliding-window Gemma 4 models. Likely direction:

- **Chunked prefill** for sliding-window Gemma 4: process the prompt in window-sized chunks so the rotating cache never drops keys that an in-flight query row still needs. Reuse the existing chunked-prefill machinery in the codebase where applicable rather than adding a parallel path.
- Or an equivalent approach that keeps a >window single-pass prefill from leaving all-masked attention rows.

## Acceptance Criteria

- [ ] A prompt longer than the sliding window (text, image, or video) on `mlx-community/gemma-4-12b-it-4bit` produces coherent output instead of `<pad>`.
- [ ] The default-fps video command from #164 (`mlxcel generate -m mlx-community/gemma-4-12b-it-4bit --video clip.mp4 -p \"Describe this video.\"`) works without manually lowering `--fps`.
- [ ] No regression for prefills at or below the window length.
- [ ] Update / remove the known-limitation note in `docs/supported-models.md` once the fix lands.

## Technical Considerations

- Affects `gemma4_unified` and `gemma4_vl`; the root cause is in the sliding-window rotating KV cache path (`src/models/gemma4.rs`), so the fix should be shared across both.
- The bug is independent of input modality; a text-only repro is the simplest regression case and avoids needing a video fixture.
- Origin: follow-up to #164 / PR #400.

inureyes:
## Summary

For Gemma 4 sliding-window models (`gemma4_unified` and `gemma4_vl`), and identically for Gemma 3, a single-pass prefill whose length exceeds the sliding window (1024 for `gemma-4-12b-it-4bit`) degenerated to `<pad>`. This makes the default-fps video path from #164 (20 frames, ~1377 tokens > 1024) and any ~1500+ token text/image/video prompt produce garbage. The fix makes the windowed prefill correctness come from the attention mask rather than from physically dropping keys, mirroring mlx-lm's `RotatingKVCache`.

## Root cause

The `RotatingKVCache` keeps every key on the first prefill pass (`update_concat` returns all keys when `self.keys.is_none()`) and only trims to the window on the subsequent decode step (`update_in_place`). But the sliding mask was built with `create_causal_mask_with_window`, which caps the key axis to `window` whenever `seq_len + offset > window`, on the assumption that the cache only ever returns the most recent `window` keys. For a single-pass prefill longer than the window that assumption is false. The capped `[seq_len, window]` mask is then discarded by gemma4's `trim_mask_to_keys` (mask shorter than the full key axis), so `attend` falls back to `causal_attention`, which slices K/V down to the trailing `window` keys (`start = k_len - window_size`). The earliest query rows (logical position `< seq_len - window`) then have no visible key at all: an all-masked attention row that softmaxes to NaN and emits `<pad>`. For Gemma 3 (whose attention passes the mask directly) the same capped mask mismatches the full key axis and degenerates to immediate EOS. PR #400 had already fixed the separate `overlay_block_bidirectional` reshape panic; this is the remaining numerical degeneration.

## The fix and why

New shared helper `create_causal_mask_with_window_full(size, offset, window)` in `mlxcel-core::utils` builds the uncapped `[size, size + offset]` windowed-causal mask (causal lower-triangular intersected with the sliding-window lower bound), so every query row attends to its own window over the full key set the rotating cache returns during prefill. `trim_mask_to_keys` still crops it on any later capped fetch, so the change is robust for cache rollover too. Both `gemma4.rs` mask sites (`forward_with_speculative_sinks` and the pipeline-stage `execute_hidden`) and both `gemma3.rs` sites now route through a small `sliding_prefill_mask` helper that uses the full mask only when `seq_len > window` and keeps the existing clamped `create_causal_mask_with_window` path verbatim for `seq_len <= window`. That preserves byte-identical greedy output at or below the window (including the batched MTP verify rounds, which always have `l <= window`), and the full mask is mathematically the uncapped form of the same windowed-causal mask, so its last `window` columns equal the old capped mask. This reuses the existing rotating-cache + `trim_mask_to_keys` machinery rather than adding a parallel chunked-prefill path, and it preserves the `gemma4_unified` vision-span bidirectional overlay (the overlay now runs on the clean `k == id_len` branch).

## Validation matrix (M1 Ultra, real checkpoints, `--temp 0`)

1) gemma-4-12b-it-4bit, text-only >window prefill (1511 tokens):

BEFORE: `...Be specific and accurate.<pad><pad><pad>...` (all `<pad>`).

AFTER: `John Smeaton's Eddystone lighthouse was a significant engineering advancement because he modeled its shape on an oak tree trunk, which directed the force of waves downward into the rock rather than sideways against the masonry. He also pioneered the use of hydraulic lime, a mortar that sets underwater, and designed granite blocks that dovetailed into one another and the bedrock to lock the structure together as a single mass.`

2) gemma-4-12b-it-4bit, the default-fps #164 video command `generate -m models/gemma-4-12b-it-4bit --video <clip>.mp4 -p "Describe this video." -n 100` (loader reports `20 total frames`, `1377 total tokens` > 1024, no `--fps` lowering):

BEFORE: `Describe this video.<pad><pad><pad>...` (all `<pad>`).

AFTER: `The video shows a parking lot with several cars. A black car is parked in the foreground, and a silver car is parked next to it. A red car is parked in the background. The black car is moving forward and hits the silver car. The silver car is pushed back slightly. The black car then continues to move forward and comes to a stop...`

3) Regression at/below the window (must be unchanged): generated text is byte-identical before vs after (only the non-deterministic model-load and tok/s timing lines differ in the captured stdout).

gemma-4-12b-it-4bit short text: `The sky appears blue because Earth's atmosphere scatters shorter wavelengths of sunlight, such as blue and violet, more effectively than longer wavelengths. This phenomenon, known as Rayleigh scattering, occurs as sunlight interacts with gas molecules and small particles in the air.` (identical before/after).

gemma-4-12b-it-4bit single image (cats.jpg, 288 tokens): `Two tabby cats are lying on a pink couch, with one cat on the left and another on the right.` (identical before/after).

4) gemma3-4b-4bit, >window text prefill (1511 tokens):

BEFORE: empty / degenerate (no generated text after the prompt).

AFTER: `John Smeaton's Eddystone lighthouse was a significant advance because he fundamentally altered the structure's design to withstand the powerful forces of the sea. He modeled the tower's shape on a tree trunk, ensuring that wave energy was directed downwards into the bedrock rather than impacting the masonry directly. Furthermore, he pioneered the use of hydraulic lime and precisely cut granite blocks that dovetailed together... Augustin Fresnel's key optical innovation was the development of a series...`

gemma3-4b-4bit short text (<=window) is byte-identical before vs after.

## Tests

- `mlxcel-core` utils: `full_windowed_mask_over_window_has_no_all_masked_row` (the [4,4] mask for window 2 has no fully-masked row, vs the capped [4,2] that strands rows 0 and 1) and `full_windowed_mask_within_window_matches_capped` (cell-for-cell equality for `size + offset <= window`).
- `mlxcel` gemma4: `sliding_prefill_mask_over_window_has_no_fully_masked_row` and `sliding_prefill_mask_within_window_matches_clamped`.
- Suites green: gemma4 (146), gemma3 (18), diffusion_gemma (33), core cache (435), core utils (27). `cargo fmt --check` clean; `cargo clippy -p mlxcel-core -p mlxcel --lib --tests -- -D warnings` clean.

## What changed

- `src/lib/mlxcel-core/src/utils.rs`: add `create_causal_mask_with_window_full` + two unit tests.
- `src/models/gemma4.rs`: add `sliding_prefill_mask` helper, route both prefill mask sites through it.
- `src/models/gemma3.rs`: add `sliding_prefill_mask` helper, route both prefill mask sites through it.
- `docs/supported-models.md`: replace the `gemma4_unified` known-limitation note (long clips needing lower `--fps`) with the fixed behavior.

## Not regressed

`diffusion_gemma` (shares `overlay_block_bidirectional`) and `gemma3` are not regressed: the `<= window` path is unchanged (proven byte-identical by test and by real-model output), and the over-window path is the strict mlx-lm-equivalent fix. Decode (`l == 1`) is untouched.

Closes #401


inureyes:

### Intent
> Fix Gemma 4 (`gemma4_unified` / `gemma4_vl`) and Gemma 3 single-pass prefill degenerating to `<pad>` when the prompt exceeds the sliding window, by making windowed-prefill correctness come from the attention mask instead of dropping keys.

### Verification (all focus areas confirmed)

- [x] **All stated requirements implemented.** New `create_causal_mask_with_window_full` is the uncapped form of the existing windowed-causal mask (`tril(offset)` ∧ `triu(offset-w+1)`, `where_cond` to 0/-inf). Both Gemma 4 prefill sites (`forward_with_speculative_sinks` non-ragged `l > 1` branch, `Gemma4StageModel::execute_hidden`) and both Gemma 3 sites (`Gemma3Model::forward`, `Gemma3StageModel`) route through `sliding_prefill_mask`.
- [x] **`seq_len > window` gating is correct and the `<= window` path is byte-identical.** The gate exactly matches the cap condition in `create_causal_mask_with_window` (for prefill `offset == 0`, the old code capped iff `l > w`). For `l <= window` the helper reproduces the prior code verbatim (`sliding_offset.min((window - l).max(0))` + same `create_causal_mask_with_window` call), so output is unchanged. The two new unit tests per layer (`*_over_window_has_no_fully_masked_row`, `*_within_window_matches_clamped`) lock both halves.
- [x] **Prefill -> first-decode boundary aligns.** The mask change does not alter cache behavior (the rotating cache always kept all prefill keys and trims only on the first decode step). Fresh prefill: full `[l, l]` mask matches the full key axis, so `trim_mask_to_keys` is a no-op and `attend` uses the explicit mask instead of the `causal_attention` K/V-slicing fallback. On any later capped fetch the trailing-`window` slice of the full mask equals the old capped mask (verified algebraically: in the trailing columns the window lower bound `triu` is trivially satisfied, leaving only `tril`), so `trim_mask_to_keys` keeps it aligned. Decode (`l == 1`) passes `None` and is untouched.
- [x] **Vision-span bidirectional overlay runs on the clean `k == id_len` branch.** For unified video/image prefill the VLM passes `mask = None` + `bidirectional_block_ids = Some`; the full `[l, l]` mask (offset 0) has `k == id_len == l`, so `overlay_block_bidirectional` takes the clean branch. The PR #400 windowed-cap (`k < id_len`) branch is retained for cache-rollover safety but is no longer exercised for over-window prefill (no contradiction).
- [x] **diffusion_gemma path unchanged.** It builds its own `dense_windowed_causal_mask` and calls the (unmodified) `overlay_block_bidirectional`; it does not use `create_causal_mask_with_window` or `sliding_prefill_mask`. The PR only adds a new utils function plus two private per-model helpers, so the diffusion path is byte-identical.
- [x] **MTP / speculative routes consistently.** Single-sequence MTP prefill flows through the same `l > 1` branch (`sliding_prefill_mask`). Batched ragged MTP is admitted only when `max_prompt_len <= sliding_window` (gated via `sliding_window_value`), and verify rounds always have `l <= window`, so neither hits the over-window cap. This matches the byte-identical `<= window` claim.
- [x] **Conventions followed.** `// Used by:` present on the new shared helper; no AI attribution; `docs/supported-models.md` note accurately describes the fixed behavior.
- [x] **Tests pass** (per orchestrator: core utils 27, core cache 435, gemma4 146, gemma3 18, diffusion_gemma 33; build + clippy clean) and real-checkpoint validation on M1 Ultra (text/video over-window coherent; <=window byte-identical).

### Remaining Items (non-blocking, LOW / informational)
- [ ] (LOW, reuse) `sliding_prefill_mask` is duplicated near-identically in `gemma4.rs` and `gemma3.rs`; it could live once in `mlxcel-core::utils` beside `create_causal_mask_with_window_full`. Tiny (5 lines), optional.
- [ ] (INFO, out of scope) Other sliding-window models that use the capped `create_causal_mask_with_window` (cohere2, gpt_oss, ministral3, exaone4, olmo3, gemma3n, mellum, step3p5) may share the same latent over-window single-pass prefill weakness depending on their attention path. Out of #401 scope; worth a follow-up audit.

No CRITICAL or HIGH findings. The fix is correct, complete, well-tested, and properly integrated; no auto-fix required.

inureyes:
## Follow-up: HIGH finding found and fixed (Codex-assisted)

The supplementary Codex review surfaced a real defect that my first pass missed, and on verification it is a reachable, PR-introduced crash. It is now fixed on this branch.

### Finding (HIGH): gemma3 over-window mask crashes on a non-fresh multi-token pass

My earlier Focus-2 note claimed `trim_mask_to_keys` keeps the mask aligned on cache rollover. That is true for Gemma 4 but NOT for Gemma 3: `Gemma3` `Attention::forward` (src/models/gemma3.rs:205-216) passes the mask straight to `attention_from_ptr` with no trim step.

Confirmed mechanism:
- `RotatingKVCache::update_concat` (src/lib/mlxcel-core/src/cache.rs:4203-4218) trims the returned K/V to `max_size` (= `window`) on a NON-fresh multi-token append (`keys.is_some()` and `total_len > max_size`). On a fresh prefill (`keys.is_none()`, cache.rs:4187-4192) it keeps all `seq_len` keys.
- When `seq_len > window` AND `sliding_offset > 0`, the old `sliding_prefill_mask` built a `[seq_len, seq_len + sliding_offset]` mask while the cache returned only `window` keys, so it could not broadcast against the `[B, H, seq_len, window]` scores: shape error / crash.
- Reachable via server chunked prefill when `--batch-size` / `--ubatch-size` / `prefill_chunk_size` exceeds the sliding window (not clamped to it, see `resolve_prefill_chunk_size`), or multi-turn cache reuse feeding a >window chunk. The validated default paths (CLI single-pass with `offset == 0`, server default chunk 512 < 1024) are unaffected, which is why validation passed. Gemma 4 was never exposed because `trim_mask_to_keys` re-crops the full mask to the window (and the trailing-window slice of the full mask equals the clamped mask).

### Fix (commit 4cb2c18a1)

`src/models/gemma3.rs`: gate the full mask on a fresh prefill only.

```rust
-    if seq_len > window {
+    if seq_len > window && sliding_offset == 0 {
```

For `seq_len > window && sliding_offset > 0` the else branch's `effective_offset` collapses to `0`, producing the clamped `[seq_len, window]` mask, which both broadcasts against the window-trimmed cache and carries the correct windowed-causal values (the offset cancels in the relative query/key relation). Gemma 4 is intentionally left unchanged (protected by `trim_mask_to_keys`). Doc comment updated; two regression tests added (`[4,0,2] -> [4,4]` fresh, `[4,3,2] -> [4,2]` non-fresh, the assertion that catches the crash).

### Re-verification (narrow selectors)
- `cargo build --release --features metal,accelerate`: clean.
- `cargo test --release -p mlxcel gemma3`: 20 passed (18 prior + 2 new), 0 failed.
- `cargo fmt --check`: clean. `cargo clippy -p mlxcel --lib --tests --features metal,accelerate -- -D warnings`: clean.

### Updated verdict
The original issue #401 fix (over-window single-pass prefill, `offset == 0`) is preserved and correct for both Gemma 3 and Gemma 4. With this follow-up, the Gemma 3 non-fresh over-window path no longer crashes. No remaining CRITICAL/HIGH findings.

inureyes:
## PR Finalization Complete

### Summary

**Docs** (`docs/supported-models.md`): verified. The implementer already replaced the stale "lower --fps / \`<pad\>`" caveat with the correct description ("over-window single-pass prefill is handled correctly... long clips decode coherently without lowering --fps"). No further change needed.

**Doc comments** (two LOW items from review):
- `gemma3.rs` `sliding_prefill_mask`: tightened the trailing sentence to state the exact gate (`seq_len > window && sliding_offset == 0`) rather than only the `seq_len <= window` case, which omitted the non-fresh over-window (`sliding_offset > 0`) path.
- `utils.rs` and `gemma4.rs`: removed two em dashes from doc comments (project style prohibition; the reviewer removed one, these were two remaining).

**Tests**: added `mask_at` helper and diagonal assertion (`mask_at(q, q) == 0.0` for all rows) to `gemma3_mask_tests::sliding_prefill_mask_fresh_over_window_spans_full_key_axis`, bringing it to parity with gemma4's equivalent test.

**Lint/format**: `cargo fmt -p mlxcel-core -p mlxcel` clean; `cargo clippy --features metal,accelerate -p mlxcel-core -p mlxcel --lib --tests -- -D warnings` passed with no warnings.

No em dashes remain anywhere in the PR diff. Branch is pushed and clean.

Issue #410, opened 2026-06-23
refactor: hoist duplicated sliding_prefill_mask helper from gemma3/gemma4 to a shared module


inureyes:
## Problem / Background

PR #405 (fix for #401) added a `sliding_prefill_mask(seq_len, sliding_offset, window)` helper to **both** `src/models/gemma3.rs` and `src/models/gemma4.rs`. The helper wraps the new `create_causal_mask_with_window_full` (in `src/lib/mlxcel-core/src/utils.rs`): it selects the uncapped full windowed mask in the over-window prefill case and otherwise falls through to the existing clamped `create_causal_mask_with_window` path (with `effective_offset = sliding_offset.min((window - seq_len).max(0))`).

Both the pr-reviewer and pr-security-checker flagged the duplication as a LOW, non-blocking note: the helper should live in a single shared location so the gating logic is defined once. This matters because the gate encodes a subtle correctness invariant — the `sliding_offset == 0` condition that avoids a shape mismatch with the trimmed rotating cache — and two independent copies will drift.

**They already have.** The two copies are not byte-identical today:

- `src/models/gemma3.rs` (lines 72-79):
  ```rust
  fn sliding_prefill_mask(seq_len: i32, sliding_offset: i32, window: i32) -> UniquePtr<MlxArray> {
      if seq_len > window && sliding_offset == 0 {
          create_causal_mask_with_window_full(seq_len, sliding_offset, Some(window))
      } else {
          let effective_offset = sliding_offset.min((window - seq_len).max(0));
          create_causal_mask_with_window(seq_len, effective_offset, Some(window))
      }
  }
  ```
- `src/models/gemma4.rs` (lines 2088-2095):
  ```rust
  fn sliding_prefill_mask(l: i32, sliding_offset: i32, window: i32) -> UniquePtr<MlxArray> {
      if l > window {
          create_causal_mask_with_window_full(l, sliding_offset, Some(window))
      } else {
          let sliding_effective_offset = sliding_offset.min((window - l).max(0));
          create_causal_mask_with_window(l, sliding_effective_offset, Some(window))
      }
  }
  ```

[...]

PR #412, opened 2026-06-23
fix: harden sliding-window over-window prefill across models

### LINKED ISSUE #408: fix: harden sliding-window >window prefill across non-Gemma models; guard causal_attention windowed K/V slice
## Problem / Background

PR #405 (issue #401) fixed Gemma 3/4 single-pass prefill degenerating to `<pad>` when `seq_len > window`. The root cause chain:

- `create_causal_mask_with_window` (in `src/lib/mlxcel-core/src/utils.rs`) CAPS the key axis to `window` when `seq_len > window`, producing a `[seq_len, window]` mask.
- A fresh `RotatingKVCache` prefill keeps ALL keys, so gemma4's `trim_mask_to_keys` discards the capped mask (it does not match the full key count).
- With the mask gone, `attend` falls back to `causal_attention` (in `src/lib/mlxcel-core/src/layers.rs`). Its `needs_window_mask` path slices K/V to the trailing `window` keys (`start = k_len - window_size`).
- That trailing-window slice is only correct for single-query decode. For a multi-token prefill (`q_len > 1`, `k_len > window`) it strands the earliest query rows (logical position `< seq_len - window`) with no visible key, producing an all-masked row that softmaxes to NaN.

The #405 fix introduced `create_causal_mask_with_window_full` and a per-model `sliding_prefill_mask` helper, but wired it into `gemma3.rs` and `gemma4.rs` ONLY. The underlying `causal_attention` trailing-window K/V slice remains a latent footgun for ANY other caller that passes `q_len > 1`, `k_len > window`, and a discarded/short mask.

PR #405's review also found and fixed a related HIGH severity bug (gemma3 over-window with `sliding_offset > 0` shape mismatch), so the corrected gate is `seq_len > window && sliding_offset == 0`. Use that as prior art for any generalized fix.

## Proposed Solution

Two concrete asks:

1. Add a defensive guard in `causal_attention` (`src/lib/mlxcel-core/src/layers.rs`): when `q_len > 1 && k_len > window` and the explicit mask was discarded, either build the correct windowed mask over the full key set (generalizing the #401 fix so it protects all models) or fail loudly instead of silently stranding query rows. The single-query decode path must stay unchanged.
2. Audit the other sliding-window models in the tree that rely on the clamped `create_causal_mask_with_window` (search for its callers and `RotatingKVCache` users beyond gemma3/gemma4) and confirm whether they can hit the same `>window` single-pass prefill degeneration; fix or document each. Likely candidates include any model combining sliding-window attention with a rotating cache.

## Acceptance Criteria

- [ ] A `>window` single-pass prefill on every sliding-window model in the tree produces coherent output (no `<pad>` / NaN), OR is explicitly documented as unsupported with a guard that fails loudly rather than silently degenerating.
- [ ] The single-query decode path through `causal_attention` is unchanged (existing decode parity tests still pass).
- [ ] A regression test covers the multi-token `q_len > 1, k_len > window` case at the `causal_attention` / mask layer.
- [ ] All sliding-window callers of `create_causal_mask_with_window` / `RotatingKVCache` are audited; each is either fixed or has a documented guard.

## Technical Considerations

- Generalized fix should reuse `create_causal_mask_with_window_full` (added by #405) rather than introduce a parallel mask builder.
- The gate must follow the #405 prior art: `seq_len > window && sliding_offset == 0` (the over-window `sliding_offset > 0` shape-mismatch case was the related HIGH fixed in #405's review).
- Decode (`q_len == 1`) must keep the trailing-window K/V slice fast path; only the multi-token prefill case needs the corrected mask / loud failure.

inureyes:
## Summary

A single-pass prefill longer than a model's sliding window degenerated to NaN / `<pad>` (or hard-crashed on a mask/key shape mismatch) on every sliding-window model except gemma3/gemma4. When `seq_len > window`, a fresh `RotatingKVCache` (and a dense `KVCache`) keep ALL prefill keys, but the shared `causal_attention` wrapper and the per-model prefill mask builders capped the windowed mask to `[seq_len, window]` and sliced K/V to the trailing `window` keys. That strands the earliest query rows (logical position `< seq_len - window`) with an all-`-inf` mask row that softmaxes to NaN; models that pass full K/V with a capped mask hit a hard `[broadcast_shapes]` crash instead. PR #405 fixed only gemma3/gemma4.

## Chosen approach: general fix + shared helper (no loud-guard fallback needed)

The general `causal_attention` fix landed cleanly and is bit-exact for every currently-correct path, so the loud-guard fallback was not required.

1. General fix in `causal_attention` (`src/lib/mlxcel-core/src/lib.rs`): for a multi-token prefill that exceeds the window (`q_len > 1 && k_len > window`), build the full-width windowed-causal mask over ALL keys via `create_causal_mask_with_window_full` instead of slicing K/V. The single-query decode path (`q_len == 1`) keeps the trailing-window K/V slice byte-for-byte; non-windowed callers (`window_size == 0`) and within-window prefills are untouched. This is the shared safety net and directly fixes baichuan (which reaches this path with `mask.is_none()`), and incidentally also fixes the tensor-parallel gemma4 path (which routes through `causal_attention`).

2. Shared helper `create_sliding_window_prefill_mask(size, sliding_offset, window)` (`src/lib/mlxcel-core/src/utils.rs`): selects the uncapped full mask for a fresh over-window prefill (`size > window && sliding_offset == 0`, the gemma3 prior-art gate) and the clamped mask otherwise. Its else-branch is byte-identical to the per-model `sliding_offset.min((window - size).max(0))` construction it replaces (the capped tril offset `window - size` is offset-independent, and the helper's non-capped triu is a no-op for `size <= window`), so greedy temp-0 output is unchanged for every non-degenerate case.

3. Per-model wiring of the helper into the prefill mask builders (see audit table). gemma3/gemma4 keep their #405 `sliding_prefill_mask` helpers as-is; #410 will consolidate them onto the shared helper.

4. Tensor-parallel gemma3 path (`src/distributed/tensor_parallel/llama_runtime.rs`, `gemma3_masks`): the TP path independently computed its own clamped offset and called `create_causal_mask_with_window`, so a TP gemma3 >window prefill hit the same NaN/crash as the single-process path. Replaced with `create_sliding_window_prefill_mask`, mirroring the single-process fix. TP gemma4 was incidentally fixed by the general `causal_attention` change in item 1.

## Per-model audit (every `RotatingKVCache` / `create_causal_mask_with_window` sliding-window caller)

| Model | Cache | Pre-fix `>window` prefill | Fix applied | How validated |
|-------|-------|----------------|-------------|---------------|
| gpt_oss | RotatingKVCache, full K/V | CRASH (mask `[N,128]` vs scores `[..,N,N]`) | helper at both mask sites (forward + pipeline stage) | real model, window 128 |
| mellum | RotatingKVCache, full K/V | CRASH (capped mask vs full K/V) | helper at prefill mask builder | unit test + code analysis (no local ckpt) |
| exaone4 | RotatingKVCache, full K/V (bf16 cast) | CRASH | helper (astype preserved) | unit test + analysis (local ckpt not sliding) |
| exaone_moe | RotatingKVCache, full K/V | CRASH | helper | unit test + analysis (no local ckpt) |
| ministral3 | RotatingKVCache, full K/V | CRASH | helper | unit test + analysis (local ckpt not sliding) |
| step3p5 | RotatingKVCache, full K/V | CRASH | helper | unit test + analysis (no local ckpt) |
| cohere2 | dense KVCache + model-level K/V slice | NaN (capped mask + trailing slice strands early rows) | helper + slice to mask key-axis (not blindly `window`) | unit test + analysis (window 4096, not quick to exceed) |
| gemma3n | dense KVCache + model-level K/V slice | NaN | helper at both mask sites + slice to mask key-axis | unit test + analysis |
| olmo3 | dense KVCache + model-level K/V slice | NaN (capped mask + trailing slice strands early rows) | helper + slice to mask key-axis (not blindly `window`) | unit test + analysis |
| baichuan | dense KVCache, prefill via `causal_attention(mask=None)` | NaN / shape issue | covered by the general `causal_attention` fix (no model edit) | unit test + analysis |
| tensor-parallel gemma3 | RotatingKVCache via `gemma3_masks` in `llama_runtime.rs` | NaN / CRASH (same clamped construction as single-process) | `create_sliding_window_prefill_mask` in `gemma3_masks` | logit-match test (`tensor_parallel_gemma3_matches_full_model_logits`) |
| tensor-parallel gemma4 | routes through `causal_attention` | NaN / CRASH | incidentally fixed by the general `causal_attention` fix (item 1 above) | covered by the general `causal_attention` fix |
| recurrent_gemma | KVCache, full causal mask, `window_size = 0` | unaffected (never caps; window not mask-enforced) | none | code analysis |
| diffusion_gemma | dense KVCache, `dense_windowed_causal_mask` (uncapped) | unaffected (already full windowed mask) | none | code analysis |
| gemma3 / gemma4 | RotatingKVCache, `sliding_prefill_mask` | already fixed by #405 | untouched | real model (gemma-4-12b) + existing tests |

ministral-3b-4bit and exaone4-1.2b-4bit ship with `sliding_window: null` (no windowed layers), so those local checkpoints do not exercise the windowed path; their fix rests on the unit test, the byte-identical helper proof, and the gpt-oss real-model result for the identical code shape.

## Real-model validation (M1 Ultra, `--temp 0`)

Bug existed on `main` (HEAD 346018a72), gpt-oss-20b-mxfp4, 298-token prompt (window 128):

```
exit=134
libc++abi: terminating due to uncaught exception of type std::invalid_argument:
[broadcast_shapes] Shapes (298,128) and (1,64,298,298) cannot be broadcast.
```

Same prompt on this branch:

```
...<|channel|>final<|message|>The bakery sits on the second hill, and the clock tower rings thirteen times at noon.
[Generated 90 tokens in 1.57s = 57.36 tok/s]
```

gemma-4-12b-it-4bit, ~1600-token prompt (window 1024), this branch (#405 not regressed):

```
The bakery is located on the second hill. At noon, the clock tower rings thirteen times as a joke that began two centuries ago.
```

Regression (byte-identical `main` vs this branch, generated text diffed):

- gpt-oss-20b short prompt (`<128`, prefill + decode): BYTE-IDENTICAL
- gemma-4-12b short prompt ("The capital of France is **Paris**."): BYTE-IDENTICAL
- qwen3-0.6b non-windowed ("Mercury / Venus / Earth"): BYTE-IDENTICAL

This confirms decode (`q_len == 1`), non-windowed (`window_size == 0`), and within-window prefill paths are unchanged; only the previously-degenerate `>window` prefill changed (crash/NaN -> coherent).

## Tests (fast, no model loads)

- `utils::tests::sliding_window_prefill_mask_selects_full_when_fresh_over_window` / `_clamps_when_rolled_over` / `_within_window_matches_capped_builder` (the selector's three regimes).
- `layers::tests::causal_attention_multi_token_over_window_uses_full_windowed_mask` (the `q_len > 1, k_len > window` case: finite, matches the full windowed-mask reference; the prior clamp+slice path produced NaN). Updated from the old test that asserted the now-removed sliced behavior.
- Existing decode/window tests, all `utils::tests` (30) and `layers::tests` (28), and gemma3/gemma4 `sliding_prefill_mask` tests still pass.
- `tensor_parallel_gemma3_matches_full_model_logits` validates the TP gemma3 mask path.

## Test plan

- [x] `cargo clippy --features metal,accelerate -p mlxcel-core -p mlxcel --lib --tests -- -D warnings` (clean)
- [x] `cargo test --release -p mlxcel-core sliding_window_prefill_mask` and `causal_attention` (all pass)
- [x] `cargo test --release -p mlxcel-core "utils::tests"` / `"layers::tests"` (30 / 28 pass)
- [x] `cargo test --release -p mlxcel sliding_prefill_mask` (gemma3/gemma4, 4 pass)
- [x] `cargo test --release -p mlxcel tensor_parallel_gemma3` (logit-match test passes)
- [x] real-model gpt-oss-20b `>window` (crash on main -> coherent here), gemma-4-12b `>window` coherent, byte-identical short/decode regression

## Known follow-ups

- #413 (pre-existing, out of scope here): dense-cache `offset > 0` rolled-over over-window path; in that case the cache already trims to `window` keys so the clamped mask is the correct shape, but this has not been confirmed safe under every model's K/V trimming logic.
- #410 consolidation: migrate gemma3/gemma4 `sliding_prefill_mask` onto the shared `create_sliding_window_prefill_mask`.
- recurrent_gemma applies its local-attention window via a full causal mask with `window_size = 0` (window not enforced by the mask); orthogonal to this issue, worth a separate look.

Closes #408

### DIFF
--- a/src/distributed/tensor_parallel/llama_runtime.rs
+++ b/src/distributed/tensor_parallel/llama_runtime.rs
@@ -2757,11 +2757,17 @@ fn gemma3_masks(
     let sliding_mask = if rank0.sliding_window_pattern > 1 {
         let sliding_offset = gemma3_cache_offset(&mut rank0_caches[0]);
         let max_cache = rank0.sliding_window as i32;
-        let effective_offset = sliding_offset.min((max_cache - seq_len).max(0));
-        Some(mlxcel_core::utils::create_causal_mask_with_window(
+        // Mirror the single-process gemma3 `sliding_prefill_mask` (issue #401/#408):
+        // a fresh single-pass prefill longer than the window keeps every key (the
+        // RotatingKVCache returns all of them) and needs the full-width windowed
+        // mask, otherwise the clamped `(seq_len, window)` mask strands the earliest
+        // query rows with an all-(-inf) row and the broadcast crashes against the
+        // full key axis. The helper's else-branch is byte-identical to the prior
+        // `sliding_offset.min((max_cache - seq_len).max(0))` clamped construction.
+        Some(mlxcel_core::utils::create_sliding_window_prefill_mask(
             seq_len,
-            effective_offset,
-            Some(max_cache),
+            sliding_offset,
+            max_cache,
         ))
     } else {
         None
--- a/src/lib/mlxcel-core/src/layers.rs
+++ b/src/lib/mlxcel-core/src/layers.rs
@@ -3461,7 +3461,14 @@ mod tests {
     }
 
     #[test]
-    fn causal_attention_window_bool_mask_matches_float_reference() {
+    fn causal_attention_multi_token_over_window_uses_full_windowed_mask() {
+        // Multi-token prefill (`q_len = 3`) whose key length (`k_len = 5`)
+        // exceeds the window. Before issue #408 `causal_attention` sliced K/V
+        // to the trailing `window_size` keys and used the clamped
+        // `(q_len, window_size)` mask, which strands the earliest query rows.
+        // The corrected behavior keeps all keys and builds the full
+        // `(q_len, k_len)` windowed-causal mask, so the reference must use the
+        // full K/V and `create_causal_mask_with_window_full`.
         let q = crate::from_slice_f32(
             &[
                 0.1, -0.2, 0.3, 0.4, 0.2, 0.0, -0.1, 0.5, -0.3, 0.1, 0.2, -0.4,
@@ -3483,26 +3490,22 @@ mod tests {
             &[1, 1, 5, 4],
         );
         let scale = 0.5_f32;
-        // Use window_size=3 (not 2) so the post-earlier cap path produces a
-        // non-degenerate mask: with q_len=3, window=3, tril_offset = w - size
-        // = 0, so q=0 attends to k=0. With window=2 the row-0 mask is fully
-        // masked, which produces NaN under the now-correct semantics (a
-        // query whose logical position predates every cache key).
-        let window_size = 3_i32;
+        // window_size=2 is the previously-degenerate case: under the old
+        // clamp+slice path query row 0 (logical position 0) had an all-`-inf`
+        // mask row and softmaxed to NaN. The full mask gives it its own
+        // window, so no row is fully masked.
+        let window_size = 2_i32;
 
         let out = crate::causal_attention(&q, &k, &v, scale, 0.0, window_size);
-        // See note in `causal_attention_single_query_window_respects_mask_when_needed`:
-        // the explicit-mask reference must slice K/V to the last `window_size`
-        // slots to line up with the post-cap mask shape `(q_len, window_size)`.
-        let mask_f = crate::utils::create_causal_mask_with_window(3, 2, Some(window_size));
-        let k_sliced = crate::slice(&k, &[0, 0, 5 - window_size, 0], &[1, 1, 5, 4]);
-        let v_sliced = crate::slice(&v, &[0, 0, 5 - window_size, 0], &[1, 1, 5, 4]);
+        // Reference: full key set + uncapped windowed mask over offset =
+        // k_len - q_len = 2.
+        let mask_full = crate::utils::create_causal_mask_with_window_full(3, 2, Some(window_size));
         let out_ref = attention(
             &q,
-            &k_sliced,
-            &v_sliced,
+            &k,
+            &v,
             scale,
-            Some(mask_f.as_ref().unwrap()),
+            Some(mask_full.as_ref().unwrap()),
             0.0,
             window_size,
         );
@@ -3511,9 +3514,21 @@ mod tests {
         let diff_abs = crate::abs(&diff);
         let diff_sum = crate::sum_all(&diff_abs);
         crate::eval(&diff_sum);
+        let total = crate::item_f32(&diff_sum);
         assert!(
-            crate::item_f32(&diff_sum) < 1e-5,
-            "bool-mask causal window path should match float-mask reference"
+            total.is_finite() && total < 1e-5,
+            "multi-token over-window causal attention should match the full \
+             windowed-mask reference (no NaN, no stranded rows); diff_sum = {total}"
+        );
+
+        // The output itself must be finite (the pre-#408 path produced NaN for
+        // the earliest query row under window_size=2).
+        let out_abs = crate::abs(out.as_ref().unwrap());
+        let out_sum = crate::sum_all(&out_abs);
+        crate::eval(&out_sum);
+        assert!(
+            crate::item_f32(&out_sum).is_finite(),
+            "multi-token over-window causal attention output must be finite"
         );
     }
 
--- a/src/lib/mlxcel-core/src/lib.rs
+++ b/src/lib/mlxcel-core/src/lib.rs
@@ -2570,6 +2570,14 @@ pub fn fast_rope_batched(
 /// explicit mask array. Other hardware keeps using MLX's native causal SDPA
 /// entry point.
 ///
+/// Sliding-window handling splits by query length: a single-query decode
+/// (`q_len == 1`) slices K/V to the trailing `window_size` keys and uses the
+/// clamped `(1, window_size)` mask, while a multi-token prefill that exceeds
+/// the window (`q_len > 1 && k_len > window_size`) keeps all keys and builds a
+/// full-width windowed-causal mask over them. Slicing K/V for a multi-token
+/// prefill would strand the earliest query rows with an all-`-inf` row -> NaN
+/// (issue #401/#408); the windowed correctness instead comes from the mask.
+///
 /// Used by: Llama, Qwen, Mixtral, Gemma, Cohere, Phi, OLMo, Exaone, GLM4,
 /// MiniCPM, DeepSeek, Hunyuan, StarCoder2 and other causal prefill call sites
 pub fn causal_attention(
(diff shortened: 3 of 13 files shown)


inureyes:

### Intent
> Harden sliding-window `>window` single-pass prefill across all models by fixing the shared `causal_attention`, adding `create_sliding_window_prefill_mask`, and wiring per-model fixes (closes #408).

### Findings Addressed
- [x] **olmo3 missed (CRITICAL)** — `src/models/olmo3.rs` is a registered, reachable sliding-window model on a dense `KVCache` with the exact over-window prefill NaN/`<pad>` bug this PR fixes for cohere2/gemma3n: it built the clamped `(l, window)` mask via `create_causal_mask_with_window` and sliced K/V to the trailing `window_size` keys, stranding the earliest query rows. It was not in the audit table and not covered by the general `causal_attention` fix (its windowed path goes through the explicit-mask branch, and it passes `window_size = 0` to `causal_attention`). Fixed by mirroring the cohere2/gemma3n pattern exactly: build the mask via `create_sliding_window_prefill_mask`, and slice K/V to the mask's key axis (`mask_klen`) instead of blindly to `window_size`. `Used by:` on the helper corrected to add `Olmo3` and drop `Baichuan` (baichuan does not call the helper).

### Verification
- [x] All stated requirements implemented (now including olmo3)
- [x] Integrated into project code flow (mask builders + attention K/V slicing wired per model)
- [x] Project conventions followed (no em dashes, no AI attribution; `// Used by:` updated on `causal_attention` and the new helper)
- [x] Existing modules reused (`create_sliding_window_prefill_mask` / `create_causal_mask_with_window_full`, no parallel builder)
- [x] Tests pass (helper selector 3/3, `causal_attention_multi_token_over_window_uses_full_windowed_mask`)

### Review notes (verified correct, no change needed)
- **Bit-exactness:** `over_window_prefill = needs_window_mask && q_len > 1` is the only new branch. For `q_len == 1` decode, `window_size == 0`, and within-window (`k_len <= window`) prefills, `over_window_prefill` is always false and both the K/V-slice selection and the mask chain are byte-identical to `main`. `softcap > 0` composes correctly: the guard is independent of softcap, and softcap is still passed through to `attention` when the full windowed mask is built.
- **Helper else-branch:** byte-identical to the prior per-model `sliding_offset.min((window - size).max(0))` construction. For step3p5 (which previously passed the raw offset) the resulting mask is still byte-identical, because the capped path's `tril` diagonal `window - size` is offset-independent and the non-capped `triu` term is a no-op for `size <= window`.
- **Dense-cache fixes (cohere2, gemma3n):** `mask_klen` correctly tracks the mask key axis (full mask keeps all keys, clamped mask drops the oldest). Both gemma3n sites (`forward` + `forward_with_embeddings`) handled.
- **baichuan:** covered by the general `causal_attention` fix on the `mask.is_none()` prefill path; no separate clamped-mask site. The M5 tile-aligned padded-mask path does not crash (full causal mask matches full K/V) and its lack of window enforcement there is pre-existing and orthogonal to #408.

### Out of scope (pre-existing, flagged for a separate chore)
- `src/lib/mlxcel-core/build.rs:90` trips `clippy::needless_borrows_for_generic_args` under the pinned toolchain 1.93.1 (`bridge.flag_if_supported(&format!(...))` should drop the `&`). It is byte-identical to `main` and unrelated to #408, but a `mlxcel-core` recompile re-emits it under `-D warnings`. Recommend a one-line `chore` fix separately rather than folding it into this PR.

PR #420, opened 2026-06-24
fix(models): size dense prefill masks from live_len under trim (#419)

### LINKED ISSUE #419: fix(server): dense sliding-window models build prefill masks from monotonic offset, not live_len, under --max-kv-size
## Problem / Background
The dense-`KVCache` sliding-window models (cohere2, gemma3n, olmo3) build their multi-token prefill attention masks (both the global `create_causal_mask(l, ga_offset)` and the sliding `create_sliding_window_prefill_mask_dense(l, sliding_offset, window)`) from the cache's MONOTONIC `offset`. PR #418 (issue #413) made the dense sliding mask a full `[l, l + sliding_offset]` mask over all retained keys (`create_sliding_window_prefill_mask_dense` in `src/lib/mlxcel-core/src/utils.rs:710`), which is correct for the normal dense path where the cache returns all `offset` keys.

Mask-build sites that read the monotonic `offset` directly:
- cohere2: `src/models/cohere2.rs:426-438` and `:488-500` (`ga_offset`/`swa_offset = caches[..].offset`).
- gemma3n: `src/models/gemma3n.rs:1199-1214` and `:1376-1391` (`global_offset`/`sliding_offset = caches[..].offset`).
- olmo3: `src/models/olmo3.rs:463-471` (`ga_offset`/`swa_offset = caches[..].offset`).

## The bug
Under the server `--max-kv-size` path, `enforce_max_kv_size_for` (`src/server/batch/scheduler.rs:2082`, trim at `:2113`) calls `KVCache::trim_front` (`src/lib/mlxcel-core/src/cache.rs:2408`), which advances `live_start` while `offset` keeps growing monotonically (to preserve RoPE relative positions). The dense `KVCache::update_and_fetch` Fp16/Int8 path (`cache.rs:2515`, `live_len = buffer_idx()` at `:2535`, sliced at `:2629-2638`) then returns only the live window of `live_len = offset - live_start` keys, NOT `offset` keys. On a later multi-token continuation chunk (`l > 1`), a mask sized from the monotonic `offset` is WIDER than the live K/V the cache actually returns, so the mask key axis and the K/V key axis disagree.

This is PREEXISTING and not introduced by #418: the untouched global-attention path in the same models builds `create_causal_mask(l, ga_offset)` from the same monotonic offset, and the global layer carries `window_size == 0` (cohere2 `:283-287`), so the attention layer's "slice K/V to the mask key axis" branch (`cohere2.rs:193`, `gemma3n.rs:529`, gated on `self.window_size > 0`) never fires for the global layer. It thus already mismatches under the exact same config (`--max-kv-size` + a dense sliding-window model + a multi-token continuation chunk). So the end-to-end config is already broken on the global path regardless of #418; the sliding path now carries the same offset-vs-`live_len` assumption.

Note on gemma3n: its attention layer already reads `cache.live_len()` to slice the KV-shared K/V buffer (`gemma3n.rs:482-489`) and its comment explicitly anticipates the `--max-kv-size` trim, but the model-level mask at `gemma3n.rs:1199-1200`/`1376-1377` is still built from the monotonic `cache.offset`. So even gemma3n, which is partly trim-aware on the K/V side, still builds a mask whose key axis can exceed the live K/V after a trim.

## Proposed direction
Align dense-cache prefill mask construction (both global and sliding) with `cache.live_len()` (the actual returned key count) rather than the monotonic `offset`, for the `--max-kv-size` / trimmed path. Because the masks are byte-identical when `live_start == 0` (no trim has happened — `live_len == offset`), this can be an isolated, behavior-preserving change for the untrimmed common case. A sliding-only partial fix is insufficient: the global path must be aligned too for the config to work end-to-end.

## Acceptance criteria
- [ ] A multi-token continuation chunk on a dense sliding-window model (cohere2 / gemma3n / olmo3) under `--max-kv-size` (after at least one `trim_front`) produces coherent output, with the mask key axis matching the live K/V key axis on both the global and sliding layers.
- [ ] The untrimmed path (`live_start == 0`) remains byte-identical (greedy output unchanged).
- [ ] A regression test exercises the trimmed (`live_start > 0`) continuation case.

## Technical Considerations
- `--max-kv-size` is an opt-in operator flag (not a default), and the mismatch additionally requires a dense sliding-window family plus a multi-token continuation chunk after at least one trim, so the trigger is config-specific rather than something every server hits.
- Turbo* cache modes are out of scope here: `trim_front` is a no-op for them (`cache.rs:2433`) and the scheduler already warns that `--max-kv-size` is not honored on the V side for those modes.

## References
- Surfaced in the PR #418 security/performance review (follow-up to #413). PR #418: https://github.com/lablup/mlxcel/pull/418

inureyes:
## Summary

The dense-`KVCache` sliding-window models (cohere2, gemma3n, olmo3) built their multi-token (`l > 1`) prefill attention masks from the cache's monotonic `offset`, for both the global mask (`create_causal_mask`) and the sliding mask (`create_sliding_window_prefill_mask_dense`). Under the server `--max-kv-size` path that assumption is wrong after a trim, and the mismatch aborts the whole server process. This sizes the masks from `cache.live_len()` instead, which is byte-identical on the common untrimmed path.

## The bug

`enforce_max_kv_size_for` (called in `continue_chunked_prefill`, among others) calls `KVCache::trim_front`, which physically slices the buffer to the live window and advances `live_start` while `offset` keeps growing monotonically (to preserve RoPE relative positions). `update_and_fetch` (Fp16 and Int8) then returns only the live window of `live_len = offset - live_start` keys, not `offset` keys. On a multi-token continuation chunk after at least one trim (`live_start > 0`), a mask sized from the monotonic `offset` is WIDER than the returned K/V, so the mask key axis and the K/V key axis disagree. MLX `broadcast_shapes` throws, the cxx bridge turns the C++ exception into `std::terminate`, and the server process ABORTS (not a recoverable per-request error).

Reproduced on `gemma3n-e2b-4bit` with `--max-kv-size 1024 --prefill-chunk-size 512` on a 2631-token prompt: crash on chunk 4 with `[broadcast_shapes] Shapes (512,2048) and (1,8,512,1536) cannot be broadcast`.

## What changed

- `src/models/cohere2.rs` (`forward_impl` and `forward_with_embeddings_impl`): build the global and sliding prefill masks from `caches[..].live_len()` instead of `caches[..].offset`.
- `src/models/gemma3n.rs` (both `forward` variants): same change for `global_mask` / `sliding_mask`; the comment notes consistency with the KV-shared attention layer, which already slices its K/V to `cache.live_len()`.
- `src/models/olmo3.rs` (`forward_impl`): same change for the global and sliding masks.
- `src/lib/mlxcel-core/src/cache.rs`: new cache-level regression test `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim`.

These `offset` reads were mask-only. RoPE keeps using each layer's own monotonic `cache.offset` (read internally by the attention layer); no positional / layer-forward path changed. `create_causal_mask`, `create_causal_mask_with_window_full`, `create_sliding_window_prefill_mask_dense`, `trim_front`, `update_and_fetch`, and the RotatingKVCache consumers are untouched.

## Why it is correct (and byte-identical when untrimmed)

For a query row `q` at monotonic abs position `offset + q`, the live key column `k` is at abs position `live_start + k`. The causal bound `live_start + k <= offset + q` reduces to `k <= q + (offset - live_start) = q + live_len`, which is exactly `create_causal_mask(l, live_len)`; the window bound reduces the same way. The mask key dim `l + live_len` then matches `update_and_fetch`'s returned `l + live_len`. When no trim has happened, `live_start == 0` so `live_len == offset` and the masks are bit-for-bit the same, so greedy output on the common path is unchanged.

## Test plan

- [x] `cargo test --release -p mlxcel-core --features metal,accelerate cache::` (402 passed, incl. the new `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim`)
- [x] `cargo check --lib --features metal,accelerate`
- [x] `cargo fmt --check`

Closes #419

### DIFF
--- a/src/lib/mlxcel-core/src/cache.rs
+++ b/src/lib/mlxcel-core/src/cache.rs
@@ -8144,6 +8144,97 @@ mod tests {
         metadata.assert_consistent().unwrap();
     }
 
+    /// Regression for #419: a dense `KVCache` multi-token continuation chunk
+    /// after a `--max-kv-size` `trim_front` must size its prefill mask from
+    /// `live_len()` (the key count the cache actually returns), not the
+    /// monotonic `offset`. The dense sliding-window models (cohere2, gemma3n,
+    /// olmo3) build their model-level prefill masks from the cache before the
+    /// chunk's `update_and_fetch`; under a trim (`live_start > 0`) a mask
+    /// sized from `offset` is wider than the returned K/V, so the attention
+    /// broadcast throws and the server aborts. This pins the cache invariant
+    /// the fix relies on so a regression back to `offset` is caught.
+    #[test]
+    fn kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim() {
+        const H: i32 = 2;
+        const D: i32 = 4;
+
+        // Build a `[1, H, n, D]` Fp16-cache K/V tensor with arbitrary values;
+        // only the key axis (n) matters for this shape invariant.
+        let make_kv = |n: i32, base: f32| {
+            let count = (H * n * D) as usize;
+            let data: Vec<f32> = (0..count).map(|i| base + i as f32).collect();
+            ffi::from_slice_f32(&data, &[1, H, n, D])
+        };
+
+        let mut cache = KVCache::new();
+
+        // Prefill `n1` tokens, then trim the oldest `trimmed` (the
+        // `--max-kv-size` cap path) so the live window starts at
+        // `live_start > 0`.
+        let n1 = 8;
+        let trimmed = 3;
+        let _ = cache.update_and_fetch(make_kv(n1, 0.0), make_kv(n1, 100.0));
+        assert_eq!(cache.offset, n1);
+        assert_eq!(cache.live_start, 0);
+
+        assert_eq!(cache.trim_front(trimmed), trimmed);
+
+        // RoPE invariant: `offset` stays monotonic and `live_start` advances,
+        // so the visible window is `live_len = offset - live_start`, NOT
+        // `offset`.
+        assert_eq!(
+            cache.offset, n1,
+            "offset must stay monotonic after trim_front"
+        );
+        assert_eq!(cache.live_start, trimmed);
+        assert_eq!(cache.live_len(), n1 - trimmed);
+
+        // Snapshot what the model-level mask builder reads just before the
+        // continuation chunk's `update_and_fetch`.
+        let live_len_before = cache.live_len(); // correct mask width source
+        let offset_before = cache.offset; // buggy mask width source
+
+        // Continuation chunk of `m` tokens (the `l > 1` prefill path).
+        let m = 4;
+        let (cont_k, _cont_v) = cache.update_and_fetch(make_kv(m, 200.0), make_kv(m, 300.0));
+
+        // The cache returns only the live window plus the new tokens.
+        let returned_key_axis = ffi::array_shape(&cont_k)[2];
+        assert_eq!(
+            returned_key_axis,
+            live_len_before + m,
+            "Fp16 update_and_fetch must return live_len + new tokens"
+        );
+        assert_eq!(
+            returned_key_axis,
+            cache.live_len(),
+            "returned key axis must equal the post-update live window"
+        );
+        assert_ne!(
+            returned_key_axis,
+            offset_before + m,
+            "returned key axis must NOT equal offset + new tokens (the buggy assumption)"
+        );
+
+        // The correct (live_len-sized) prefill mask key axis matches the
+        // returned K/V key axis.
+        let correct_mask = crate::utils::create_causal_mask(m, live_len_before);
+        let correct_klen = *ffi::array_shape(&correct_mask).last().unwrap();
+        assert_eq!(
+            correct_klen, returned_key_axis,
+            "create_causal_mask(m, live_len) key axis must match the returned K/V"
+        );
+
+        // The buggy (offset-sized) mask is strictly wider than the returned
+        // K/V; that mismatch is what crashes the attention broadcast.
+        let buggy_mask = crate::utils::create_causal_mask(m, offset_before);
+        let buggy_klen = *ffi::array_shape(&buggy_mask).last().unwrap();
+        assert!(
+            buggy_klen > returned_key_axis,
+            "offset-sized mask ({buggy_klen}) must be wider than the returned K/V ({returned_key_axis})"
+        );
+    }
+
     /// End-to-end batch grow + shrink sequence: start with one sequence,
     /// admit a second mid-decode (extend), then evict the first
     /// (filter_in_place). Per-row offsets must remain coherent throughout.
--- a/src/models/cohere2.rs
+++ b/src/models/cohere2.rs
@@ -423,17 +423,28 @@ impl Cohere2Model {
 
         // Create masks for full and sliding attention
         let (full_mask, sliding_mask) = if l > 1 {
-            let ga_offset = caches[self.ga_idx].offset;
-            let swa_offset = caches[self.swa_idx].offset;
+            // Size the prefill masks from the cache's live window
+            // (`live_len() = offset - live_start`), not the monotonic
+            // `offset`. Under `--max-kv-size`, `trim_front` slices the buffer
+            // to the live window and advances `live_start` while `offset`
+            // keeps growing to preserve the RoPE relative positions, so
+            // `update_and_fetch` returns only `live_len` keys. A mask sized
+            // from `offset` would be wider than the returned K/V and break the
+            // attention broadcast. With no trim (`live_start == 0`),
+            // `live_len == offset`, so this is byte-identical to the untrimmed
+            // path. See issue #419.
+            let ga_live_len = caches[self.ga_idx].live_len();
+            let swa_live_len = caches[self.swa_idx].live_len();
 
-            let full = Some(create_causal_mask(l as i32, ga_offset));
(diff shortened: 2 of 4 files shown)


inureyes:
## Real-model validation across all three dense sliding-window models (before/after)

The bug is a full server-process abort, reproduced on all three affected models with `--max-kv-size 1024 --prefill-chunk-size 512` and a ~2.6k-token prompt: chunked prefill drives a `trim_front` (live window capped at 1024), and the following continuation chunk builds an `offset`-sized mask wider than the trimmed K/V, so MLX `broadcast_shapes` throws and cxx hits `std::terminate` (`Abort trap: 6`).

Before (main `22c9694e3`, the bug is preexisting) vs after (this PR, masks sized from `live_len()`):

| Model | window | before (`--max-kv-size 1024`) | after |
|---|---|---|---|
| gemma3n-e2b-4bit | 512 | abort: `broadcast_shapes (512,2048) vs (1,8,512,1536)` | HTTP 200, coherent summary |
| command-r7b-4bit (cohere2) | 4096 | abort: `(512,2048) vs (1,32,512,1536)` | HTTP 200, no crash |
| olmo3-32b-4bit | 4096 | abort: `(512,2048) vs (1,40,512,1536)` | HTTP 200, coherent summary |

The mask shape `(512, 2048)` is `[l, l + offset]` (l=512, offset=1536); after the fix it is `[l, l + live_len]` = `(512, 1536)`, matching the live K/V key axis. Identical for all three (the shared failure mode), differing only in head count.

Semantic-correctness control: cohere2 with the fix binary and NO trim (`--max-kv-size 8192`, the full 2632-token prompt retained, the untrimmed path that is byte-identical to main) returns a coherent summary ("The history of computing is a long chain of incremental ideas, each building on the last..."). The degenerate text cohere2 emits at `--max-kv-size 1024` is the expected consequence of capping its context to 1024 tokens, far below its 4096 sliding window (so it loses ~1600 of 2632 prompt tokens); the fix makes that opted-into truncation run correctly instead of crashing. gemma3n (window 512 <= 1024) and olmo3 stay coherent under the same trim.

inureyes:
## Security / performance review

Reviewed the `offset` -> `live_len()` change on cohere2, gemma3n, olmo3.

### In scope (the 3 models): correct and complete, no findings

- **DoS / abort path closed.** Both the global (`create_causal_mask`) and sliding (`create_sliding_window_prefill_mask_dense`) prefill masks now size from `live_len()`, which matches the `live_len + l` key axis that `update_and_fetch` returns after a `trim_front`. The helper produces `[l, l + live_len]`, so the mask key axis equals the returned K/V key axis. The only remaining `cache.offset` reads on these models' prefill path are RoPE position inputs (correct: positional, not a shape), and the K/V comes from `update_and_fetch` (the live window). No other mask or K/V slice on these paths can exceed the live K/V, so a client cannot induce the `broadcast_shapes` abort on these 3 models under `--max-kv-size` after this change. The `--max-kv-size` trim loop trims every per-sequence cache against its own `live_len()` in lockstep, so the representative-cache `live_len()` used to build the shared global/sliding masks is valid for every layer of that class.
- **Performance: no regression.** `live_len() = offset - live_start` is an `#[inline]` field subtraction, not a recompute. `live_len() <= offset`, so the mask is the same size or smaller than before, with no extra allocations or graph nodes. The change is confined to the `l > 1` prefill branch; decode (`l == 1`) is untouched.
- **Integer arithmetic.** `trim_front` clamps `live_start <= offset` (at most `live_start == offset`, giving `live_len() == 0`), so `live_len()` cannot underflow.
- No untrusted-input parsing, I/O, or auth surface is touched. The regression test pins the cache invariant the fix relies on.

### Out of scope, pre-existing (HIGH, recommend follow-up)

The same crash signature (a model-level prefill mask sized from a dense `KVCache`'s monotonic `offset`, passed to a non-rebuilding SDPA path, reachable under `--max-kv-size` chunked prefill with a multi-token continuation chunk after a trim) still exists in other dense models. Confirmed by inspection:

- `nemotron_h`: `create_causal_mask(seq_len, caches[idx].offset())`
- `mistral4`: `create_causal_mask(l, caches[0].offset)`
- `jamba`: `create_causal_mask(seq_len, c[attn_idx].offset())`

For contrast, `glm4_moe_lite` and `minicpm3` already build this mask from `seq_len()` (= live window) and are safe, which shows the migration is partial. These models are not part of #419; a separate issue should apply the same `offset` -> `live_len()`/`seq_len()` change across the remaining dense models, validated per family.

Not merging.

inureyes:
## PR Finalization Complete

### Summary

- **Tests**: Adequate. The cache-level regression test `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim` precisely pins the invariant this fix relies on. The 3-model E2E validation (abort to HTTP 200) was already recorded in the PR. No additional tests added; no gaps found.

- **Documentation**: No doc update warranted. The `--max-kv-size` flag and the `live_len` / `trim_front` mask-sizing behavior are not documented in any file under `docs/`. The relevant explanation lives in the inline code comments, which were updated in all three model files and in the new test.

- **Code quality**: No debug artifacts, no leftover TODOs. Comments are accurate and reference the correct issue numbers (#408, #413, #419). All checks pass: `cargo fmt --check` (clean), `cargo check --lib` (clean), `cargo clippy --lib --tests -- -D warnings` (clean), `cargo test --release -p mlxcel-core cache::` (402 passed, 0 failed).

PR #422, opened 2026-06-24
fix(models): size dense prefill masks from live_len for mistral4/nemotron_nas/qwen-vl under --max-kv-size (#421)

### LINKED ISSUE #421: fix(models): complete offset->live_len prefill-mask migration for dense-KVCache models under --max-kv-size (mistral4/nemotron_nas/qwen*-vl)
## Problem / Background

Follow-up to #419 (PR #420), which fixed the dense-`KVCache` sliding-window trio (cohere2, gemma3n, olmo3). The same crash class exists in other models that build their multi-token (`l > 1`) prefill attention mask from the cache's monotonic `offset`.

Mechanism (unchanged from #419): under the server `--max-kv-size` path, `enforce_max_kv_size_for` (`src/server/batch/scheduler.rs:2082`) calls `KVCache::trim_front` (`src/lib/mlxcel-core/src/cache.rs:2408`), which advances `live_start` while `offset` keeps growing monotonically (RoPE invariant). `KVCache::update_and_fetch` (`cache.rs:2515`, `live_len = buffer_idx()` at `:2535`) then returns only the live window of `live_len = offset - live_start` keys, not `offset` keys. A model-level mask sized from the monotonic `offset` is therefore WIDER than the live K/V after a trim. MLX `broadcast_shapes` throws; because the throw crosses the cxx FFI boundary as a non-`Result` C++ exception it reaches `std::terminate` and ABORTS the whole server process. A remote client can crash all sessions with one long prompt plus a multi-token continuation chunk.

### Scope is gated by cache ownership (important)

`enforce_max_kv_size_for` trims only the caches returned by `CachePool::get_caches_mut`, whose signature is `Option<&mut [KVCache]>` (`cache.rs:5928`). It iterates that slice and calls `cache.trim_front(...)`. Two consequences:

- Models that store their attention K/V as a plain `Vec<KVCache>` in the scheduler cache pool (the standard dense path: `make_caches() -> Vec<KVCache>`, `forward(caches: &mut [KVCache])` used directly) ARE trimmed. These are the affected models.
- Models that use `ModelOwnedSequenceState<CustomCacheEnum>` get `SequenceCacheSet::model_owned`, which sets `caches: Vec::new()` (`cache.rs:5699-5707`). `get_caches_mut` returns an empty slice, so the trim loop runs zero iterations and `trim_front` is never called on their attention caches. Likewise, models that keep their real caches in a custom enum or nested array (`Cache`, `AnyKVCache`, `Qwen3NextCache`, `KimiLinearCache`, `[[KVCache; 2]]`, ...) and ignore the scheduler-provided `&mut [KVCache]` (the `_caches` parameter) are never trimmed through this path.

So an offset-based mask is only a live (remotely triggerable) bug for dense `[KVCache]` scheduler-pool models. For model-owned and enum/rotating models it is a latent inconsistency, not currently reachable via `--max-kv-size`.

The batched/serving path confirms the trigger: `LanguageModel::forward_batched` calls `self.forward(input_ids, batch_caches[0], None)` (`generate.rs:591,600`), so the model builds its own causal mask from `cache.offset` rather than receiving a scheduler-built one.

## Confirmed affected (verified)

Each builds an `l > 1` prefill `create_causal_mask(seq_len, cache.offset)` from the monotonic `offset`, stores its attention K/V as a plain `Vec<KVCache>` in the scheduler pool (so `enforce_max_kv_size_for` trims it), and passes the mask straight to `attention_from_ptr` against `update_and_fetch` keys with no rebuild against the live K/V:

- `src/models/mistral4.rs:639-640` (`Mistral4Model::transformer_body`, `offset = caches[0].offset`). `transformer_body` always builds its own mask and ignores any caller-supplied mask, so the trait `forward(.., mask)` argument does not save it.
- `src/models/nemotron_nas.rs:649-651` (`LanguageModel::forward`, `offset = caches.first().map(|c| c.offset)`; built only when the caller passes `mask = None`, which is the server path).
- `src/models/qwen2_vl.rs:637,693` (`cache_offset = caches[0].offset`; `auto_mask = create_causal_mask(seq_len, cache_offset)` when the caller passes no mask).
- `src/models/qwen3_vl.rs:829,898` (same pattern).
- `src/models/qwen3_vl_moe.rs:1022,1091` (same pattern).

The VLM three build the mask only on the no-mask branch; the text path (`forward_batched -> forward(.., None)`) takes that branch. End-to-end triggering on a VLM additionally requires the text decode to run under the trimmed scheduler pool with a multi-token continuation chunk, which should be validated per family.

## Originally cited but NOT affected via `--max-kv-size` (corrected)

These build an offset-based prefill mask but use `ModelOwnedSequenceState`, so their attention caches are never trimmed by `enforce_max_kv_size_for` (latent inconsistency only, not remotely triggerable today). Dropping them from the confirmed list:

- `src/models/nemotron_h.rs:1629` and `:1983` (`ModelOwnedSequenceState<NemotronLayerCache>`).
- `src/models/jamba.rs:855` and `:988` (`ModelOwnedSequenceState<JambaLayerCache>`).

## Already-correct reference (the migration pattern)

- `src/models/glm4_moe_lite.rs:830-831` and `src/models/minicpm3.rs:344-345` already size from `caches.first().map(|c| c.seq_len())` (the live window). Both are dense `[KVCache]` scheduler-pool models, so they would be trimmed; using `seq_len()` is exactly what keeps them correct. Use them as the pattern.
- cohere2/gemma3n/olmo3 (fixed in #420) now size from `*_live_len` / `live_len()`.

## Audited candidates that are NOT affected (model-owned or internal enum/rotating caches; never trimmed by `enforce_max_kv_size_for`)

- Model-owned (`ModelOwnedSequenceState`): `qwen3_5`, `falcon_h1`, `granitemoehybrid`, `gemma3`, `gemma4`, `llama4`, `lfm2`, `plamo2`, `recurrent_gemma`.
- Custom enum / nested-array caches, scheduler `&mut [KVCache]` ignored: `qwen3_next` (`Qwen3NextCache`), `kimi_linear` (`KimiLinearCache`), `longcat_flash_ngram` (`[[KVCache; 2]]`), `gpt_oss` / `exaone4` / `ministral3` / `mellum` / `step3p5` (`Cache` with `Standard` + `Rotating`), `exaone_moe` (`AnyKVCache`).

For all of these the offset-based mask is latent: correct today because the trim never runs against the cache the model actually reads. If any later gains a dense `[KVCache]` scheduler-pool trim path, it would inherit the same crash, so migrating them defensively is reasonable but not required to stop the current abort.

## Proposed fix

Mirror #420: for MASK construction only, replace `cache.offset` with `cache.live_len()` (alias `seq_len()`); leave RoPE / position inputs reading the monotonic `cache.offset`. Byte-identical when untrimmed (`live_start == 0`, so `live_len == offset`). Apply to the five confirmed models. Optionally migrate the latent model-owned / enum sites in the same pass for consistency.

## Acceptance criteria

- [ ] Each confirmed model (mistral4, nemotron_nas, qwen2_vl, qwen3_vl, qwen3_vl_moe), under `--max-kv-size` small enough to trigger a `trim_front` plus a multi-token continuation chunk, completes without the `broadcast_shapes` abort, with the mask key axis aligned to the live K/V.
- [ ] Untrimmed paths remain byte-identical (greedy output unchanged).
- [ ] Per-family real-checkpoint validation (CLAUDE.md requires it for inference changes); list the checkpoint used per family.
- [ ] A regression test pins the trimmed (`live_start > 0`) invariant, mirroring #420's `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim`.

## References

- Follow-up to #419 / PR #420 (see the PR #420 security-review comment).

inureyes:
## Summary

Completes the `offset` -> `live_len` prefill-mask migration started in #419 / PR #420, applying it to the five remaining dense `Vec<KVCache>` scheduler-pool models that still built their multi-token (`l > 1`) prefill causal mask from the cache's monotonic `offset`.

## The bug

Under the server `--max-kv-size` path, `enforce_max_kv_size_for` (`src/server/batch/scheduler.rs`) calls `KVCache::trim_front`, which advances `live_start` while `offset` keeps growing monotonically (the RoPE invariant). `KVCache::update_and_fetch` then returns only the live window of `live_len = offset - live_start` keys, not `offset` keys. A model-level mask sized from `offset` is therefore WIDER than the live K/V after a trim. MLX `broadcast_shapes` throws, and because the throw crosses the cxx FFI boundary as a non-`Result` C++ exception it reaches `std::terminate` and ABORTS the whole server process. This is a remote-triggerable full-server abort: one long prompt plus a multi-token continuation chunk crashes every session.

## The fix (mask-only)

For mask construction only, size the prefill causal mask from `cache.live_len()` (alias of `seq_len()`, both `= offset - live_start`) instead of `cache.offset`. RoPE, `position_ids`, and the Llama-4 attention scale keep reading the monotonic `offset`. With no trim (`live_start == 0`), `live_len == offset`, so greedy output on the normal untrimmed path is byte-identical.

## What changed

- `src/models/mistral4.rs`: `Mistral4Model::transformer_body` builds the prefill mask from `caches[0].live_len()`. The separate `let offset = caches[0].offset` that feeds the Llama-4 attention scale (`get_llama4_attn_scale`) is left monotonic and unchanged.
- `src/models/nemotron_nas.rs`: `LanguageModel::forward` no-mask branch (the server path) builds the mask from `caches.first().map(|c| c.live_len())`. The per-layer RoPE `offset` in the attention block is untouched.
- `src/models/qwen2_vl.rs`, `src/models/qwen3_vl.rs`, `src/models/qwen3_vl_moe.rs`: only the `auto_mask = create_causal_mask(seq_len, ...)` argument switches to `caches[0].live_len()`. The `cache_offset` variable stays `caches[0].offset` and continues to drive `position_ids` / MRoPE grid slicing and the deepstack `cache_offset == 0` checks, because position_ids and RoPE need absolute (monotonic) positions.

## Scope: confirmed affected vs excluded

The five models above each store attention K/V as a plain `Vec<KVCache>` in the scheduler pool, so `enforce_max_kv_size_for` trims them and the offset-sized mask is a live (remotely triggerable) bug. `nemotron_h` and `jamba` were investigated and EXCLUDED: they use `ModelOwnedSequenceState`, so `CachePool::get_caches_mut` returns an empty slice and the trim loop never runs against their attention caches. For them the offset-based mask is a latent inconsistency, not reachable via `--max-kv-size`, and is left unchanged. `glm4_moe_lite` and `minicpm3` already size from `seq_len()` and are the reference pattern.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo check --lib --features metal,accelerate` (compiles all five changed models)
- [x] `cargo test --release -p mlxcel-core --features metal,accelerate cache::` (402 passed) including the shared invariant guard `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim` (added in #420), which pins the trimmed-cache (`live_start > 0`) invariant these five models now rely on.

No per-model unit test is added: exercising each model's mask construction requires a fully loaded real checkpoint (heavy, not in CI), and the underlying primitive is already pinned by the shared cache-level invariant test above.

Real-model E2E validation status will be added by the orchestrator: nemotron_nas / qwen2_vl / qwen3_vl / qwen3_vl_moe validated locally under a small `--max-kv-size` with a multi-token continuation chunk; mistral4 (Mistral-Small-4-119B) validated when its checkpoint is available.

Closes #421

### DIFF
--- a/src/models/mistral4.rs
+++ b/src/models/mistral4.rs
@@ -634,10 +634,18 @@ impl Mistral4Model {
         let shape = mlxcel_core::array_shape(&h);
         let l = shape[1];
 
-        // Create causal mask for prefill
+        // Create causal mask for prefill. Size it from the cache's live
+        // window (`live_len() = offset - live_start`), not the monotonic
+        // `offset`. Under `--max-kv-size`, `trim_front` advances `live_start`
+        // while `offset` keeps growing for RoPE; `update_and_fetch` then
+        // returns only `live_len` keys, so an `offset`-sized mask would be
+        // wider than the live K/V and trip `broadcast_shapes`. With no trim
+        // (`live_start == 0`), `live_len == offset`, so this is byte-identical
+        // to the untrimmed path. The Llama-4 attention scale below keeps the
+        // monotonic `offset`. See issue #421 (mirrors #419/#420).
         let mask = if l > 1 {
-            let offset = caches[0].offset;
-            Some(create_causal_mask(l, offset))
+            let live_len = caches[0].live_len();
+            Some(create_causal_mask(l, live_len))
         } else {
             None
         };
--- a/src/models/nemotron_nas.rs
+++ b/src/models/nemotron_nas.rs
@@ -642,13 +642,20 @@ impl LanguageModel for NemotronNASModel {
     ) -> UniquePtr<MlxArray> {
         let mut h = self.embeddings.forward(input_ids);
 
-        // Create causal mask if not provided
+        // Create causal mask if not provided. Size it from the cache's live
+        // window (`live_len() = offset - live_start`), not the monotonic
+        // `offset`. Under `--max-kv-size`, `trim_front` advances `live_start`
+        // while `offset` keeps growing for RoPE; `update_and_fetch` then
+        // returns only `live_len` keys, so an `offset`-sized mask would be
+        // wider than the live K/V and trip `broadcast_shapes`. With no trim
+        // (`live_start == 0`), `live_len == offset`, so this is byte-identical
+        // to the untrimmed path. See issue #421 (mirrors #419/#420).
         let computed_mask = if mask.is_none() {
             let shape = array_shape(&h);
             let seq_len = shape[1];
-            let offset = caches.first().map(|c| c.offset).unwrap_or(0);
+            let live_len = caches.first().map(|c| c.live_len()).unwrap_or(0);
             if seq_len > 1 {
-                Some(create_causal_mask(seq_len, offset))
+                Some(create_causal_mask(seq_len, live_len))
             } else {
                 None
             }
--- a/src/models/qwen2_vl.rs
+++ b/src/models/qwen2_vl.rs
@@ -685,12 +685,21 @@ impl Qwen2VLModel {
             }
         });
 
-        // Create causal mask if needed
+        // Create causal mask if needed. Size it from the cache's live window
+        // (`live_len() = offset - live_start`), not the monotonic
+        // `cache_offset`. Under `--max-kv-size`, `trim_front` advances
+        // `live_start` while `offset` keeps growing; `update_and_fetch` then
+        // returns only `live_len` keys, so an `offset`-sized mask would be
+        // wider than the live K/V and trip `broadcast_shapes`. With no trim
+        // (`live_start == 0`), `live_len == cache_offset`, so this is
+        // byte-identical to the untrimmed path. `cache_offset` stays monotonic
+        // above because position_ids/RoPE need absolute positions. See issue
+        // #421 (mirrors #419/#420).
         let auto_mask;
         let mask = if mask.is_some() {
             mask
         } else {
-            auto_mask = mlxcel_core::utils::create_causal_mask(seq_len, cache_offset);
+            auto_mask = mlxcel_core::utils::create_causal_mask(seq_len, caches[0].live_len());
             Some(auto_mask.as_ref().unwrap() as &MlxArray)
         };
 
--- a/src/models/qwen3_vl.rs
+++ b/src/models/qwen3_vl.rs
@@ -890,12 +890,21 @@ impl Qwen3VLModel {
             }
         });
 
-        // Create causal mask if needed
+        // Create causal mask if needed. Size it from the cache's live window
+        // (`live_len() = offset - live_start`), not the monotonic
+        // `cache_offset`. Under `--max-kv-size`, `trim_front` advances
+        // `live_start` while `offset` keeps growing; `update_and_fetch` then
+        // returns only `live_len` keys, so an `offset`-sized mask would be
+        // wider than the live K/V and trip `broadcast_shapes`. With no trim
+        // (`live_start == 0`), `live_len == cache_offset`, so this is
+        // byte-identical to the untrimmed path. `cache_offset` stays monotonic
+        // above because position_ids/RoPE need absolute positions. See issue
+        // #421 (mirrors #419/#420).
         let auto_mask;
         let mask = if mask.is_some() {
             mask
         } else {
-            auto_mask = mlxcel_core::utils::create_causal_mask(seq_len, cache_offset);
+            auto_mask = mlxcel_core::utils::create_causal_mask(seq_len, caches[0].live_len());
             Some(auto_mask.as_ref().unwrap() as &MlxArray)
         };
 
--- a/src/models/qwen3_vl_moe.rs
+++ b/src/models/qwen3_vl_moe.rs
@@ -1083,12 +1083,21 @@ impl Qwen3VLMoeModel {
             }
         });
 
-        // Create causal mask if needed
+        // Create causal mask if needed. Size it from the cache's live window
+        // (`live_len() = offset - live_start`), not the monotonic
+        // `cache_offset`. Under `--max-kv-size`, `trim_front` advances
+        // `live_start` while `offset` keeps growing; `update_and_fetch` then
+        // returns only `live_len` keys, so an `offset`-sized mask would be
+        // wider than the live K/V and trip `broadcast_shapes`. With no trim
+        // (`live_start == 0`), `live_len == cache_offset`, so this is
+        // byte-identical to the untrimmed path. `cache_offset` stays monotonic
+        // above because position_ids/RoPE need absolute positions. See issue
+        // #421 (mirrors #419/#420).
         let auto_mask;
         let mask = if mask.is_some() {
             mask
         } else {
-            auto_mask = mlxcel_core::utils::create_causal_mask(seq_len, cache_offset);
+            auto_mask = mlxcel_core::utils::create_causal_mask(seq_len, caches[0].live_len());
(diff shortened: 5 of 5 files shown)


inureyes:
## PR Finalization Complete

### Summary

- **Tests**: No new tests added. The shared invariant `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim` (added in #420, lives in `src/lib/mlxcel-core/src/cache.rs:8157`) pins the `trim_front` / `live_len` / `create_causal_mask` contract all five models now rely on; it passes. None of the five model files have any existing unit tests (consistent with all inference model files in the codebase), and per-model mask-construction tests require a fully loaded checkpoint, so no cheap unit test exists. No gap to fill.
- **Documentation**: No docs in `docs/` reference `--max-kv-size`, `live_len`, `trim_front`, or dense-cache mask sizing behavior. The change is internal to model forward passes. No doc update warranted.
- **Lint/Format**: Clean. `cargo check --lib`, `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --check`, and `cargo test --release -p mlxcel-core cache::` (402 passed) all green on current HEAD `af05dc748`.

No commits were added; the branch is already production-ready. Not merged.

inureyes:
## Real-model validation (all 5 models, before/after under `--max-kv-size 1024 --prefill-chunk-size 512`)

Same bug class as #419/#420: a `trim_front` mid chunked-prefill leaves an `offset`-sized mask wider than the live K/V → MLX `broadcast_shapes` throw → cxx `std::terminate` → server abort. `before` = the merged `main` binary (loads all models but mask-buggy for these 5); `after` = this branch.

- **qwen2-vl-2b**: before = abort (`broadcast_shapes (512,2048) vs (1,12,512,1536)`), after = HTTP 200. Directly reproduced crash→fix. Untrimmed control coherent ("...evolution of computing from early mechanical calculators...").
- **qwen3-vl-2b / qwen3-vl-30b-a3b**: no-regression. These route text-only generation through `forward_text_only`; the fixed `forward_impl` auto-mask is the multimodal path. Byte-identical before==after; untrimmed controls coherent. They share qwen2_vl's exact one-line mask fix.
- **nemotron-nas-30b**: no-regression, coherent, byte-identical (crash not reproduced even at `--max-kv-size 600 --prefill-chunk-size 200`; it does not hit a mid-prefill trim + continuation, so the fix is defensive).
- **mistral4 (Mistral-Small-4-119B)**: the mask fix is byte-identical and correct, and mistral4 now loads and generates text (via #424). Its mask crash→fix cannot be demonstrated because a SEPARATE bug aborts every mistral4 `offset>0` forward first (`[reshape] ... size 2097152 into (1,1,1)`, reachable both via image tokens and chunked prefill), tracked as #425. mistral4 text-only without `--max-kv-size` (single-pass, `offset==0`) generates coherently, confirming the untrimmed path is unaffected.

Net: the fix is correct and byte-identical for all 5; qwen2_vl is the directly-reproduced crash→fix; the remaining 4 are no-regression (defensive / multimodal / blocked-by-#425). The shared cache invariant test from #420 pins the primitive.

PR #431, opened 2026-06-24
fix: size gemma3/gemma4/exaone_moe prefill mask from live_len under --max-kv-size

### LINKED ISSUE #430: fix: gemma3/gemma4/exaone_moe size prefill mask from monotonic offset, remote-crash under --max-kv-size (missed by #419-#422)
## Summary

gemma3, gemma4, and exaone_moe build their multi-token (l > 1) prefill attention mask from the cache's monotonic `offset` instead of `live_len()`/`seq_len()`. Under the server `--max-kv-size` path the global (full-attention) cache is trimmed, so the mask becomes wider than the live K/V, MLX `broadcast_shapes` throws, and because the op is a non-Result cxx function the throw becomes `std::terminate` and aborts the whole server. A remote client can crash all sessions with one long prompt plus a multi-token continuation chunk. This is the exact remote-DoS class fixed for the other dense Vec<KVCache> models in #419/#420 and #421/#422, which named cohere2/gemma3n/olmo3 and mistral4/nemotron_nas/qwen-vl but missed these three.

## Root cause (static trace, airtight)

1. `enforce_max_kv_size_for` (`src/server/batch/scheduler.rs:2082-2115`) iterates EVERY cache in the sequence and trims any whose `live_len() > max_kv_size` via `KVCache::trim_front`, which advances `live_start` while `offset` keeps growing (the RoPE invariant). After a trim, `update_and_fetch` returns only `live_len = offset - live_start` keys. The global (full-attention) Standard cache is the unbounded one this is designed to trim.
2. These three size their multi-token prefill mask from the monotonic `offset`, not `live_len`:
   - **gemma3** `src/models/gemma3.rs:766` — `let global_offset = caches[global_idx].as_interface().offset();` then `create_causal_mask(seq_len, global_offset)` at :767 (and the sliding mask at :769 also uses `offset()`). This is the seq_len > 1 batched prefill branch (no alternate live_len path).
   - **gemma4** `src/models/gemma4.rs` — `first_cache_offset(caches, "full_attention")` (`:3279`) returns `Cache::Standard(c).offset`, used to size the global prefill mask at `:2750`, `:2790`, `:2833`, `:3786`.
   - **exaone_moe** `src/models/exaone_moe.rs:674-679` — `ga_offset` from the `AnyKVCache::Standard` cache's `.offset()`, used in `create_causal_mask(seq_len, ga_offset)` (and the sliding `swa_offset` from the Rotating cache at :684).
3. After the global Standard cache is trimmed, `offset > live_len`, so `create_causal_mask(seq_len, offset)` is wider than the live K/V (which is `live_len`), the attention score broadcast mismatches, MLX throws `std::invalid_argument`, and the non-Result cxx boundary turns it into `std::terminate`.

Contrast the fixed sibling: `src/models/mistral4.rs:648` uses `create_causal_mask(l, live_len)`.

## Why these were missed

cohere2/gemma3n/olmo3 use the `_dense` mask builder (`create_sliding_window_prefill_mask_dense`) that #413/#419 migrated to `live_len`. gemma3/gemma4 use the older non-dense `create_sliding_window_prefill_mask` + `create_causal_mask` with `offset`, and `gemma3.rs`/`gemma4.rs` appear in NONE of the live_len migration commits (#413/#418, #419/#420, #421/#422); their earlier gemma fixes (#405/#412/#415) addressed the sliding-window over-window shape, not the global-mask offset->live_len change. exaone_moe uses the default `supports_batching` (no override) so it is batched and trimmed, and reads `AnyKVCache::Standard.offset()`.

## Audit (the set is exhaustive)

AT-RISK (size mask from monotonic offset AND batch with a trimmable dense cache):
| Model | mask source | file:line |
|---|---|---|
| gemma3 | `offset()` | `src/models/gemma3.rs:766` (global), `:769` (sliding) |
| gemma4 | `offset` via `first_cache_offset` | `src/models/gemma4.rs:2750/2790/2833/3786` |
| exaone_moe | `offset()` via AnyKVCache::Standard | `src/models/exaone_moe.rs:674-679` |

Confirmed SAFE:
- **llama4, qwen3_5**: `supports_batching() == true` but use `ModelOwnedSequenceState`, which `get_caches_mut` returns empty for, so `enforce_max_kv_size_for` never trims them (offset stays == live_len).
- **minicpm3**: its `offset` local is `caches.first().seq_len()` (== live_len), not the monotonic offset.
- **supports_batching() == false** (never trimmed): gpt_oss, step3p5, ministral3, mellum, exaone4, nemotron_h, jamba, lfm2, plamo2, falcon_h1, granitemoehybrid, qwen3_next, kimi_linear, mamba, mamba2, rwkv7, recurrent_gemma, longcat_flash_ngram, nemotron_nas.
- Already fixed by #419-#422: cohere2, gemma3n, olmo3, mistral4, qwen2_vl, qwen3_vl, qwen3_vl_moe.

TO VERIFY during implementation: `gemma4_mtp_target` (the MTP draft target uses `create_causal_mask_with_left_padding`; confirm its mask length is `live_len`-based or apply the same fix).

## Fix plan

Mirror #421/#422 exactly: size the multi-token prefill masks from `cache.live_len()` / `seq_len()` (== offset - live_start) instead of `offset` in gemma3, gemma4, and exaone_moe, for BOTH the global causal mask and the sliding-window prefill mask. RoPE, position_ids, and any attention scale keep reading the monotonic `offset` (that is correct and must not change). Byte-identical behavior when untrimmed (live_start == 0, so live_len == offset).

## Tests

Mirror the test approach used by #419-#422. At minimum, assert that the prefill mask key-length is derived from `live_len()`/`seq_len()` for each of the three models, and add (or extend) a regression that exercises a trimmed cache (live_len < offset) so the mask width equals the live K/V rather than the monotonic offset.

## Acceptance criteria

- With `--max-kv-size` set, a long prompt that trims the global cache followed by a multi-token continuation chunk no longer aborts the server for gemma3, gemma4, or exaone_moe.
- The three models size their prefill masks from `live_len`/`seq_len`; RoPE/position_ids still use the monotonic offset.
- `gemma4_mtp_target` is confirmed safe or fixed.
- No behavior change when untrimmed.

## Validation note

The fix is device-independent shape logic; the primary gate is `cargo test` on the mask-sizing plus, ideally, a `mlxcel serve --max-kv-size <small>` run with a long prompt and a multi-token continuation (this dev host can build and run real models on the GB10 GPU, so the runtime repro is feasible).

## Related

Same `std::terminate` crash class as the conv-layout fix (#428) and the FFI-fallibility follow-up (#427); making the mask/conv MLX ops Result-returning at the cxx boundary (#427, generalized) would convert this entire class from a process abort into a recoverable per-request error.

inureyes:
## Summary

gemma3, gemma4, and exaone_moe sized their multi-token (l > 1) prefill attention masks from the KV cache's monotonic `offset` instead of the live window, for both the global (full-attention) causal mask and the sliding-window prefill mask. Under the server `--max-kv-size` path, `enforce_max_kv_size_for` trims the global Standard cache via `KVCache::trim_front`, which advances `live_start` while `offset` keeps growing for RoPE, so `update_and_fetch` returns only `live_len = offset - live_start` keys. A mask sized from `offset` is then wider than the live K/V, MLX `broadcast_shapes` throws, and because the op is a non-Result cxx function the throw becomes `std::terminate` and aborts the whole server. This completes the `offset` -> `live_len` prefill-mask migration started in #419/#420 and #421/#422 for the three Gemma-family / ExaOne dense-cache models they missed.

The fix is mask-only and byte-identical when untrimmed (`live_start == 0`, so `live_len == offset`). RoPE, position_ids, and the batched-MTP coordinate math keep reading the monotonic `offset`.

## What changed

- `src/models/gemma3.rs`: the `seq_len > 1` prefill branch sizes the global mask from `caches[global_idx].live_len()` and the sliding mask from `caches[0].live_len()`. Added a `live_len()` accessor to the model's `CacheInterface` trait (`KVCache` -> `live_len()`, `RotatingKVCache` -> `seq_len()`).
- `src/models/gemma4.rs`: added a `first_cache_live_len` helper (free function plus method form) that mirrors `first_cache_offset` and returns the live window (`live_len()` for the full-attention Standard cache, `seq_len()` for the sliding Rotating cache). It feeds the two genuine prefill-mask sites: the non-ragged `l > 1` branch in `forward_with_speculative_sinks` and the stage-forward `execute_hidden` path. The batched-MTP sites that use the offset as a `per_row_valid_end` column coordinate (divergent verify, `has_padding`, `mask_stale_key_gap`) intentionally keep the monotonic `first_cache_offset`; that regime enforces `live_len == offset` (`slot_base == 0`) anyway, and the value there is a coordinate, not a pure mask width. The now-unused `first_cache_offset` method form was removed (the free function remains in use).
- `src/models/exaone_moe.rs`: the global and sliding prefill masks size from a new `AnyKVCache::live_len()` (`Standard` -> `live_len()`, `Rotating` -> `seq_len()`) instead of `offset()`.

## Mask sites switched to `live_len` vs offset uses kept

Switched to `live_len`/`seq_len` (prefill mask width):

- gemma3 global mask (`gemma3.rs`, seq_len > 1 branch) and sliding mask.
- gemma4 non-ragged `l > 1` prefill (`create_causal_mask` + `create_sliding_window_prefill_mask`) and the `execute_hidden` stage forward (`seq_len > 1`).
- exaone_moe global causal mask and sliding-window prefill mask.

Intentionally kept on the monotonic `offset`:

- All per-layer RoPE / `fast_rope` offsets (attention reads each layer's own `cache.offset()`).
- gemma4 batched-MTP coordinate sites: `first_present_cache_offset` divergence detection, `build_divergent_verify_mask`, the `has_padding` left-padding masks, and `mask_stale_key_gap` (these compare against `per_row_valid_end`, which lives in monotonic-offset space, and the MTP regime requires `slot_base == 0`).
- gemma4 `speculative_cache_offset` drafter RoPE rebind.

## gemma4_mtp_target verdict

Verified safe, no change needed. `gemma4_mtp_target` does not implement `LanguageModel` and is not registered in `model_metadata.rs`; its adapters drive forward through `Gemma4Wrapper::forward_with_speculative_sinks_explicit_cache` on caller-owned `Vec<Cache>` that never enters the scheduler pool, so `enforce_max_kv_size_for` cannot reach it. Its only mask builder (`create_causal_mask_with_left_padding`) sizes from a fixed `offset 0` prefill. The Gemma 4 prefill-mask sites it routes into are already covered by the `gemma4.rs` edits.

## Test plan

- [x] `cargo fmt`
- [x] `cargo check --lib --tests` (clean, no warnings)
- [x] `cargo test --lib gemma3` (22 passed; new `global_prefill_mask_sizes_from_live_len_not_offset_after_trim` and `sliding_cache_live_len_matches_returned_keys` pass)
- [x] `cargo test --lib gemma4` (new `first_cache_live_len_sizes_global_mask_from_live_window_after_trim` and `first_cache_live_len_sliding_matches_returned_keys` pass; one pre-existing unrelated failure `load_and_sanitize_weights_dequantizes_nvfp4_gemma4_checkpoint` reproduces identically on the clean base commit and is NVFP4 weight-dequant, not mask logic)
- [x] `cargo test --lib exaone_moe` (2 passed: new `global_mask_sizes_from_live_len_not_offset_after_trim` and `sliding_cache_live_len_matches_returned_keys`)

The new per-model tests trim a Standard cache so `live_start > 0`, then assert the `live_len`-sized prefill mask key axis matches the continuation chunk's returned K/V (`live_len + m`) while the `offset`-sized mask is strictly wider, plus a Rotating-cache case where `live_len` equals the returned key axis. The shared cache invariant is already pinned by `kv_cache_continuation_mask_sizes_from_live_len_not_offset_after_trim` in `cache.rs` (#419).

Closes #430


inureyes:

### Intent
> Size the multi-token (l > 1) prefill attention mask in gemma3/gemma4/exaone_moe from the cache's live window (`live_len()`/`seq_len()`) instead of the monotonic `offset`, for both the global and sliding masks, completing the #419/#420 + #421/#422 migration. Defensive/consistency hardening (not live-crashable today; the issue's remote-DoS framing was corrected), byte-identical on the untrimmed path.

### Findings Addressed
- [x] gemma3 `Gemma3StageModel::execute_hidden` (pipeline-stage path) still sized both the global and sliding prefill masks from `offset()` while the gemma4 stage analog and the gemma3 single-process path were switched (HIGH, incomplete migration). Fixed in this branch by switching both sites to `as_interface().live_len()`, mirroring `Gemma3Model::forward_with_caches` and `Gemma4StageModel::execute_hidden`. RoPE kept on the monotonic offset.

### Remaining Items
- None. exaone_moe discards/recreates its internal `AnyKVCache` per `forward` (only `offset` round-trips to the external `KVCache`) is a pre-existing design unrelated to this PR and out of scope.

### Regression verdict
- The untrimmed gemma3/gemma4 path is byte-identical. At the cache level `KVCache::live_len() == seq_len() == offset - live_start`, and untrimmed `live_start == 0`, so `live_len == offset` exactly. Confirmed at `src/lib/mlxcel-core/src/cache.rs:720-731`.
- Both the GLOBAL causal mask and the SLIDING-window prefill mask were switched in every file (gemma3 `forward_with_caches` 795/799 + stage 1074/1080 after fix; gemma4 non-ragged `l > 1` 2806/2807 + stage 3833/3834; exaone_moe 702/710).
- Cache accessor usage is correct: Standard -> `live_len()`, Rotating -> `seq_len()` (gemma3 `CacheInterface`, gemma4 `first_cache_live_len`, exaone_moe `AnyKVCache::live_len`).
- No RoPE/position/coordinate site was switched. Verified offset is kept at: gemma3 per-layer `fast_rope`/`fast_rope_batched` (141/223), gemma4 per-layer `fast_rope` (1399), exaone_moe per-layer `fast_rope` (506); gemma4 MTP coordinate sites (`first_present_cache_offset` 2697, `build_divergent_verify_mask` 2727/2732/2739, `has_padding` left-pad masks 2759/2760, `mask_stale_key_gap` upper bounds 2849/2869, `speculative_cache_offset` 4252). Each is a column coordinate compared against `per_row_valid_end` (monotonic-offset space) or a RoPE position, not a bare prefill-mask width, and the MTP regime enforces `slot_base == 0` so `live_len == offset` there anyway.
- `gemma4_mtp_target` verdict confirmed safe: it does not `impl LanguageModel` (not registered for scheduling), and routes through `Gemma4Wrapper::forward_with_speculative_sinks[_explicit_cache]` on caller-owned caches that never enter the scheduler pool. The Gemma 4 prefill-mask sites it reaches are already covered.

### Tests
- New per-model tests genuinely exercise the trimmed invariant (construct cache, `trim_front` so `live_start > 0` and `live_len < offset`, then assert a `live_len`-sized mask key axis equals the continuation chunk's returned K/V `live_len + m` while an `offset`-sized mask is strictly wider), plus an untrimmed Rotating-cache equality case. Not trivially passing.

### Verification
- [x] All stated requirements implemented (offset -> live_len for both masks in all three models; one stage site was missed and is now fixed)
- [x] Integrated into project code flow (model `forward` and pipeline-stage paths)
- [x] Project conventions followed (reuses existing `live_len()`/`seq_len()`/`as_interface()` accessors; no new restructuring)
- [x] Tests pass

### Narrow gate results (this dev host, MLX CPU build cached)
- `cargo fmt --check`: clean
- `cargo check --lib`: clean
- `cargo test --lib gemma3`: 22 passed (incl. 2 new)
- `cargo test --lib gemma4`: 148 passed (incl. 2 new); 1 pre-existing unrelated failure `load_and_sanitize_weights_dequantizes_nvfp4_gemma4_checkpoint` (NVFP4 dequant), confirmed failing identically on the clean base `22002d1`
- `cargo test --lib exaone_moe`: 2 passed (2 new)


The fix commit (`fix(models): size gemma3 stage prefill mask from live_len`) is on the PR branch and not pushed.

PR #433, opened 2026-06-24
feat: add N-gram loop detection with Gemma 4 family default-on

### LINKED ISSUE #432: Add generation loop/repetition detection to break Gemma 4 token-repetition collapse under tool/grammar-constrained decoding
## Summary

Gemma 4 models (31B Dense and 26B-A4B MoE, including the QAT 4-bit checkpoints) have an upstream, weights-level token-repetition collapse: during generation the output degenerates into a single repeated token (or short fragment) that fills the token budget. This is documented upstream in google-deepmind/gemma#622 and reproduced across Ollama, LMStudio, vLLM, and Cloudflare Workers AI. mlxcel 0.3.2 is also affected.

The collapse is strongly amplified when the prompt carries tool declarations and/or when decoding is grammar-constrained (structured output / tool-call JSON). mlxcel currently has no countermeasure that reliably works for this failure mode, so we request an engine-side repetition/loop detection mechanism.

## Environment

- mlxcel-server 0.3.2 (package 26.6.21), macOS arm64, Metal
- Model: mlx-community/gemma-4-26b-a4b-it-qat-4bit
- Launch flags: `--ctx-size 262144 --jinja --parallel 2 --cache-type-k int8 --cache-type-v int8 --prefill-chunk-size 1024` (no `--repeat-penalty`)

## Reproduction

Direct OpenAI-compatible requests to the running mlxcel socket, single user turn.

- No tools in the request: always clean.
- System prompt + 10 tool declarations, `chat_template_kwargs.enable_thinking=true`, fresh first turn: the reasoning/thought channel collapses into a repeated garbage token (observed: CJK such as `様様様` / `婕婕婕`, box-drawing `◸◸◸` / `╿╽╿╽`, and control characters) up to the thinking budget, after which a tool call still emerges. `prompt_tokens` is about 2150 against a 262144 context, so this is not context overflow.

Determinism and seed dependence (this proves the prompt is well-formed and the failure is a sampling trajectory, not a truncated/malformed prompt):

- Fixed `seed=42`, same payload, 3 runs: byte-identical output every time.
- Same payload, varying seed: clean at some seeds, degenerate at others.

Sampling comparison on the worst case (10 tools, enable_thinking=true, first turn):

| sampling | degeneration rate |
|---|---|
| temp1.0 / top_p0.9 / top_k40 / min_p0.1, no penalty | ~3-4 / 5 |
| Gemma recommended: temp1.0 / top_p0.95 / top_k64 / min_p0.0, no penalty | 3 / 6 |
| + repeat_penalty 1.1 / repeat_last_n 64 | 1 / 5 |
| + repeat_penalty 1.3 / repeat_last_n 256 | 0 / 4 (small sample; upstream reports flat penalties unreliable for sustained generation) |
| DRY 0.8 / 1.75 / 2 | 5 / 5 (no effect) |

## Why existing mitigations are insufficient

- The Gemma-4 channel priming in `server/chat_template.rs` (`patch_gemma4_generation_prompt`) only applies when `enable_thinking=true` AND the rendered prompt ends exactly in `<|turn>model\n`. Under heavier real prompts it does not reliably fire, and even when it does the collapse still occurs (see the 3-4/5 row).
- Flat `repeat_penalty` and the DRY sampler shape the probability distribution. Once the logits collapse, the top-k candidates are themselves garbage, so distribution shaping cannot recover. Upstream google-deepmind/gemma#622 confirms repeat_penalty 1.0 to 1.5 has no reliable effect.

## Proposed mitigation

Add N-gram repetition detection that ends a generation early when the output stream contains a short pattern that repeats. This is exactly the mechanism vLLM exposes in `SamplingParams`:

- `max_pattern_size` (int, default 0 = disabled): maximum N-gram pattern size to scan for. Must be set together with `min_count`.
- `min_pattern_size` (int, default 0, treated as 1): minimum N-gram pattern size to check. Must be <= `max_pattern_size`.
- `min_count` (int, default 0, must be >= 2): minimum number of repeats of the pattern that triggers detection.
- Action: when a pattern whose size is in [`min_pattern_size`, `max_pattern_size`] repeats at least `min_count` times, generation is ended early. The vLLM docs state: "If such repetition is detected, generation will be ended early." With `min_pattern_size=1` this catches single-token loops such as the observed `様様様` collapse, and larger sizes catch short multi-token loops such as `the the the` or `abcdabcd...`.

In vLLM these fields are opt-in (all default 0). For mlxcel we suggest either defaulting them on for the Gemma 4 family, or auto-enabling them when `tools` or `response_format: json_schema` is present, with a per-request override. A starting point that covers the loops we observed is roughly `min_pattern_size=1`, `max_pattern_size=20`, `min_count=4`.

Important: the detector must run on the raw generated token stream (around the reasoning-channel and tool-call parsing), so it also catches loops inside the `<|channel>thought` span, not just the final answer. In our reproduction the collapse most often filled the thought channel up to the thinking budget.

## Preferred activation surface (for downstream serving apps)

mlxcel is bundled as a serving engine inside a desktop app, and many users download Gemma 4 and derivative checkpoints independently. To keep behavior consistent without per-user configuration or per-model file edits, we would prefer the detection to be activated at the engine level rather than only as a per-request field:

- Best: auto-enable the detection for the Gemma 4 model family by default, with sane built-in parameters, so a downstream serving app needs no configuration and end users see no toggle. This is a safety mechanism that ends generation only when a real repetition loop is detected, so a conservative default-on for this family is low risk.
- Acceptable: expose it as server-level defaults via CLI flags (the same way `--repeat-penalty` is exposed today) and/or auto-enable it whenever `tools` or `response_format: json_schema` is present, with a per-request override. A single launch default per family is easy for an integrator to set.
- Please avoid making a per-request `SamplingParams` field the only surface. That forces every integrating client to add request plumbing, and settings UI if it is user-tunable, for what is really an engine-level safety default.

Rationale: an engine-side default keeps the fix consistent across every integrating app and every independently downloaded checkpoint, with no per-user options and no per-model template edits.

## References

- google-deepmind/gemma#622 (root cause: Gemma 4 token repetition collapse; reports flat repeat_penalty 1.0 to 1.5 as ineffective)
- vllm-project/vllm#40080 (Gemma 4 infinite repetition with structured output)
- vLLM `SamplingParams` N-gram repetition detection fields `max_pattern_size` / `min_pattern_size` / `min_count` (docs.vllm.ai/en/latest/api/vllm/sampling_params/)
- ollama/ollama#15502, google-ai-edge/LiteRT-LM#2249
- huggingface google/gemma-4-12B-it discussion #38 (chat-template prior-turn reasoning re-injection amplifies the loop on multi-turn tool use)

inureyes:
## Summary

Adds vLLM-style N-gram repetition / loop detection that ends a generation early when the raw generated token stream collapses into a short repeated pattern, and turns it on by default for the Gemma 4 family. This breaks the upstream, weights-level Gemma 4 token-repetition collapse (google-deepmind/gemma#622, vllm-project/vllm#40080) where generation degenerates into a single repeated token or short fragment that fills the token budget, most often inside the thought channel and amplified by tool declarations or grammar-constrained decoding. Sampling penalties (repeat_penalty, DRY) cannot recover once the logits collapse, so distribution shaping is not enough; this is a stop condition on the raw stream instead.

## Design

The detector is generic and model-family agnostic; the activation policy lives in the server control plane.

- `mlxcel-core` gains a `loop_detection` module: `LoopDetectionConfig` (vLLM field names `max_pattern_size` / `min_pattern_size` / `min_count`, all default `0` = disabled) and the pure, unit-testable `detect_repetition_loop(generated, cfg)`. For each pattern size `p` in `min_pattern_size..=max_pattern_size` it checks whether the last `p * min_count` tokens are a single `p`-length block repeated `min_count` times. This catches single-token loops (`p=1`) and short multi-token loops (`p=2,3,...`). It runs only when enabled, so a model that does not opt in pays nothing.
- `SamplingConfig` gains a `loop_detection: LoopDetectionConfig` field defaulting to a zero-overhead no-op, mirroring how `token_bias` / `stop_token_ids` preserve the bit-exact baseline.
- The detector is invoked at every decode stop site (the four decode loops in `generate.rs` and the three scheduler decode paths) right after each token is pushed to `generated_tokens`, alongside the existing eos/length checks, so it also catches loops inside the reasoning channel and tool-call JSON, not just the final answer. A new `FinishReason::RepetitionLoop` is added with its `SequenceState` transitions; the OpenAI wire `finish_reason` stays `"stop"` (an early, non-error stop, matching vLLM), and the scheduler logs an info line when it fires.

## Activation policy (issue "Preferred activation surface")

Default-on for the Gemma 4 family with no configuration required, implemented in `request_options::resolve_loop_detection` (the server request to SamplingConfig assembly), keyed off the loaded model type resolved once at startup (`ModelType::Gemma4 | Gemma4VLM | Gemma4Unified`). Precedence, highest first:

1. Explicit per-request fields (`max_pattern_size` / `min_pattern_size` / `min_count` on the OpenAI chat `SamplingParams`); a client may tune or disable (set `max_pattern_size=0`) but never has to send anything.
2. Global operator override `MLXCEL_LOOP_DETECTION` (off / on / `MIN,MAX,COUNT`), usable on any model to force-enable, tune, or force-disable.
3. Gemma 4 family default-on with the conservative built-in threshold `min_pattern_size=1, max_pattern_size=20, min_count=4`, unconditional (does not require tools or a structured-output request, so plain Gemma 4 chat is covered).
4. Disabled for every non-Gemma-4 model, preserving the exact baseline output.

## What changed

- `src/lib/mlxcel-core/src/loop_detection.rs` (new) + `loop_detection_tests.rs` (new): generic detector and config, with thorough unit tests.
- `src/lib/mlxcel-core/src/generate.rs`, `lib.rs`: `SamplingConfig.loop_detection` field, re-exports, and the detector call at all four decode loops.
- `src/server/batch/sequence.rs`: `FinishReason::RepetitionLoop` + transition-table entries + integration-style tests that drive the real scheduler guard to the new finish reason.
- `src/server/batch/scheduler.rs`: the loop guard at the three decode paths plus an info log when it fires; `RepetitionLoop` treated as a healthy cache-donation finish.
- `src/server/request_options.rs` (+ tests), `config.rs`, `startup.rs`: the resolution helper, the `loop_detection` global override (`MLXCEL_LOOP_DETECTION`), and the `model_is_gemma4_family` startup flag.
- `src/server/types/request.rs`, `src/server/routes/chat.rs`, `src/server/routes/native_completion.rs`, `src/execution/sampling.rs`, `src/distributed/disaggregated/serving_protocol.rs`: thread the new fields through the request and sampling-config surfaces (the disaggregated handoff keeps the disabled baseline; serializing the config across the prefill/decode split is a noted follow-up).
- `docs/environment-variables.md`, `docs/supported-models.md`: document the params, the default-on policy, the override surfaces, and the precedence.

## Test plan

- [x] `cargo test -p mlxcel-core --lib loop_detection::` (14 passed): detector boundaries (fires at exactly `min_count`, not before; multi-token loops; clamp / normalization; disabled no-op).
- [x] `cargo test --lib server::request_options` (12 passed): resolution precedence, family default-on without amplifier, non-Gemma disabled, per-request and global overrides.
- [x] `cargo test --lib server::batch::sequence` (30 passed): includes a deterministic integration test that drives the real scheduler stop guard (`detect_repetition_loop` + `transition_to`) to `FinishReason::RepetitionLoop`.
- [x] `cargo check --lib --tests` and `cargo clippy --lib --tests -- -D warnings` clean; `cargo fmt` applied.
- [x] Real-model smoke on `models/gemma-4-e2b-it-qat-4bit` (release `mlxcel-server`, `--jinja --parallel 1`):
  - (a) Family default-on, no per-request loop fields and no tools: a forced repeated tail ended early with `finish_reason=stop`, `completion_tokens=173` of 300, and the server logged `loop detection: ending generation early (repetition loop) generated=173`. A normal tool-laden generation completed coherently (no regression).
  - (b) Non-Gemma `models/qwen3-0.6b-4bit`, same prompt: detection disabled by default, ran to `finish_reason=length` (300 tokens), no firing. Baseline preserved.
  - (c) Gemma 4 with per-request `max_pattern_size: 0`: detection disabled even for the family, ran to `finish_reason=length` (300 tokens), no firing. Override precedence confirmed.

Closes #432

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -128,6 +128,31 @@ The OpenAI audio endpoints (`/v1/audio/speech`, `/v1/audio/transcriptions`, `/v1
 | `MLXCEL_ENABLE_MTP_BATCH` | truthy value | off | **Advanced.** Forces the batched Gemma 4 MTP burst path for parity/debug testing. Not governed by the adaptive policy (issue #333), which scopes to the validated, byte-identical B=1 path. |
 | `MLXCEL_ENABLE_MTP_DEFERRED` | `1` | off | **Advanced.** Enables the deferred greedy verifier path for Gemma 4 MTP when sampling settings allow it. |
 
+## Generation loop detection (issue #432)
+
+N-gram tail repetition detection ends a generation early when the raw generated token stream collapses into a short repeated pattern (a single token such as `様様様様`, or a short block such as `abcdabcd...`). It runs on the raw stream, so it also catches loops inside the reasoning/thought channel and tool-call JSON, not just the final answer. The wire `finish_reason` is `stop`, the same as vLLM. Sampling penalties (`repeat_penalty`, DRY) cannot recover once the logits collapse, which is why this is a stop condition rather than a logit reshaper.
+
+The detector mirrors vLLM's `SamplingParams` fields, with the same JSON names on the OpenAI chat surface:
+
+| Field | Meaning |
+|-------|---------|
+| `max_pattern_size` | Largest N-gram pattern size to scan. `0` (default) disables detection. |
+| `min_pattern_size` | Smallest N-gram pattern size to scan. `0` (default) is treated as `1`; clamped to `<= max_pattern_size`. |
+| `min_count` | Minimum consecutive repeats of a pattern that ends generation. Must be `>= 2`; any smaller value disables detection. |
+
+The preferred activation surface is engine-level: detection is **default-on for the Gemma 4 family with no configuration required**, so a downstream serving app needs no setup and end users see no toggle. The per-request fields and the global env var below are additional tuning surfaces, not the only way to turn it on.
+
+| Variable | Values | Default | Notes |
+|----------|--------|---------|-------|
+| `MLXCEL_LOOP_DETECTION` | `off`/`0`/`none`/`false`/`disabled`, `on`/`default`/`true`/`enabled`, or `MIN,MAX,COUNT` (also `MIN:MAX:COUNT`) | unset | Global operator override for any model. Unset lets the Gemma 4 family default-on apply (non-Gemma models stay disabled). `off` force-disables for every request, including the Gemma 4 family; `on` forces the recommended threshold `1,20,4` for every model; an explicit triple (e.g. `1,20,4`) sets exact values. A malformed value warns and is ignored. |
+
+Resolution precedence, highest first:
+
+1. **Explicit per-request fields.** If a chat request sets any of `max_pattern_size` / `min_pattern_size` / `min_count`, those values are used verbatim, including an explicit disable (`max_pattern_size=0`). A client never has to send anything; the fields are only for tuning or opting out.
+2. **Global override.** `MLXCEL_LOOP_DETECTION`, which an operator can use to force-enable, tune, or force-disable for any model.
+3. **Gemma 4 family default-on.** When the loaded model is in the Gemma 4 family (`Gemma4`, `Gemma4VLM`, `Gemma4Unified`), the conservative threshold `min_pattern_size=1, max_pattern_size=20, min_count=4` is applied unconditionally. This does not require tools or a structured-output request, so plain Gemma 4 chat is covered too. Detection only ends generation when a real repetition loop is present, so a conservative default-on for this family is low risk.
+4. **Disabled.** The default for every non-Gemma-4 model, preserving the exact baseline output.
+
 ## KV cache and TurboQuant variables
 
 Use CLI flags such as `--cache-type-k`, `--cache-type-v`, `--kv-cache-mode`,
--- a/docs/python-client.md
+++ b/docs/python-client.md
@@ -109,7 +109,7 @@ print(llm.chat(messages, max_tokens=128, temperature=0.3))
 
 ## Sampling parameters
 
-Generation methods accept OpenAI sampling fields directly: `max_tokens`, `temperature`, `top_p`, `stop`, `seed`, `presence_penalty`, `frequency_penalty`, `logit_bias`, and `response_format`. Server-specific knobs that are not part of the OpenAI schema (`top_k`, `min_p`, `repetition_penalty`, and DRY settings) are forwarded in the request body. You can also pass an explicit `extra_body={...}` for arbitrary server fields; values you set there win on conflict.
+Generation methods accept OpenAI sampling fields directly: `max_tokens`, `temperature`, `top_p`, `stop`, `seed`, `presence_penalty`, `frequency_penalty`, `logit_bias`, and `response_format`. Server-specific knobs that are not part of the OpenAI schema (`top_k`, `min_p`, `repetition_penalty`, DRY settings, and the vLLM-compatible loop-detection fields `max_pattern_size` / `min_pattern_size` / `min_count`) are forwarded in the request body. You can also pass an explicit `extra_body={...}` for arbitrary server fields; values you set there win on conflict.
 
 ```python
 llm.generate("Once upon a time", max_tokens=200, top_p=0.9, top_k=40, min_p=0.05)
--- a/docs/supported-models.md
+++ b/docs/supported-models.md
@@ -110,6 +110,12 @@ Qwen-style `<think>` models. To turn thinking off, pass
 default via `--chat-template-kwargs` or `LLAMA_ARG_CHAT_TEMPLATE_KWARGS`. A
 per-request value always wins over the server default.
 
+### Loop-detection default-on for the Gemma 4 family
+
+Gemma 4 (31B Dense and 26B-A4B MoE, including the QAT 4-bit checkpoints) has an upstream, weights-level token-repetition collapse documented in [google-deepmind/gemma#622](https://github.com/google-deepmind/gemma/issues/622) and reproduced across other engines, including [vllm-project/vllm#40080](https://github.com/vllm-project/vllm/issues/40080). Generation can degenerate into a single repeated token (or short fragment) that fills the token budget, most often inside the thought channel. Tool declarations and `json_schema` structured-output requests amplify it, but it is not limited to those cases. Sampling penalties do not reliably recover it, because once the logits collapse the top-k candidates are themselves garbage.
+
+mlxcel applies vLLM-style N-gram loop detection to break this, default-on for the Gemma 4 family with no configuration required. For any model in the family (`Gemma4`, `Gemma4VLM`, `Gemma4Unified`), the engine applies the conservative threshold `min_pattern_size=1, max_pattern_size=20, min_count=4` by default; a degenerate run then ends early with `finish_reason` of `stop`. The default-on applies to plain Gemma 4 chat too, so a downstream serving app and its users need no setup and see no toggle. Detection only ends generation when a real repetition loop is present, so this conservative default is low risk. Every non-Gemma-4 model defaults to disabled (bit-exact baseline preserved). The behavior is still tunable on top of the default: a per-request override (the vLLM `max_pattern_size` / `min_pattern_size` / `min_count` fields, including `max_pattern_size=0` to opt out) wins, and a global operator override (`MLXCEL_LOOP_DETECTION`) can tune, force-enable for any model, or force-disable. See [Generation loop detection](environment-variables.md#generation-loop-detection-issue-432) for the field semantics and the full precedence order.
+
 ## Speech-to-text (ASR)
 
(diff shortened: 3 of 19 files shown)


inureyes:

Reviewed the N-gram loop-detection change (detector, server resolution, request/env override surfaces, scheduler/sequence integration, finish-reason wiring). One HIGH finding addressed; all other concerns verified safe.

### HIGH (fixed): unbounded effective `max_pattern_size` -> CPU-DoS
`detect_repetition_loop` runs once per decode step. A per-request `max_pattern_size` (OpenAI chat field) and the `MLXCEL_LOOP_DETECTION` global-env triple both build a `LoopDetectionConfig` with no upper bound. With `min_count=2` and a large `max_pattern_size`, the per-step tail scan is O(n^2) in the generated length and O(n^3) across a long generation, so a single crafted request can pin the worker thread. The Gemma 4 family default (max=20) was already safe.

Fix (commit `91d499d21`): cap the effective scanned pattern size at `MAX_EFFECTIVE_PATTERN_SIZE = 64` inside the detector, the single chokepoint every path flows through (per-request, global-env, family-default), so the bound applies uniformly without touching each entry point. `effective_min_pattern_size` now clamps to the capped maximum so a huge configured max paired with a min above the cap still yields a well-formed scan range. Patterns longer than 64 tokens (well beyond the 1-20 token collapses reported in google-deepmind/gemma#622) are no longer scanned. Added three detector tests: huge-max still fires and terminates, the cap boundary (length-64 fires, length-65 does not), and the min-above-cap no-panic case.

### Verified safe (no change needed)
- **No-op when disabled**: `is_enabled()` short-circuits before touching the slice; the token stream is passed by reference (no clone/alloc). Non-Gemma models and `max_pattern_size=0` keep the bit-exact baseline.
- **Bounded tail / no panic**: the scan reads `generated[len-window..]` only after a `len < window` guard (no underflow), `p.checked_mul(count)` guards overflow, and `p >= 1` so `chunks_exact` never panics.
- **Finish-reason wiring**: `FinishReason::RepetitionLoop` maps to the wire `finish_reason="stop"` via `completion_tokens < max_tokens` (matches vLLM), has its state-transition entries, and is correctly grouped as a healthy cache-donation finish (prompt-cache reuse keys on the exact token prefix and the KV up to the committed loop tokens is mathematically valid, so no poisoning).
- **Input validation / fail-safe**: request fields are `Option<usize>` (malformed -> clean 422); the `MLXCEL_LOOP_DETECTION` parser returns `None` on malformed input with a warning (no panic), failing safe to the per-family policy.

### Verification (narrow targets only)
- `cargo test -p mlxcel-core --lib loop_detection::` -> 17 passed
- `cargo clippy --lib --tests -- -D warnings` -> clean

No remaining CRITICAL/HIGH items.

inureyes:
## PR finalization complete

### Summary

**Tests added** (7 new, in `src/server/startup_tests.rs`): covered all branches of the `MLXCEL_LOOP_DETECTION` env parser (`resolve_loop_detection_env`), which was the only public-surface gap. Branches: unset returns `None`; `off`/`0`/`none`/`false`/`disabled` force-disable; `on`/`default`/`true`/`enabled` apply the recommended threshold; comma-separated and colon-separated numeric triples are parsed and ordered correctly; a malformed value and a wrong-field-count value both return `None` without panicking.

**Documentation updated** (`docs/python-client.md`): added the vLLM-compatible per-request loop-detection fields (`max_pattern_size` / `min_pattern_size` / `min_count`) to the "Sampling parameters" discoverability list alongside the other non-OpenAI knobs. The `MLXCEL_LOOP_DETECTION` env var section, Gemma 4 default-on policy, and precedence table in `docs/environment-variables.md` and `docs/supported-models.md` were already complete and accurate.

**Lint/format**: `cargo fmt` made no changes; `cargo clippy --lib --tests -- -D warnings` was clean before and after.

**`// Used by:` annotations**: `detect_repetition_loop` already carries the correct annotation covering both call sites (`generate.rs` and `server/batch/scheduler.rs`).

**No translations to update**: `docs/en/` and `docs/ko/` do not exist in the repo (the mkdocs config references them as a future destination).

### Verification output

- `cargo check --lib --tests`: finished in ~10 s, 0 errors
- `cargo clippy --lib --tests -- -D warnings`: finished in ~0.2 s, 0 warnings
- `cargo fmt`: no changes
- `cargo test -p mlxcel-core --lib loop_detection::`: 17 passed
- `cargo test --lib server::request_options`: 12 passed
- `cargo test --lib server::batch::sequence`: 30 passed
- `cargo test --lib server::startup::tests::loop_detection`: 7 passed

Benchmark CSVs under `benchmarks/pylm_*.csv` were not staged, stashed, or touched.

Commit: 39be67d8d9bff3c56e2dbe9b22243c954ce2ec39

PR #434, opened 2026-06-25
fix: make conv2d/conv1d fallible at the FFI boundary so conv shape faults don't abort the server

### LINKED ISSUE #427: fix: make conv2d/conv1d fallible at the FFI boundary so conv shape faults don't abort the server
## Summary

`conv2d` (and `conv1d`) are declared in the cxx bridge as returning `UniquePtr<MlxArray>`, not `Result`. When MLX throws inside them (for example a shape mismatch), cxx converts the uncaught C++ exception into `std::terminate`, which aborts the whole server process instead of failing the one request. This is the same crash class that #384 fixed for the audio-synthesis matmul path. This issue extends that fix to the convolution ops so a conv fault on the audio (or vision) path degrades to a recoverable per-request error.

## Background

The Gemma 4 audio conv crash surfaced as:

```
libc++abi: terminating due to uncaught exception of type std::invalid_argument: [conv] Expect the input channels in the input and weight array to match but got shapes - input: (1,12,130,1) and weight: (128,3,1,3)
```

The shape mismatch itself is a weight-layout bug fixed separately (the conv-weight layout-detection issue). This issue is the defense-in-depth half: even with correct layouts, any future data-dependent conv shape fault should not be able to take down the server.

## Mechanism

- `src/lib/mlxcel-core/src/lib.rs:953` declares `fn conv2d(...) -> UniquePtr<MlxArray>` in the cxx bridge (non-Result). The C++ side (`src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:2375`) calls `mlx::core::conv2d` directly.
- Per the #384 commit (`09e5c3d`) rationale: "An MLX C++ exception thrown from a non-Result cxx op aborts the process via std::terminate, which the catch_unwind backstop in the audio worker cannot intercept regardless of the release panic setting." So a throw from `conv2d` aborts the process; the audio worker's `catch_unwind` cannot catch a C++ `std::terminate`.
- MLX validates conv shapes eagerly at op construction, so the throw happens at graph build, exactly where a `Result`-returning cxx wrapper can catch it.

## Scope

- Add `try_conv2d` (and `try_conv1d`) declared `-> Result<...>` in the cxx bridge, mirroring `try_matmul` from #384, so cxx catches an MLX throw at the FFI boundary and returns a recoverable `Err`.
- Route the audio encoder's conv calls through the fallible variants: `SSCPConvBlock` conv2d (`src/audio/encoder.rs:125`) and `ConformerLightConv1d` depthwise conv1d (`src/audio/encoder.rs:374`), since their input shapes are data-dependent (derived from the runtime audio length).
- Follow #384's scoping rationale: leave conv calls on hot paths with statically fixed shapes unchanged to avoid per-op overhead; convert the data-dependent audio-path calls. Vision-encoder conv calls can be evaluated for inclusion but are lower priority (their patch-embed shapes are fixed by config).
- Map a returned `Err` to the existing audio worker error handling so the request fails cleanly with a message instead of aborting.

## Tests

- FFI regression beside the bridge, mirroring #384's `try_matmul` tests: `try_conv2d` on mismatched input/weight channels returns `Err` (the graph-build throw is caught, the process is not aborted), and the happy path matches `conv2d`.

## Acceptance criteria

- A conv shape mismatch on the audio path produces a per-request `Err` (surfaced as a request error), not a process abort.
- An FFI test demonstrates `try_conv2d` returns `Err` on a channel mismatch rather than terminating.
- No behavior change on the happy path; no measurable per-op overhead added to statically-shaped conv hot paths.

## Related

Companion to the conv-weight layout-detection fix (which removes the specific Gemma 4 audio trigger). This issue ensures the broader conv path is crash-isolated. References #384 (`09e5c3d`), which established the `try_*` fallible-op pattern for the audio path.

inureyes:
## Summary

The Gemma 4 Conformer audio encoder calls `conv2d`/`conv1d`, which are declared in the cxx bridge returning `UniquePtr<MlxArray>` (not `Result`). An MLX shape-mismatch throw from a non-`Result` cxx op is converted to `std::terminate` and aborts the whole server process instead of failing the single request. This extends the `try_*` fallible-op pattern established by #384 (`try_matmul`) to convolution so a conv fault on the audio path degrades to a recoverable per-request error. The real crash this isolates: `libc++abi: terminating due to uncaught exception of type std::invalid_argument: [conv] Expect the input channels in the input and weight array to match but got shapes - input: (1,12,130,1) and weight: (128,3,1,3)`.

## Changes

- Add `try_conv2d`/`try_conv1d` to the cxx bridge: Rust decls in `src/lib/mlxcel-core/src/lib.rs` and C++ decls/impls in `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.h` and `mlx_cxx_bridge.cpp`. The C++ bodies are identical to `conv2d`/`conv1d` but the Rust signature is `-> Result<UniquePtr<MlxArray>>`, so cxx wraps the call in try/catch and surfaces MLX's eager graph-build throw as a Rust `Err` instead of letting it cross the FFI boundary uncaught. This mirrors `try_matmul` exactly.
- Route the two data-dependent audio-encoder conv calls through the fallible variants and thread `Result` up the encoder forward chain in `src/audio/encoder.rs`: `SSCPConvBlock::forward` (conv2d), `ConformerLightConv1d::forward` (depthwise conv1d), and their transitive callers `SubSampleConvProjection::forward`, `ConformerBlock::forward`, and the public `AudioEncoder::forward` now return `Result<_, String>` and propagate with `?`.
- Make the Gemma 4 audio embedding builders `Gemma4VLModel::get_input_embeddings_with_audio` and `get_input_embeddings_with_audio_and_cache` return `Result` and propagate the error at every request-path call site: `src/server/model_worker.rs`, `src/multimodal/vlm_runtime.rs`, and `src/commands/generate_vlm.rs` (two sites). All four call sites already sit inside `Result`-returning functions, so propagation is a `.map_err(|e| anyhow::anyhow!(e))?`.
- Keep the vision-only and video builders (`get_input_embeddings`, `get_input_embeddings_with_videos`) infallible: they pass `audio_features = None`, so the fallible audio branch in `..._and_cache` is never reached and the result is always `Ok` (asserted via `.expect` documenting the invariant). The conv-bearing `AudioEncoder` is reached only through `gemma4_vl.rs`; `gemma4_unified` embeds audio features directly via `embed_audio.forward` with no Conformer encoder, so it is untouched.
- Leave statically-shaped conv calls unchanged, matching #384's scoping rationale of converting only the data-dependent audio-path calls: the Mamba fused `conv1d` in `mlx_cxx_kernels.cpp`, `avg_pool2d`'s internal `conv2d`, and the vision patch-embed conv (shapes fixed by config).

## Tests

FFI regression tests beside the bridge in `src/lib/mlxcel-core/src/ffi_tests.rs`, mirroring the `try_matmul` tests: `try_conv2d_returns_err_on_channel_mismatch` (the real Gemma 4 crash shapes, input `(1,12,130,1)` vs weight `(128,3,1,3)`) and `try_conv1d_returns_err_on_channel_mismatch` both assert the throw is caught and returned as `Err` rather than aborting; `try_conv2d_ok_matches_conv2d` asserts the happy path matches the non-fallible `conv2d` in shape `(1,4,4,3)` and value (sum `384.0`).

## Validation

`cargo test -p mlxcel-core --release --features metal,accelerate ffi_tests::try_conv` passes all 3 new tests on Apple Silicon (M1 Ultra). `cargo check --release --features metal,accelerate --all-targets` is clean, confirming the `Result` threading compiles across the whole workspace including test targets. This is a defense-in-depth change: the happy path is byte-identical to the existing ops (the `try_*` C++ bodies match the non-fallible ops), so the FFI `Err` test is the correct validation that the crash is isolated, rather than a real-checkpoint run.

Closes #427


inureyes:

### Intent
> Make the MLX conv2d/conv1d FFI ops fallible so a data-dependent conv shape fault on the Gemma 4 Conformer audio path returns a per-request error instead of std::terminate-aborting the server (mirrors #384's try_matmul).

### Findings Addressed
- None. No CRITICAL/HIGH/MEDIUM findings; no auto-fix needed.

### Remaining Items
- None.

### Verification
- [x] All stated requirements implemented (try_conv2d/try_conv1d added to the bridge in lib.rs/.h/.cpp; audio-encoder conv calls routed through them; Result threaded to request boundaries; FFI tests added)
- [x] Integrated into project code flow (error propagates through SSCPConvBlock, SubSampleConvProjection, ConformerLightConv1d, ConformerBlock, AudioEncoder::forward, the gemma4_vl audio builders, and the four request-path call sites in model_worker / vlm_runtime / generate_vlm)
- [x] Project conventions followed (cxx decls placed next to non-fallible siblings; commit/PR English; no AI attribution; no em dashes; no local references/ paths)
- [x] Existing modules reused where applicable (follows the established try_matmul pattern exactly)
- [x] No unintended structural changes (only the conv data-dependent audio path touched)
- [x] Tests pass (orchestrator confirmed: 3 new ffi_tests::try_conv tests pass; cargo check --all-targets clean)

### Notes
- Happy-path parity confirmed: the C++ try_conv2d/try_conv1d bodies are byte-for-byte equivalent to conv2d/conv1d (same mlx::core call, same arg order); only the Rust `-> Result` decl differs, so inference output is unchanged.
- The three `.expect(...)` calls in the vision-only and video builders are genuinely unreachable: all pass `audio_features = None`, and the fallible audio block in `get_input_embeddings_with_audio_and_cache` is gated on `Some(audio_feat)`, so the function returns `Ok` on those paths.
- Scope discipline verified: the Mamba fused conv1d (mlx_cxx_kernels.cpp), avg_pool2d's internal conv2d, the vision patch-embed conv calls (pixtral/gemma3n/siglip/minicpmo/internvl/rt_detr), the gemma4_unified audio path (embed_audio.forward), and other SSM conv1d calls all remain non-fallible. No per-op overhead added to hot paths.
- Out of scope (not a finding): the nemotron_h_nano_omni audio encoder also runs data-dependent conv calls that remain non-fallible; issue #427 is scoped to the Gemma 4 Conformer path only, so converting it would exceed scope. Possible follow-up.

inureyes:
## PR Finalization Complete

### Tests

Added `try_conv1d_ok_matches_conv1d` in `src/lib/mlxcel-core/src/ffi_tests.rs`. The PR already had error-path tests for both `try_conv2d` and `try_conv1d`, and a happy-path parity test for `try_conv2d`; this fills the symmetric gap for `try_conv1d`. The test uses a valid NLC input `[1,10,2]` with weight `[4,3,2]` at stride 1 / no padding, checks output shape `[1,8,4]`, verifies value identity against the non-fallible `conv1d`, and asserts the expected sum (192.0 = 32 outputs * 6 ones per kernel footprint).

Also applied `cargo fmt` to the pre-existing `try_conv2d_ok_matches_conv2d` test (one `.expect()` chain exceeded the line width limit).

Verification: `cargo test -p mlxcel-core --release --features metal,accelerate ffi_tests::try_conv` passed 4/4.

### Documentation

Updated two files that previously described the `try_matmul` / `try_array_to_raw_bytes` fallible-FFI pattern from PR #384, to record that PR #434 extends the same pattern to `try_conv2d` / `try_conv1d`:

- `docs/audio-api.md` (Fault isolation section)
- `docs/adr/0003-release-panic-unwind-with-core-thread-abort.md` (Consequences bullet and References list)

No new doc files were created; no other docs were touched.

### Lint / Format

`cargo clippy -p mlxcel-core --lib --tests -- -D warnings`: clean, no warnings.
`cargo fmt -p mlxcel-core -- --check`: clean after applying fmt.

### Commit

`80063abc6` on branch `fix/issue-427-fallible-conv-ffi`.

Issue #437, opened 2026-06-25
fix: Gemma 4 server audio (input_audio) places <|audio|> outside the user turn (0-token output)


inureyes:
## Summary

Follow-up to #436 (CLI Gemma 4 audio). The server (`mlxcel-server` / `/v1/chat/completions`) Gemma 4 audio path still places the `<|audio|>` block outside the user turn, the same placement defect #436 fixed for the CLI. #436 fixed the server **resampling** (Fix B, in `prepare_gemma4_audio_embeddings`) but deliberately deferred the server **placement** fix (Fix A) because it requires reworking the prefix-cache-sensitive prompt rendering path, which is out of scope for that bug fix.

## Root cause

The server renders chat messages **text-only**. `build_chat_messages_with_thinking` / `build_raw_json_messages_with_thinking` (`src/server/chat_request.rs`) flatten each message to `m.content.text()` and never emit `{"type":"audio"}` (or image/video) content parts. So for an `input_audio` request the chat template never renders an `<|audio|>` marker into the user turn.

`expand_gemma4_audio_tokens_for_server` (`src/multimodal/vlm_runtime.rs`) then finds no placeholder and uses its "insert before the last token" fallback, which for the Gemma generation prompt (`<bos><start_of_turn>user\n{text}<end_of_turn>\n<start_of_turn>model\n`) lands the `BOA + audio*N + EOA` block inside the model turn. As shown empirically in #436, audio in the model turn forces an immediate EOS, so the server emits 0 tokens for Gemma 4 audio understanding.

The OpenAI content type on the wire is `input_audio` (`src/server/types/request.rs`), while the gemma-4 template expects a content item of type `audio` (`chat_template.jinja`: `{%- elif item['type'] == 'audio' -%} {{- '<|audio|>' -}}`), so even a multimodal rendering path must map `input_audio` to an `audio` content item.

## Evidence (from #436)

- 24 kHz input on the CLI before the fix: encoder output `[1, 64, 1536]` vs 43 placeholders -> 0 tokens, even with the audio token correctly placed (that was the resample defect, now fixed on both CLI and server).
- Audio placed in the model turn (no `<|audio|>` in the user turn) -> immediate EOS. `BOA@index 17` while the two `<start_of_turn>` markers are at `[1, 11]`; the audio block (17..61) sits after the model-turn marker.
[...]

PR #438, opened 2026-06-25
fix: Gemma 4 CLI audio resample to 16kHz and place <|audio|> in user turn

### LINKED ISSUE #436: fix: Gemma 4 audio understanding (CLI --audio) produces no output (missing 16kHz resample + audio token placed outside the user turn)
## Summary

`mlxcel generate -m <gemma-4-e2b/e4b-it> --audio <file.wav> -p "..."` runs the Conformer audio encoder successfully but the model generates 0 tokens (immediate EOS), so Gemma 4 audio understanding is effectively non-functional from the CLI. Root-caused to two independent pre-existing defects; fixing both makes the model produce a coherent transcription. This is unrelated to PR #434 (verified: identical behavior before and after #434).

## Reproduction

`./target/release/mlxcel generate -m models/gemma-4-e2b-it-4bit --audio <24kHz wav> -p "Transcribe this audio." -n 120 -t 0`

Output: encoder loads and runs ("Loaded Gemma4 audio encoder (12 Conformer layers)", "expanded audio into 43 soft tokens"), then `[Generated 0 tokens]`. Text-only and image inputs on the same model work fine.

## Root cause A: the audio placeholder is never placed in the user turn

The CLI chat-template path (`src/commands/generate.rs` `resolve_cli_prompt` / `apply_vlm_chat_template`) only emits image and video content parts, never audio, and for an audio-only request it does not use the VLM template path at all (it calls `apply_user_chat_template`, plain text). The gemma-4 checkpoint chat template DOES support audio content (`chat_template.jinja`: `{%- elif item['type'] == 'audio' -%} {{- '<|audio|>' -}}`), but mlxcel never passes a `{"type":"audio"}` content item, so no `<|audio|>` (id 258881) marker is rendered into the user turn.

Consequently `expand_gemma4_audio_tokens` (`src/commands/generate_vlm.rs`) cannot find an `<|audio|>` placeholder and falls back to inserting `BOA + audio*N + EOA` "before the last token". After the Gemma generation prompt that tail is the model turn, so the audio block lands inside the model's own turn instead of the user turn. Evidence (instrumented prompt_tokens, default prompt): `BOA@index 17`, while the two `<start_of_turn>` (105) markers are at indices `[1, 11]` — the audio block (17..61) is after the model-turn marker (11). The model sees audio in its own turn with nothing to answer and emits EOS immediately.

The server path has the analogous helper `expand_gemma4_audio_tokens_for_server` (`src/multimodal/vlm_runtime.rs`) with the same "insert before last token" fallback; verify whether the server's `input_audio` content (`src/server/types/request.rs`) is rendered into the user turn as `<|audio|>` (the OpenAI content type is `input_audio`, the gemma-4 template expects `audio`), and fix if it has the same defect.

## Root cause B: the waveform is not resampled to 16 kHz before mel extraction (more impactful)

`audio::load_wav_file` returns samples at the file's native sample rate (the test clips are 24000 Hz). The Gemma 4 audio path in `compute_gemma4_audio_embeddings` (`src/commands/generate_vlm.rs`) passes those native-rate samples straight into `AudioFeatureExtractor::extract(waveform, max_length)`, which does NOT resample (its second arg is `max_length`, not a source rate) and assumes `config.sampling_rate = 16000`. So mel extraction runs on 24 kHz audio with a 16 kHz hop (160 samples = 6.67 ms instead of 10 ms), producing ~1.5x too many frames: the Conformer encoder outputs 64 frames for a 1.7 s clip instead of the expected 43. The placeholder count (43 = ceil(duration_ms / 40), correct) no longer matches the encoder output length (64), so `masked_scatter` fills the 43 audio-token positions with the first 43 of the 64 time-distorted frames. The model receives temporally-garbled audio embeddings and emits EOS even when the audio token is correctly placed.

Evidence:
- 24 kHz input: instrumented encoder output shape `[1, 64, 1536]` (sane values, no NaN) vs 43 placeholders -> 0 tokens, even with the audio token correctly in the user turn.
- The same clip resampled to 16 kHz: encoder output `[1, 43, 1536]` (matches the 43 placeholders) and the model engages ("Transcription...").
- e4b + a 3.6 s clip resampled to 16 kHz + audio token in the user turn: a full coherent transcription is produced ("...I can make out: 'Well, honestly, I don't know...'").

A `resample_to_16k(samples, src_rate)` helper already exists in `src/audio/whisper_mel.rs` (used by the Whisper STT path); the Gemma 4 audio path should call it before mel extraction. The same resample is likely needed wherever the server builds Gemma 4 audio features.

## Fix scope

1. Resample the Gemma 4 audio waveform to 16 kHz before mel feature extraction (CLI `compute_gemma4_audio_embeddings` and the image+audio path `compute_gemma4_multimodal_embeddings`, plus the equivalent server feature-building path), reusing `resample_to_16k`. Keep the placeholder count computed from true duration (already correct).
2. Render the `<|audio|>` marker into the USER turn for CLI audio: emit a `{"type":"audio"}` chat-template content part and route audio-bearing requests through the VLM chat template (`resolve_cli_prompt` currently only takes the VLM path when num_images>0 || num_videos>0), so `expand_gemma4_audio_tokens` finds and expands the placeholder in-place instead of the model-turn fallback. Verify/fix the server `input_audio` -> user-turn `<|audio|>` rendering analogously.
3. Keep the "insert before last token" fallback only as a true last resort, or make it insert into the user turn (before the final `<end_of_turn>` of the user turn, not before the generation prompt).

## Acceptance criteria

- [ ] `mlxcel generate -m models/gemma-4-e2b-it-4bit --audio <native-rate wav> -p "Transcribe this audio." -n 120 -t 0` (and the e4b model) produces a coherent transcription, with no need for the user to manually type `<|audio|>` and no external resampling. Validate on a real 24 kHz clip (e.g. `references/mlx-vlm/computer_use/audio/ask.wav` and a longer one).
- [ ] The encoder output frame count matches the audio-token placeholder count for a representative clip after the fix.
- [ ] Text-only and image inputs on the same model are unchanged (no regression).
- [ ] If the server audio path shares either defect, it is fixed too (or a follow-up is filed with the same evidence if out of scope).

## Notes

Found while regression-testing PR #434 (conv FFI fallibility). #434 is confirmed NOT the cause: the 0-token behavior is byte-identical before and after #434. These two defects are pre-existing.

inureyes:
## Summary

Gemma 4 audio understanding from the CLI (`mlxcel generate --audio`) ran the Conformer encoder but generated 0 tokens (immediate EOS) because of two independent pre-existing defects. Fixing both makes the model produce a coherent transcription. Root-caused and reproduced in #436 (this PR is unrelated to #434; the 0-token behavior is byte-identical before and after #434).

## What changed

- Fix B (resampling): the Gemma 4 Conformer audio paths now resample the waveform to 16 kHz via `whisper_mel::resample_to_16k` before mel extraction, because `AudioFeatureExtractor::extract` assumes 16 kHz (160-sample hop = 10 ms) while `load_wav_file` / `load_wav_from_bytes` return native-rate samples (the reference clips are 24 kHz). Without it the encoder emitted ~1.5x too many frames (64 vs 43 for a 1.7 s clip) and desynced from the duration-based placeholder count, garbling the embeddings. Fixed in CLI `compute_gemma4_audio_embeddings` and `compute_gemma4_multimodal_embeddings` (`src/commands/generate_vlm.rs`) and server `prepare_gemma4_audio_embeddings` (`src/server/model_worker.rs`). Duration (and thus `num_audio_tokens`) is rate-invariant, so the placeholder count stays correct and now matches the encoder output length.
- Fix A (placement): a `num_audios` count is threaded through `load_cli_prompt` -> `resolve_cli_prompt` -> `apply_vlm_chat_template` (`src/commands/generate.rs`), which now emits `{"type":"audio"}` content parts into the user turn, gated on a new `ChatTemplateProcessor::supports_audio_content()` (`src/server/chat_template.rs`). The gemma-4 template renders these as `<|audio|>` in the user turn so `expand_gemma4_audio_tokens` finds and expands the placeholder in place, instead of the old "insert before the last token" fallback that landed the audio block inside the model turn and forced an immediate EOS. Routing only switches to the VLM template when the template actually handles audio content, so non-audio models keep the byte-identical text-only path (no regression).
- Server placement deferred: the server renders chat messages text-only (`build_chat_messages_with_thinking` flattens to `content.text()`), so `input_audio` never becomes a user-turn `<|audio|>`. Fixing that requires reworking the prefix-cache-sensitive rendering path, so it is tracked as a focused follow-up in #437 with the evidence from #436. The server resampling half is fixed in this PR.
- Tests: `expand_gemma4_audio_tokens` in-place and fallback expansion; `apply_vlm_chat_template` / `resolve_cli_prompt` audio rendering and routing (and the no-regression text path for non-audio templates); `supports_audio_content` detection; a 24 kHz -> 16 kHz resample length check.

## Test plan

- [x] `cargo check --bin mlxcel --features metal,accelerate` (compiles)
- [x] `cargo fmt --check` and `cargo clippy --bin mlxcel --features metal,accelerate` (clean)
- [x] `cargo test --bin mlxcel -- gemma4_audio vlm_chat_template resolve_cli_prompt` (10 passed)
- [x] `cargo test --lib -- resample_scales_length_from_24k supports_audio_content` (passed)
- [ ] Full `cargo build --release --features metal,accelerate` (run by the orchestrator)
- [ ] Real-model E2E (orchestrator): `mlxcel generate -m models/gemma-4-e2b-it-4bit --audio references/mlx-vlm/computer_use/audio/ask.wav -p "Transcribe this audio." -n 120 -t 0` produces a coherent transcription with no manual `<|audio|>` and no external resampling (the 24 kHz clip is resampled internally), and the encoder frame count equals the placeholder count; gemma-4-e2b text and image inputs remain coherent (no regression).

Closes #436


inureyes:

### Intent
> Make Gemma 4 CLI audio understanding (`mlxcel generate --audio`) produce output instead of 0 tokens, by resampling the waveform to 16 kHz before mel extraction (Fix B) and rendering `<|audio|>` into the user turn (Fix A); server resampling fixed, server placement deferred to #437.

### Findings Addressed
None requiring a fix. The implementation is correct, complete, and integrated.

### Verification
- [x] All stated requirements implemented (Fix A placement + Fix B resampling, CLI; server resampling)
- [x] Integrated into project code flow (`load_cli_prompt` call site threads `num_audios`; E2E confirmed by orchestrator)
- [x] Project conventions followed (no `references/` paths, GitHub URL used; no `// Used by:` needed for CLI-local helpers; no em dashes; no AI attribution; `Closes #436`)
- [x] Existing modules reused (`whisper_mel::resample_to_16k`, `compute_audio_num_tokens`, `supports_*_content` heuristic family)
- [x] Tests pass (orchestrator-run: 10 + 2; logic hand-verified)

### Detailed checks
- Resampling: `resample_to_16k` is applied BEFORE `AudioFeatureExtractor::extract` in all three Conformer paths (`compute_gemma4_audio_embeddings`, `compute_gemma4_multimodal_embeddings`, server `prepare_gemma4_audio_embeddings`). The placeholder count (`compute_audio_num_tokens`) is computed from the native `sample_rate` before resample and is duration-based (rate-invariant), so it still equals the post-resample encoder frame count. A true 16 kHz input is a no-op via the `if sample_rate != 16_000` guard. No Gemma 4 Conformer path was missed: the three are the only request-path `AudioFeatureExtractor` users. The Nemotron H Nano Omni path is a separate extractor that hard-errors on a rate mismatch (out of scope), and the `gemma4_unified` encoder-free path derives its placeholder count from the same `process_audio` chunking it feeds, so it is correctly left untouched.
- Placement: `num_audios` threads `load_cli_prompt -> resolve_cli_prompt -> apply_vlm_chat_template`. The `{"type":"audio"}` part is emitted only when `num_audios > 0 && supports_audio_content()`, and routing only switches to the VLM template under the same gate, so non-audio and non-gemma models keep a byte-identical text-only (and image/video) prompt. Tests cover routing, rendering, omission-when-unsupported, and the no-regression text path. `supports_audio_content()` matches the template's `type == 'audio'` branch and mirrors `supports_image_content` / `supports_video_content`. `expand_gemma4_audio_tokens` expands the first `<|audio|>` marker in place and keeps the model-turn fallback only as a last resort; both new tests match the implementation.
- Server deferral: #437 exists (OPEN) and tracks the server placement (Fix A) half; the server resampling (Fix B) half is fixed in this PR.

### Remaining Items
- None blocking. One non-blocking observation (LOW, informational, no change requested): the audio-routing in `resolve_cli_prompt` reaches `apply_vlm_chat_template`, which early-returns text-only if the template lacks image support. A hypothetical template that handles `type == 'audio'` but not `type == 'image'` would drop the audio marker. This does not affect Gemma 4 (its template handles image content, and image input is unregressed), so it is theoretical only.

inureyes:
## PR Finalization Complete

### Summary

**Tests**: 8 tests added by the developer; all pass. No additional test was added (the server-path resample and `compute_audio_num_tokens` rate-invariance cases would test the same `resample_to_16k` function already covered, so they would add no coverage value). Verification:

- `cargo test --bin mlxcel -- vlm_chat_template_renders_audio_content_part_in_user_turn vlm_chat_template_omits_audio_when_template_lacks_audio_support resolve_cli_prompt_routes_audio_through_vlm_template resolve_cli_prompt_keeps_text_path_for_audio_on_plain_template` passed (4/4)
- `cargo test --bin mlxcel -- gemma4_audio_expands_placeholder_in_place gemma4_audio_fallback_inserts_before_last_token` passed (2/2)
- `cargo test --lib -- supports_audio_content_detects_audio_branch resample_scales_length_from_24k` passed (2/2)

**Documentation**: Updated `docs/supported-models.md` Gemma 4 Unified entry to state that audio understanding is available from the CLI (`--audio <path>`) and that input audio is resampled to 16 kHz automatically. `docs/audio-api.md` was not touched (it covers the Whisper/Kokoro server endpoints, a separate feature).

**Lint/Format**: `cargo clippy --bin mlxcel --features metal,accelerate -- -D warnings` produced zero warnings. Applied rustfmt brace-placement fix in `generate_tests.rs` (the assert! over the pipeline-stage iterator; the three files with let chains could not be formatted by the installed rustfmt 1.8.0, which is a pre-existing limitation unrelated to this PR).

**Commit**: `eaf156748`

PR #582, opened 2026-07-01
feat(phi3-small): implement blocksparse attention

### LINKED ISSUE #511: feat(phi3-small): implement blocksparse attention to replace the dense fallback
## Summary

Phi-3-small uses a blocksparse attention pattern (local blocks plus a vertical stride) on its non-dense layers, with periodic fully-dense layers. mlxcel currently runs **dense attention everywhere** as a simplicity fallback. This is correct at short context but diverges from the trained sparsity pattern at long context and forgoes the intended long-sequence efficiency.

## Current state in mlxcel

- `src/models/phi3small.rs:21` documents it: *"Blocksparse attention pattern (with dense fallback)"*.
- The blocksparse config is parsed but unused for actual sparsity: `blocksparse_block_size` (`phi3small.rs:66`), `blocksparse_num_local_blocks` (`phi3small.rs:69`), `blocksparse_vert_stride` (`phi3small.rs:72`), `dense_attention_every_n_layers` (`phi3small.rs:38`).
- A per-layer `block_sparse` flag is computed (`phi3small.rs:257`: `!(layer_idx + 1).is_multiple_of(dense_attention_every_n_layers)`) and stored on the layer (`phi3small.rs:140`, `:269`), but the attention path ignores it: `phi3small.rs:207-208` notes *"block sparse attention would be more efficient for long sequences, but we use dense attention for simplicity (fallback)"*.

**Consequence:** outputs match a dense reference at short context but diverge from the Phi-3-small reference once the sequence exceeds the local-block window, and the long-sequence speedup the architecture targets is unrealized.

## What needs to be built

1. **Blocksparse mask / gather** for `block_sparse` layers: attend to (a) the local window of `blocksparse_num_local_blocks` blocks of size `blocksparse_block_size` ending at the current query block, plus (b) earlier blocks selected by `blocksparse_vert_stride`. Keep the existing dense path for layers where `block_sparse == false` (the every-`dense_attention_every_n_layers` dense layers).
2. Prefer a gather-based KV selection for the decode (`L==1`) path over building a full sparse mask each step, consistent with how other long-context paths avoid per-step mask allocation.
3. Thread the per-layer `block_sparse` flag (already computed) into attention so dense and sparse layers diverge correctly.

## Acceptance criteria

- [ ] Blocksparse pattern implemented for non-dense layers; dense layers unchanged.
- [ ] Mask/selection matches the Phi-3-small reference pattern (local blocks + vertical stride) on a constructed case.
- [ ] Short-context parity: when the sequence fits within the local window, output is RMS-equivalent to the current dense fallback.
- [ ] Long-context output matches the reference sparsity (not dense).
- [ ] Validated on a real Phi-3-small checkpoint: generation runs and stays coherent beyond the local-block window.

## Files

- `src/models/phi3small.rs` (attention path, mask construction)
- a shared blocksparse-mask helper if warranted

## Context

Discovered during an upstream-sync review while auditing sparse-attention deferrals (the DeepSeek-V3.2 indexer deferral comment explicitly references this Phi-3-small blocksparse deferral as a sibling case).

inureyes:
## Summary

Replace the dense-attention fallback on Phi-3-small's non-dense layers with the real blocksparse pattern (local block window + per-head vertical stride), while keeping the periodic fully-dense layers on the plain causal path. This makes long-context output follow the trained sparsity instead of diverging toward dense.

## What changed

- `src/models/phi3small.rs`
  - Added `build_blocksparse_mask(...)`: builds the additive `[1, n_heads, q_len, kv_len]` mask on-device (broadcasting ops only, nothing of size `q_len * kv_len` materialized on the host). Faithful port of `mlx_lm/models/phi3small.py` (`Attention._block_sparse_mask` + `_block_sparse_attention`), including the reference's right-aligned query-block assignment and the per-head vertical stride `(kb + head + 1) % vert_stride == 0`. Token-level causality is folded into the same mask so it reproduces the reference `mask + dense_mask` combination exactly.
  - Split the attention dispatch into `dense_attention` (unchanged causal path, used by the periodic dense layers) and `blocksparse_attention` (non-dense layers), threading the already-computed per-layer `block_sparse` flag into the attention path.
  - In-window fast path: when the whole key length fits inside the local window (`kv_len <= num_local_blocks * block_size`) the pattern is exactly causal, so the dense path is reused. This gives byte-identical short-context output and, for in-window `L == 1` decode, allocates no per-step mask (the maskless single-query fast path selects the whole cache).
  - Stored `blocksparse_block_size` / `_num_local_blocks` / `_vert_stride` on `Attention`.
- `src/models/phi3small_tests.rs` (new): config parsing/defaults, `block_sparse` flag pattern, device-mask vs host-reference agreement (bulk + explicit key-block set), short-context mask==causal and RMS parity against the dense fallback, long-context sparser-than-causal + runtime divergence, and a decode single-row mask check.

## Decode (`L == 1`) note

The issue asks to prefer gather-based KV selection over a full per-step mask. In-window decode does exactly that (no mask; trailing-cache selection via the maskless fast path). Beyond the window, the union of attended blocks across all heads spans the full causal range for Phi-3-small's head count (`vert_stride` divides the head set), so a per-head gather cannot shrink the fused-SDPA work and would require a ragged per-head kernel; we build a single `[1, n_heads, 1, kv_len]` additive mask (one row) and let the fused kernel consume it.

## Test plan

- [x] `cargo test --release phi3small --features metal,accelerate` (mask pattern, short-context dense parity, long-context sparsity, decode row)

## Validation scope (honest)

Validated: unit tests above, including short-context RMS parity with the dense fallback and mask-pattern agreement with the mlx-lm reference formula on constructed cases, run on Apple Silicon Metal. NOT run: generation on a real Phi-3-small checkpoint (no Phi-3-small checkpoint available in this environment; Phi-3-small is a ~7B model and a multi-GB download was out of scope here). Long-context behavior is validated at the mask/selection and SDPA-output level, not via a real-model long-context generation.

Closes #511

### DIFF
--- a/src/models/phi3small.rs
+++ b/src/models/phi3small.rs
@@ -18,7 +18,8 @@
 //! - Fused query_key_value projection
 //! - GeGELU activation (gelu * (linear + 1))
 //! - mup scaling (attention, embedding, width multipliers)
-//! - Blocksparse attention pattern (with dense fallback)
+//! - Blocksparse attention pattern (local block window + per-head vertical
+//!   stride) on non-dense layers; periodic fully-dense layers stay causal
 //! - LayerNorm (not RMSNorm)
 //! - bias=True for all projections
 //! - Tied word embeddings
@@ -138,6 +139,9 @@ pub struct Attention {
     pub rope_dims: i32,
     pub rope_base: f32,
     pub block_sparse: bool, // Whether this layer uses block sparse attention
+    pub blocksparse_block_size: i32,
+    pub blocksparse_num_local_blocks: i32,
+    pub blocksparse_vert_stride: i32,
 }
 
 impl Attention {
@@ -203,18 +207,16 @@ impl Attention {
         // Update KV cache and get sliced views
         let (cache_k, cache_v) = cache.update_and_fetch(k, v);
 
-        // Scaled dot-product attention
-        // Note: block sparse attention would be more efficient for long sequences,
-        // but we use dense attention for simplicity (fallback)
-        let attn_out = if l > 1 && mask.is_none() {
-            mlxcel_core::causal_attention(&q, &cache_k, &cache_v, self.scale, 0.0, 0)
+        // Scaled dot-product attention.
+        //
+        // Non-dense layers use the Phi-3-small blocksparse pattern (local window
+        // of blocks plus a per-head vertical stride); the periodic dense layers
+        // (`block_sparse == false`) keep the plain causal path. See
+        // `blocksparse_attention` for the mask/selection details.
+        let attn_out = if self.block_sparse {
+            self.blocksparse_attention(&q, &cache_k, &cache_v, l, offset, mask)
         } else {
-            let mask_ptr = mask.map(|m| m as *const _).unwrap_or(std::ptr::null());
-            unsafe {
-                mlxcel_core::layers::attention_from_ptr(
-                    &q, &cache_k, &cache_v, self.scale, mask_ptr, 0.0, 0,
-                )
-            }
+            self.dense_attention(&q, &cache_k, &cache_v, l, mask)
         };
 
         // Transpose back and reshape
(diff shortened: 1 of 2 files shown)


inureyes:
Duplicate of #581, which already squash-merged this exact change (commit 92396a8d9, phi3-small blocksparse attention) into main. #582 was opened automatically on the same branch after #581 merged and carries no content difference. Closing as superseded.

PR #653, opened 2026-07-04
test(qwen3.5): preserve MoE expert stack ordering

### LINKED ISSUE #670: fix(qwen3.5-moe): sanitize renames stacked experts to w1/w2/w3 but SparseMoeBlock now loads gate_proj/up_proj/down_proj (load regression since #588)
## Summary
Every qwen3.5-MoE / qwen3.6-MoE checkpoint fails to load with `Missing weight: model.layers.0.mlp.switch_mlp.gate_proj`. Found by the 2026-07-06 full benchmark sweep: `qwen3.5-35b-a3b-4bit` and `qwen3.6-35b-a3b-4bit` benched fine on 2026-06-12 (70 and 67 tok/s decode) and are FAIL:bench now.

## Root cause
- PR lablup/mlxcel#588 (2026-06-16, "make real checkpoints load and run") renamed the shared `qwen3_next` `SwitchGLU`/`SparseMoeBlock` expert projections from `w1/w2/w3` to `gate_proj/up_proj/down_proj` (`src/models/qwen3_next.rs`, `SwitchGLU::from_weights` now requests `{prefix}.gate_proj` etc.).
- `src/models/qwen3_5.rs` reuses that `SparseMoeBlock`, but its `sanitize_weights` still normalizes toward the OLD naming: step 6 stacks `experts.{e}.{w1,w2,w3}` into `switch_mlp.{w1,w2,w3}`, and step 8 renames pre-stacked `switch_mlp.{gate_proj,up_proj,down_proj}` into `switch_mlp.{w1,w3,w2}`. The block then asks for `switch_mlp.gate_proj` and finds nothing.
- PR lablup/mlxcel#587 (2026-07-01) added a per-expert `gate_proj/up_proj/down_proj` stacking pass, but it also targets the stale `w1/w3/w2` slot names. Its unit tests assert sanitize OUTPUT keys only and never run `Qwen35Model::from_weights`, so they pass against the wrong convention: the same "synthetic passes, real checkpoint fails" class as lablup/mlxcel#588 itself.

## Fix
Align `qwen3_5::sanitize_weights` with the post-#588 block naming: stack `experts.{e}.w1/w2/w3` into `switch_mlp.gate_proj/down_proj/up_proj` (w1 to gate, w2 to down, w3 to up), keep pre-stacked `switch_mlp.{gate,up,down}_proj` untouched, REVERSE step 8 to rename legacy stacked `w1/w2/w3` into the gate/up/down names, and retarget the #587 per-expert pass at the gate/up/down slots. Update the #587 tests to the new expected keys, and add the missing gate: a test that runs sanitize + `Qwen35Model::from_weights` end to end on tiny synthetic MoE weights in each naming convention (pre-stacked gate/up/down, pre-stacked w1/w2/w3, per-expert w1/w2/w3, per-expert gate/up/down).

## Acceptance Criteria
- [ ] `qwen3_5::sanitize_weights` produces `switch_mlp.gate_proj/up_proj/down_proj` for all four input naming conventions (pre-stacked gate/up/down, pre-stacked w1/w2/w3, per-expert w1/w2/w3, per-expert gate/up/down).
- [ ] New end-to-end test runs sanitize + `Qwen35Model::from_weights` on tiny synthetic MoE weights for each convention; #587 unit tests updated to the new expected keys.
- [ ] Real-checkpoint load + generate succeeds on `models/qwen3.5-35b-a3b-4bit` and `models/qwen3.6-35b-a3b-4bit` (both are VLM exports with `language_model.*` nesting, routed through `load_qwen3_5_moe_vlm`).
- [ ] Dense qwen3.5 regression check passes (e.g. qwen3.5-4b-4bit).
- [ ] Both MoE models re-added to the 2026-07-06 benchmark CSV.

ukint-vs:
## Summary

- keep this PR as a focused test-only follow-up now that the loader/sanitizer implementation landed through #671
- remove the redundant ignored output-key-only test from `src/models/qwen3_5_tests.rs`
- add a sanitizer unit test that verifies per-expert `gate_proj`, `up_proj`, and `down_proj` tensors are stacked into the current `switch_mlp.*_proj` loader names while preserving expert order and projection values

## Audit result

The original implementation delta is already present on current `main` via #671, which closed #670. This branch now leaves the main `qwen3_5::sanitize_weights` implementation unchanged and only contributes additional regression coverage aligned with the current `SwitchGLU`/`SparseMoeBlock` naming convention.

## Validation

- `cargo test qwen3_5::sanitize_tests -- --nocapture`

Refs #670 and #671


inureyes:
Thanks for the contribution and for tracking down the Qwen3.5 MoE switch projection naming mismatch. I audited the PR against current `main` and the related fix that landed in #671.

The main loader/sanitizer implementation is now already equivalent on `main`: `src/models/qwen3_5.rs` has no remaining delta for this PR after #671, and #670 is closed. The useful part still worth preserving here is the extra test coverage, but it should be adjusted to the current implementation style so it complements the stronger load-through regression test that #671 added.

I will push a small follow-up commit to this PR that drops the now-redundant implementation delta and keeps only a cleaned-up, focused test. After that I will rerun the relevant qwen3_5 sanitizer regression and merge if the branch remains clean.

PR #659, opened 2026-07-05
feat(models): port LLaDA-2 MoE (masked-diffusion LLM, DeepSeek-V3 MoE FFN)

### LINKED ISSUE #546: feat(models): port LLaDA-2 MoE (diffusion LLM (MoE))
## Summary

Add LLaDA-2 MoE support to mlxcel. LLaDA-2 is a masked-diffusion language model with a DeepSeek-V3-style MoE FFN: checkpoints ship `config.json` with `model_type: llada2_moe`, and text is generated by iterative block-wise unmasking of `<|mask|>` tokens rather than autoregressive decode. mlxcel currently has no arm for it in `src/models/detection.rs`, so a `llada2_moe` checkpoint fails with "Unsupported model type".

Verified real checkpoint: `inclusionAI/LLaDA2.0-mini` (bf16, 7 safetensors shards + `model.safetensors.index.json`, ~32.5 GB, 20 layers, 256 experts). All config values and weight-key patterns below were checked against it.

## Current state in mlxcel

mlxcel already serves one diffusion LM end to end, so the second architecture plugs into existing seams:

- **Engine precedent**: `src/models/diffusion_gemma/` (`mod.rs` model + config parsing, `generate.rs` iterative denoising loop with pure host-side helpers, co-located `tests.rs`). Its generation algorithm does NOT match LLaDA-2 (random-token canvas + entropy-bound acceptance + self-conditioning vs. mask-token canvas + confidence-threshold unmasking), but the structural pattern does: model-owned batch-1 loop, `LanguageModel` trait impl with `supports_batching() == false`, read-only prefix KV plus a per-step transient block forward (`forward_canvas` + `KVCache::visible_state()` in `src/lib/mlxcel-core/src/cache.rs`).
- **Model registration**: `LoadedModel::DiffusionGemma` variant + `LanguageModel` delegation macro in `src/loaded_model.rs`; `ModelType` enum, supported-types list, and display-name table in `src/models/mod.rs`; metadata row in `src/model_metadata.rs`; construction arm in `src/loading/nonstandard.rs` (`ModelType::DiffusionGemma => load_from_path(...).map(LoadedModel::DiffusionGemma)`).
- **Server dispatch**: `src/server/model_worker.rs` has two match points (batched-worker setup, ~line 286, and the legacy worker, ~line 805) that branch `LoadedModel::DiffusionGemma` into `run_diffusion_worker_loop` before any scheduler setup. `src/server/diffusion_worker.rs` (+ `diffusion_worker_tests.rs`) implements the batch-1 serving loop over the shared `mpsc` `ModelRequest` channel: tokenization, `StreamingDecodeState` incremental detokenization, per-step cooperative cancellation, `GenerateEvent::Token/Done/Error` streaming, finish-reason mapping. This plumbing is generic; only its model parameter (`&DiffusionGemmaModel`) and options mapping (`DiffusionGenerateOptions`, `DiffusionSamplerKind`) are gemma-specific.
- **Flags**: `--max-denoising-steps`, `--diffusion-sampler`, `--diffusion-threshold` are threaded through `src/main.rs` (`DiffusionCliOptions`), `src/server/cli_input.rs`, `src/server/config.rs`, `src/server/startup.rs`, `src/server/model_provider.rs`. The sampler/threshold pair is semantically gemma-only; LLaDA-2 takes its sampling knobs from the request (temperature/top-k/top-p) plus new loop knobs.
- **CLI**: `src/commands/generate.rs` (~line 1760) routes `LoadedModel::DiffusionGemma` to `src/commands/generate_diffusion.rs`; `src/commands/chat.rs` (~line 177) rejects interactive chat for diffusion models.
- **MoE machinery**: `src/models/switch_layers.rs` provides `SwitchGLU`/`SwitchLinear` (including automatic stacking of per-expert weights, see Weight loading), `moe_weighted_sum`, and `group_mask_scores`. `src/models/deepseek_v3.rs` has the closest gate precedent (`MoEGate`: sigmoid scoring in f32, additive bias for selection only, group-limited top-k, `norm_topk_prob`, `routed_scaling_factor`) and the `first_k_dense_replace` dense/MoE layer mix.

## Architecture

Decoder-only transformer, pre-norm RMSNorm residual blocks, bidirectional attention (no causal mask anywhere), untied LM head. All linear weights in the checkpoint use the (out_features, in_features) layout, which is what mlxcel's `UnifiedLinear`/matmul convention consumes directly; no transposition is needed at load time.

### config.json keys (defaults apply when the key is absent)

| Key | Default | `LLaDA2.0-mini` |
|---|---|---|
| `model_type` | `"llada2_moe"` | same |
| `vocab_size` | 157184 | 157184 |
| `hidden_size` | 2048 | 2048 |
| `intermediate_size` (dense layers) | 5120 | 5120 |
| `num_hidden_layers` | 20 | 20 |
| `num_attention_heads` | 16 | 16 |
| `num_key_value_heads` | 4 | 4 |
| `head_dim` | `hidden_size / num_attention_heads` | 128 |
| `partial_rotary_factor` | 0.5 | 0.5 |
| `rotary_dim` | `head_dim * partial_rotary_factor` | 64 |
| `rope_theta` | 600000 | 600000 |
| `rope_scaling` | null | null |
| `use_qk_norm` | **true** | key absent (defaults on; the checkpoint ships `query_layernorm`/`key_layernorm` weights) |
| `use_qkv_bias` | false | false |
| `use_bias` | false | false |
| `rms_norm_eps` | 1e-6 | 1e-6 |
| `hidden_act` | `"silu"` | silu |
| `tie_word_embeddings` | false | false |
| `num_experts` | 256 | 256 |
| `num_experts_per_tok` | 8 | 8 |
| `n_group` | 8 | 8 |
| `topk_group` | 4 | 4 |
| `num_shared_experts` | 1 | 1 |
| `moe_intermediate_size` | 512 | 512 |
| `first_k_dense_replace` | 1 | 1 |
| `routed_scaling_factor` | 2.5 | 2.5 |
| `norm_topk_prob` | true | true |
| `score_function` | `"sigmoid"` | sigmoid |
| `router_dtype` | `"fp32"` | fp32 |
| `moe_router_enable_expert_bias` | true | true |
| `pad_token_id` | 156892 | 156892 |
| `eos_token_id` | **defaults to `pad_token_id`** | key absent |
| `mask_token_id` | **156895** (`<|mask|>`) | key absent |
| `max_position_embeddings` | 32768 | 32768 |
| `use_sliding_window` | false | false (ignore `sliding_window`/`max_window_layers`) |

The bolded defaults are load-bearing: the real config omits `eos_token_id`, `mask_token_id`, and `use_qk_norm`, so serde defaults must supply them. The repo also has **no `generation_config.json`**, so `crate::read_eos_token_ids` contributes nothing and the EOS set comes from the default above plus request stop ids. The tokenizer (`tokenizer_config.json`) defines `eos_token = "<|endoftext|>"`, `mask_token = "<|mask|>"`, and a chat template (role-tag style, `<role>SYSTEM</role>`...), which mlxcel's existing chat-template path applies unchanged.

### Layer structure

Per layer `i` in `0..num_hidden_layers`, with input `x` of shape (batch=1, seq, hidden):

```
h   = x + Attention(RMSNorm_input(x))
out = h + FFN(RMSNorm_post_attention(h))
```

FFN is a dense SwiGLU MLP for `i < first_k_dense_replace` (layer 0 only on the mini) and the MoE block otherwise. Final: `RMSNorm(model.norm)` then `lm_head` (separate weight, embeddings not tied).

### Attention

- Fused QKV: one weight `attention.query_key_value.weight` of shape ((num_heads + 2 * num_kv_heads) * head_dim, hidden) = (3072, 2048) on the mini. Row order along the out_features axis: Q rows first (num_heads * head_dim = 2048 rows), then K (num_kv_heads * head_dim = 512), then V (512). After `y = x @ W^T`, reshape y from (1, seq, 3072) to (1, seq, num_heads + 2 * num_kv_heads, head_dim) = (1, seq, 24, 128) and split along axis 2 at boundaries [16, 20]: q (1, seq, 16, 128), k (1, seq, 4, 128), v (1, seq, 4, 128). This split is correct because the reshape groups consecutive `head_dim`-sized row blocks into heads in the same order as the weight rows.
- QK-norm (`use_qk_norm`): per-head RMSNorm with weight shape (head_dim,) applied to q and k over the last axis, before the head transpose.
- Transpose to (1, heads, seq, head_dim), then partial RoPE: rotate only channels `0..rotary_dim` (64 of 128) of the last axis, pass channels `rotary_dim..head_dim` through unrotated, non-traditional (split-halves) pairing, base `rope_theta`, scale 1.0. The position offset for a token at sequence position `p` is `p` (during block denoising: prefix length + index within the block).
- GQA: 4 KV heads shared by 16 query heads; scale `head_dim^-0.5`; output projection `attention.dense` of shape (hidden, num_heads * head_dim).
- **Masking semantics during diffusion: attention is fully bidirectional.** No causal mask exists at any stage (prefill, denoising, commit). With batch 1 and no padding, no mask tensor is needed at all: every query position attends to every key position visible at that step (committed prefix + current block).

### MoE routing (exact math, per token hidden vector x of shape (hidden,))

```
logits  = f32(x) @ f32(gate.weight)^T                # gate.weight (num_experts, hidden) -> logits (num_experts,), router runs in float32
scores  = sigmoid(logits)                            # (num_experts,), score_function "sigmoid"
biased  = scores + expert_bias                       # expert_bias (num_experts,) f32; used for SELECTION only

# group-limited top-k over `biased`:
groups        = reshape(biased, (n_group, num_experts / n_group))     # (8, 32)
group_score_g = sum of the top-2 values within group g (over the 32-expert axis)
keep the topk_group=4 groups with the largest group_score; set every expert
of a non-kept group to -inf in `biased`

topk_idx    = indices of the num_experts_per_tok=8 largest entries of masked `biased`  (over the expert axis)
topk_weight = scores[topk_idx]                       # gathered from the UNBIASED sigmoid scores
if norm_topk_prob and k > 1:
    topk_weight = topk_weight / (sum(topk_weight) + 1e-20)
topk_weight = topk_weight * routed_scaling_factor    # 2.5

y = sum_j topk_weight[j] * Expert_{topk_idx[j]}(x)   # expert = down(silu(gate(x)) * up(x)), dims 2048 -> 512 -> 2048
y = y + SharedExpert(x)                              # dense SwiGLU, intermediate = moe_intermediate_size * num_shared_experts = 512
return cast(y, dtype of x)
```

Notes for the implementation: `group_mask_scores` in `src/models/switch_layers.rs` zeroes non-kept groups instead of setting them to -inf; that shortcut is only equivalent when all biased scores are non-negative, and `expert_bias` can be negative, so either mask with a large negative constant or verify the invariant on the loaded bias. The `+ 1e-20` in the normalization is part of the spec (the DeepSeek-V3 gate in `src/models/deepseek_v3.rs` divides without it). The bias key is `expert_bias`, not `e_score_correction_bias`.

### Safetensors weight keys (complete set, verified against the real index)

```
model.word_embeddings.weight                                   (vocab_size, hidden)   # NOT model.embed_tokens
model.layers.{i}.input_layernorm.weight                        (hidden,)
model.layers.{i}.attention.query_key_value.weight              ((num_heads + 2*num_kv_heads) * head_dim, hidden)
model.layers.{i}.attention.query_layernorm.weight              (head_dim,)
model.layers.{i}.attention.key_layernorm.weight                (head_dim,)
model.layers.{i}.attention.dense.weight                        (hidden, num_heads * head_dim)
model.layers.{i}.post_attention_layernorm.weight               (hidden,)
# dense layers (i < first_k_dense_replace):
model.layers.{i}.mlp.gate_proj.weight                          (intermediate_size, hidden)
model.layers.{i}.mlp.up_proj.weight                            (intermediate_size, hidden)
model.layers.{i}.mlp.down_proj.weight                          (hidden, intermediate_size)
# MoE layers (i >= first_k_dense_replace):
model.layers.{i}.mlp.gate.weight                               (num_experts, hidden)
model.layers.{i}.mlp.gate.expert_bias                          (num_experts,)
model.layers.{i}.mlp.experts.{e}.gate_proj.weight              (moe_intermediate_size, hidden)      # e in 0..num_experts
model.layers.{i}.mlp.experts.{e}.up_proj.weight                (moe_intermediate_size, hidden)
model.layers.{i}.mlp.experts.{e}.down_proj.weight              (hidden, moe_intermediate_size)
model.layers.{i}.mlp.shared_experts.gate_proj.weight           (moe_intermediate_size * num_shared_experts, hidden)
model.layers.{i}.mlp.shared_experts.up_proj.weight             (moe_intermediate_size * num_shared_experts, hidden)
model.layers.{i}.mlp.shared_experts.down_proj.weight           (hidden, moe_intermediate_size * num_shared_experts)
model.norm.weight                                              (hidden,)
lm_head.weight                                                 (vocab_size, hidden)
```

## Generation loop

Semi-autoregressive block diffusion. The sequence lives on axis 1 of an int32 canvas `x` of shape (batch=1, total_len); position index `p` counts from 0 at the first prompt token. Defaults: `block_length B = 32`, `steps S = 32` (clamped to `min(S, gen_length)`, at least 1), `threshold T = 0.95`, `min_threshold = T` (constant threshold unless configured lower), greedy sampling at temperature 0 with optional temperature/top-k/top-p.

```
P          = prompt length; gen_length = requested new tokens (server: max_tokens)
num_blocks = ceil((P + gen_length) / B)
total_len  = num_blocks * B
x[0, 0..P)         = prompt ids
x[0, P..total_len) = mask_token_id

# 1. Prefill: commit full prompt blocks into per-layer prefix KV caches.
prefill_blocks = floor(P / B)
for pb in 0..prefill_blocks:
    forward x[:, pb*B..(pb+1)*B), APPENDING K/V to the prefix caches   # bidirectional

# 2. Block loop. The first generation block may straddle the prompt tail.
for nb in prefill_blocks..num_blocks:
    s = nb * B; e = (nb + 1) * B                 # sequence positions [s, e) on axis 1
    # prompt guard: block position j (0-based, j in 0..B) is a prompt token iff s + j < P
    for t in 1..=S:
        active[j] = (x[0, s+j] == mask_token_id)                    # (B,) bool
        if no j has active[j]: break                                # block fully unmasked
        logits = forward(x[:, s..e), prefix caches READ-ONLY)       # (1, B, vocab_size)
            # block K/V are computed fresh each iteration and discarded;
            # RoPE position of block index j is s + j
        for each j: x0[j]   = sample(logits[0, j, :]) over the vocab axis (axis 2)
                    conf[j] = softmax_f32(logits[0, j, :])[x0[j]]   # prob of the chosen token
        thr = T - (T - min_threshold) * (t - 1) / max(1, S - 1)     # linear decay across steps
        transfer = { j : active[j] and conf[j] > thr }
        if transfer is empty:
            transfer = { argmax over j of conf[j] restricted to active[j] }   # exactly one: guarantees progress
        for j in transfer: x[0, s+j] = x0[j]     # unmask; committed tokens are NEVER re-masked or rewritten
    # 3. Commit: one more forward over [s, e), this time APPENDING its K/V to the prefix caches.
    forward x[:, s..e), prefix caches APPEND
    stream tokens x[0, max(s, P)..e) to the caller
    if x[0, P..e) contains no mask tokens and contains an EOS id: break

# 4. Output: x[0, P..P+gen_length) truncated at the first EOS.
```

Unmasking criterion recap: per iteration, reveal every still-masked block position whose chosen-token probability clears the (linearly decaying) threshold; when none clears it, reveal exactly the single highest-confidence masked position. Because every iteration reveals at least one position, a block always finishes within `B` iterations; keep `S >= B` (the defaults pair `S == B == 32`) or leftover mask tokens would be committed. Stopping: the EOS set is `{eos_token_id}` (default `pad_token_id` = 156892) unioned with request stop ids; on the server path EOS stops generation without being emitted, matching the existing diffusion worker convention. There is no confidence-based re-masking and no rewriting of committed tokens.

## Implementation plan

1. `src/models/llada2_moe/mod.rs` (new): config struct with the exact serde defaults above (including the derived `head_dim`/`rotary_dim`/`eos_token_id` fallbacks in a post-parse step), attention (fused-QKV runtime split, QK-norm, partial RoPE), dense MLP, MoE block. Reuse `SwitchGLU` from `src/models/switch_layers.rs` for the routed experts and `moe_weighted_sum` for the combine; model the gate on `MoEGate` in `src/models/deepseek_v3.rs` with the three deltas noted above (key name `expert_bias`, `+ 1e-20` normalization, group-mask sign caveat). Implement `LanguageModel` with `supports_batching() == false` and `supports_padded_prefill() == false`, like `DiffusionGemmaModel`.
2. `src/models/llada2_moe/generate.rs` (new): the block-unmasking engine with its own options struct (block_length, steps, threshold/min_threshold, temperature/top-k/top-p, max_new_tokens, extra EOS ids, cancellation flag). Follow `src/models/diffusion_gemma/generate.rs`: keep the schedule and transfer-mask selection as pure host-side functions so they are unit-testable without a model; use the prefix-cache pattern (`KVCache` append for prefill/commit, read-only prefix + transient block K/V for denoising, as `forward_canvas` does via `visible_state()`).
3. `src/models/llada2_moe/tests.rs` (new, co-located via `#[path = "tests.rs"]`).
4. Registration: `src/models/detection.rs` arm `"llada2_moe" => Ok(ModelType::Llada2Moe)`; `src/models/mod.rs` (`pub mod` + `pub use`, `ModelType` variant, supported-types list, display-name table, variant-list macro); `src/model_metadata.rs` row (kind Text; adapter unsupported message like DiffusionGemma's); constructor arm where the metadata routes it, mirroring the `ModelType::DiffusionGemma` arm in `src/loading/nonstandard.rs`; `src/loaded_model.rs` variant `Llada2Moe(models::Llada2MoeModel)` plus the delegation-macro arm.
5. CLI: extend the diffusion branch in `src/commands/generate.rs` and add a driver (either generalize `src/commands/generate_diffusion.rs` or a sibling module); keep the `src/commands/chat.rs` interactive-chat rejection for diffusion models.
6. Server: extend both `LoadedModel::DiffusionGemma` match points in `src/server/model_worker.rs` to also route `LoadedModel::Llada2Moe` into the diffusion worker. In `src/server/diffusion_worker.rs`, introduce a small internal dispatch (an enum over the two model refs, or a narrow trait covering "generate streaming from prompt ids + options") so `run_diffusion_worker_loop` keeps one copy of the tokenize/stream/cancel/usage plumbing; add an options mapper (`ServerGenerateOptions` to LLaDA-2 options: max_tokens, temperature, top-p/top-k, stop ids union) next to `diffusion_options_from_server`. Do not change any DiffusionGemma behavior: its options mapping, samplers, and flags stay as they are, and `--diffusion-sampler`/`--diffusion-threshold` remain gemma-only. `--max-denoising-steps` may map to the LLaDA-2 `steps` override.
7. Docs: add the entry to `docs/supported-models.md`.

## Weight loading and sanitize

- No key renames are needed: the checkpoint keys are flat under `model.` / `lm_head.` exactly as listed above, sharded with a standard `model.safetensors.index.json`.
- **Per-expert stacking**: the checkpoint stores 256 separate expert tensors per projection, each (moe_intermediate_size, hidden) = (512, 2048) for gate/up and (hidden, moe_intermediate_size) = (2048, 512) for down. mlxcel's `SwitchLinear::from_weights` (`src/models/switch_layers.rs`) already handles this: given the prefix `model.layers.{i}.mlp.switch_mlp`, `stack_individual_experts` finds `model.layers.{i}.mlp.experts.{e}.{proj}.weight`, stacks along a NEW leading expert axis in ascending `e`, and produces the (num_experts, out_features, in_features) layout the gather paths expect: (256, 512, 2048) for gate/up, (256, 2048, 512) for down. No permutation of the inner two axes is required. The same helper stacks `.scales`/`.biases` for pre-quantized exports.
- **Fused QKV**: keep the (3072, 2048) tensor fused as one `UnifiedLinear` and split the activation at runtime along the head axis (axis 2 after the (1, seq, 24, 128) reshape, boundaries [16, 20]); no load-time splitting needed.
- Router precision: keep `mlp.gate.weight` and `mlp.gate.expert_bias` in float32 for routing (`router_dtype: fp32`); do not quantize the gate.
- `tie_word_embeddings` is false: `lm_head.weight` is a real (157184, 2048) tensor, never aliased to `model.word_embeddings.weight`. Handle a hypothetical tied config by dropping `lm_head.weight` and using the embedding as the output projection.
- All linear weights are (out_features, in_features); the embedding is (vocab_size, hidden). Nothing needs transposing.

## Validation and acceptance criteria

- [x] Co-located unit tests in `src/models/llada2_moe/tests.rs`, including loop-schedule unit tests: threshold decay endpoints (`t = 1` gives `T`, `t = S` gives `min_threshold`), transfer-mask selection (all-above-threshold, none-above-threshold forces exactly the single argmax-confidence masked position, ties prefer the lower block index), progress guarantee (a block of `B` masked positions fully unmasks within `B` iterations), prompt-tail guard in the first generation block, EOS truncation of the output window. Plus config-default tests for the omitted-key cases (`eos_token_id` falls back to `pad_token_id`, `mask_token_id` to 156895, `use_qk_norm` to true, `head_dim`/`rotary_dim` derivation) and gate-math tests on a tiny synthetic gate (group top-2 selection, bias affecting selection but not weights, `+ 1e-20` normalization, scaling factor).
- [x] A synthetic parity test: build a tiny checkpoint on disk (2 layers: 1 dense + 1 MoE; e.g. 8 experts, `n_group` 4, `topk_group` 2, k 2), load it through the real detection + loading path, and check (a) the MoE block output equals a naive per-expert loop computed on host, and (b) a fixed-logits unmasking run reproduces the hand-computed reveal order and final ids.
- [ ] A real-checkpoint smoke test (DEFERRED to orchestrator Phase 3.5 validation; the ~32.5 GB `inclusionAI/LLaDA2.0-mini` download + real-runtime generation was NOT executed on this Linux/GB10 host): download `inclusionAI/LLaDA2.0-mini` (or another real `llada2_moe` checkpoint), load it, and generate through mlxcel's actual runtime on both surfaces: `mlxcel generate` one-shot, and `mlxcel-server` where the request must flow through the diffusion worker loop (verify streaming SSE output, `finish_reason`, and usage accounting on `/v1/chat/completions`).
- [ ] Known failure mode, and why the real-checkpoint test is the completion bar: synthetic parity can pass while real checkpoints break on weight-key layout (e.g. `word_embeddings` vs `embed_tokens`, per-expert vs pre-stacked tensors), config defaults (this model's config omits `eos_token_id`, `mask_token_id`, and `use_qk_norm`), tokenizer/chat-template details (role-tag template, `<|mask|>`/`<|endoftext|>` specials, no `generation_config.json`), and MoE fusion. Real-checkpoint load + generate through the server/CLI runtime is required before this issue closes.
- [x] Completion requires integration into the working runtime (detection, `LoadedModel`, CLI generate, server diffusion worker, docs), not standalone modules.
- [x] `cargo fmt` / clippy / full local test suite green (CI does not build this crate's tests; local `cargo test` is the merge gate).

## Out of scope

- Re-masking or editing of already-committed tokens (confidence-based editing extensions).
- Batched serving (the model stays `supports_batching() == false`, batch-1 by design), paged KV, and scheduler integration.
- LoRA adapters and tensor-parallel inference (exclude in `src/distributed/tensor_parallel/inference.rs` like DiffusionGemma).
- Multimodal variants (`LLaDA2.0-Uni`); this issue is text-only.
- Sliding-window attention (`use_sliding_window` is false on real checkpoints) and KV-cache quantization for the diffusion path.

inureyes:
## Summary

Ports **LLaDA-2 MoE** (`model_type: llada2_moe`, e.g. `inclusionAI/LLaDA2.0-mini`): a decoder-only masked-diffusion language model with a DeepSeek-V3-style MoE FFN. It generates by iterative block-wise unmasking of `<|mask|>` tokens rather than autoregressive decode, so like DiffusionGemma it is a model-owned batch-1 generator (`supports_batching() == false`) served on the diffusion worker loop.

## What changed

New module `src/models/llada2_moe/`:

- **`mod.rs`** — config with the exact serde defaults from the spec, including derived `head_dim`/`rotary_dim` and the load-bearing omitted-key fallbacks (`eos_token_id` -> `pad_token_id`, `mask_token_id` -> 156895, `use_qk_norm` -> true). Bidirectional attention (no causal mask anywhere) with a fused-QKV runtime split (boundaries `[num_heads, num_heads + num_kv_heads]`), per-head QK-norm, and partial RoPE (`rotary_dim` of `head_dim`, non-traditional). Dense SwiGLU MLP for `i < first_k_dense_replace`, MoE block otherwise. Routed experts reuse `switch_layers::SwitchGLU` + `moe_weighted_sum`; the gate follows `deepseek_v3::MoEGate` with the three LLaDA-2 deltas: bias key `expert_bias`, `+ 1e-20` top-k normalization, and a group mask that fills non-kept groups with a large negative constant (not zero, since `expert_bias` may be negative). Router runs in float32.
- **`generate.rs`** — the semi-autoregressive block-unmasking engine: prefill commits full prompt blocks into per-layer KV caches; each generation block is denoised by repeatedly forwarding it against the frozen prefix (read-only `KVCache::visible_state`) and revealing positions whose chosen-token probability clears a linearly decaying threshold, with a single forced reveal when none clears it (progress guarantee). Schedule and transfer-mask selection are pure host-side functions.
- **`tests.rs`** — loop-schedule, config-default, and gate-math unit tests, plus the synthetic parity test and an on-disk tiny checkpoint loaded through the real detection + loading path.

Runtime integration (completion requires the working runtime, not standalone modules): `detection.rs`, `models/mod.rs` (variant + registry + display + exhaustiveness), `model_metadata.rs`, `loaded_model.rs` (variant + delegation), `loading/nonstandard.rs`, `distributed/tensor_parallel/inference.rs` (exclusion), CLI `commands/generate.rs` + new `commands/generate_llada2.rs`, `commands/chat.rs` rejection, and both `LoadedModel::DiffusionGemma` dispatch points in `server/model_worker.rs`. The diffusion worker's tokenize/stream/cancel/usage plumbing is factored into one `serve_streaming_request` helper shared by DiffusionGemma and LLaDA-2 (DiffusionGemma behavior byte-identical); a LLaDA-2 options mapper + worker loop sit beside the gemma ones, with `--max-denoising-steps` mapping to the per-block step count. Docs row added to `docs/supported-models.md`.

Also wraps one pre-existing constant assertion in `switch_layers` tests in a `const` block so `clippy --tests -D warnings` stays clean under the pinned 1.93.1 toolchain (byte-identical otherwise; unrelated to LLaDA-2).

## ⚠️ Real-checkpoint smoke test NOT run (deferred to Phase 3.5)

The issue's stated closure bar includes downloading `inclusionAI/LLaDA2.0-mini` (~32.5 GB) and generating through the real runtime on both `mlxcel generate` and `mlxcel-server`. **This was NOT executed on this Linux/GB10 host** (it cannot complete within the build/watchdog budget) and remains for orchestrator **Phase 3.5** validation. This PR does not imply that bar is met; it delivers everything else (implementation + full integration + synthetic-parity + unit tests, all green locally).

## Test plan

- [x] `cargo test --lib models::llada2_moe` — 20 tests green (loop-schedule, config defaults, gate math, MoE-vs-naive-loop parity, fixed-logits reveal order, on-disk synthetic checkpoint detect + load + generate)
- [x] `cargo test --lib server::diffusion_worker` — 11 green (6 DiffusionGemma unchanged + 5 new LLaDA-2 option-mapper / media-reject / finish-reason)
- [x] `cargo test --lib -- every_variant all_model_types_covers` — registry exhaustiveness guards green
- [x] `cargo clippy --lib --tests -- -D warnings` and `cargo clippy --bin mlxcel -- -D warnings` clean
- [x] `cargo fmt` clean
- [ ] Real-checkpoint smoke test on `inclusionAI/LLaDA2.0-mini` — deferred to Phase 3.5 (see caveat above)

Closes #546


inureyes:
## Real-checkpoint validation (orchestrator-run on GB10)

Validated end-to-end against the real `inclusionAI/LLaDA2.0-mini` checkpoint (30.29 GiB, bf16), through the actual runtime:

- **Load:** config parsed with the omitted-key defaults, all weight keys mapped (word_embeddings, per-expert stacking, fused QKV), tokenizer and chat template loaded. `Model loaded` with no key or dtype errors.
- **Generate (CUDA backend, sm_121):** prompt `What is the capital of France? Answer in one word.` produced `Paris` followed by the chat-template end-of-turn token. Coherent and correct.

Notes:
- The CPU backend cannot run this model: MLX's gathered-MoE matmul (`gather_mm` / `gather_qmm`) only supports float32 on CPU, while the experts are bf16. This is a general MLX CPU-backend limitation, not specific to this port. Running on an NVIDIA host requires a `--features cuda` build.
- The synthetic parity test used f32 weights, so it did not exercise the bf16 MoE path. The real-checkpoint run on the CUDA backend covers that gap.

Real-checkpoint closure bar met; moving to status:done.

Issue #688, opened 2026-07-07
bug(gemma4/cuda): decode collapses to <pad> after 1-3 tokens (single-query maskless attention + quantized GEMV); mlx_lm decodes coherently


inureyes:
## Summary
On CUDA (GB10), gemma-4 (gemma-4-12b-it-4bit and the uniform-4bit requant) greedy decode collapses to `<pad>` (id 0) spam after only 1 to 3 generated tokens, even when the prompt is byte-identical to a reference that decodes correctly. This is a decode-path (single-query, KV-cache) forward bug specific to mlxcel: the prefill and first-token logits are correct, but the incremental decode diverges. Discovered while fixing the gemma-4 chat-template rendering (#686, PR #687); that fix made the first token correct but the decode collapse is a separate, still-open defect and is the remaining blocker to gemma-4 usability.

## Evidence (GB10, release, greedy temp-0)
On the exact same model and byte-identical chat prompt ("Explain in two sentences why the sky is blue."):
- Apple mlx_lm gemma4_text decodes 50 tokens coherently: "The sky appears blue because of a phenomenon called Rayleigh scattering, where Earth's atmosphere scatters shorter wavelengths of light more effectively. Because blue light travels in shorter, smaller waves, it is scattered in all directions by gas molecules more than colors".
- mlxcel gemma-4-12b decodes "The" then all `<pad>`.
- mlxcel gemma-3-4b (control) decodes coherently, so the decode path is fine for non-gemma-4 models.

The forward is NOT the cause of the first token: mlxcel fed the reference's exact rendered prompt generates the correct first token ("Paris" for a capital-of-France prompt), and mlxcel's teacher-forced prefill perplexity matches mlx_lm within 0.05 nats. So the bug is in the INCREMENTAL DECODE path (single-query attention over the growing KV cache), not prefill.

## Localization (already narrowed)
- `MLXCEL_DISABLE_SINGLE_QUERY_MASKLESS=1` lengthens the coherent run from ~2 to ~24 tokens (uniform-4bit), so the single-query maskless decode-attention path is the dominant contributor.
- The 8-bit MLP GEMV (the shipped mixed-quant gemma-4-12b) shortens the coherent run further (~24 to ~6 tokens vs the uniform-4bit ~24), so a quantized-GEMV precision interaction is a secondary contributor.
- Collapse onset at cache length ~25 is far below the sliding_window of 1024, so it is not sliding-window overflow.

## Suspects
[...]

PR #695, opened 2026-07-08
fix(gemma4): diagnose malformed per-module quant overrides

### LINKED ISSUE #691: bug(gemma4): diagnose malformed per-module quant overrides
## Summary

PR #690 adds Gemma 4 per-module quantization override support for the #683 frontier variants. The implementation intentionally accepts arbitrary flattened keys under `config.quantization` and applies an override only when the matching value is an object with unsigned-integer `group_size` and/or `bits`. Malformed values, wrong scalar types, negative values, and typoed prefixes currently fall back to the root defaults without a direct diagnostic.

## Problem

This is not a security issue because model configs are local/trusted inputs, but it is a debugging hazard for quantization frontier work. A typo in a key such as `language_model.model.layers.0.mlp.gate_proj`, or a string value like `"bits": "4"`, can silently route the loader through the root `(group_size, bits)` and leave the user to infer the mistake from downstream shape reconciliation warnings, load failures, or degraded quality.

## Proposed fix

When an exact per-module override key is present but malformed, emit an actionable warning or return a load-time error before constructing the affected projection. At minimum, diagnose non-object override values, non-integer `group_size`/`bits`, unsupported bit widths, and non-positive group sizes. Prefer a focused helper around `QuantizationArgs::quant_params_for()` so future model paths that reuse the override map get the same behavior.

## Acceptance criteria

- [ ] Gemma 4 config parsing reports an actionable diagnostic when an override value is not an object.
- [ ] Gemma 4 config parsing reports an actionable diagnostic when `group_size` or `bits` is present with the wrong type, negative value, zero group size, or unsupported bit width.
- [ ] Valid root-only quantization configs and valid per-module overrides continue to load unchanged.
- [ ] Add unit coverage for valid overrides, malformed override values, and typo/wrong-type cases.
- [ ] The diagnostic mentions the offending override key/prefix.

## Related

- PR #690
- Issue #683

inureyes:
## Summary

PR #690 added Gemma 4 per-module MLP quantization overrides through a flattened `overrides` map on `QuantizationArgs`, but a malformed override value (a non-object value, a wrong-typed/negative/zero `group_size`, or an unsupported `bits` width) silently fell back to the root defaults with no indication anything was wrong, which is a debugging hazard for quantization frontier work.

## What changed

- `QuantizationArgs::quant_params_for` in `src/models/gemma4.rs` now validates the override entry matched by the queried module `prefix` before applying it, reporting an actionable `tracing::warn!` naming the offending key for a non-object override value, a wrong-typed or negative `group_size`/`bits`, a zero group size, or an unsupported bit width (mlxcel only supports the affine `{2,3,4,5,6,8}` set), then falls back to the root defaults for that field exactly as before.
- Added two small validation helpers, `validate_override_group_size` and `validate_override_bits`, that produce the actionable reason strings used in the warnings.
- Metadata keys that legitimately live alongside per-module overrides in the same flattened map (`mode`, `quant_method`, `quant_algo`, `config_groups`) are untouched because they never coincide with a real module-prefix lookup, so they are never mistaken for malformed overrides.
- Added a `quant_override_diagnostics_tests` module in `src/models/gemma4.rs` covering valid full and partial overrides, absent overrides, a non-object override value, string-typed and negative/zero/unsupported-width numeric fields, sibling metadata keys, and a typo'd prefix that simply never matches.

## Test plan

- [x] `cargo check -p mlxcel --features cuda --lib`
- [x] `cargo test -p mlxcel --features cuda --lib models::gemma4::quant` (24 tests, all pass, including the 11 new diagnostics tests and the pre-existing `quant_scheme_tests`)
- [x] `cargo test -p mlxcel --features cuda --lib models::gemma4_tests` (9 pass, 8 ignored for serial MLX execution as before, including the existing `gemma4_root_quantization_preserves_per_module_overrides` regression test, confirming valid configs load unchanged)
- [x] `cargo clippy -p mlxcel --features cuda --lib --tests` (no warnings touching `gemma4.rs`; the crate has one pre-existing unrelated `redundant_closure` warning in `src/models/sanitize.rs` that predates this change)
- [x] `cargo fmt -p mlxcel -- --check`

Closes #691

### DIFF
--- a/src/models/gemma4.rs
+++ b/src/models/gemma4.rs
@@ -67,18 +67,96 @@ fn default_quant_bits() -> usize {
     4
 }
 
+/// Bit widths mlxcel's affine quantization understands, mirroring
+/// `mlxcel_core::layers::infer_quantization_bits`'s `{2,3,4,5,6,8}` set.
+const SUPPORTED_OVERRIDE_BITS: &[u64] = &[2, 3, 4, 5, 6, 8];
+
+/// Validates a per-module override's `group_size` value, returning an
+/// actionable reason string when it is not a positive integer.
+fn validate_override_group_size(raw: &serde_json::Value) -> Result<i32, String> {
+    let Some(group_size) = raw.as_u64() else {
+        return Err(format!("must be a positive integer, got {raw}"));
+    };
+    if group_size == 0 {
+        return Err("must be a positive integer, got 0".to_string());
+    }
+    i32::try_from(group_size).map_err(|_| format!("overflows a 32-bit group size ({group_size})"))
+}
+
+/// Validates a per-module override's `bits` value, returning an actionable
+/// reason string when it is not an integer or not a supported bit width.
+fn validate_override_bits(raw: &serde_json::Value) -> Result<i32, String> {
+    let Some(bits) = raw.as_u64() else {
+        return Err(format!("must be a positive integer, got {raw}"));
+    };
+    if !SUPPORTED_OVERRIDE_BITS.contains(&bits) {
+        return Err(format!(
+            "must be one of {{2, 3, 4, 5, 6, 8}} bits, got {bits}"
+        ));
+    }
+    i32::try_from(bits).map_err(|_| format!("overflows a 32-bit bit width ({bits})"))
+}
+
 impl QuantizationArgs {
+    /// Resolves the quantization parameters for the module weight registered
+    /// under `prefix`, applying a per-module override when one is present.
+    ///
+    /// `overrides` is the flattened remainder of the parent `quantization`
+    /// object (everything besides the well-known root `group_size` / `bits`
+    /// pair, see PR #690). Only the entry keyed by the module's own `prefix`
+    /// is consulted here, so unrelated metadata keys that legitimately live
+    /// alongside per-module overrides (e.g. `mode`, `quant_method`,
+    /// `quant_algo`, `config_groups`, all validated separately by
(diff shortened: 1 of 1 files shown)


inureyes:
## PR Finalization Complete

### Summary
- **Tests**: Verified the 11 tests in `quant_override_diagnostics_tests` cover every bullet in issue #691's acceptance criteria (non-object override value, wrong-typed/negative/zero `group_size`, unsupported `bits` width, valid full/partial overrides, absent overrides, sibling metadata keys, typo'd prefix). No gap found; no additional tests added.
- **Documentation**: Checked `docs/*.md` (single-source English docs; `en/`/`ko/` are build-generated and untouched). Per-module quantization override behavior from PR #690 was never user-documented anywhere, so this diagnostics-only change needs no doc update. `CHANGELOG.md` is not touched by recent comparable small fix PRs (#688, #689, #690, #692), so it was left as-is.
- **Lint/Format**: `cargo fmt -p mlxcel -- --check` is clean. `cargo clippy -p mlxcel --features cuda --lib --tests -- -D warnings` shows exactly one warning, `redundant_closure` at `src/models/sanitize.rs:290`, confirmed identical on `main` and untouched by this PR's diff (out of scope, left as-is).
- **Test suite**: Re-ran `models::gemma4::` (28 tests) and the new `quant_override_diagnostics_tests` module (11 tests) locally; all pass.

No source changes were needed; working tree is unchanged from the prior commit. Ready for merge.

PR #697, opened 2026-07-08
perf(nvfp4): direct-transcode ModelOpt triplets to MLX native

### LINKED ISSUE #693: perf(nvfp4): direct-transcode ModelOpt triplets to MLX native
## Context

PR #692 repairs the CUDA ModelOpt NVFP4 path for `gemma-4-31b-it-nvfp4` by accepting the narrow NVIDIA ModelOpt triplet format and repacking it at load time. The final CUDA path repacks `weight/weight_scale/weight_scale_2` into MLX native NVFP4 tensors under `--features cuda`; non-CUDA builds intentionally keep the affine fallback until Apple Silicon is measured separately.

The current repair path is intentionally conservative, but cold-load expensive. It reconstructs dense f16 values first, then calls MLX `quantize(..., mode="nvfp4")`: `ModelOpt NVFP4 -> dense f16 -> MLX native NVFP4`. This is safe because MLX owns scale rounding and packing, but it pays a large one-time load cost and temporarily materializes dense tensors.

## Goal

Implement a direct ModelOpt triplet -> MLX native NVFP4 transcode path that avoids materializing a dense f16 weight matrix for each linear weight group while preserving the source two-level scale semantics.

## Technical notes

MLX native NVFP4 stores weights as `uint32` with eight FP4 E2M1 nibbles per word and stores per-block scales as `uint8` FP8 E4M3 with `group_size=16`. The ModelOpt FP4 nibble encoding used by the local Gemma 4 checkpoint matches the MLX E2M1 LUT order (`0, 0.5, 1, 1.5, 2, 3, 4, 6` with sign bit), so the weight payload can likely be repacked directly by combining four source U8 bytes into one little-endian U32 per eight input columns.

The scale side should prefer preserving `weight_scale_2` as an explicit global-scale sidecar rather than folding it into every E4M3 block scale. Folding `weight_scale * weight_scale_2` into each block scale is smaller, but it can introduce avoidable overflow or rounding drift when the folded value exceeds what one E4M3 byte can represent. A sidecar keeps the source ModelOpt two-level scale model intact and lines up with MLX/CUDA NVFP4 paths that already understand optional global scales.

## Implementation plan

1. Extend the weight representation and `UnifiedLinear` loading path so a quantized weight can carry an optional per-linear global scale sidecar for native NVFP4.
2. Thread that sidecar through `dequantize`, `quantized_matmul`, and any qmv/qmm bridge surfaces that currently pass `mode="nvfp4"` without a global scale.
3. Add a direct transcode helper for ModelOpt NVFP4 triplets: pack four source U8 bytes into one little-endian U32, preserve per-block `weight_scale` as E4M3 bytes, and preserve `weight_scale_2` as the global-scale sidecar.
4. Keep the dense f16 -> MLX `quantize(..., mode="nvfp4")` path behind a debug/test fallback until direct transcode parity is validated.
5. Measure cold-load wall time, MLX peak memory, short-prompt throughput, and 2048-prompt throughput on GB10.

## Acceptance criteria

- [x] Direct transcode avoids materializing a full dense f16 matrix for each ModelOpt triplet.
- [x] `weight_scale_2` is preserved as a global-scale sidecar and used by native NVFP4 dequantize/matmul rather than folded into each E4M3 block scale.
- [x] A focused fixture compares dequantized direct-transcode values against the dense f16 -> MLX native NVFP4 path and documents the expected FP8 scale rounding tolerance.
- [x] Greedy continuation parity is spot-checked against the dense f16 -> MLX native NVFP4 path for at least 40 generated tokens, allowing only expected FP8 rounding drift.
- [x] Cold-load wall time or peak load-time memory improves by at least 20% versus the dense f16 -> MLX quantize path on `gemma-4-31b-it-nvfp4`.
- [x] Results are recorded under `benchmarks/` and summarized under `docs/benchmark_results/`.

## Relationship to Apple Silicon

This issue should land before enabling native NVFP4 repack broadly on Apple Silicon. Once direct transcode and sidecar plumbing are validated, a separate Apple Silicon issue should benchmark whether Metal/non-CUDA should switch from the affine fallback to the native NVFP4 path.

inureyes:
## Summary

Replaces the conservative dense repack for CUDA ModelOpt NVFP4 (introduced in PR #692) with a direct triplet transcode that never materializes a dense f16 matrix per linear weight. The packed FP4 U8 bytes reinterpret straight into MLX native NVFP4 U32 words, the per-block E4M3 scales are preserved verbatim, and `weight_scale_2` is kept as a per-linear global-scale sidecar applied as an exact scalar multiply on the matmul output. The transcode is bit-exact to the checkpoint, so it is both faster to load and more faithful than the dense path (which re-derives block scales and drifts).

## Why the global scale is applied on the output, not folded into the block scales

MLX native NVFP4 already stores per-block E4M3 scales, but its `quantize`/`dequantize` `global_scale` parameter uses a different convention than ModelOpt: MLX re-derives the block scales from `global_scale` during quantize, so handing the checkpoint's E4M3 scales plus `weight_scale_2` as MLX `global_scale` produces wrong values (verified empirically). Because `weight_scale_2` is a single per-tensor f32 scalar, applying it on the matmul output is mathematically exact: `y = x @ (W * s2)^T = (x @ W^T) * s2`. Folding it into every E4M3 block scale is rejected by the issue for overflow/underflow drift (many `block_scale * weight_scale_2` products underflow E4M3), and was independently confirmed lossy here.

## What changed

- `src/models/sanitize.rs`: add `f32_to_f8_e4m3` (round-to-nearest-even inverse of the existing `f8_e4m3_to_f32`, exhaustively round-trip tested); add the direct transcode branch (default under CUDA) and the `MLXCEL_NVFP4_DENSE_REPACK=1` escape hatch that forces the old dense f16 -> MLX `quantize(mode="nvfp4")` path; emit a `{prefix}.global_scale` sidecar.
- `src/lib/mlxcel-core/src/layers.rs`: `QuantizedWeight` gains an optional `global_scale` sidecar and `apply_global_scale`; `UnifiedLinear::from_weights_with_mode` loads the sidecar; `UnifiedLinear::forward` runs the qmm without a fused bias, applies the scalar, then adds the bias separately so a (rare) linear bias is never scaled.
- `src/models/gemma4.rs`: `MLP::forward` falls through to the op-at-a-time path when any projection carries a global scale, so each `UnifiedLinear::forward` applies its own scalar; the activation `gelu_tanh_approx(gate) * up` is byte-identical to the fused NVFP4 fallback.
- `src/bin/bench_decode.rs`: report cold-load wall time and MLX peak memory for the load phase.
- Tests: exhaustive E4M3 round-trip, a fixture comparing direct-transcode dequant against the dense path with a documented FP8/FP4 tolerance, and the checkpoint fixture now asserts the global-scale sidecar and raw E4M3 U8 scales.

## Validation

The direct transcode dequantizes bit-exactly to the ModelOpt reference (`fp4 * e4m3_decode(weight_scale) * weight_scale_2`); the dense fallback drifts by one E4M3 block-scale plus one FP4 rounding step. Nibble order and scale re-encode were verified against the real `gemma-4-31b-it-nvfp4` checkpoint before trusting the reinterpret.

## Test plan

- [x] `cargo test -p mlxcel --features cuda --lib f32_to_f8` (E4M3 round-trip + specials)
- [x] `cargo test -p mlxcel --features cuda --lib nvfp4` (direct-vs-dense fixture + checkpoint fixture)
- [x] `cargo clippy -p mlxcel-core --features cuda --lib --tests -- -D warnings`
- [x] `cargo build --release --features cuda --bin mlxcel --bin mlxcel-bench-decode`
- [x] GB10 load + throughput benchmark, direct vs dense (table below)
- [x] Greedy continuation parity, direct vs dense, on the real checkpoint

## Benchmark (GB10, `gemma-4-31b-it-nvfp4`, release, same binary)

| Run | Repack | Cold-load wall (s) | Load peak (GB) | Prefill tok/s | Decode tok/s |
|-----|--------|-------------------:|---------------:|--------------:|-------------:|
| short (20 tok) | direct | 58.77 | 36.60 | 76.13 | 4.84 |
| short (20 tok) | dense | 190.72 | 36.60 | 79.51 | 5.26 |
| 2048-token | direct | 58.87 | 36.60 | 443.85 | 5.01 |
| 2048-token | dense | 190.69 | 36.60 | 395.20 | 5.38 |

Cold-load wall time drops 69.2% (190.72 s -> 58.77 s), so the >=20% acceptance gate passes on load time. Peak load memory is identical (36.60 GB): the dense f16 transients are freed between weight groups and never raise the high-water mark. Prefill at 2048 tokens is 12% faster on the direct path; decode is about 7% slower because the global scale forces the Gemma 4 dense MLP onto the op-at-a-time path plus a per-linear scalar multiply while Gemma 4 decode runs with CUDA graphs disabled. Folding the three scales into the fused `compiled_gelu_approx_mlp_forward` kernel would recover that and is a reasonable follow-up. Raw CSV: `benchmarks/cuda_gb10_issue693_nvfp4_direct_vs_dense_2026-07-08.csv`; summary: `docs/benchmark_results/nvfp4-direct-transcode-gb10-2026-07-08.md`.

## Parity

Greedy (`--temp 0`) continuation on the real checkpoint, 48 tokens, prompt "Explain in a few sentences why the sky appears blue during the day.". Both paths produce coherent, semantically identical explanations of Rayleigh scattering. Direct: "The sky appears blue due to a phenomenon called Rayleigh scattering. As sunlight enters Earth's atmosphere, it collides with gas molecules and scatters in all directions. Because blue light travels in shorter, smaller waves, it is scattered". Dense: "The sky appears blue because of a phenomenon called Rayleigh scattering. As sunlight reaches Earth's atmosphere, it is scattered in all directions by the gases and particles in the air. Because blue light travels in shorter, smaller waves, it". The token streams diverge early ("due to" vs "because of", "enters" vs "reaches") because the dense fallback re-quantizes and drifts about 10% mean relative weight error while the direct transcode is bit-exact to the checkpoint. Token-identical output between the two paths is therefore not expected; the bit-exactness that matters is direct-vs-checkpoint, covered by the `nvfp4_direct_transcode_is_exact_and_bounds_dense_drift` fixture.

## Review fixes

Post-review hardening in a356fe3 addressed one MEDIUM and three LOW findings from the security review of the direct transcode.

- `src/models/gemma4.rs`: the compiled per-layer-input-gate fused path now checks for a `global_scale` sidecar on both the gate and projection weights (mirroring `MLP::forward`) before taking the fused C++ path, so a per-layer-input ModelOpt-NVFP4 checkpoint no longer silently drops `weight_scale_2`. Three new tests cover the no-sidecar case and both sidecar-present cases.
- `src/models/sanitize.rs`: `MLXCEL_NVFP4_DENSE_REPACK` now matches its truthy values case-insensitively; `weight_scale_2` is validated as a single-element scalar before being read, with malformed triplets falling back like the other malformed-shape guards; the direct-transcode weight repack checks 4-byte alignment before reinterpreting packed bytes as `u32`, falling back to an aligned `from_slice_u32` decode when the allocation is misaligned.
- `docs/environment-variables.md`: documents `MLXCEL_NVFP4_DENSE_REPACK` alongside the other Gemma 4 rollback flags.

Validated with `cargo test -p mlxcel --features cuda --lib nvfp4` and `cargo test -p mlxcel --features cuda --lib per_layer_input_gate`, both green, and `cargo clippy -p mlxcel --features cuda --lib --tests -- -D warnings` clean.

Closes #693

### DIFF
--- a/benchmarks/cuda_gb10_issue693_nvfp4_direct_vs_dense_2026-07-08.csv
+++ b/benchmarks/cuda_gb10_issue693_nvfp4_direct_vs_dense_2026-07-08.csv
@@ -1,0 +1,5 @@
+variant,repack,model,model_path,prompt_tokens,generated_tokens,load_wall_s,load_peak_gb,prefill_ms,prefill_tok_s,decode_ms,decode_tok_s,run_peak_gb,date,hardware,mlx_version,build_type,max_tokens,prompt,prompt_target_len
+direct_short,direct,gemma-4-31b-it-nvfp4,models/gemma-4-31b-it-nvfp4,20,26,58.774,36.60,262.72,76.13,5376.07,4.84,36.60,2026-07-08,NVIDIA_GB10_122GB,0.3.3,release,100,"Hello, how are you today?",
+dense_short,dense,gemma-4-31b-it-nvfp4,models/gemma-4-31b-it-nvfp4,20,42,190.716,36.60,251.55,79.51,7990.42,5.26,36.60,2026-07-08,NVIDIA_GB10_122GB,0.3.3,release,100,"Hello, how are you today?",
+direct_2048,direct,gemma-4-31b-it-nvfp4,models/gemma-4-31b-it-nvfp4,2048,32,58.874,36.60,4614.17,443.85,6391.37,5.01,36.60,2026-07-08,NVIDIA_GB10_122GB,0.3.3,release,32,"Hello, how are you today?",2048
+dense_2048,dense,gemma-4-31b-it-nvfp4,models/gemma-4-31b-it-nvfp4,2048,32,190.687,36.60,5182.13,395.20,5950.56,5.38,36.60,2026-07-08,NVIDIA_GB10_122GB,0.3.3,release,32,"Hello, how are you today?",2048
--- a/docs/benchmark_results/nvfp4-direct-transcode-gb10-2026-07-08.md
+++ b/docs/benchmark_results/nvfp4-direct-transcode-gb10-2026-07-08.md
@@ -1,0 +1,106 @@
+# ModelOpt NVFP4 direct transcode benchmark (issue #693)
+
+This note records the CUDA load and throughput comparison between the direct
+ModelOpt NVFP4 transcode (issue #693) and the dense f16 requantize fallback it
+replaces (issue #692) on NVIDIA GB10. The local `gemma-4-31b-it-nvfp4`
+checkpoint is NVIDIA ModelOpt NVFP4 metadata with `quant_method=modelopt`,
+`quant_algo=NVFP4`, and per-linear `weight/weight_scale/weight_scale_2`
+triplets. The direct path reinterprets the packed FP4 U8 bytes into MLX native
+NVFP4 U32 words, preserves the per-block E4M3 scales verbatim, and keeps
+`weight_scale_2` as a per-linear global-scale sidecar. It never materializes a
+dense f16 matrix, so it loads far faster than the dense fallback and is
+bit-exact to the checkpoint. `MLXCEL_NVFP4_DENSE_REPACK=1` forces the dense
+path for A/B comparison.
+
+## Environment
+
+| Item | Value |
+|------|-------|
+| Hardware | NVIDIA GB10 (DGX Spark), 122 GB unified LPDDR5x |
+| Backend | CUDA |
+| Build | `cargo build --release --features cuda --bin mlxcel --bin mlxcel-bench-decode` |
+| Harness | `target/release/mlxcel-bench-decode` |
+| Test date | 2026-07-08 |
+| Prompt | `Hello, how are you today?` (short) and a synthesized 2048-token prompt |
+| Raw CSV | `benchmarks/cuda_gb10_issue693_nvfp4_direct_vs_dense_2026-07-08.csv` |
+
+Each run is a separate process, so the reported cold-load wall time and MLX
+peak are measured against a fresh MLX allocator. `mlxcel-bench-decode` now
+resets the MLX high-water mark before the load and prints `[Load] wall` and the
+peak after load completes.
+
+## Results
+
+| Run | Repack | Prompt tokens | Cold-load wall (s) | Load peak (GB) | Prefill tok/s | Decode tok/s |
+|-----|--------|--------------:|-------------------:|---------------:|--------------:|-------------:|
+| direct, short | direct triplet transcode | 20 | 58.77 | 36.60 | 76.13 | 4.84 |
+| dense, short | dense f16 requantize | 20 | 190.72 | 36.60 | 79.51 | 5.26 |
+| direct, 2048 | direct triplet transcode | 2048 | 58.87 | 36.60 | 443.85 | 5.01 |
+| dense, 2048 | dense f16 requantize | 2048 | 190.69 | 36.60 | 395.20 | 5.38 |
+
+## Acceptance
+
+The acceptance gate is a 20% improvement in cold-load wall time or peak load
+memory. Cold-load wall time drops from 190.72 s (dense) to 58.77 s (direct), a
+69.2% reduction, so the gate passes on load time by a wide margin. The dense
+path spends most of that time reconstructing a dense f16 matrix for each of the
+180 NVFP4 MLP weight groups and re-quantizing it; the direct path only
+reinterprets bytes and re-encodes the small per-block scale tensors.
+
+Peak load memory is identical at 36.60 GB for both paths. The dense f16
+transients are freed between weight groups, so they do not raise the MLX
+high-water mark above the loaded model. The load-time win, not peak memory, is
+the improvement here.
+
+## Throughput
+
+Prefill at 2048 tokens is 443.85 tok/s for the direct path versus 395.20 tok/s
+for the dense path, a 12% gain, because the direct weights carry the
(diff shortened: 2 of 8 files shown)


inureyes:
Applied fixes from the security review (commit a356fe3).

MEDIUM: the compiled per-layer-input-gate fused path in gemma4.rs did not check for a native-NVFP4 global_scale sidecar the way MLP::forward already does, so a per-layer-input ModelOpt-NVFP4 checkpoint would silently drop weight_scale_2 and compute wrong outputs. Added a per_layer_input_gate_fused_path_eligible guard requiring global_scale.is_none() on both projections before taking the fused path, with three new unit tests in gemma4::quant_scheme_tests.

LOW fixes in sanitize.rs: MLXCEL_NVFP4_DENSE_REPACK now matches its truthy values case-insensitively (with a new test); weight_scale_2 is validated as a single-element scalar before item_f32 reads it, with the triplet skipped and a warning logged on a malformed shape (new test); the direct-transcode weight repack now checks 4-byte alignment before reinterpreting the packed U8 buffer as U32, falling back to a properly aligned from_slice_u32 decode on the rare misaligned allocation.

Docs: added MLXCEL_NVFP4_DENSE_REPACK to docs/environment-variables.md.

Validated with cargo test -p mlxcel --features cuda --lib nvfp4, cargo test -p mlxcel --features cuda --lib per_layer_input_gate, and cargo clippy -p mlxcel --features cuda --lib --tests -- -D warnings, all clean.

PR #701, opened 2026-07-08
perf(gemma4): fold NVFP4 global scales into fused MLP kernel

### LINKED ISSUE #698: perf(gemma4): fold NVFP4 global scales into fused MLP kernel
## Problem / Background

PR #697 (merged `21db6fe`, closes #693) direct-transcodes ModelOpt NVFP4 triplets and preserves `weight_scale_2` as a per-linear `global_scale` sidecar, applied as a scalar multiply on the matmul output in `UnifiedLinear::forward` (`src/lib/mlxcel-core/src/layers.rs`).

The compiled fused paths cannot apply this sidecar. As a result, `MLP::forward` and `compiled_per_layer_input_gate` in `src/models/gemma4.rs` bypass the fused `compiled_gelu_approx_mlp_forward` kernel whenever any participating weight carries a sidecar, falling back to op-at-a-time projections.

Measured cost on GB10 with gemma-4-31b-it-nvfp4: decode is roughly 7% slower than the dense-repack baseline (4.84 vs 5.26 tok/s at short prompt, 5.01 vs 5.38 tok/s at 2048). The gap is amplified because gemma4 decode runs with CUDA graphs disabled (#688), so every extra element-wise dispatch is paid at full launch cost.

Benchmarks:
- `benchmarks/cuda_gb10_issue693_nvfp4_direct_vs_dense_2026-07-08.csv`
- `docs/benchmark_results/nvfp4-direct-transcode-gb10-2026-07-08.md`

## Proposed Solution

Teach the fused MLP path (and the per-layer-input gate fused path) to accept up to three per-linear f32 global scales, one each for the gate, up, and down projections, and apply them inside the compiled kernel:

- Fold the gate and up scales into their respective pre-activation matmul products.
- Apply the down scale on the fused output.

This restores fused execution for sidecar-carrying NVFP4 weights instead of falling back to op-at-a-time projections.

**Mathematical care (GeGLU is nonlinear):** the MLP computes `down(gelu(gate(x)) * up(x))`. Because `gelu` is nonlinear, the gate projection's global scale must be applied to the gate product BEFORE the activation: `gelu(s_gate * (x @ W_gate))`. It cannot be factored out past the activation, i.e. `s_gate * gelu(x @ W_gate)` is wrong. The up scale multiplies the up product (linear, so placement before or after the elementwise product is equivalent), and the down scale can be applied to the final output since the down projection is linear.

## Acceptance Criteria

- [ ] Fused MLP path (`compiled_gelu_approx_mlp_forward`) and `compiled_per_layer_input_gate` accept sidecar-carrying weights, with each per-projection global scale applied at the mathematically correct point (gate scale before the activation, up scale on the up product, down scale on the output)
- [ ] Greedy output parity vs the current op-at-a-time sidecar path on gemma-4-31b-it-nvfp4 is token-exact
- [ ] Decode throughput on GB10 recovers to within ~2% of the dense-repack baseline (short prompt and 2048)
- [ ] Guard tests updated: the fused path now accepts sidecar weights, while non-gemma4 paths remain unaffected
- [ ] Benchmark CSV and `docs/benchmark_results/` summary updated with the recovered numbers

## Technical Considerations

- The sidecar is produced by the NVFP4 direct transcode in PR #697; `UnifiedLinear::forward` currently applies it as a scalar multiply on the matmul output, which is what the fused kernel must reproduce internally.
- With CUDA graphs disabled for gemma4 decode (#688), dispatch count dominates; folding the scalar multiplies into the compiled kernel removes the per-projection element-wise launches that the op-at-a-time fallback incurs.
- Keep the scale inputs optional (default 1.0 or absent) so dense-repack and non-sidecar weights compile to the same kernel behavior as today.

## Related

- #693 (NVFP4 direct transcode)
- PR #697 (introduced the `global_scale` sidecar and the fused-path bypass)
- #683 (CUDA decode kernel profiling)
- #688 (gemma4 CUDA graphs disabled for decode)
- #694 (Metal native NVFP4 path)

inureyes:
## Summary

Folds the NVFP4 `weight_scale_2` `global_scale` sidecar (from the direct ModelOpt transcode, #693/#697) back into the fused gemma-4 MLP and per-layer-input-gate kernels, at the mathematically correct points, so a sidecar-carrying MLP no longer bypasses the fused path and runs three op-at-a-time projections.

The GeGLU is nonlinear, so the gate scale is applied to the gate product BEFORE `gelu_tanh_approx`; the up scale is a linear factor on the up product; the down scale multiplies the fused output. Each fold reproduces `apply_global_scale` byte-for-byte (multiply by the f32 scalar, then astype back to the activation dtype), so the fused result is bit-close to the op-at-a-time path. A null scale means no multiply for that projection, so mixed sidecar sets stay exact for the unscaled projections.

## What changed

- `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp` / `.h`: new `compiled_gelu_approx_mlp_forward_global_scale` (compiles the scaled graph on the single-token decode call, keyed on group_size/bits/mode and the gate/up/down presence flags so the unscaled graph is unchanged and each sidecar pattern gets its own graph; folds eagerly on prefill, mirroring the decode-only compile gate of the unscaled path). `compiled_per_layer_input_gate` gains two optional scale operands folded into its eager fallback (NVFP4 and the affine compiled branch are mutually exclusive, so the compiled affine path is untouched). Shared `apply_global_scale_eager` helper mirrors `QuantizedWeight::apply_global_scale`.
- `src/lib/mlxcel-core/src/lib.rs`: FFI declarations for the new function and the two extra scale operands.
- `src/lib/mlxcel-core/src/layers.rs`: `QuantizedWeight::global_scale_ptr()` (null when absent), mirroring `biases_ptr()`.
- `src/models/gemma4.rs`: `MLP::forward` and the per-layer-input gate call the scaled fused path only for single-token inputs when any sidecar is present; `MLXCEL_DISABLE_FUSED_GLOBAL_SCALE` (read once) restores the op-at-a-time bypass; `per_layer_input_gate_fused_path_eligible` takes `fused_scale_enabled` and `single_token` flags, and the new `dense_mlp_scaled_fused_path_eligible` mirrors it for the dense MLP branch (review fix, see below).
- `docs/environment-variables.md`: documents `MLXCEL_DISABLE_FUSED_GLOBAL_SCALE`.
- `benchmarks/cuda_gb10_issue698_nvfp4_fused_scale_2026-07-09.csv` and `docs/benchmark_results/nvfp4-direct-transcode-gb10-2026-07-08.md`: the GB10 numbers below, including the post-fix re-measurement.

## Review fixes

Review found that the initial fold gated only on sidecar presence and the `MLXCEL_DISABLE_FUSED_GLOBAL_SCALE` kill switch, with no check on query length. The C++ fused-scale function compiles the scaled graph only for single-token calls and falls back to an eager, uncompiled `multiply(gelu_tanh_approx(gate), up)` otherwise, so every multi-token prefill call silently downgraded from the compiled op-at-a-time activation to that eager fold. On the 2048-token prompt this regressed prefill from 449.90 tok/s (op-at-a-time) to 412.48 tok/s, an 8.3% loss, for no decode benefit; the 20-token short prompt showed no measurable change because a matmul that small is too cheap for the eager-versus-compiled cost to clear the noise floor.

The fix gates both fused call sites in `src/models/gemma4.rs` (`MLP::forward`'s dense MLP branch and the per-layer-input gate) to single-token inputs (`mlxcel_core::array_shape(x)[1] == 1`), so a sidecar-carrying multi-token call now bypasses the fused path unconditionally, regardless of the kill switch, and takes the identical op-at-a-time projections plus `compiled_geglu_approx_activation` that non-sidecar prefill has always used. The C++ bridge itself is unchanged; its eager-fold branch remains correct and is still exercised by the multi-token FFI test, just no longer reached from gemma-4 prefill. Re-measured on the same host and release binary, 2048-token prefill recovers to 447.24 tok/s (within 0.6% of op-at-a-time) while decode stays fused at 5.05 tok/s, and greedy `--temp 0` output is still byte-identical between the default and `MLXCEL_DISABLE_FUSED_GLOBAL_SCALE=1` on the short prompt.

Review also flagged the benchmark doc's "no regression" and weight-layout-attribution claims. `src/models/sanitize.rs` shows the dense and direct-transcode routes produce identical MLX native NVFP4 layout (same mode, group_size, bits, weight dtype, and scale dtype), differing only in scale values (dense re-derives with about 10% drift, direct is bit-exact) plus a negligible f32 `[1]` sidecar, so the layout attribution was wrong. The doc now documents the prefill regression and fix honestly, and notes the residual dense-vs-direct decode gap is not conclusively isolated: the dense short run generated 42 tokens against 26 for direct, and the dense 2048 row is a cross-day carry-over rather than a same-run measurement.

`dense_mlp_scaled_fused_path_eligible` was extracted as a testable mirror of `per_layer_input_gate_fused_path_eligible` (now taking a `single_token` parameter) so both fused call sites have unit coverage without needing a full compiled forward pass. New guard tests pin: sidecar + single-token + fold enabled routes to the fused path; sidecar + multi-token bypasses the fused path regardless of the fold toggle; the existing kill-switch tests are preserved.

## Parity outcome

Greedy `--temp 0` decode is byte-identical between the fold (default) and the op-at-a-time bypass (`MLXCEL_DISABLE_FUSED_GLOBAL_SCALE=1`) on the 26-token short prompt `Hello, how are you today?`, both before and after the single-token gate fix.

On a 120-token generation (`List five interesting facts about the planet Jupiter.`) the two paths agree for the first ~68 tokens and then diverge at a near-tie fact selection. That divergence is baseline CUDA nondeterminism, not the fold: two runs of the unmodified default path diverge from each other as well (at ~115 tokens, `so large that Earth could` vs `so large that it is`), so the NVFP4 qmm FP-reduction order is not run-to-run reproducible on this backend for either path. The fused output stays inside that envelope rather than introducing divergence. This mirrors the documented behavior of `MLXCEL_FUSED_QK_NORM` on CUDA.

Synthetic `compiled_qgelu_mlp_global_scale_*` unit tests confirm the fold matches the op-at-a-time `apply_global_scale` reference to 1e-5 on the compiled single-token branch, the eager multi-token fallback, and a mixed (down-only) sidecar set, and that non-unit scales genuinely change the output (the sidecar is not silently dropped).

## Benchmarks (GB10, gemma-4-31b-it-nvfp4, same release binary, 2026-07-09)

| Run | Scale path | Prompt tokens | Prefill tok/s | Decode tok/s |
|-----|-----------|--------------:|--------------:|-------------:|
| short | dense (folded into block scales) | 20 | 78.55 | 5.25 |
| short | op-at-a-time bypass | 20 | 74.62 | 4.86 |
| short | fused fold (this PR, default) | 20 | 74.63 | 4.90 |
| 2048 | dense (folded into block scales) | 2048 | 395.20 | 5.38 (2026-07-08) |
| 2048 | op-at-a-time bypass | 2048 | 449.90 | 5.01 |
| 2048 | fused fold (initial, pre-fix) | 2048 | 412.48 | 5.07 |
| 2048 | fused fold (this PR, post-fix) | 2048 | 447.24 | 5.05 |

The `~2% of dense` throughput gate is NOT met: the fold is throughput-neutral on GB10 decode (+0.8% short, +0.8% at 2048 post-fix vs op-at-a-time) and stays 6.7% below dense short, 6.1% below dense at 2048 post-fix. GB10 gemma-4 decode is weight-bandwidth-bound (~94% GPU-busy streaming the NVFP4 MLP weights), so removing per-layer element-wise dispatches is throughput-neutral there, exactly as documented for the sibling `MLXCEL_COMPILED_QGELU_MLP` fusion. The fold rules out scale-op dispatch count as the cause of the residual dense-vs-direct gap (folding the scales into one compiled graph does not close it), but the gap itself is not conclusively isolated by this benchmark: the comparison is confounded by mismatched generation lengths (42 tokens dense-short vs 26 direct-short) and a cross-day dense-2048 row. The fold still removes three FFI crossings and three element-wise dispatches per MLP layer for single-token decode, which reduces CPU dispatch work and helps op-count-bound backends, so it ships default-on for decode with the kill switch retained; multi-token prefill takes the unaffected op-at-a-time path unconditionally. Full rationale in `docs/benchmark_results/nvfp4-direct-transcode-gb10-2026-07-08.md`.

## Test plan

- [x] `cargo test -p mlxcel-core --features cuda --lib global_scale` (4 tests: compiled single-token, eager multi-token, mixed down-only, all-null == unscaled; all pass, fold matches reference to 1e-5)
- [x] `cargo test -p mlxcel --features cuda --lib per_layer_input_gate` (5 guard tests, including the new multi-token-bypass test; all pass)
- [x] `cargo test -p mlxcel --features cuda --lib dense_mlp_scaled_fused_path_eligible` (4 new guard tests for the dense MLP branch's single-token gating; all pass)
- [x] `cargo clippy -p mlxcel-core --features cuda --lib --tests -- -D warnings` (clean)
- [x] `cargo clippy -p mlxcel --features cuda --lib --tests -- -D warnings` (clean)
- [x] `cargo fmt -p mlxcel -p mlxcel-core`
- [x] Real-checkpoint greedy parity (fused default vs `MLXCEL_DISABLE_FUSED_GLOBAL_SCALE=1`) on gemma-4-31b-it-nvfp4, byte-identical on the short prompt; long-generation divergence attributed to baseline GPU nondeterminism (fused-vs-fused also diverges)
- [x] Decode benchmarks (short + 2048, pre-fix and post-fix) fused vs op-at-a-time vs dense, recorded in the CSV and doc

Closes #698


inureyes:
Applied review fixes in `a39c8d6`.

**HIGH: unconditional fused-scale routing regressed prefill.** Both fused call sites in `src/models/gemma4.rs` (`MLP::forward`'s dense MLP branch and the per-layer-input gate) now gate on `mlxcel_core::array_shape(x)[1] == 1` in addition to sidecar presence and the kill switch, so a sidecar-carrying multi-token (prefill) call bypasses the fused path unconditionally and falls through to the existing op-at-a-time projections plus the compiled GeGLU activation. The C++ bridge is unchanged; its eager-fold branch stays correct for the multi-token case and is exercised by the existing FFI test, it's just no longer reached from gemma-4 prefill. `per_layer_input_gate_fused_path_eligible` now takes a `single_token` parameter, and the new `dense_mlp_scaled_fused_path_eligible` mirrors it for the dense MLP branch so both call sites have unit coverage. New guard tests pin: sidecar + single-token + fold enabled routes to the fused path; sidecar + multi-token bypasses regardless of the fold toggle; the kill-switch tests are preserved.

**MEDIUM: benchmark doc wording.** Removed the "no regression" claim and the weight-layout attribution for the residual dense-vs-direct decode gap, which `src/models/sanitize.rs` contradicts (both routes produce identical MLX native NVFP4 layout, differing only in scale values and a negligible sidecar). The doc now documents the prefill regression and fix honestly, and notes the dense-vs-direct comparison is confounded by mismatched generation lengths and a cross-day carry-over row.

Post-fix re-measurement, same GB10 host, same release binary, minimal reload (3 loads total):

| Run | Scale path | Prompt tokens | Prefill tok/s | Decode tok/s |
|-----|-----------|--------------:|--------------:|-------------:|
| 2048 | op-at-a-time bypass (reference) | 2048 | 449.90 | 5.01 |
| 2048 | fused fold (pre-fix, regressed) | 2048 | 412.48 | 5.07 |
| 2048 | fused fold (post-fix, default) | 2048 | 447.24 | 5.05 |

Prefill recovers to within 0.6% of the op-at-a-time reference while decode stays on the fused path. Greedy `--temp 0` output on the 26-token short prompt is byte-identical between the default and `MLXCEL_DISABLE_FUSED_GLOBAL_SCALE=1`.

Verification: `cargo test -p mlxcel-core --features cuda --lib global_scale` (4 pass), `cargo test -p mlxcel --features cuda --lib per_layer_input_gate` (5 pass) and `dense_mlp_scaled_fused_path_eligible` (4 new, pass), `cargo clippy -p mlxcel -p mlxcel-core --features cuda --lib --tests -- -D warnings` (clean), `cargo fmt` (clean). No CUDA errors during any of the three real-model loads.

Issue #624, opened 2026-07-02
perf(bench): long-prompt prefill benchmark coverage and serving TTFT/decode-rate telemetry


inureyes:
Part of #623.

## Context

All existing benchmark CSVs (`benchmarks/cuda_gb10_2026-06-17.csv` and the Metal sets) use prompts of 8-66 tokens. At that length prefill numbers are dominated by graph-build and kernel-launch overhead, not by matmul throughput, so they cannot validate any prefill optimization (MoE prefill fixes, sm_121 GEMM work, chunked prefill tuning). Separately, the server exposes only aggregate counters (`src/server/batch/observability.rs`); there is no TTFT or decode-rate telemetry, so serving-level changes (batching defaults, pipelining) cannot be measured except by external tooling.

This issue is the measurement foundation for the rest of the epic and must land first.

## Scope

1. Long-prompt prefill benchmarking in the existing harness.
2. TTFT and decode-rate metrics in the server's Prometheus surface.
3. A concurrency load-generation script for serving-throughput measurements.

### 1. Long-prompt prefill benches

- Extend `src/bin/bench_decode.rs` (binary `mlxcel-bench-decode`) and `scripts/bench_decode.sh` to accept a target prompt token length (e.g. `--prompt-tokens N`). Synthesize the prompt deterministically (repeat a fixed corpus paragraph, then truncate to N tokens with the model's tokenizer) so runs are reproducible across models.
- Standard lengths: 512, 2048, 8192, 32768 (cap at the model's max context; record the actual length used).
- Emit the same CSV schema as today (`prefill_ms`, `prefill_tok_s`, `decode_ms`, `decode_tok_s`) plus a `prompt_target_len` column. Keep the short-prompt default intact so historical CSV comparability is preserved.
- Add a wrapper mode in `scripts/bench_all_models.sh` (or a sibling script) that runs a representative subset (llama-3.1-8b-4bit, qwen2.5-7b-4bit, qwen3-30b-a3b-4bit, mixtral-8x7b-4bit, gemma-4-31b-it-4bit) across the length ladder, since a full 147-model sweep at 32k is too slow.

### 2. Server TTFT / decode-rate telemetry

[...]

PR #707, opened 2026-07-09
fix(router): emit usage on the disaggregated /v1/chat/completions responses

inureyes:
## Summary

The disaggregated router's `/v1/chat/completions` path never emitted a `usage` object, diverging from the single-node chat route (which reports `usage` in both streaming and non-streaming responses) and from the router's own `/v1/completions` path. PR #392 (issue #387) had already made the router resolve the worker's authoritative generated-token count for the chat path, but only used it for `finish_reason`; this PR wires that already-resolved count, plus the previously-discarded prompt token count, into the chat `usage` block.

## What changed

- `src/server/router_front.rs`: non-streaming `route_chat` now passes `prompt_tokens` (previously discarded as `_prompt_tokens`) and the resolved `completion_tokens` into `chat_completion_json`, which now serializes a `usage` object (`prompt_tokens`, `completion_tokens`, `total_tokens`) matching the single-node `ChatCompletionResponse` shape.
- `src/server/router_front.rs`: streaming `route_chat` now reads `stream_options.include_usage` the same way the router's `/v1/completions` streaming path does, resolves `completion_tokens` once for both `finish_reason` and the usage chunk, and emits a new `chat_chunk_usage` SSE chunk (empty `choices`, populated `usage`) after the finish chunk, only on a successful handoff — a handoff failure still emits no usage chunk, matching the existing completions-path behavior.
- `src/server/router_front_tests.rs`: unit tests for `chat_completion_json`'s new `usage` field and `chat_chunk_usage`'s chunk shape.
- `tests/disaggregated_router_e2e.rs`: a new E2E test, `disaggregated_router_chat_usage_matches_single_node_byte_fallback`, that spins up a three-process disaggregated stack plus a single-node reference and asserts `usage.prompt_tokens` / `usage.completion_tokens` / `usage.total_tokens` (and `finish_reason`) match between router and single-node, for both the non-streaming response and the streaming final usage chunk. Gated behind the same checkpoint-presence guard as the other real-model tests in that file.

## E2E fixture note (issue #708)

[...]

PR #713, opened 2026-07-09
perf(server): incremental detokenization and streaming overhead cuts

### LINKED ISSUE #633: perf(server): incremental detokenization and per-token streaming overhead reduction
Part of #623. Independent of the CUDA work; measured impact is largest on fast small models (200-650 tok/s on GB10).

## Context

Three host-side per-token costs on the serving path:

1. **O(N) detokenization per token.** `StreamingDecodeState::on_token` -> `emit_regular_token` calls `tokenizer.decode(&self.all_ids, false)` on the FULL token history for every generated token (`src/server/model_worker.rs:2014`, byte-fallback branch `:1979`). At 500 tok/s and 2k-token outputs this is quadratic work sitting on the single generation thread.
2. **Per-token streaming overhead in the chat route.** For every token the callback locks three mutexes (accumulator, logprobs buffer, stream filter), builds a `ChatCompletionChunk`, serializes JSON, and does two mpsc hops (`src/server/routes/chat.rs:640-810`, sends at `:762,:781,:806`; `src/server/streaming.rs:133-147`).
3. **Prompt tokenization on the generation thread.** `enqueue_request` tokenizes inside the scheduler thread (`src/server/batch/scheduler.rs:2275`), so a long prompt arriving stalls decode ticks for every active sequence.

## Scope

All three items; they are small individually and share test surface (byte-exact streaming).

### 1. Incremental detokenizer

- Replace full-history decode with an incremental scheme in `StreamingDecodeState` (`src/server/model_worker.rs:1936-2060`): keep the previously decoded string (or byte offset) plus a small tail window of token ids; on each token, decode only `tail_window + new_id` and emit the suffix past the previously emitted boundary. The tail window (a few tokens) handles tokenizers whose piece merging changes the tail rendering; the existing UTF-8 partial-character handling (byte-fallback branch at `:1979`) must be preserved: never emit an incomplete UTF-8 sequence, buffer until complete.
- Mirror the proven approach in mlx-lm/llama.cpp streaming detokenizers rather than inventing merge rules.
- The same state is used by both the batched and single decode emit paths (`scheduler.rs:4513-4519`, `:4830-4836`); no call-site changes should be needed beyond the struct internals.

### 2. Streaming-path diet

- Collapse the three per-token mutex acquisitions in the chat callback into one small state struct behind a single lock, or restructure so the callback owns its state (it runs on one thread; check whether the mutexes exist only for Sync bounds and can become `RefCell`-style or plain ownership).
- Reuse a serialization buffer for chunk JSON (serde_json `to_writer` into a reused `Vec<u8>`), and skip constructing unused optional fields on the common path.

### 3. Tokenize off the generation thread

- Move `tokenizer.encode` from `enqueue_request` (`scheduler.rs:2275`) to the HTTP handler side (`src/server/routes/chat.rs`, where the chat template is already applied on the tokio worker at `:365`) and pass token ids through the request struct. The tokenizer is already shared/thread-safe enough to construct encodings outside; verify `add_special` handling stays identical (`scheduler.rs:2275` semantics).

## Acceptance criteria

- [ ] Detokenization cost is O(tail) per token: demonstrated by a benchmark generating 4096 tokens on qwen2.5-0.5b-bf16 where server tok/s no longer degrades with output length (record tok/s for tokens 0-512 vs 3584-4096; current code degrades measurably).
- [ ] Server streaming throughput on qwen2.5-0.5b-bf16 within 10% of `mlxcel-bench-decode` for the same generation on GB10.
- [ ] Byte-exact stream output vs the old implementation across: Korean/CJK text, emoji, byte-fallback tokens, stop-string truncation mid-multibyte, logprobs enabled. Add a regression test comparing accumulated stream text to a one-shot `tokenizer.decode` of the final ids.
- [ ] Long-prompt request (8k tokens) no longer stalls concurrent decode: verified with the #624 concurrency script (TTFT of a concurrent short request stays bounded while an 8k-prompt request is admitted).
- [ ] `cargo test --features cuda` (or default features on a dev box; this issue is backend-agnostic) passes with the new streaming tests.

## Validation

```
cargo build --release
./target/release/mlxcel-server --model ./models/qwen2.5-0.5b-bf16 &
python3 scripts/bench_serving_concurrency.py --concurrency 1 --max-tokens 4096
```

## References

- Detok: `src/server/model_worker.rs:1936-2132`.
- Route callback: `src/server/routes/chat.rs:640-810`; SSE plumbing `src/server/streaming.rs:117-165`.
- Enqueue tokenize: `src/server/batch/scheduler.rs:2275`.

inureyes:
## Summary

Reduces three host-side per-token costs on the serving path (issue #633): O(N)-per-token detokenization, three per-token mutex acquisitions in the chat streaming callback, and prompt tokenization on the scheduler thread. The dominant win is the incremental detokenizer, which turns the quadratic-with-output-length detok cost into O(window) per token while staying byte-exact.

### 1. Incremental detokenizer (`src/server/model_worker.rs`)

`StreamingDecodeState` previously called `tokenizer.decode(&all_ids, false)` over the full token history on every generated token (regular emit path and byte-fallback re-sync), which is O(N) per token and O(N^2) over a full generation, on the single generation thread. It now re-decodes only a bounded suffix window: two decodes anchored at the same `prefix_offset` (the already-emitted window prefix, and that prefix plus the newest tokens), emitting the delta past the shared prefix. Since both decodes share the left context, the shared prefix cancels out and the delta is byte-identical to a whole-history decode. `read_offset` only advances to a UTF-8-complete boundary, so an incomplete multibyte sequence (byte-fallback `<0xXX>` runs or split byte-level pieces) is held until it completes and is never streamed as U+FFFD. Mirrors the mlx-lm `NaiveStreamingDetokenizer` and the llama.cpp/vLLM incremental detokenizers.

A `MAX_HELD_TOKENS` cap bounds the window on degenerate never-completing output, and a `strip_prefix` guard keeps the delta slice panic-safe when a decoder retroactively rewrites already-emitted text (the `tokenizers` `ByteFallback` decoder renders an entire byte run as U+FFFD once it holds an incomplete byte). The public `StreamingDecodeState` surface is unchanged, so the batched, single, speculative, diffusion, XLA, and CLI decode call sites are untouched. The now-redundant `token_piece` byte-fallback buffer, `parse_byte_fallback_token`, and `safe_emit_boundary` helpers are removed; their behavior (never emit incomplete UTF-8) is preserved by the window guard and covered end-to-end by the streaming tests.

### 2. Streaming-path diet (`src/server/routes/chat.rs`)

The chat streaming callback locked three separate mutexes per token (tool-call accumulator, logprobs buffer, stream filter). They are collapsed into one `StreamCallbackState` behind a single `Mutex`, so the hot callback locks once per token; emitted chunks are collected and sent after the lock is released. The lock is required only for the `Send` bound on the `spawn_blocking` closure (an `Arc<RefCell>` would not be `Send`) and is never contended. The SSE wire format is byte-identical.

### 3. Tokenize off the generation thread (`src/server/model_provider.rs`, `src/server/batch/scheduler.rs`)

Prompt tokenization moved out of `BatchScheduler::enqueue_request` (scheduler thread, which stalled concurrent decode ticks for a long prompt) to the request-dispatch path. `ModelProvider` holds an optional pre-tokenizer and encodes the prompt in `send_generate_request_with_cancellation` (HTTP-side thread), passing the ids through `ModelRequest::Generate`. The scheduler uses them when present and falls back to tokenizing itself otherwise. `tokenize_prompt_for_generation` is the single source of truth for the `add_special` convention (suppressed on a literal `<bos>`/`<s>` prefix), so both paths are byte-identical. Legacy, XLA, diffusion, and test paths keep the `None` fallback.

## Test plan

Hardware: Apple M1 Ultra (macOS), features `metal,accelerate`. Model: `models/qwen2.5-0.5b-bf16` (bf16, ByteLevel tokenizer, `clean_up_tokenization_spaces=false`).

Measured here:

- **Detok is O(window), not O(N).** Single greedy 4096-token stream, tok/s bucketed by output position (custom SSE arrival timer):

  | | first 512 tok/s | last 512 tok/s | end/start ratio | overall decode tok/s |
  |---|---|---|---|---|
  | BEFORE (`main` b73be08d6) | 196.9 | 160.4 | 0.815 | 163.0 |
  | AFTER (this branch) | 191.8 | 179.6 | 0.936 | 181.2 |

  The length-dependent degradation drops from 18.5% to 6.4% (the residual is GPU warmup/thermal, not detok). Tail throughput +12% (160.4 -> 179.6), overall decode +11% (163.0 -> 181.2). `scripts/bench_serving_concurrency.py --concurrency 1 --max-tokens 4096` corroborates the overall figure (162.8 -> 180.9 aggregate tok/s).

- **Byte-exact streaming.** New unit tests beside the code assert the accumulated streamed text equals a one-shot `tokenizer.decode` of the final ids across Korean, mixed CJK/emoji/ASCII, regular-piece + multibyte, a non-empty prompt seam, mid-multibyte truncation, and a 300-case fuzz. All existing byte-fallback streaming regressions and the `server::routes::chat` / `server::streaming` tests pass.
- `cargo test --release --features metal,accelerate --lib -- server::model_provider::model_worker::tests` -> 24 passed, 0 failed.
- Combined `server::model_provider::model_worker::tests server::routes::chat::tests server::streaming::tests` -> all pass.
- `cargo check --lib --tests --features metal,accelerate`, `cargo clippy --lib --tests --features metal,accelerate -- -D warnings`, `cargo fmt --all -- --check` -> clean.

Pending (not cheaply reproducible on this hardware; recorded per acceptance criteria):

- **#624 long-prompt admission (8k-prompt request does not stall a concurrent short request's TTFT).** Item 3 moves tokenization off the scheduler thread by construction, but the mixed-workload scenario needs a client that fires one 8k-prompt request alongside short requests; `scripts/bench_serving_concurrency.py` fires N identical requests and does not model the mixed case. Suggested command for a follow-up run: `python3 scripts/bench_serving_concurrency.py --concurrency 8 --prompt-tokens 8192 --max-tokens 64` while separately timing a short request's TTFT.

## Review follow-up: stream the held detokenizer tail

The incremental detokenizer holds back the whole window while its tail is an incomplete UTF-8 sequence, and a single byte-level token (Qwen-style) can carry complete text plus a trailing incomplete byte. If generation stopped on such a token, `StreamingDecodeState::flush` appended that complete text to `generated_text` (seen by the non-streaming `result.text`) but returned nothing, so streaming clients dropped it. `flush` now returns the flushed delta (`Option<String>`, `#[must_use]`) and every streaming finish site forwards it as one final `GenerateEvent::Token` before `Done`: both `BatchScheduler` finish paths, the speculative-burst finalize, the diffusion worker, and the XLA worker's `Finished` branch (after the stop-matcher tail). The XLA stop-string finalize still drops its post-stop tail and the CLI print paths already recover the tail via a separate `decode(generated_ids)` + `strip_prefix`, so both are unchanged.

New/updated tests: a ByteLevel stub whose single token decodes to `abc` + the first byte of a 3-byte char proves `flush` returns the held `abc` tail on stop and stays byte-exact when the char completes mid-stream; the byte-fallback stream helper now folds the `flush` return into the streamed chunks and the incomplete-at-end regression asserts the U+FFFD tail is streamed exactly once and equals `result.text`. `cargo test --release --features metal,accelerate --lib -- server::model_provider::model_worker::tests server::routes::chat::tests server::streaming::tests` -> 66 passed, 0 failed. `cargo check --lib --tests --bins`, `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --all -- --check` clean.

Closes #633

### DIFF
--- a/src/server/batch/scheduler.rs
+++ b/src/server/batch/scheduler.rs
@@ -4044,7 +4044,13 @@ impl BatchScheduler {
             {
                 tracing::error!("State transition error: {err}");
             }
-            seq.decode_state.flush(&self.tokenizer);
+            // Forward any tail the incremental detokenizer held back (a final
+            // token carrying complete text plus a trailing incomplete UTF-8
+            // byte) as one last token event before Done, so streaming clients
+            // receive it (issue #633).
+            if let Some(tail) = seq.decode_state.flush(&self.tokenizer) {
+                let _ = seq.response_tx.send(GenerateEvent::Token(tail));
+            }
             let cached = seq.already_cached_tokens;
             let result = seq.decode_state.finish_with_cache(
                 seq.created_at,
@@ -4967,7 +4973,12 @@ impl BatchScheduler {
             if let Some(mut seq) = self.active_batch.remove(id) {
                 let tokens_generated = seq.generated_tokens.len();
 
-                seq.decode_state.flush(&self.tokenizer);
+                // Forward the incremental detokenizer's held tail as one final
+                // token event before Done, so streaming clients are not missing
+                // text the non-streaming result.text still carries (issue #633).
+                if let Some(tail) = seq.decode_state.flush(&self.tokenizer) {
+                    let _ = seq.response_tx.send(GenerateEvent::Token(tail));
+                }
                 let cached = seq.already_cached_tokens;
                 let result = seq.decode_state.finish_with_cache(
                     seq.created_at,
--- a/src/server/batch/scheduler.rs
+++ b/src/server/batch/scheduler.rs
@@ -2236,6 +2236,7 @@ impl BatchScheduler {
         match req {
             ModelRequest::Generate {
                 prompt,
+                prompt_token_ids,
                 options,
                 images,
                 audio,
@@ -2245,6 +2246,7 @@ impl BatchScheduler {
             } => {
                 self.enqueue_request(
                     prompt,
+                    prompt_token_ids,
                     options,
                     images,
                     audio,
@@ -2264,6 +2266,7 @@ impl BatchScheduler {
     fn enqueue_request(
         &mut self,
         prompt: String,
+        prompt_token_ids: Option<Vec<i32>>,
         options: ServerGenerateOptions,
         images: Vec<Vec<u8>>,
         audio: Vec<Vec<u8>>,
@@ -2271,16 +2274,24 @@ impl BatchScheduler {
         response_tx: mpsc::Sender<GenerateEvent>,
         cancelled: Arc<AtomicBool>,
     ) {
-        let add_special = !prompt.starts_with("<bos>") && !prompt.starts_with("<s>");
-        let token_ids = match self.tokenizer.encode(&prompt, add_special) {
-            Ok(ids) => ids,
-            Err(err) => {
-                let _ =
-                    response_tx.send(GenerateEvent::Error(format!("Tokenization error: {err}")));
-                return;
-            }
+        // Prefer the ids tokenized on the request-dispatch thread (issue #633);
+        // fall back to scheduler-side tokenization when the dispatcher had no
+        // pre-tokenizer. `tokenize_prompt_for_generation` is the shared
+        // `add_special` convention so both paths are byte-identical.
+        let mut prompt_tokens: Vec<i32> = match prompt_token_ids {
+            Some(ids) => ids,
+            None => match crate::server::model_provider::tokenize_prompt_for_generation(
+                &self.tokenizer,
+                &prompt,
+            ) {
+                Ok(ids) => ids,
+                Err(err) => {
+                    let _ = response_tx
+                        .send(GenerateEvent::Error(format!("Tokenization error: {err}")));
+                    return;
+                }
+            },
         };
-        let mut prompt_tokens: Vec<i32> = token_ids.iter().map(|&x| x as i32).collect();
 
         // Empty-prompt guard (null/empty-cache safety):
         //
--- a/src/server/batch/speculative_burst.rs
+++ b/src/server/batch/speculative_burst.rs
@@ -1773,7 +1773,11 @@ fn finalize_burst_success(
     }
 
     let tokens_generated = seq.generated_tokens.len();
-    seq.decode_state.flush(ctx.tokenizer);
+    // Forward the incremental detokenizer's held tail as one final token event
+    // before Done so streaming clients receive it (issue #633).
+    if let Some(tail) = seq.decode_state.flush(ctx.tokenizer) {
+        let _ = seq.response_tx.send(GenerateEvent::Token(tail));
+    }
     let cached = seq.already_cached_tokens;
     let result = seq.decode_state.finish_with_cache(
         seq.created_at,
--- a/src/server/batch/xla_worker.rs
+++ b/src/server/batch/xla_worker.rs
@@ -233,6 +233,10 @@ impl XlaServeWorker {
         match req {
             ModelRequest::Generate {
                 prompt,
+                // The XLA worker tokenizes on its own engine thread; the
+                // dispatch-thread pre-tokenized ids (issue #633) are unused here
+                // (the XLA provider path never sets a pre-tokenizer).
+                prompt_token_ids: _,
                 options,
                 images,
                 audio,
--- a/src/server/batch/xla_worker.rs
+++ b/src/server/batch/xla_worker.rs
@@ -354,7 +354,13 @@ impl XlaServeWorker {
                         if !tail.is_empty() {
                             let _ = response_tx.send(GenerateEvent::Token(tail));
                         }
-                        detok.flush(&self.tokenizer);
+                        // Then release any tail the incremental detokenizer held
+                        // back (a final token carrying complete text plus a
+                        // trailing incomplete UTF-8 byte). It is the last output,
+                        // so it follows the stop-matcher tail (issue #633).
+                        if let Some(detok_tail) = detok.flush(&self.tokenizer) {
+                            let _ = response_tx.send(GenerateEvent::Token(detok_tail));
+                        }
                         let mut result = detok.finish(start, prompt_token_count, max_tokens);
                         // The engine knows the authoritative reason; prefer it over
                         // the count-based inference in `finish`.
--- a/src/server/diffusion_worker.rs
+++ b/src/server/diffusion_worker.rs
@@ -368,7 +368,11 @@ fn serve_streaming_request<G>(
 
     match result {
         Ok(finish_reason) => {
-            decode_state.flush(tokenizer);
+            // Forward the incremental detokenizer's held tail as one final token
+            // event before Done so streaming clients receive it (issue #633).
+            if let Some(tail) = decode_state.flush(tokenizer) {
+                let _ = response_tx.send(GenerateEvent::Token(tail));
+            }
             let mut gen_result: GenerationResult = decode_state.finish_with_cache(
                 start,
                 engine_prompt.len(),
--- a/src/server/diffusion_worker.rs
+++ b/src/server/diffusion_worker.rs
@@ -198,6 +198,9 @@ pub(crate) fn run_diffusion_worker_loop(
             }
             ModelRequest::Generate {
                 prompt,
+                // Diffusion workers tokenize internally; the dispatch-thread
+                // pre-tokenized ids (issue #633) are not used here.
+                prompt_token_ids: _,
                 options,
                 images,
                 audio,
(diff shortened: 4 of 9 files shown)


inureyes:
Orchestrator E2E verification on the branch build (Apple M1 Ultra, `models/qwen2.5-0.5b-bf16`, real `mlxcel-server`): accumulated SSE stream deltas vs the non-streamed response content for five cases — Korean, emoji, mixed CJK/Korean/English, byte-fallback (café), and a 2554-byte long generation.

Result: korean / cjk-mixed / byte-fallback / long are byte-identical between stream and non-stream. The emoji case differed, and both differences trace to pre-existing behavior, not this PR:

- Leading space: the non-stream chat path passes the text through `clean_structural_tokens` → `strip_thinking`/`clean_content_markers`, which `.trim()`s (src/server/tool_calls/parser.rs, untouched by this PR), while the stream path emits raw deltas. Reproduced 3/3 deterministically on both paths; same asymmetry class as the router-vs-single-node leading space noted in PR #707's E2E. Pre-existing main behavior.
- Tail divergence: repeating the identical non-stream request three times produced two different generations (md5 1864c0232d49 vs 01a3b170b22d) — greedy run-to-run jitter on this bf16 checkpoint (Metal reduction-order nondeterminism, the known f16-jitter class), so comparing two separate requests cannot be token-stable for this model.

Net: no streaming byte-exactness regression attributable to this PR in E2E; the strong invariant (accumulated stream text == one-shot decode of final ids) is pinned by the new unit/fuzz tests.

inureyes:

### Intent
Cut three host-side per-token serving costs (issue #633): O(N)-per-token detokenization, three per-token chat-stream mutexes, and prompt tokenization on the scheduler thread. The core is a windowed incremental detokenizer mirroring vLLM/mlx-lm, kept byte-exact.

### Findings

**HIGH: streaming loses trailing complete text when the last content token ends mid-multibyte.**
`on_token` holds the entire window whenever `new_text` ends in U+FFFD (`src/server/model_worker.rs:2161`). When a single token carries new complete text plus an incomplete trailing byte (common for ByteLevel/byte-fallback tokenizers) and generation stops right after (max_tokens, EOS mid-character, or a stop-string), that complete text is never emitted as a `GenerateEvent::Token`. It is only appended to `generated_text` by `flush()` (`:2183`, returns nothing) and surfaces solely in `GenerateEvent::Done(result.text)`. The streaming drain consumes only `Token`/`TokenWithLogprobs` and ignores `Done.text` (`src/server/model_provider.rs:1214-1225`), so streaming clients miss the final characters while non-streaming clients keep them. This is a regression from the old `safe_emit_boundary`, which emitted the complete prefix eagerly and held only the incomplete tail. The new tests do not catch it because the byte-fallback stub emits one byte per token, so complete text can never be stranded together with an incomplete tail in the same held window; only a multi-byte token that spans a UTF-8 boundary triggers it.

Recommended fix: have `flush()` return the flushed delta and emit it as a final `GenerateEvent::Token` before `Done` at every streaming finish site (`scheduler.rs:4047` and `:4970`, `speculative_burst.rs:1776`, `xla_worker.rs:357`, `diffusion_worker.rs:371`); mirror it into the CLI print paths. Validate with the cohort/handoff parity suites and a real checkpoint whose `max_tokens` lands mid-multibyte. Not auto-fixed here because a correct fix spans hot serving paths outside the narrow test scope this review could execute.

**LOW (fixed): stale `// Used by:` discovery comment** on `MlxcelTokenizer::token_piece` still named `StreamingDecodeState`, which no longer calls it. Corrected in d4fb044e0.

### Verified clean
- Removed helpers (`parse_byte_fallback_token`, `safe_emit_boundary`, byte-fallback buffer) have no remaining non-test callers; all `StreamingDecodeState` call sites (batched, single, speculative, diffusion, XLA, CLI) compile against the unchanged public surface (`cargo check --lib --tests` and `--bins` clean).
- Window seam is byte-exact: both decodes share `prefix_offset`, so leading-space/SentencePiece context cancels; legitimate U+FFFD stays byte-exact (delayed up to `MAX_HELD_TOKENS`).
- Chat mutex collapse is per-request, holds one std mutex only within the synchronous `spawn_blocking` callback (no lock across `.await`), releases before sending, and preserves reasoning -> content -> placeholder chunk order and logprob wire format.
- Logprob token strings still come from `build_single_token_chat_logprobs`, independent of the removed `token_piece` path.
- Pre-tokenization funnels through a single `ModelRequest::Generate` construction with a shared `tokenize_prompt_for_generation` (`add_special` identical to the old scheduler call); `None` fallback intact; no double-tokenization; Moondream text-only re-tokenization unaffected.
- Fuzz test is deterministically seeded (xorshift64), not flaky.

### Checks
- `cargo fmt --all -- --check`, `cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings`: clean.
- `server::model_provider::model_worker::tests` + `server::routes::chat::tests` + `server::streaming::tests`: 64 passed, 0 failed.

PR #740, opened 2026-07-11
perf(cuda/quant): amortize small-M quantized matmul with a multirow qmv path

inureyes:
## Summary

Implements the amortizing small-M quantized-matmul path from #725. On Blackwell sm_120/121 the CUDA dispatch routes `M*B < 8` to per-row `qmv`, which maps one input row per `grid.y`/`grid.z` block and re-reads the full weight matrix once per row. That made batched decode aggregate throughput flat in B and made the speculative-verify `[1, K]` forward cost ~K classic forwards, the root cause of the GB10 MTP regression (#638 / PR #733, corrected framing in #738 / PR #739).

The fix is a multirow qmv variant in the `qmv.cu` overlay (`src/lib/mlx-cpp/patches/mlx/backend/cuda/quantized/qmm/qmv.cu`): when the weights are broadcast and `2 <= M*B <= 8`, one warp keeps per-row accumulator arrays and applies each dequantized weight tile to every input row, so weight DRAM traffic is O(1) in the row count instead of O(R). The per-row arithmetic (dequant order, fma order, accumulator types, final float reduction) is kept exactly equal to the stock kernel, so per-row outputs are bit-identical to the per-row launches it replaces. `M*B == 1` (classic single-token decode) keeps the stock kernel unchanged, and `gather_qmv` (MoE, per-row expert indices) is untouched. Kill switch: `MLXCEL_QMV_MULTIROW=0`.

## Measurements (GB10, sm_121)

MTP speculative decoding, `gemma-4-12b-it-4bit` + `gemma-4-12b-it-assistant-4bit`, same invocations as the #638 matrix (baseline 14.5 tok/s):

| Kind | K | tok/s | speedup | pre-#725 | acceptance | mean accepted len |
|------|---|------:|--------:|---------:|-----------:|------------------:|
| mtp | 2 | 19.0 | 1.31x | 0.77x | 56.6% | 0.57 |
| mtp | 4 | 21.2 | 1.46x | 0.52x | 35.8% | 1.07 |
| mtp | 8 | 20.4 | 1.41x | 0.52x | 34.1% | 1.02 |
| mtp | 4 (`MLXCEL_QMV_MULTIROW=0`) | 7.7 | 0.53x | — | 35.8% | 1.07 |

[...]

Issue #725, opened 2026-07-10
perf(cuda/quant): amortizing small-M quantized GEMM for batched decode on Blackwell


inureyes:
## Context

This is the follow-up carved out of #637 (part of epic #623). #637 fixed the large-M quantized *prefill* GEMM on sm_120/121 by raising the CTA tile M cap to 128 (merged in PR #723, +38% prefill on llama-3.1-8b-4bit on GB10). But it does NOT address small-M **batched decode**.

## The gap

On GB10 (Blackwell sm_121), batched decode does not amortize weight bandwidth across the batch:

- The qmm dispatch (`patches/mlx/backend/cuda/quantized/quantized.cpp`) routes `M*B < 8` (batched decode with B in [2,8)) to per-row `qmv` (matrix-vector), which re-reads the full weight matrix for each of the B sequences -> O(B) weight reads.
- `M*B >= 8` routes to `qmm_sm80`, but the Ampere 128-wide tile is ~94% wasted at M=8, so it is *worse* than per-row qmv (measured).
- Net: aggregate decode throughput is flat (~45 tok/s at B=1/2/4 on llama-3.1-8b-4bit on GB10); batched and sequential decode are a throughput wash. Measurements and dispatch map are in `docs/benchmark_results/qmm-sm121-tile-tuning-gb10-2026-07-10.md`, and the operator-facing caveat is in `docs/CONTINUOUS_BATCHING.md`.

## Impact

This is why #714's `--parallel 4` serving default does not improve aggregate decode throughput on CUDA (it is a TTFT/concurrency knob there, not a throughput lever). Also caps #638 (speculative continuous-batching) benefit on CUDA.

## What's needed

An amortizing small-M path on Blackwell: either a weight-reusing batched `qmv` (load the weight tile once, apply across the small batch of B columns) or an sm_120/121 small-M quantized GEMM tile that keeps weights resident across the batch. This is likely upstream MLX territory (the qmm kernels are MLX JIT templates); an upstream MLX issue with the ncu evidence would be the path, tracked here.

## Acceptance Criteria

- [ ] Small-M batched decode (B in [2,8)) on sm_120/121 amortizes weight reads across the batch instead of O(B) full-matrix re-reads.
- [ ] Aggregate decode throughput on llama-3.1-8b-4bit (GB10) scales with batch size B rather than staying flat at ~45 tok/s.
[...]

Issue #638, opened 2026-07-02
perf(speculative): CUDA drafter pairing matrix and continuous-batching integration


inureyes:
Part of #623. Depends on #625 (baseline) and #632 (scheduler pipelining; the burst integration builds on the reworked decode loop).

## Context

Dense single-stream decode on GB10 sits at 83-87% of the memory-bandwidth roofline (epic table), so the only large single-stream lever left is executing fewer weight-read passes per emitted token: speculative decoding. GB10 is a favorable target: its compute-to-bandwidth ratio is much higher than Apple Silicon, so verifying K draft tokens in one pass is comparatively cheaper.

The infrastructure exists:
- MTP drafters: `src/lib/mlxcel-core/src/speculative/mtp/` (B=1 `MtpGenerator`, batched `MtpBatchedGenerator`, adaptive policy `adaptive.rs`); gemma4 MTP target `src/models/gemma4_mtp_target.rs`.
- DFlash drafters: `src/lib/mlxcel-core/src/drafter/dflash/` (B=1 and batched round loops; uses `async_eval_all`).
- Server dispatch: `SpeculativeDispatch { Disabled, Mtp, DFlash }` (`src/server/speculative_dispatch.rs`), env gates `MLXCEL_ENABLE_MTP_B1`, `MLXCEL_ENABLE_MTP_BATCH`, `MLXCEL_MTP_ADAPTIVE`, `MLXCEL_ENABLE_MTP_BATCH_RAGGED` (`docs/environment-variables.md:126-129`).
- Bench harness: `src/bin/speculative_bench.rs`, which currently captures no-drafter baselines and scaffolds the speculative numerator rows "for follow-up to fill in" (its module docs); deferral matrix in `docs/model_tests.md` (Speculative drafters section).

Two gaps:
1. The pairing matrix is unfilled: no measured accepted-length/speedup numbers for drafter-target pairs on CUDA (or largely anywhere).
2. Server integration head-of-line-blocks: a speculative request runs as a "burst" that completes the entire request at prefill time, stalling every concurrent classic decode row for its duration (documented at `src/server/model_worker.rs:400-438`; burst impl `src/server/batch/speculative_burst.rs`, dispatched from `execute_prefill` via `try_speculative_burst`, `scheduler.rs:2803-2811,2889`).

## Scope

1. Fill the drafter pairing matrix on GB10 CUDA and tune round length for CUDA cost ratios.
2. Remove the burst head-of-line block so speculative and continuous batching compose.

[...]

Issue #165, opened 2026-06-05
perf: validate B=1 MTP default-on on lower-bandwidth Apple Silicon (M1/M2)


inureyes:
Validate the B=1 MTP default-on decision (#159) on lower-bandwidth Apple Silicon (M1 / M2 / base M-series), and add a per-hardware safeguard if the default regresses there.

## Background

#159 flipped `MLXCEL_ENABLE_MTP_B1` to **on by default** for all MTP targets, based on **Apple M5 Max** measurements (~1.2-1.4x on the 31B, ~1.87x on the 12B Unified, byte-identical at temperature 0). The guard that previously declined B=1 MTP cited earlier "B=1 is slower than classic decode" measurements, which were likely taken on **lower-bandwidth Apple Silicon**.

B=1 speculative decoding pays off only when the target verify forward has spare bandwidth/compute to absorb the extra draft positions. On a lower-bandwidth chip (M1 / M2 / base M-series) the single target forward is closer to bandwidth-saturated, so B=1 MTP may be break-even or slower than classic decode, i.e. the new default could be a regression there. `MLXCEL_ENABLE_MTP_B1=0` opts out, but the **default** should be correct per hardware.

## Goal

Measure B=1 MTP (12B Unified + 4-bit assistant, and 31B + bf16 assistant) on M1 / M2-class hardware, classic vs B=1 MTP at temperature 0 (byte-identical expected). Then either:

- confirm B=1 MTP is still beneficial there and keep the unconditional default-on, or
- add a per-hardware / heuristic default (e.g. gate the default-on a memory-bandwidth or compute estimate, or a runtime micro-probe) so lower-bandwidth chips fall back to classic by default while M5 Max-class chips keep the speedup. Keep `MLXCEL_ENABLE_MTP_B1` as the override either way.

## Touchpoints

- `src/server/batch/speculative_burst.rs` — `mtp_b1_burst_enabled()` (currently env-only default-on, M5 Max rationale in the doc comment).
- `src/server/batch/scheduler.rs` — the singleton B=1 decline guard.
- `docs/benchmark_results/gemma4-mtp-speculative-decoding.md` / `docs/environment-variables.md` — document per-hardware behavior.

## Acceptance criteria

- B=1 MTP measured on at least one M1/M2-class chip for the 12B Unified and 31B pairs (classic vs MTP tok/s, temperature-0 byte-identity).
[...]

PR #742, opened 2026-07-11
perf(speculative): settle the adaptive MTP policy from measured round cost

### LINKED ISSUE #736: perf(speculative): replace the sqrt(K) verify-cost heuristic with measured round cost in the adaptive MTP policy
## Context

PR #733 (issue #638) made `mtp_policy::estimate_speedup_scaled` de-rate the verify cost by `sqrt(K)` on compute-bound hosts, calibrated against a measured "~1.9x round-cost growth from K=2 to K=8" on GB10. That calibration point was mislabeled: `--block-size 8` ran at an effective K=4, because the drafter's configured block size is 4 and the adaptive block controller (`effective_mtp_block_size`) never expanded at 35% acceptance. The true measurement is 2x verify width for 1.94x time, i.e. verify cost is LINEAR in K on GB10.

The 2026-07-11 investigation explains why it is linear: the MTP verify is a `[1, K]` sequence-axis forward, and the CUDA quantized dispatch (`src/lib/mlx-cpp/patches/mlx/backend/cuda/quantized/quantized.cpp:229-245`) routes `M*B < 8` to per-row qmv (the sm90 CUTLASS path is Hopper-gated), so a K-wide verify re-reads the weights K times. Measured on GB10: classic forward 69 ms, K=2 round 140 ms, K=4 round ~270 ms; the measured 0.77x/0.52x speedups from PR #733 (`docs/benchmark_results/speculative-pairing-gb10-2026-07-10.md`) equal tokens-per-round divided by round-cost-in-forwards exactly. Same root cause as the flat batched-decode throughput tracked in #725.

## Impact

At K=4 the policy assumes a verify-cost multiple of sqrt(4) = 2 while the measured round cost is ~3.9 classic-forward equivalents, so the estimator under-costs rounds by ~2x on CUDA. The Gemma pairing was declined only because its accepted_len was very low (the estimate landed at ~1.0 against the <=1.0 decline threshold); a moderately better pairing could be wrongly enabled and regress live serving. This is a policy-correctness fix, not a tuning tweak.

## What's needed

Stop inferring verify cost from a shape heuristic; use what the profiling window already measures.

- The profiling window already records per-round `verify_ms` and `drafter_ms` (`src/server/batch/mtp_policy.rs`, `ProfileAccumulator`).
- Add a measured classic single-token step time for the same target, either from the scheduler's own decode timing or a short probe.
- Estimate speedup directly as `(accepted_len + 1) / ((verify_ms + drafter_ms + per-round overhead) / classic_step_ms)`.
- Apple Silicon verdicts must remain correct: either byte-identical decisions, or re-validated against M5 Max measurements.

## Acceptance Criteria

- [ ] Unit tests pinning the measured-cost estimator.
- [ ] The GB10 Gemma pairing declines with clear margin, not borderline.
- [ ] A bandwidth-bound favorable pairing still enables.
- [ ] The sqrt(K) path removed, or demoted to a fallback used only when no classic-step measurement exists.
- [ ] Docs updated: the policy section of `docs/benchmark_results/speculative-pairing-gb10-2026-07-10.md`, and `docs/environment-variables.md` if flags change.

## Technical Considerations

- Refs #638, PR #733, #725, and the docs correction issue #738.

inureyes:
## Summary

Replaces the adaptive MTP policy's shape-heuristic speedup estimator with a measured one (issue #736). The old estimator assumed the K-wide verify costs `verify_cost_multiple` classic forwards (1.0 on Apple Silicon, `sqrt(K)` elsewhere, issue #638). That multiple was calibrated against the pre-#725 per-row-qmv verify and is wrong in both directions across kernel eras: it under-costed the pre-#725 linear-in-K verify (the borderline ~1.0 decline that #733 shipped) and over-costs the post-#725 multirow verify, wrongly declining the now-favorable GB10 Gemma pairing (measured 1.46x in `speculative_bench`, estimated ~1.0 by the heuristic). The bench bypasses the policy, which is why PR #740's win did not reach serving.

## How it works

While the policy is profiling a pairing, each B=1 burst runs `PROFILE_PROBE_ROUNDS_PER_BURST = 2` classic-step probe rounds: drafterless rounds whose `[1, 1]` verify forward is shape-identical to a classic decode step. Each probe emits one real greedy token (the walk over an empty draft returns the target argmax), so no compute is wasted, temperature-0 output stays byte-identical to classic decode, and the total profiling overhead is 8 classic-paced tokens per pairing lifetime. Probes are excluded from the round/acceptance aggregates and from the adaptive block controller's window, so they cannot dilute the speculative signal.

The estimator is then fully measured: `speedup = (accepted_len + 1) * classic_step_ms / round_ms`, with `round_ms = verify + drafter + walk/finalize/re-arm overhead` and `classic_step_ms` the MEDIAN of per-burst probe means. The median matters: the first burst's probes absorb the one-time CUDA kernel/graph compilation for the `[1, 1]` verify shape (measured ~1.6 s per probe vs the ~74 ms steady state on GB10), and the plain mean reported ~450 ms and inflated the estimate ~6x in the first validation run; the median lands on the steady state (a regression test pins this). The `sqrt(K)` heuristic remains only as a fallback for profiling windows that collected no probe signal. The persisted hint format bumps to v3, so verdicts settled by the old estimator (or against the pre-#725 kernel) re-profile once.

Because the estimator now measures rather than models, it derives the right verdict on every host without hardware heuristics: unit tests reproduce the M5 Max enable (measured 1.87x) and the M1 Ultra decline (measured 0.75-0.96x in #165) purely from their measured shapes, and the pre-#725 GB10 shape declines with clear margin (0.53) instead of at the boundary.

## End-to-end validation (GB10, serving path)

`mlxcel serve -m gemma-4-12b-it-4bit --draft-model gemma-4-12b-it-assistant-4bit --draft-kind mtp`, six greedy 128-token requests:

- The profile settles after 4 samples to **ENABLE** with `classic_step_ms=76.26`, `est_speedup=1.86 [measured]` (accepted_len 1.36, verify 89.6 ms, drafter 7.1 ms), and the v3 hint persists the verdict.
- Post-settle requests run **22.9-24.4 tok/s** end-to-end vs **13.2-13.4 tok/s** with `MLXCEL_ENABLE_MTP_B1=0` on the same binary: a ~1.8x serving speedup that the pre-#736 policy declined. The measured estimate (1.86, decode-only) matches the end-to-end ratio (1.77, includes prefill).

Applies identically to `mlxcel serve` and the standalone `mlxcel-server` (both drive the same `server::batch` scheduler and policy).

## What changed

- `src/lib/mlxcel-core/src/speculative/mtp/generator.rs`: `with_profile_probe_rounds` builder; probe branch in the round loop (empty draft, `[1, 1]` verify, separate `probe_rounds`/`probe_ms` diagnostics); `MtpAcceptanceSummary` gains `overhead_ms`, `probe_rounds`, `probe_ms`.
- `src/server/batch/mtp_policy.rs`: `estimate_speedup_measured`; accumulator gains overhead/probe folding, `classic_step_ms()` (median of per-burst means), `round_cost_ms()`; `record_b1_sample` prefers the measured estimate and logs which estimator decided; `profile_probe_rounds()` accessor (non-zero only while profiling); hint v3.
- `src/server/batch/speculative_burst.rs` + `scheduler.rs`: `profile_probe_rounds` plumbed from the policy through `BurstContext` into the generator (B=1 only; batched and DFlash arms unaffected); probe-only summaries pass the profile filter.
- Tests: probe round-loop test (core), measured-estimator closed form + guards, GB10 post-#725 enable, pre-#725 decline with margin, M5/M1U verdicts from measured shapes, JIT-pollution median regression, probe-only sample accounting, probe-request lifecycle (10 new tests total).
- Docs: `environment-variables.md` `MLXCEL_MTP_ADAPTIVE` row rewritten for the measured estimator and v3 hints; `qmv-multirow-gb10-2026-07-11.md` policy note marked resolved; `CONTINUOUS_BATCHING.md` gains the "keep `--n-parallel` at 7 or below on CUDA" caveat (a full batch of 8+ decode rows crosses into the under-tiled `qmm_sm80` shape until the upstream small-M tile lands).

## Test plan

- [x] `cargo test --release -p mlxcel-core --features cuda --lib speculative::mtp -- --test-threads=1` — 42 passed.
- [x] `cargo test --release --features cuda --lib batch::mtp_policy -- --test-threads=1` — 40 passed.
- [x] `cargo test --release --features cuda --lib batch::speculative_burst -- --test-threads=1` — 50 passed.
- [x] `cargo check --features cuda --lib --tests -p mlxcel-core -p mlxcel` clean; `cargo fmt` clean.
- [x] GB10 serving E2E: profile settles ENABLE with measured estimator; ~1.8x end-to-end vs classic-forced on the same binary; v3 hint persisted.

Closes #736


inureyes:
## PR Finalization Complete

### Checked
- `cargo fmt --check`: clean, no diff, no changes needed.
- Reviewer and security passes already green (57c49bc, fe5a301); test plan checkboxes match the state reported by the orchestrator (core mtp 42, policy 40, burst 50, plus the GB10 serving E2E), left unedited since accurate.
- Swept `docs/` for stale `#736`/`sqrt(K)`/`MLXCEL_ENABLE_MTP_B1=1`-workaround guidance now that the measured estimator has landed.
- Confirmed `environment-variables.md`'s `MLXCEL_MTP_ADAPTIVE` row covers probes, fallback, and v3 re-profile, but was missing the median-of-per-burst-probe-means detail from the PR description (guards against the first burst's CUDA JIT-compilation outlier).
- No Korean docs mirror exists for any of the touched files; nothing to translate.

### Changed (97798c3)
- `docs/benchmark_results/model_tests.md`: the "Serving caveat" sentence described `sqrt(K)` as the live de-rate and pointed at `MLXCEL_ENABLE_MTP_B1=1` as a manual workaround pending #736. Rewritten to state #736 is resolved: the policy now settles from the measured comparison and `sqrt(K)` is only the no-probe-signal fallback.
- `docs/benchmark_results/speculative-pairing-gb10-2026-07-10.md`: the "Known limitation (#736)" section and its Follow-ups bullet framed #736 as open work. Marked resolved in place, past-tensed the limitation, and pointed the follow-up item at `qmv-multirow-gb10-2026-07-11.md`.
- `docs/environment-variables.md`: added the median-vs-mean aggregation detail to the `MLXCEL_MTP_ADAPTIVE` row so the env var doc alone explains why the estimator resists the first-burst compilation outlier, matching the PR description's rationale.

### Left open
- Nothing outstanding from this pass. No new env var was needed since the probe count is a compile-time constant, matching the task's expectation.

Issue #632, opened 2026-07-02
perf(server): port lookahead async_eval pipelining into BatchScheduler decode


inureyes:
Part of #623. Can proceed in parallel with Phase 3; rebase measurements on the post-#625 baseline.

## Context

The CLI generation loop overlaps graph construction with GPU execution: it builds step n+1's graph (reshape -> forward -> fused sample) before reading back step n's token, schedules it with `ffi::async_eval_pair`, and only then syncs the current token with `item_i32` (`src/lib/mlxcel-core/src/generate.rs:1157-1260`; first token at `:1246`; `MLXCEL_FORCE_SYNC` opt-out at `:1227`). A dedicated thread-local generation stream supports the overlap (`src/lib/mlxcel-core/src/streams.rs`, installed at `generate.rs:1059`).

The server scheduler has none of this. Both decode paths synchronize every step:
- `decode_single_step` (`src/server/batch/scheduler.rs:4699`): forward at `:4724`, then synchronous `eval` + `item_i32` at `:4778-4779`.
- `execute_batched_decode` (`scheduler.rs:4300`): forward at `:4363`, fused batch sample at `:4384-4388`, then a host copy of the `[B]` token ids.

On CUDA, where kernel-launch and graph-build latency is higher than Metal, this makes server decode structurally slower than CLI decode for identical models. Overhead-bound models (<= 2B: 67-75% of roofline) lose the most; the effect also scales the whole batched path since one tick's build time is serialized against the GPU.

## Scope

Bring lookahead pipelining to the scheduler's decode, for both B=1 and batched decode, without breaking continuous-batching semantics (admission, completion, chunked-prefill interleave, cancellation).

## Implementation plan

1. Mechanics (mirror the CLI loop): after sampling step n's token array, immediately build step n+1's graph feeding the token array back as the next `[B,1]` input WITHOUT reading it to host, schedule with `async_eval` (`async_eval_pair` when logprobs are also needed), and only then extract step n's ids to host for stop checks and streaming. The sampled-token MLX array is a valid device-side input; the CLI loop already relies on this.
[...]

Issue #518, opened 2026-07-01
feat(server): let APC prompt-cache prefixes use the MTP/DFlash speculative burst (currently mutually exclusive)


inureyes:
## Summary

mlxcel implements **both** an APC / prompt-cache prefix-adoption path **and** an MTP/DFlash speculative-decoding burst, but they are currently **mutually exclusive**. When a request adopts a prompt-cache prefix, the speculative burst is declined and the request falls back to the **classic non-speculative** decode path. A cache-hit request therefore loses speculative decoding entirely. This issue tracks letting an APC-adopted prefix coexist with the speculative burst so cache hits keep the speculative speedup.

## Current state in mlxcel

- The burst gate declines speculation when a prompt-cache prefix was adopted: `src/server/batch/speculative_burst.rs:103-105` — *"The request adopted a prompt-cache prefix (`prefill_start_offset > 0`): a burst owns the cache for its full lifetime; mixing an adopted prefix would double-prefill the leading tokens."* It is enforced in `should_burst_for_sequence` (the `prefill_start_offset > 0` decline near `speculative_burst.rs:370-377`).
- Both subsystems exist independently: prompt-cache adoption in `src/server/prompt_cache/` plus scheduler adoption (`src/server/batch/scheduler.rs`, around the adopt path ~line 1245), and the MTP (Gemma 4) / DFlash (Qwen 3.5) bursts in `speculative_burst.rs`.
- Speculation happens **only** via the burst (the classic scheduler pipeline `prefill → finish_prefill → decode_single_step` is non-speculative), so declining the burst means a cache-hit request gets no speculation at all.

## Why it is gated today

The burst driver owns the full prefill + decode lifecycle and prefills the prompt from scratch. If a prefix is already materialized in the adopted cache, the burst would re-prefill those leading tokens, causing a double-prefill (wasted compute and cache corruption). The current gate avoids that by declining the burst.

## What needs to be built

Teach the burst drivers (`run_mtp_burst` → `MtpGenerator`, `run_dflash_burst` → `DFlashGenerator`) to start from a non-zero adopted cache offset:

1. Thread `prefill_start_offset` (and the adopted cache handle) into the burst so it prefills only the suffix `[offset..]`, reusing the cached KV for `[..offset]`.
2. Seed the drafter/target adapter per-round state (hidden / bonus bookkeeping) consistently with the adopted offset.
[...]

PR #745, opened 2026-07-11
perf(server): tick-cooperative speculative decoding to remove the burst head-of-line block

### LINKED ISSUE #734: perf(server): tick-cooperative speculative decoding to remove the burst head-of-line block
## Context

#638 part 2 asked to replace the run-to-completion speculative burst (`src/server/batch/speculative_burst.rs`, dispatched from `execute_prefill`) with tick-cooperative slices so a speculative request no longer stalls every concurrent classic decode row for its whole duration (HOL block documented at `src/server/model_worker.rs:400-438`). PR #733 landed the measurement side (pairing matrix, adaptive-policy de-rate so GB10 declines regressing pairings, and a `BurstFinalized.burst_wall_ms` observability counter that measures the HOL stall) but deferred the slice itself.

## Blocker found (by #733)

`MtpGenerator<T>` borrows the model through its adapter, so the generator cannot be stashed across scheduler ticks (self-referential borrow); this is exactly why the burst runs to completion today. A resumable design needs a generator step API with stored per-round state (feasible: `MtpVerifyOutput` owns MLX handles, not model borrows) plus drafter re-seed on resume.

## Design sketch (from #638's plan and #733's analysis)

One speculative round (draft K + verify) per scheduler tick, re-entering the tick loop between rounds (`round_loop.rs` / `round_loop_batched.rs` are already per-round structured); per-sequence KV slots preserve state across interleaved classic decodes; the cross-tick lookahead pipeline (#632) is already excluded under speculative dispatch (`lookahead_params` returns `None`), and any new action type must call `discard_lookahead` like the `Prefill` action does; preferred end state is speculative rounds as first-class `decide_action` outcomes.

## Acceptance criteria

- [ ] With one speculative and one classic request active, the classic request's inter-token latency stays bounded (scheduler-level regression test with a mock/small model, plus manual E2E via `scripts/bench_serving_concurrency.py --concurrency 2`).
- [ ] `burst_wall_ms` drops to about one round length.
- [ ] Speculative-vs-greedy output parity holds for B=1 and batched.
- [ ] Env gate semantics unchanged.
- [ ] `cargo test --features cuda` passes.

Refs: deferred from #638 / PR #733.

inureyes:
## Summary

Lands the tick-cooperative slice deferred from #638 / PR #733: a B=1 MTP speculative request on the Gemma 4 family is now served one speculative round per scheduler tick instead of running its whole prefill + decode lifecycle inside a single tick, so concurrent classic-decode rows advance between rounds and the head-of-line stall a speculative request imposes drops from the whole burst to about one round.

## Design

PR #733 identified the blocker: `MtpGenerator<T>` borrows the model through its target adapter, so the generator cannot be stashed across scheduler ticks. The sanctioned resumable design is implemented in two layers.

**Core (mlxcel-core)**: `MtpGenerator` gains a session API. `begin_session` runs the prefill + seed and emits the first bonus; `step_session` runs exactly one round (probe or draft + verify + walk + finalize + emit); `finish_session` stamps the diagnostics and acceptance summary. All cross-round state lives in the new owned `MtpSessionState` (`MtpVerifyOutput` holds MLX array handles and plain integers, no model borrows), so the borrowing generator is dropped and reconstructed per tick while the state and the un-reset drafter survive. The drafter's shared-KV arm moved from the end of a round to the start of the next one; the arm is a stateless overwrite consumed only by `draft_block`, so per-round computation is unchanged and greedy output is identical, while making the session resumable from the state alone. `generate()` is reimplemented as a tight loop over the same session methods, so the run-to-completion path (offline CLI, benches, DFlash-adjacent flows) and the sliced path compute identical rounds by construction. `into_parts()` recovers target + drafter per tick, and a blanket `impl MtpTarget for &T` supports by-value callers over one long-lived target.

**Server**: new `speculative_slice.rs` holds the cross-tick `MtpSliceJob` (owned `SequenceInfo`, session state, stream state, and the bound drafter held WITHOUT reset between slices; the resetting `return_drafter` runs once at session end) plus generic begin/step drivers. The scheduler gains a first-class `BatchSchedulerAction::SpeculativeRound` outcome that calls `discard_lookahead` exactly like the `Prefill` arm (a guaranteed no-op under speculative dispatch, kept for the #632 invariant), and a strict round/classic alternation (`slice_takes_tick` plus a fairness flag): under contention a classic row waits at most about one speculative round, and the slice waits at most one classic action; without contention the slice takes every tick and never Idle-blocks. Slice 0 is the prefill + seed; the same gates as `run_mtp_burst` apply in the same order (variant gate before drafter IO, #518 adopted-prefix suffix resolution, drafter lazy-load + compat + bind), declines fall back to classic decode, and at most one slice is in flight at a time (a second speculative request arriving mid-slice runs classic, matching the legacy worker's effective capacity of one speculative request).

**Streaming**: `finalize_burst_success` is split into `begin_burst_stream` / `stream_burst_tokens` / `finalize_burst_stream`, shared by both paths. Each slice streams its round's accepted tokens as they land (the legacy burst emitted one end-of-burst lump; the design doc note is updated), with identical thinking-budget enforcement, EOS-vs-Length classification, and logprobs payloads.

**Fallback kept, as stated per the issue guidance**: the legacy run-to-completion burst remains for the DFlash B=1 arm and every B>1 batched arm (their generators expose no resumable step API yet), and `MLXCEL_MTP_TICK_SLICE=0` is a new operator escape hatch that forces the MTP B=1 arm back onto the legacy burst. The MTP B=1 path on the Gemma family is tick-cooperative by default.

## Observability

`BurstFinalized::burst_wall_ms` is re-scoped honestly: it now reports the maximum wall-clock any single scheduler tick spent on the request, which is the realized HOL bound (about one round on the slice path; unchanged whole-burst meaning for the one-slice legacy arms). New `burst_active_ms` (cumulative worker occupancy) and `slices` fields accompany it in the finalize log line. The acceptance-summary counters feeding the adaptive MTP policy (#333) are identical whether the rounds ran in one tick or across many, and the #736 probe rounds run in the first slices exactly as they ran in the first rounds.

## Env gates

`MLXCEL_ENABLE_MTP_B1`, `MLXCEL_MTP_ADAPTIVE`, `MLXCEL_ENABLE_MTP_BATCH`, `MLXCEL_ENABLE_MTP_BATCH_RAGGED`, and `MLXCEL_QMV_MULTIROW` keep their existing meanings; the adaptive policy's gates run before the slice dispatch unchanged. `MLXCEL_MTP_TICK_SLICE` (default on) is the only new gate and is documented in `docs/environment-variables.md`.

## What changed

- `src/lib/mlxcel-core/src/speculative/mtp/generator.rs`: `MtpSessionState` / `MtpStepOutput`, `begin_session` / `step_session` / `finish_session`, `into_parts`; `generate()` reimplemented over the session API.
- `src/lib/mlxcel-core/src/speculative/mtp/target.rs`: blanket `impl MtpTarget for &T`.
- `src/lib/mlxcel-core/src/speculative/mtp/tests.rs`: session-vs-generate identity tests (full-accept, partial-accept, EOS-mid-round, probe-rounds-first, cancel-between-rounds, finished no-op), all with per-round generator reconstruction.
- `src/server/batch/speculative_slice.rs` (new): `MtpSliceJob`, begin/step slice drivers, `slice_takes_tick`, `MLXCEL_MTP_TICK_SLICE` gate.
- `src/server/batch/speculative_slice_tests.rs` (new): slice-vs-run-to-completion stream/event identity, per-round commits, cross-slice cancellation, one re-arm per round, probe slices.
- `src/server/batch/scheduler.rs`: `SpeculativeRound` action wiring in `run()` / `decide_action`, slice start/step/finalize methods, factored `finish_speculative_b1` bookkeeping shared with the legacy arm, single-slice admission guard.
- `src/server/batch/speculative_burst.rs`: streaming split, `BurstFinalized` fields + re-scoped docs, `WorkerDrafterSlot::restore_unused`, module doc update.
- `src/server/batch/sequence.rs`: `BatchSchedulerAction::SpeculativeRound` variant.
- `src/server/model_worker.rs`, `docs/environment-variables.md`, `docs/benchmark_results/gemma4-mtp-speculative-decoding.md`: HOL and streaming docs updated to the new semantics.

## Test plan

- [x] `cargo test --release -p mlxcel-core --features cuda --lib speculative::mtp -- --test-threads=1` (49 passed; includes 7 new session tests)
- [x] `cargo test --release --features cuda --lib batch::speculative -- --test-threads=1` (59 passed; includes 9 new slice/policy tests)
- [x] `cargo test --release --features cuda --lib batch::mtp_policy -- --test-threads=1` (40 passed)
- [x] `cargo test --release --features cuda --lib scheduler -- --test-threads=1` (188 passed)
- [x] `cargo check --features cuda --lib --tests --bins -p mlxcel-core -p mlxcel` clean; `cargo fmt`
- [ ] Real-model E2E: `scripts/bench_serving_concurrency.py --concurrency 2` (classic inter-token gap bounded, `burst_wall_ms` about one round), greedy-parity on hardware, and the full `cargo test --features cuda` suite: deferred to the orchestrator lane per task scope.

Closes #734


inureyes:
## Orchestrator hardware validation (GB10)

Full root lib suite: 4024 passed, 0 failed (release, cuda, --test-threads=1). Release binary built clean.

**AC1, bounded classic latency under mixed load** (gemma-4-31b-it-4bit + gemma-4-31b-it-assistant-bf16, --n-parallel 2, bench_serving_concurrency --concurrency 2 --prompt-tokens 256 --max-tokens 96, warm kernels):

| mode | ttft_ms mean | ttft_ms p95 | aggregate tok/s |
|---|---:|---:|---:|
| tick slices (default) | 527.7 | 677.4 | 8.6 |
| MLXCEL_MTP_TICK_SLICE=0 (legacy) | 11846.3 | 15678.2 | 11.8 |

The second request no longer waits out the first request's whole burst (p95 TTFT 23x better); hol_waiters=1 observed in the slice run. The cost is ~27% aggregate throughput in this 2-concurrent scenario (strict round/classic alternation plus per-tick adapter reconstruction and shape switching between the two model paths); the legacy escape hatch restores the old behavior where aggregate matters more than latency. The legacy row's per-request decode rate is a measurement artifact of lump delivery (documented pre-#734); the slice row's 4.6 tok/s per request is a real inter-token rate, i.e. per-round streaming works.

**AC2, burst_wall_ms drops to ~one round** (gemma-4-12b-it-4bit + assistant, single request, 128 tokens): legacy single tick 5361 ms for the whole request; sliced max single tick 99.1 ms steady-state (238 ms on the first request of the process, first-slice seed) across 54 slices, burst_active_ms ~5150. Round structure identical to legacy (rounds=53, accepted_draft_tokens=74 in both modes).

**AC3, parity**: slice-on vs MLXCEL_MTP_TICK_SLICE=0 completions are byte-identical at temperature 0 (the change's own invariant). MTP vs MLXCEL_ENABLE_MTP_B1=0 classic diverges at word-choice level ('(small, fast)' parenthetical), and the legacy burst diverges from classic at the same point, so the divergence pre-dates this PR: it is the documented CUDA [1,K]-vs-[1,1] near-tie argmax jitter class (issues #165/#203). Batched B>1 arm untouched.

**AC4/AC5**: env-gate semantics unchanged (new MLXCEL_MTP_TICK_SLICE documented as the escape hatch); full cuda lib suite green as above.

PR #758, opened 2026-07-12
ci: generate and publish a CycloneDX SBOM on release

### LINKED ISSUE #719: ci: generate and publish a CycloneDX SBOM on release
## Summary

Add SBOM (Software Bill of Materials) generation to the release packaging workflow so every published release ships a CycloneDX SBOM as a release asset. Reuse the approach already in production in `lablup/backend.ai-go` (`.github/workflows/packaging.yml`, the `finalize` job): generate the SBOM with [Syft](https://github.com/anchore/syft) in CycloneDX JSON format, gzip it, name it by version, and upload it to the GitHub release.

## Motivation

- Supply-chain transparency and provenance: consumers can see exactly which dependencies a given release was built from.
- Vulnerability scanning: a CycloneDX SBOM feeds directly into scanners (Grype, Trivy, Dependency-Track).
- Parity with `backend.ai-go`, which already publishes a CycloneDX SBOM on every release.

## Reference implementation (backend.ai-go)

From `lablup/backend.ai-go` `.github/workflows/packaging.yml` (`finalize` job, "Generate SBOM" step):

```bash
VERSION="${{ needs.prepare.outputs.version }}"
TAG="${{ needs.prepare.outputs.tag }}"

# Install syft
curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin

# Generate CycloneDX JSON SBOM
syft dir:. -o cyclonedx-json > sbom.cyclonedx.json

# Compress with gzip
gzip -9 sbom.cyclonedx.json

# Rename with version for release asset
SBOM_FILE="sbom-${VERSION}.cyclonedx.json.gz"
mv sbom.cyclonedx.json.gz "$SBOM_FILE"

# Upload to release
gh release upload "$TAG" "$SBOM_FILE" --clobber
```

## Current state in mlxcel

`.github/workflows/release.yml` triggers on `release: published` and `workflow_dispatch`. Jobs: `build-macos`, `build-linux-cuda`, `build-linux-x86_64-cuda`, `notify-teams`, `promote-release`. Each build job uploads only its own platform zip plus a per-platform `.sha256` checksum via `softprops/action-gh-release@v3`. There is **no** repository-wide SBOM and no central finalize/aggregation job.

## Proposed change

Add a dedicated SBOM job to `.github/workflows/release.yml` (do not fold it into `promote-release`, which only runs when a prerelease is being promoted). The job should:

1. `needs: [build-macos, build-linux-cuda, build-linux-x86_64-cuda]` and run after the builds so the SBOM reflects the released tag.
2. Gate to real release uploads only, mirroring the existing guards: run on `github.event_name == 'release'` for `lablup/mlxcel`, and (optionally) on `workflow_dispatch` when a `release_tag` input is provided. Skip on build-only verification runs where there is no tag to upload to.
3. `permissions: contents: write` (needed for `gh release upload`).
4. `actions/checkout` the repo, install Syft, run `syft dir:. -o cyclonedx-json`, gzip `-9`, rename to `sbom-<version>.cyclonedx.json.gz`, and `gh release upload "<tag>" "<file>" --clobber`.

## Considerations / notes

- **What Syft catalogs:** mlxcel is a Rust workspace, so `syft dir:.` primarily catalogs the Rust dependencies via `Cargo.lock`. Confirm the generated SBOM actually lists the crate graph (Cargo.lock is committed).
- **MLX C++ pin:** mlxcel links MLX C++ fetched at build time (FetchContent), pinned by `MLX_EXPECTED_COMMIT`. `syft dir:.` will not automatically capture that native dependency's commit. Decide whether to note the MLX pin in the release (out of scope for a first cut, but call it out) so the SBOM does not silently omit the largest native dependency.
- **Version/tag source:** unlike backend.ai-go there is no `prepare` job exposing `version`/`tag` outputs; derive the tag from `github.event.release.tag_name` (or the `release_tag` dispatch input) and the version by stripping a leading `v`.
- **Aggregate checksums (optional, out of scope):** backend.ai-go's `finalize` job also downloads all assets and writes a single `SHA256SUMS.txt`. mlxcel already emits per-platform `.sha256` files, so an aggregate checksum file is optional and can be a follow-up.
- Pinning the Syft installer/version (rather than tracking `main`) is preferable for reproducibility.

## Acceptance criteria

- [ ] A CycloneDX JSON SBOM is generated during the release workflow and uploaded as a `sbom-<version>.cyclonedx.json.gz` release asset.
- [ ] The SBOM job runs only for real release uploads (not build-only verification runs) and is scoped to `lablup/mlxcel`.
- [ ] The uploaded SBOM lists the Rust dependency graph from `Cargo.lock`.
- [ ] Job permissions are minimal (`contents: write` only) consistent with the workflow's default-deny `permissions: {}`.
- [ ] Validated on a real (or pre-)release tag that the asset appears and is a valid CycloneDX document.

inureyes:
## Summary

Add a `generate-sbom` job to `.github/workflows/release.yml` so every real release publishes a CycloneDX JSON SBOM (`sbom-<version>.cyclonedx.json.gz`) as a release asset, matching the approach already in production in `lablup/backend.ai-go`.

## What changed

- New job `generate-sbom` in `.github/workflows/release.yml`, placed after `promote-release`, `needs: [build-macos, build-linux-cuda, build-linux-x86_64-cuda]` so it only runs once every platform build has succeeded.
- Checks out the exact released ref (`github.event.release.tag_name || github.event.inputs.release_tag || github.sha`) with `persist-credentials: false`, since this job only reads source.
- Installs Syft pinned to `v1.46.0` (both the installer script and the resolved binary version, for reproducibility instead of tracking `main`).
- Runs `syft dir:. -o cyclonedx-json`, gzips with `-9`, renames to `sbom-<version>.cyclonedx.json.gz`, and uploads with `gh release upload "$TAG" ... --clobber --repo "${{ github.repository }}"` (the explicit `--repo` is needed because `persist-credentials: false` leaves no git remote credential context for `gh` to infer the repo from).

## Key design decisions

- `runs-on: ubuntu-latest` (GitHub-hosted), matching `notify-teams` and `promote-release`. Syft catalogs source, not binaries, so no self-hosted GPU runner is needed, and `gh` is preinstalled on `ubuntu-latest`.
- Plain `needs` (no `if: always()`), matching `notify-teams`: the SBOM only publishes after a successful build, so a broken release never ships an SBOM.
- The job-level `if` gate mirrors the existing per-platform "Upload release artifacts" guard exactly (`github.event_name == 'release'` or `workflow_dispatch` with a non-empty `release_tag`), scoped to `github.repository == 'lablup/mlxcel'`, so build-only verification runs are skipped and forks cannot trigger an upload.
- `permissions: contents: write` under the file's default-deny top-level `permissions: {}`; the only permission `gh release upload` needs.
- Syft version `v1.46.0` was the latest stable tag at implementation time; pinning avoids catalog drift from tracking `main`.
- Known limitation: partial `workflow_dispatch` runs that build only a subset of platforms (e.g. `targets: macos`) will skip `generate-sbom`, because a skipped `needs` dependency blocks a downstream job by default. This is acceptable: real `release` events always build every platform, so the SBOM job always has all three dependencies satisfied on an actual release. Recording the MLX C++ FetchContent pin (`MLX_EXPECTED_COMMIT`) in the SBOM is also out of scope for this first cut, as called out in the issue.

## Validation

This is a workflow-only change; nothing here compiles, so no `cargo` commands were run.

- YAML sanity: `python3 -c "import yaml,sys; yaml.safe_load(open('.github/workflows/release.yml')); print('yaml ok')"` → `yaml ok`. `actionlint` was not available in this environment, so it was not run.
- Proved Syft will catalog the Rust dependency graph without picking up untracked files or the gitignored `./models` symlink: exported the tracked tree only via `git archive --format=tar HEAD | tar -x -C "$WORK"`, installed Syft `v1.46.0` into a scratch `bin/` directory, and ran `syft dir:"$WORK" -o cyclonedx-json > "$WORK/sbom.json"` against that export.
- Confirmed the export excluded the 20 untracked `benchmarks/*.csv` files present in the working tree and had no `models` entry.
- `jq -e '.bomFormat=="CycloneDX"'` → `true` (specVersion `1.7`).
- `jq '[.components[]? | select((.purl // "")|startswith("pkg:cargo/"))] | length'` → `439` cargo components found (out of 1163 total components), confirming the Cargo.lock crate graph is captured.

Closes #719

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -855,3 +855,58 @@ jobs:
           gh release edit "${{ github.event.release.tag_name }}" \
             --repo "${{ github.repository }}" \
             --prerelease=false
+
+  # ============================================================================
+  # Software Bill of Materials (CycloneDX SBOM) generation
+  # ============================================================================
+  generate-sbom:
+    name: Generate CycloneDX SBOM
+    needs: [build-macos, build-linux-cuda, build-linux-x86_64-cuda]
+    # Real release uploads only, scoped to the canonical repo. Skips build-only
+    # verification runs (workflow_dispatch without a release_tag), mirroring the
+    # upload guards on the build jobs above.
+    if: |
+      github.repository == 'lablup/mlxcel' && (
+        github.event_name == 'release' ||
+        (github.event_name == 'workflow_dispatch' && github.event.inputs.release_tag != '')
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+        with:
+          # Don't leave GITHUB_TOKEN in .git/config; this job only reads source.
+          persist-credentials: false
+          # Catalog the source of the released tag, not the triggering branch, so
+          # the SBOM reflects exactly what shipped.
+          ref: ${{ github.event.release.tag_name || github.event.inputs.release_tag || github.sha }}
+
+      - name: Install Syft
+        run: |
+          set -euo pipefail
+          # Pin both the installer script and the syft binary to a tag for
+          # reproducibility (tracking main would drift the catalog over time).
+          SYFT_VERSION="v1.46.0"
+          curl -sSfL "https://raw.githubusercontent.com/anchore/syft/${SYFT_VERSION}/install.sh" \
+            | sh -s -- -b /usr/local/bin "${SYFT_VERSION}"
+          syft version
+
+      - name: Generate and upload CycloneDX SBOM
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ github.event.release.tag_name || github.event.inputs.release_tag }}
+        run: |
+          set -euo pipefail
+          VERSION="${TAG#v}"
+          # syft dir:. catalogs the Rust workspace via the committed Cargo.lock.
+          # Note: it does NOT capture the MLX C++ dependency fetched at build time
+          # via FetchContent (pinned by MLX_EXPECTED_COMMIT at the top of this
+          # file); recording that native pin in the SBOM is out of scope here.
+          syft dir:. -o cyclonedx-json > sbom.cyclonedx.json
+          gzip -9 sbom.cyclonedx.json
+          SBOM_FILE="sbom-${VERSION}.cyclonedx.json.gz"
+          mv sbom.cyclonedx.json.gz "$SBOM_FILE"
+          gh release upload "$TAG" "$SBOM_FILE" --clobber --repo "${{ github.repository }}"
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -177,8 +177,10 @@ and GB10 (`121`) in a single build (`90a;100;121`), and x86_64 covering Ampere
 through Blackwell (`80;86;89;90a;100;120`). For each target the `mlxcel` CLI and the
 `mlxcel-server` are published as separate archives (`mlxcel-...` and
 `mlxcel-server-...`, each roughly 347 MB) so a consumer downloads only the one
-it needs. Treat other GPU/OS combinations as source builds that need local
-validation.
+it needs. Every published release also ships a CycloneDX SBOM named
+`sbom-<version>.cyclonedx.json.gz` for supply-chain transparency and
+vulnerability scanning. Treat other GPU/OS combinations as source builds that
+need local validation.
 
 ### Prebuilt CUDA artifact: runtime requirements
 


inureyes:

### Intent
> Add a `generate-sbom` job to `.github/workflows/release.yml` that publishes a CycloneDX JSON SBOM (`sbom-<version>.cyclonedx.json.gz`) as a release asset on every real release, scoped to `lablup/mlxcel`.

### Findings Addressed
- None. No blocking (CRITICAL/HIGH) defects found; no changes made.

### Remaining Items (advisory, non-blocking)
- [ ] (MEDIUM) Targeted `workflow_dispatch` re-uploads (`release_tag` set, `targets` != `all`) skip `generate-sbom`, because a skipped `needs` dependency plus a status-function-free `if` yields an effective `success(all needs) && <if>` gate. This is the exact limitation the PR body already documents and accepts. It is harmless in practice: a same-tag targeted re-upload does not change `Cargo.lock`, so the existing SBOM stays accurate, and real `release: published` events always build all three platforms. Not changed, to avoid churning an intentional, documented design decision. Running `targets: all` regenerates everything including the SBOM if ever needed.

### Verification
- [x] `if:` gating restricts to real release uploads and the canonical repo: `github.repository == 'lablup/mlxcel' && (release || workflow_dispatch with non-empty release_tag)`, matching the per-platform "Upload release artifacts" step guard exactly, plus the repo scope carried by the build jobs. Forks and build-only verification runs are skipped.
- [x] Tag/version derivation correct: `TAG = release.tag_name || inputs.release_tag`, `VERSION = ${TAG#v}`. TAG is guaranteed non-empty whenever the upload step runs (the job `if` requires a tag source); the checkout `ref`'s extra `|| github.sha` fallback is unreachable for the upload path.
- [x] `needs: [build-macos, build-linux-cuda, build-linux-x86_64-cuda]` with plain `needs` matches `notify-teams`; a failed or skipped build correctly suppresses the SBOM so a broken release never ships one. Skipped-dependency behavior on partial dispatch matches the PR body claim.
- [x] `gh release upload "$TAG" "$SBOM_FILE" --clobber --repo "${{ github.repository }}"` invoked with `GH_TOKEN` from `secrets.GITHUB_TOKEN`; `--clobber` is idempotent for re-runs; explicit `--repo` is correct under `persist-credentials: false`. Release exists by ordering (build jobs create-if-missing via `softprops/action-gh-release@v3` before this job runs).
- [x] Permissions minimal (`contents: write` only) under the file's default-deny top-level `permissions: {}`.
- [x] YAML parses cleanly (`yaml.safe_load` ok); indentation and the block-scalar `if:` pattern are consistent with the existing sibling jobs.
- [x] Reuses established workflow conventions: `runs-on: ubuntu-latest` (as `notify-teams`/`promote-release`), `actions/checkout@v7` with `persist-credentials: false`, and pinned tooling (Syft `v1.46.0`).
- [x] No unintended structural changes: additive only (+55 lines, one new job); no existing jobs, guards, or steps modified.

inureyes:
## PR Finalization Complete

### Tests
Skipped intentionally. This is a YAML-only CI addition (a new `generate-sbom` job in `.github/workflows/release.yml`); there is nothing unit-testable in a GitHub Actions job definition, and no Rust code changed. The implementation was already validated locally per the PR description (syft v1.46.0 -> CycloneDX 1.7, 439 pkg:cargo/ components from Cargo.lock).

### Documentation
Added one sentence to `docs/installation.md`, in the Linux CUDA release-archive paragraph that already describes the published `mlxcel`/`mlxcel-server` platform archives:

> Every published release also ships a CycloneDX SBOM named `sbom-<version>.cyclonedx.json.gz` for supply-chain transparency and vulnerability scanning.

English only; docs are single-sourced under `docs/*.md` with no separate Korean file to update.

### Lint / Format
- `.github/workflows/release.yml`: unchanged in this pass; the workflow's own YAML parse check (`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"`) still passes.
- No `cargo build`/`test`/`clippy` run, since no Rust files are touched by this PR or by this finalization pass.

Issue #766, opened 2026-07-13
feat(server): Kimi-K2 tool-call parser (tool_calls_section markers)


inureyes:
## Summary
Kimi models (mlxcel family kimi_linear; also the Kimi VL families if they share the template) emit tool calls in a sectioned marker format not handled by `src/server/tool_calls/formats.rs`.

## Format specification
- Section: `<|tool_calls_section_begin|>` ... `<|tool_calls_section_end|>`.
- Each call: `<|tool_call_begin|>` ... `<|tool_call_end|>`; if no per-call wrappers are present, parse the whole section as one call.
- Inside a call: `functions.NAME:INDEX<|tool_call_argument_begin|>{json arguments}`. The call id is the full `functions.NAME:INDEX` string; the function name is the middle segment. Name regex: `^\s*((?:functions\.)?(.+?):\d+)\s*<\|tool_call_argument_begin\|>` (group 1 id, group 2 name).
- Arguments: everything after `<|tool_call_argument_begin|>`, trimmed; try JSON parse, then loose literal parse, then raw string.
- Note: `ParsedToolCall` currently carries only name and arguments; either regenerate the id via `generate_tool_call_id` or extend the struct to preserve the model-provided id. Pick one and test it.

## Implementation plan
1. `try_kimi_k2` in `formats.rs`, `ToolCallFormat::KimiK2` in `types.rs`, registry entry in `parser.rs` (the marker is specific, so ordering is uncontentious; group it with the other marker-specific parsers).
2. Streaming: add `<|tool_calls_section_begin|>` / `<|tool_calls_section_end|>` enter/exit rows to `CHAT_DELIMITERS` in `stream_filter.rs`.
3. Unit tests: single call, multiple calls, missing per-call wrappers, malformed JSON args fallback, streaming split across the long markers.

## Acceptance criteria
- Unit tests pass; end-to-end tools request against a Kimi checkpoint returns structured `tool_calls` in streaming and non-streaming modes.

Effort: low-medium.

PR #785, opened 2026-07-13
fix(server): parse the bracketed Mistral tool-call format

### LINKED ISSUE #767: fix(server): parse the bracketed Mistral tool-call format (NAME[ARGS]{json})
## Summary
`try_mistral_nemo` (`src/server/tool_calls/formats.rs:152`) only understands the older `[TOOL_CALLS] [{"name": ..., "arguments": ...}]` JSON-array format. Newer Mistral-family models (Ministral 2410 and later, Mistral Small 3, Magistral, Devstral; in mlxcel terms the ministral3, mistral4, and pixtral-backbone families) emit `[TOOL_CALLS]NAME[ARGS]{json}` instead. Today that output fails JSON parsing after the `[TOOL_CALLS]` strip and the response is returned as plain text with no `tool_calls`, so tool calling is silently broken for these models.

## Format specification
- Trigger: text contains `[TOOL_CALLS]`.
- After stripping the marker, match `^\s*(\w+)\[ARGS\]\s*(\{.*\})` with dot matching newline; group 1 is the function name, group 2 is the JSON arguments object. Single call per message.
- Disambiguation from the older format: presence of the literal `[ARGS]`. Both variants must keep working.

## Implementation plan
1. Either extend `try_mistral_nemo` with an `[ARGS]` branch or add `try_mistral_bracket` plus `ToolCallFormat::Mistral`; select the variant by checking for the `[ARGS]` literal first, then falling back to the JSON-array parse. Preserve the existing MistralNemo tests unchanged.
2. Streaming: confirm `[TOOL_CALLS]` handling in `CHAT_DELIMITERS` (`src/server/tool_calls/stream_filter.rs`) covers the new shape; add rows if missing.
3. Unit tests: new format single call, arguments JSON containing nested braces and newlines, old-format regression, content preceding the marker.

## Acceptance criteria
- Old-format tests keep passing; new-format tests pass.
- End-to-end tools request against a current Mistral-family checkpoint produces structured `tool_calls`.

Effort: low.

inureyes:
## Summary

Adds support for the bracketed Mistral tool-call format (`[TOOL_CALLS]NAME[ARGS]{json}`) used by Ministral 2410+, Mistral Small 3, Magistral, and Devstral, which previously fell back to plain text with no `tool_calls` because `try_mistral_nemo` only understood the older `[TOOL_CALLS] [{"name": ..., "arguments": ...}]` JSON-array shape.

## What changed

- `src/server/tool_calls/types.rs`: added `ToolCallFormat::Mistral` (`#[non_exhaustive]` enum) documenting the bracketed format and its model coverage.
- `src/server/tool_calls/formats.rs`: added `pub fn try_mistral_bracket`, which strips the `[TOOL_CALLS]` prefix, matches `^\s*(\w+)\[ARGS\]\s*(\{.*\})` with a dotall `fancy_regex`, and coerces the captured arguments to a JSON string (unwrap a JSON-string value, re-serialize an object, or keep the raw text on parse failure), matching the convention used by the other format parsers in this module. Returns `None` when `[ARGS]` is absent so the older JSON-array format is left untouched.
- `src/server/tool_calls/parser.rs`: registered `try_mistral_bracket` in the dispatch list immediately before `try_mistral_nemo`, with a comment explaining the `[ARGS]` disambiguation and fall-through.
- `src/server/tool_calls/stream_filter.rs`: added a streaming test confirming the bracketed shape is already suppressed by the existing one-shot `[TOOL_CALLS]` `EnterToolCall` delimiter (no new delimiter rows were needed).

`try_mistral_nemo` and its existing tests are untouched.

## Test plan

- [x] `cargo test --lib server::tool_calls::formats` (new: single-call bracket parse, nested-brace/newline arguments, old-format regression returning `None`, leading-content handling)
- [x] `cargo test --lib server::tool_calls::parser` (new: end-to-end dispatch for the bracketed format with `ToolCallFormat::Mistral`, unknown-tool filtering, and a regression confirming the old array format still resolves to `ToolCallFormat::MistralNemo`)
- [x] `cargo test --lib server::tool_calls::stream_filter` (new: bracketed payload suppressed from `delta.content`)

Closes #767

### DIFF
--- a/src/server/tool_calls/formats.rs
+++ b/src/server/tool_calls/formats.rs
@@ -1644,6 +1644,49 @@ mod tests {
         assert!(try_mistral_bracket(text).is_none());
     }
 
+    #[test]
+    fn mistral_bracket_no_args_marker_no_quadratic_blowup() {
+        // Adversarial body: `[TOOL_CALLS]` followed by a long run of word
+        // characters with no `[ARGS]` marker anywhere. The `(\w+)` capture in
+        // `^\s*(\w+)\[ARGS\]\s*(\{.*\})` backtracks one character at a time
+        // looking for the literal `[ARGS]` that never appears, so this must
+        // stay linear rather than hang.
+        let mut text = String::from("[TOOL_CALLS]");
+        text.push_str(&"a".repeat(50_000));
+
+        let start = std::time::Instant::now();
+        let result = try_mistral_bracket(&text);
+        let elapsed = start.elapsed();
+
+        // Should complete in well under one second on any reasonable hardware.
+        assert!(
+            elapsed.as_secs() < 2,
+            "parsing took {elapsed:?}; suggests catastrophic backtracking"
+        );
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn mistral_bracket_unclosed_args_no_quadratic_blowup() {
+        // Adversarial body: `[TOOL_CALLS]fn[ARGS]{` followed by a long run of
+        // word characters with no closing brace. The dotall `(\{.*\})` group
+        // cannot match without a closing `}`, so the regex must fail fast
+        // rather than degrade while `.*` backtracks across the whole buffer.
+        let mut text = String::from("[TOOL_CALLS]fn[ARGS]{");
+        text.push_str(&"a".repeat(50_000));
+
+        let start = std::time::Instant::now();
+        let result = try_mistral_bracket(&text);
+        let elapsed = start.elapsed();
+
+        // Should complete in well under one second on any reasonable hardware.
+        assert!(
+            elapsed.as_secs() < 2,
+            "parsing took {elapsed:?}; suggests catastrophic backtracking"
+        );
+        assert!(result.is_none());
+    }
+
     // -- Functionary v3.1 --
 
     #[test]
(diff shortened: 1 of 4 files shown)


inureyes:
## PR Finalization Complete

### Summary
- **Tests**: added two ReDoS regression tests for `try_mistral_bracket` in `src/server/tool_calls/formats.rs`, mirroring the existing `minimax_m2_malformed_parameter_no_quadratic_blowup` precedent: `mistral_bracket_no_args_marker_no_quadratic_blowup` (50k word characters after `[TOOL_CALLS]` with no `[ARGS]` marker) and `mistral_bracket_unclosed_args_no_quadratic_blowup` (`[TOOL_CALLS]fn[ARGS]{` followed by 50k characters with no closing brace). Both assert a prompt `None` return (elapsed < 2s) rather than a hang. `cargo test --lib server::tool_calls::formats` now passes 86/86, `parser` 60/60, `stream_filter` 60/60.
- **Documentation**: none updated. Re-confirmed (as with #766) that this repo has no doc listing supported tool-call formats or Mistral model families for tool-calling; `docs/` (architecture.md, supported-models.md, adding-models.md, code-guidelines.md, responses-api.md, etc.) has no such section, and `docs/en/`/`docs/ko/` referenced by the mkdocs configs do not exist yet. No doc was invented.
- **Lint/Format**: `cargo fmt` on the changed file produced no additional diff, `cargo clippy --lib --tests -- -D warnings` is clean.

PR #783, opened 2026-07-13
feat(server): add Kimi K2 sectioned tool-call parser

inureyes:
## Summary

Kimi models emit tool calls wrapped in `<|tool_calls_section_begin|>...<|tool_calls_section_end|>` section markers, a format none of the existing `tool_calls::formats` parsers recognized, so those calls fell through to plain content instead of structured `tool_calls`.

## What changed

- `src/server/tool_calls/formats.rs`: added `try_kimi_k2`, which locates the section, splits on the optional per-call `<|tool_call_begin|>`/`<|tool_call_end|>` wrappers (falling back to treating the whole section body as one call when the wrappers are absent), and extracts each call's function name from the `functions.NAME:INDEX<|tool_call_argument_begin|>` header via a regex compiled once per section. Uses `fancy-regex`, already a project dependency, so no new crate was added. Arguments are parsed as strict JSON first, then a loose literal reparse tolerating Python-dict-repr spellings (single quotes, `True`/`False`/`None`), then fall back to the raw trimmed text so a malformed call still produces a call instead of being dropped.
- `src/server/tool_calls/types.rs`: added the `KimiK2` variant to `ToolCallFormat` with a doc comment describing the marker format.
- `src/server/tool_calls/parser.rs`: registered `formats::try_kimi_k2` in the ordered parser dispatch list, ahead of the generic fallback parsers. `ParsedToolCall` still carries only `name` and `arguments` (unchanged struct); the model-provided id/index is discarded and a fresh id is generated downstream via `generate_tool_call_id`, matching every other parser.
- `src/server/tool_calls/stream_filter.rs`: added the section begin/end markers to `CHAT_DELIMITERS` (exit before enter, mirroring the existing Hermes convention) so streaming responses suppress the tool-call section from `delta.content` instead of leaking the raw markers.

## Test plan

- [x] `cargo test --lib server::tool_calls::formats` — 76 passed (includes single call with wrappers, multiple calls in one section, missing per-call wrappers, malformed JSON argument fallback, name extraction with/without the `functions.` prefix, missing section-end marker, content-before-section preservation)
- [x] `cargo test --lib server::tool_calls::parser` — 58 passed (end-to-end `parse_tool_calls` coverage: single call, multiple calls, unknown-tool filtering)
- [x] `cargo test --lib server::tool_calls::stream_filter` — 59 passed (single-fragment suppression, content-before-section, markers split across fragment boundaries, no confusion with Hermes/Gemma 4 tool-call delimiters)
[...]

PR #786, opened 2026-07-13
feat(server): add pythonic tool-call parser ([func(arg=value)] format)

### LINKED ISSUE #768: feat(server): pythonic tool-call parser ([func(arg=value)] format)
## Summary
Several chat templates (Llama-family pythonic tool use, some Nemotron templates) emit tool calls as a bracketed Python-like call, for example `[get_weather(city="Paris", days=2)]`, wrapped in `<|tool_call_start|>` ... `<|tool_call_end|>` markers. No parser in `src/server/tool_calls/formats.rs` handles this.

## Format specification
- Markers: `<|tool_call_start|>` ... `<|tool_call_end|>` (the corresponding tool-list template marker is `<|tool_list_start|>`).
- Call: regex `\[(\w+)\((.*?)\)\]` with dot matching newline, first match only (single call).
- Arguments: regex `(\w+)=(?:"([^"]*)"|([^,]+))(?:,\s*|$)`. Quoted values are strings; unquoted values parse as numbers, booleans, or bracketed lists, falling back to the raw string.
- Known limitation to preserve: nested parentheses or commas inside unquoted values are not supported; match the format's real usage rather than building a full expression parser.

## Implementation plan
1. `try_pythonic` in `formats.rs`, `ToolCallFormat::Pythonic` in `types.rs`, registry entry ordered after the marker-specific parsers and before the generic JSON parsers.
2. Streaming delimiters for the start/end markers in `stream_filter.rs`.
3. Unit tests: quoted and unquoted args, ints, floats, bools, lists, empty args, non-matching text, marker-wrapped and bare bodies.

## Acceptance criteria
- Unit tests pass; end-to-end tools request against a checkpoint whose template uses these markers yields structured `tool_calls`.

Effort: low.

inureyes:
## Summary

Adds a parser for the pythonic bracketed tool-call format (`[func(arg=value, ...)]` wrapped in `<|tool_call_start|>`/`<|tool_call_end|>` markers) used by Llama-family pythonic tool use and some Nemotron chat templates, and wires it into the non-streaming parser registry and the streaming delimiter table.

## What changed

- `src/server/tool_calls/types.rs`: added the `ToolCallFormat::Pythonic` variant.
- `src/server/tool_calls/formats.rs`: added `try_pythonic`, which matches `(?s)\[(\w+)\((.*?)\)\]` (first match only) and builds the JSON `arguments` object from `(\w+)=(?:"([^"]*)"|([^,]+))(?:,\s*|$)`, coercing unquoted values to integer, float, bool, a bracketed list, or a raw-string fallback. A comma-masking helper (`mask_bracketed_commas` / `unmask_bracketed_commas`) hides commas inside `[...]` list literals from the per-argument regex before it runs so multi-element lists such as `tags=[1, 2, 3]` survive intact, then restores them for list-item coercion. Nested parentheses or commas inside a non-bracketed unquoted value remain unsupported by design, matching the format's real usage rather than a full expression parser.
- `src/server/tool_calls/parser.rs`: registered `formats::try_pythonic` in the dispatch list right after `try_functionary_v32` and before `try_llama3`, ahead of the generic JSON parsers, since the pythonic regex requires `[word(` and cannot false-positive on JSON or any other registered format.
- `src/server/tool_calls/stream_filter.rs`: added `<|tool_call_start|>` to `CHAT_DELIMITERS` as an enter-only `EnterToolCall` delimiter. No matching `<|tool_call_end|>` exit row was added on purpose: that marker is Kimi K2's per-call terminator inside a `<|tool_calls_section_begin|>...<|tool_calls_section_end|>` section (added in #783), and registering it as a global exit would fire on every Kimi K2 per-call terminator and prematurely exit tool-call suppression mid-section, leaking the second and later Kimi calls into `delta.content` (a regression of #766). Pythonic calls are terminal in practice, so enter-only fully suppresses the payload, mirroring Mistral Nemo's one-shot `[TOOL_CALLS]` entry.

## Test plan

- [x] `cargo test --lib server::tool_calls::formats` (94 passed, including 9 new `pythonic_*` cases: quoted string, unquoted int/float/bool, bracketed list, mixed quoted+unquoted, empty args, non-matching text, marker-wrapped vs bare body, first-match-only)
- [x] `cargo test --lib server::tool_calls::parser` (62 passed, including `parse_pythonic_format` and `parse_pythonic_filters_unknown_tools`)
- [x] `cargo test --lib server::tool_calls::stream_filter` (64 passed, including `pythonic_tool_call_suppressed`, `pythonic_content_before_call_emitted`, `pythonic_start_marker_not_confused_with_gemma4_or_kimi_k2`, and a new `kimi_k2_multi_call_section_all_calls_suppressed` regression test proving `<|tool_call_end|>` still does not exit tool-call suppression mid-section)

Closes #768

### DIFF
--- a/src/server/tool_calls/formats.rs
+++ b/src/server/tool_calls/formats.rs
@@ -1517,6 +1517,145 @@ fn is_ident_byte(b: u8) -> bool {
     b.is_ascii_alphanumeric() || b == b'_'
 }
 
+/// Try parsing the pythonic bracketed-call tool-call format:
+/// `<|tool_call_start|>[func(arg=value, ...)]<|tool_call_end|>`
+///
+/// Used by Llama-family "pythonic" tool use chat templates and some
+/// Nemotron templates. The call regex `\[(\w+)\((.*?)\)\]` (dotall, so it
+/// spans embedded newlines) only requires `[word(...)]` somewhere in the
+/// text, so it naturally ignores the surrounding `<|tool_call_start|>` /
+/// `<|tool_call_end|>` markers: marker-wrapped and bare bodies share this
+/// one code path. Only the FIRST match is used — the format carries a
+/// single call per message in practice.
+///
+/// Known limitation, preserved on purpose: nested parentheses, or commas
+/// inside a plain (non-bracketed) unquoted value, are not supported —
+/// matching the format's real usage rather than building a full expression
+/// parser. A single level of `[...]` list literal in a value IS supported.
+pub fn try_pythonic(text: &str) -> Option<ToolCallParseResult> {
+    let call_re = Regex::new(r"(?s)\[(\w+)\((.*?)\)\]").ok()?;
+    let caps = call_re.captures(text).ok()??;
+
+    let name = caps.get(1)?.as_str().to_string();
+    let args_raw = caps.get(2).map(|m| m.as_str()).unwrap_or_default();
+
+    Some(ToolCallParseResult {
+        format: Some(ToolCallFormat::Pythonic),
+        tool_calls: vec![ParsedToolCall {
+            name,
+            arguments: pythonic_arguments(args_raw),
+        }],
+        content: String::new(),
+        reasoning_content: None,
+    })
+}
+
+/// Placeholder used to temporarily hide commas that sit inside a `[...]`
+/// list literal from the per-argument regex in [`pythonic_arguments`]
+/// (whose unquoted alternative `[^,]+` would otherwise stop at the first
+/// comma, breaking multi-element lists such as `tags=[1, 2, 3]`). Restored
+/// by [`unmask_bracketed_commas`] once the surrounding key=value span has
+/// been captured. A Unicode Private Use Area codepoint is used so it cannot
+/// collide with a comma that legitimately appears in real argument text.
+const MASKED_COMMA: char = '\u{E000}';
+
+/// Hide commas that occur inside an unquoted `[...]` list literal (bracket
+/// depth > 0, outside a quoted string) behind [`MASKED_COMMA`] so the
+/// per-argument regex's unquoted alternative can span the whole list.
+fn mask_bracketed_commas(raw: &str) -> String {
+    let mut out = String::with_capacity(raw.len());
+    let mut depth: i32 = 0;
+    let mut in_quotes = false;
+    for c in raw.chars() {
+        match c {
+            '"' => {
+                in_quotes = !in_quotes;
+                out.push(c);
+            }
+            '[' if !in_quotes => {
+                depth += 1;
+                out.push(c);
+            }
+            ']' if !in_quotes => {
+                depth = depth.saturating_sub(1);
+                out.push(c);
+            }
+            ',' if !in_quotes && depth > 0 => out.push(MASKED_COMMA),
+            _ => out.push(c),
+        }
+    }
+    out
+}
+
+/// Restore commas hidden by [`mask_bracketed_commas`].
+fn unmask_bracketed_commas(masked: &str) -> String {
+    masked.replace(MASKED_COMMA, ",")
+}
+
+/// Build the JSON `arguments` object from a pythonic call's raw argument
+/// text (the `(.*?)` capture between the call's parentheses), using the
+/// per-argument regex `(\w+)=(?:"([^"]*)"|([^,]+))(?:,\s*|$)`: group 1 is
+/// the key; a matched quoted alternative (group 2) becomes a JSON string
+/// verbatim; a matched unquoted alternative (group 3) is coerced via
+/// [`coerce_pythonic_value`]. Returns `"{}"` for empty args (e.g. `ping()`).
+fn pythonic_arguments(args_raw: &str) -> String {
+    let masked = mask_bracketed_commas(args_raw);
+    let Ok(args_re) = Regex::new(r#"(\w+)=(?:"([^"]*)"|([^,]+))(?:,\s*|$)"#) else {
+        return "{}".to_string();
+    };
+
+    let mut map = serde_json::Map::new();
+    for caps in args_re.captures_iter(&masked) {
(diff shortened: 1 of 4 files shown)


inureyes:

### Intent
> Add a pythonic tool-call parser for `[func(arg=value)]` calls wrapped in `<|tool_call_start|>` / `<|tool_call_end|>` markers (Llama pythonic / Nemotron), integrated into the parser registry and streaming filter.

### Findings Addressed
Initially none; a later pass then confirmed one HIGH regression (JSON tool call stolen by the pythonic parser), fixed in commit 60265a6. See the follow-up comment below.

### Remaining Items (informational, LOW, intentionally out of scope)
- [ ] (LOW) A `[...]` list literal whose item is a quoted string containing a comma, e.g. `tags=["a,b", "c"]`, is mis-parsed: the intra-string comma is correctly not masked, but the per-argument regex `[^,]+` stops at it. This is a narrow extension of the documented KNOWN LIMITATION and is beyond the format's real usage, so it is left as-is per the "match real usage, not a full expression parser" design.
- [ ] (LOW) Spaces around `=` (`key = value`) yield empty args. This is faithful to the spec regex `(\w+)=...` and matches the real template output, so it is intentional.

### Verification
- [x] All stated requirements implemented (call regex first-match-only, quoted/int/float/bool/list coercion, marker-wrapped and bare bodies, empty args, recursion + quoted-list-item quote stripping)
- [x] Integrated into project code flow (enum variant, registry entry after `try_functionary_v32` and before `try_llama3`/`try_generic_json`, streaming delimiter, filter-by-tools safety net)
- [x] Project conventions followed (`fancy_regex::Regex`, `content: String::new()` mirroring `try_mistral_nemo`, doc comments consistent with siblings)

### Cross-format safety (the crux)
- [x] Only `("<|tool_call_start|>", EnterToolCall)` is registered in `CHAT_DELIMITERS`; `<|tool_call_end|>` is NOT registered as a global `ExitToolCall` anywhere.
- [x] The explanatory comment on both the delimiter table and the entry is present.
- [x] `kimi_k2_multi_call_section_all_calls_suppressed` exercises a two-call Kimi K2 section split right after the first `<|tool_call_end|>` and asserts the second call does not leak into `delta.content`. This test genuinely fails if `<|tool_call_end|>` were registered as an exit, and it passes.

### Masking correctness
- [x] `mask_bracketed_commas` masks only commas inside a `[...]` list literal at depth > 0 outside quotes, using PUA `\u{E000}`; commas inside quoted strings are not masked, so top-level quoted values keep working.
- [x] `unmask_bracketed_commas` is applied to unquoted values before coercion; the quoted branch stores the value verbatim (safe, since quoted commas are never masked).
- [x] `coerce_pythonic_value` recurses for list items and strips quotes from quoted list items.

### Tests
- `cargo test --lib server::tool_calls::formats`: 94 passed (11 new pythonic tests)
- `cargo test --lib server::tool_calls::parser`: 62 passed (2 new pythonic tests)
- `cargo test --lib server::tool_calls::stream_filter`: 64 passed (4 new tests incl. the #766 regression guard)
- `cargo clippy --lib --tests -- -D warnings`: clean (only unrelated C++ build notes from `mlxcel-core`)

Update: a later review pass found one HIGH regression that has since been fixed on this branch (commit 60265a6). See the follow-up comment below; the "Findings Addressed" and "Remaining Items" sections above are superseded by it.

inureyes:
## Implementation Review Update (supersedes the initial summary)

A second review pass surfaced two findings after my first summary was posted. One is a real regression that I fixed on this branch; the other is intentional-convention and left as-is.

### Fixed (HIGH): commit `60265a6`
**Pythonic parser could steal a JSON tool call.** `try_pythonic` runs before `try_llama3` / `try_generic_json`, and its call regex `[(\w+)\((.*?)\)]` scans anywhere in the text. A valid JSON tool call whose string argument merely contains a `[word(...)]` substring was re-routed to the bracketed inner name. Concretely, `{"name": "search", "arguments": {"query": "[calc(x=1)]"}}` with both `search` and `calc` registered parsed as `calc`, dropping the real `search` call. Pre-PR this parsed correctly as `search`, so it is a genuine regression of existing JSON tool-call handling.

Fix: guard `try_pythonic` to decline when the whole trimmed body is valid JSON. A genuine pythonic call (bare or marker-wrapped) is never itself valid JSON, while a JSON tool call always is, so the JSON parsers that run next claim it. The prescribed registry placement (before the JSON parsers) is preserved. Added a parser-level regression test (`parse_pythonic_does_not_steal_json_tool_call`) that fails on the pre-fix code and passes after, plus a formats-level unit test covering object and array JSON shapes. All pre-existing pythonic tests still pass, confirming genuine pythonic parsing is unaffected.

### Not changed (LOW, intentional convention)
**Non-streaming path drops preamble text** (`content: String::new()`). This matches the plurality of parsers and, specifically, `try_mistral_nemo`, the one-shot terminal format this parser explicitly mirrors; `try_llama3`, `try_generic_json`, `try_mistral_bracket`, and `try_minimax_m2` do the same. The streaming path already preserves the preamble. This is a project-wide pattern for the JSON-family and terminal parsers, not a deviation introduced here, so it is left as-is. If preamble preservation is desired it should be applied consistently across that parser family rather than to pythonic alone.

### Tests after fix
- `cargo test --lib server::tool_calls::formats`: 95 passed (+1 guard test)
- `cargo test --lib server::tool_calls::parser`: 63 passed (+1 regression test)
- `cargo test --lib server::tool_calls::stream_filter`: 64 passed (incl. the #766 regression guard)
- `cargo clippy --lib --tests -- -D warnings`: clean (only unrelated C++ build notes from `mlxcel-core`)

The critical cross-format safety (only `<|tool_call_start|>` registered, `<|tool_call_end|>` never a global exit; the `kimi_k2_multi_call_section_all_calls_suppressed` guard) is unchanged and still passing.

PR #808, opened 2026-07-17
feat(bench): monotonic OOM backstop and persistent OOM record for decode benchmark sweeps

### LINKED ISSUE #807: feat(bench): monotonic OOM backstop and persistent OOM record for decode benchmark sweeps
## Summary

On the GB10 (DGX Spark, CUDA unified memory) machine, benchmark cells that OOM are worse than wasted time: near-OOM CUDA runs have repeatedly left the driver in a degraded state (NVRM `NV_ERR_NO_MEMORY` accumulating for hours, eventually wedging the kernel and requiring a reboot), so cells that are certain to OOM should never be launched at all. An earlier WIP gate paired a bash-formula skip with an analytic single-pass memory estimator (`MLXCEL_ESTIMATE_SINGLE_PASS_PREFILL` extending `src/execution/memory_estimate.rs`), but the estimator was measured to under-predict allocator-driven OOMs by more than 2.5x (estimated ~42 GiB vs actual >115 GiB for gemma-4-31b at 32768 prompt tokens, before PR #676), so it was judged unreliable as a gate and that WIP was discarded (the maintainer keeps the patch locally for reference). This issue asks only for the two empirical layers that remained sound, implemented fresh as shell-level bookkeeping in the bench harness.

Reference context: #648 and #672 (the July long-prompt failures), PR #676 (the chunked-prefill fix that removed the original motivating gemma-4 32k OOM).

## Design

Two layers, both purely empirical (they react to observed OOMs, they never predict):

1. **Monotonic backstop (within one sweep run).** When a (model, prompt-length) cell fails with a true OOM signature, skip every remaining cell for that same model whose prompt length is greater than or equal to the OOM'd length. Each skipped cell is logged as an explicit CSV row with a new trailing classification token (`SKIP:oom_backstop`), never silently.
2. **Persistent OOM record (across runs).** Append each true-OOM cell as a tuple to a gitignored record file at `benchmarks/.oom-record` (one line per event: `model,prompt_target_len,date,commit,peak_bytes,source_csv`, with `peak_bytes` blank when unknown since the harness does not currently capture peak memory, and `commit` from `git rev-parse --short HEAD` which is cheap). On later runs, pre-skip any cell whose model has a recorded OOM at a length less than or equal to the cell's length, logging the skip as `SKIP:oom_record`. Setting `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` disables the pre-skip (new OOMs are still appended) so cells can be re-tested after a memory fix lands.

**Staleness.** Records go stale whenever memory behavior improves: the gemma-4-31b 32768-token OOM that motivated this design was fixed by chunked prefill in PR #676, so a record of it taken before that merge must be prunable. The per-line date and commit exist precisely to make manual pruning easy (`grep -v` the model, or delete lines older than the fix), and the override env var covers one-off re-tests without editing the file.

**Non-goal.** No analytic memory estimator, in the harness or in Rust. The discarded estimator under-predicted allocator-driven OOMs by more than 2.5x; a gate that under-predicts is worse than no gate because it launches exactly the runs that wedge the driver. The existing weight-size check in `bench_decode.sh` (`model_fits_in_memory` / `SKIP:oom_estimate`) stays as-is; it is length-independent and out of scope here.

## Scope

First pass: **`scripts/bench_longprompt.sh` only.** It is the only script with a (model x prompt-length) grid (model-major, length-minor nested loop) and it is where the GB10 driver-wedge incidents happened. `scripts/bench_decode.sh` runs one prompt length per invocation, so the monotonic backstop has no meaning there; it needs no changes beyond an optional hardening item (below). Other bench scripts (`bench_kv_cache.sh`, `bench_m5_*.sh`, `bench_na_attention.sh`) are Metal/feature-specific harnesses without the length ladder; extending the persistent-record pre-skip to `bench_decode.sh --prompt-tokens N` invocations can be a follow-up if single-cell ad-hoc runs turn out to re-trigger known OOMs in practice.

## Current harness facts (from inspection)

- `scripts/bench_decode.sh` runs one cell per invocation via `bench_one()`, which classifies failures using `is_oom_failure(rc, err_text)`: exit 124 (`timeout(1)` expiry) is explicitly never OOM, exit 137 (SIGKILL, OS OOM-killer) is OOM on the exit code alone, and exit 134 (SIGABRT) or 1 (cxx-propagated exception) is OOM only when the captured combined stdout+stderr (`raw`) matches allocator text (`out of memory`, `failed to allocate`, `bad_alloc`, `metal::malloc`, etc.). True OOM emits a row ending in `SKIP:oom`; non-OOM failures emit `FAIL:bench`. This already distinguishes OOM from the exit-134 non-OOM aborts seen before (e.g. the #648 `grid.z > 65535` launch-config crash was correctly `FAIL:bench`).
- `scripts/bench_longprompt.sh` drives the grid: `for model_name in $MODELS; do for len in $LADDER; do ...` (default ladder `512 2048 8192 32768`), invoking `"$BENCH_DECODE" "$model_path" --prompt-tokens "$len" ... --output "$TMP_CELL"` per cell and appending `tail -n +2 "$TMP_CELL"` to the aggregate CSV. The cell's classification is therefore already recoverable in the sweep script from the trailing CSV field of the data row in `$TMP_CELL`.
- The trailing classification token is CSV field 16 (after `prompt_target_len`); successful rows have no trailing token, so the check must be "last field matches `^(SKIP|FAIL):`", not "field 16 is non-empty" (the quoted prompt field contains a comma, so parse from the end, e.g. `awk -F, '{print $NF}'`).
- `benchmarks/` CSVs are git-tracked (135 files), so `benchmarks/.oom-record` needs an explicit `.gitignore` entry.
- Benches read `./models`, a gitignored symlink to a local model store; if the symlink is missing, `bench_longprompt.sh` prints `[miss]` per model and produces zero rows. Infra failures (missing symlink, missing model dir, watchdog timeout exit 124) must never be recorded as OOM; only cells classified `SKIP:oom` by `is_oom_failure` feed the backstop and the record. `SKIP:oom_estimate` is also excluded: it is a length-independent weight-size decision that re-derives instantly on every run, so persisting it adds nothing.

## Implementation plan

All changes in `scripts/bench_longprompt.sh` unless noted.

1. Add a `cell_classification` helper that extracts the trailing token of the data row in `$TMP_CELL` (`tail -n +2 "$TMP_CELL" | head -n 1 | awk -F, '{print $NF}'`) and returns it only when it matches `^(SKIP|FAIL):`, empty otherwise.
2. Add an `emit_skip_row` helper that composes a schema-compatible 16-field row for a skipped cell (`model,model_path` filled, profiling fields empty, `date` and `max_tokens` from the sweep's own variables, `prompt_target_len` set to the skipped length, trailing token `SKIP:oom_backstop` or `SKIP:oom_record`) and appends it to `$OUTPUT`. The sweep script does not compute `hardware`/`mlx_version` today; either reuse the values from the most recent real cell row or leave those two fields blank, and document the choice in a comment (downstream consumers filter on the `SKIP:` prefix, so blanks are acceptable).
3. Monotonic backstop: `declare -A OOM_MIN_LEN` keyed by `model_name`. In the inner loop, before invoking `$BENCH_DECODE`, if `OOM_MIN_LEN[$model_name]` is set and `len >= OOM_MIN_LEN[$model_name]`, emit the `SKIP:oom_backstop` row, log to stderr, and `continue`. After each cell, if `cell_classification` returned `SKIP:oom`, update `OOM_MIN_LEN[$model_name]` to `min(current, len)` (use `>=` and min so an unsorted `--lengths` list still behaves).
4. Persistent record: define `OOM_RECORD_FILE="${BENCHMARKS_DIR}/.oom-record"`. Before invoking a cell, unless `MLXCEL_BENCH_IGNORE_OOM_RECORD=1`, scan the file for any line whose model equals `model_name` and whose recorded length is `<= len`; on match, emit the `SKIP:oom_record` row (stderr log includes the matching record line so the operator can see why). After a `SKIP:oom` cell, append `model_name,len,$(date '+%Y-%m-%d'),$(git rev-parse --short HEAD 2>/dev/null || echo unknown),,$OUTPUT` to the record file. Tolerate a missing or empty record file and malformed lines (skip them with a stderr warning).
5. Add `benchmarks/.oom-record` to `.gitignore` and document the file format, the pre-skip semantics, the `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` override, and the staleness/pruning guidance (PR #676 precedent) in the header comment of `bench_longprompt.sh` and in its `usage()` text.
6. Optional hardening in `scripts/bench_decode.sh` (small, same PR is fine): extend the `is_oom_failure` pattern list with CUDA-specific markers (`CUDA_ERROR_OUT_OF_MEMORY`, `cudaErrorMemoryAllocation`, `cuMemAlloc failed`) alongside the existing generic and Metal patterns, keeping the same "specific and grounded" rule stated in that function's comment.

## Tests / validation

No 100 GB run is needed to exercise the logic:

1. **Injectable failure hook.** Add `MLXCEL_BENCH_FAKE_OOM="model:len[,model:len...]"` to `bench_longprompt.sh`: when the current cell matches an entry, skip invoking the runner and synthesize a `$TMP_CELL` containing a `SKIP:oom` row. Then run the sweep against any tiny local model (e.g. qwen2.5-0.5b) with a short ladder and assert: the faked cell row says `SKIP:oom`, every longer cell for that model says `SKIP:oom_backstop`, other models are unaffected, and `benchmarks/.oom-record` gained exactly one line with the expected fields.
2. **Hand-written record file.** Write a record line for (tiny-model, 512) by hand, run the sweep, and assert all cells for that model at lengths >= 512 emit `SKIP:oom_record` without launching the runner; re-run with `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` and assert the cells execute normally.
3. **Non-OOM isolation.** Fake a `FAIL:bench` cell (or point one model at a missing directory) and assert neither the backstop nor the record file is touched.
4. **Real smoke (optional, cheap).** One real sweep on GB10 with the default ladder over a small model set to confirm the aggregate CSV schema is unchanged for successful rows.

## Acceptance criteria

- [ ] Within one `bench_longprompt.sh` run, a `SKIP:oom` cell causes all subsequent cells for the same model with `len >=` the OOM'd length to be skipped before launch, each logged as an explicit `SKIP:oom_backstop` CSV row and a stderr message.
- [ ] True OOM cells are appended to gitignored `benchmarks/.oom-record` with model, length, date, short commit, peak-if-known (blank allowed), and source CSV path.
- [ ] On later runs, recorded (model, length) cells and all longer lengths for that model are pre-skipped with explicit `SKIP:oom_record` rows.
- [ ] `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` bypasses the pre-skip and is documented in the script header and `usage()`.
- [ ] Only `SKIP:oom` classifications feed the backstop and the record; `FAIL:bench`, timeout (exit 124), `SKIP:oom_estimate`, missing `./models` symlink, and missing model dirs never do.
- [ ] Successful-row CSV schema is byte-compatible with the current 16-column output; skip rows use the same column layout with a `SKIP:` trailing token.
- [ ] The fake-OOM and hand-written-record validation scenarios above pass.
- [ ] Staleness pruning guidance (with the PR #676 precedent) is documented in the script header.

**Effort:** low. Shell-only changes, one script plus a `.gitignore` line and an optional pattern-list extension.

inureyes:
## Summary

On the GB10 (DGX Spark, CUDA unified memory) machine, benchmark cells that OOM are worse than wasted time: near-OOM CUDA runs have repeatedly left the driver in a degraded state (NVRM NV_ERR_NO_MEMORY accumulating for hours, eventually wedging the kernel and requiring a reboot), so cells that are certain to OOM should never be launched at all. This PR adds two purely empirical, reactive-only OOM-avoidance layers to `scripts/bench_longprompt.sh`, plus a small hardening item in `scripts/bench_decode.sh`, plus three review-driven hardening fixes to the OOM-avoidance layers themselves.

## What changed

- `scripts/bench_longprompt.sh`: monotonic within-sweep backstop, keyed per model, that skips every remaining cell at length `>=` an observed `SKIP:oom` length for the rest of the run, emitting an explicit `SKIP:oom_backstop` CSV row plus a stderr message.
- `scripts/bench_longprompt.sh`: persistent cross-run OOM record at gitignored `benchmarks/.oom-record` (`model,prompt_target_len,date,commit,peak_bytes,source_csv`, `peak_bytes` always blank today). On later runs, any cell whose model has a recorded OOM at a length `<=` the cell's length is pre-skipped and logged as `SKIP:oom_record`, with the matching record line echoed to stderr. `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` disables the pre-skip for one run (new `SKIP:oom` cells are still appended). Missing/empty record files and malformed lines are tolerated with a stderr warning, not a hard failure.
- `scripts/bench_longprompt.sh`: `MLXCEL_BENCH_FAKE_OOM="model:len[,model:len...]"` testing hook that skips invoking the runner for a listed cell and synthesizes a `SKIP:oom` row, so the backstop and record logic can be exercised deterministically without a real OOM.
- `scripts/bench_longprompt.sh`: only the exact `SKIP:oom` classification feeds either layer; `FAIL:bench`, timeout (exit 124), `SKIP:oom_estimate`, a missing `./models` symlink, and a missing model directory never do, since the gate is a strict string-equality check on the trailing classification token.
- `scripts/bench_longprompt.sh`: the record-file format, pre-skip semantics, the override env var, and staleness/pruning guidance (the PR #676 chunked-prefill fix is used as the concrete precedent: a record taken before that merge is stale and should be pruned with `grep -v`) are documented in both the script's header comment block and its `usage()` text.
- `scripts/bench_decode.sh`: extended the `is_oom_failure` pattern list with three CUDA-specific allocator markers (`CUDA_ERROR_OUT_OF_MEMORY`, `cudaErrorMemoryAllocation`, `cuMemAlloc failed`), kept alongside the existing generic and Metal patterns under the same "specific and grounded" rule.
- `.gitignore`: added `/benchmarks/.oom-record` since it is a locally accumulated, host-specific record and not sweep output meant to be committed.

### Review-driven hardening (3 follow-up fixes)

- `scripts/bench_longprompt.sh`: bash-version guard. The `OOM_MIN_LEN` / `FAKE_OOM_SET` associative arrays (`declare -A`) are a bash 4+ feature, but the script explicitly supports Darwin/Metal and stock macOS ships bash 3.2 as `/bin/bash`, where `declare -A` aborts the whole script at startup (including `--help`). The script now re-execs into a bash `>= 4` found on `PATH` (Homebrew `/opt/homebrew/bin` or `/usr/local/bin`) if one exists, and otherwise exits with an actionable "install a newer bash" message instead of the cryptic `declare: -A: invalid option` failure; the guard cannot loop since it only re-execs into a version that passes the check.
- `scripts/bench_longprompt.sh`: `load_oom_record` field-count validation. It previously validated only the first two fields (non-empty model, numeric length), so a truncated or hand-edited entry like `model,32768` was accepted and pre-skipped indefinitely, contradicting the header comment's promise that wrong-field-count lines are skipped with a warning. It now validates the full field count (`>= 6`, the documented `model,prompt_target_len,date,commit,peak_bytes,source_csv` format) so truncated records are rejected while a `source_csv` path that itself contains a comma stays accepted.
- `scripts/bench_longprompt.sh`: `--lengths` ladder validation. Both the monotonic backstop and the persistent-record pre-skip gate on `[[ "$len" -ge N ]]`, an arithmetic comparison; a non-numeric `--lengths` entry (a typo such as `2o48` or `32k`) made that comparison error out and silently evaluate false, so the cell the backstop was supposed to skip launched anyway and the OOM was never recorded, i.e. the safety net could fail open on a config typo. Every ladder length is now rejected up front, before the output CSV is truncated, if it is not a plain non-negative integer, naming the offending token.

Successful-row CSV output is unchanged; the trailing classification token remains an unnamed field after `prompt_target_len`, consistent with the existing `SKIP:oom_estimate` / `SKIP:oom` / `FAIL:bench` convention in `bench_decode.sh`.

## Test plan

- [x] `bash -n scripts/bench_longprompt.sh` and `bash -n scripts/bench_decode.sh` (syntax check)
- [x] `shellcheck scripts/bench_longprompt.sh scripts/bench_decode.sh` (no new warnings; one pre-existing unrelated SC2002 note in `bench_decode.sh`)
- [x] Fake-OOM sweep: `MLXCEL_BENCH_FAKE_OOM="qwen2.5-0.5b-4bit:512"` with `--lengths "512 2048 8192"` produced `SKIP:oom` at 512 and `SKIP:oom_backstop` at 2048/8192, with no real launch for the backstopped cells; `benchmarks/.oom-record` gained exactly one correctly-formatted line
- [x] Cross-model isolation: a real launch of a second model (`llama-3.2-1b-4bit`) in a separate sweep produced a normal successful row and left the first model's `.oom-record` entry untouched, confirming the backstop and record are keyed per model
- [x] Hand-written record: a manually written `(qwen2.5-0.5b-4bit, 512)` record line pre-skipped all three ladder lengths as `SKIP:oom_record` with no launch and left the record file unchanged; re-running with `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` executed the cell for real instead of pre-skipping it
- [x] Non-OOM isolation: pointing `--models` at a nonexistent model directory produced the existing `[miss]` skip path with zero CSV rows and no `benchmarks/.oom-record` file created
- [x] Malformed-record rejection: a truncated two-field line (`model,32768`) appended to `.oom-record` is skipped with an explicit stderr warning and does not count toward loaded entries, while the valid six-field line alongside it still loads and pre-skips
- [x] Ladder validation: `--lengths "512 2o48"` exits 1 with an explicit "must be space-separated non-negative integers" message before any cell launches or the output CSV is truncated, while a valid ladder still runs and the backstop still fires
- [x] Working tree verified clean of test artifacts afterward; the ~20 pre-existing untracked `benchmarks/*.csv` files were left untouched throughout, and the test-only `benchmarks/.oom-record` file was removed

Closes #807

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,8 @@ __pycache__/
 webpage/site/.next/
 webpage/site/out/
 webpage/site/node_modules/
+
+# Persistent cross-run OOM record for scripts/bench_longprompt.sh (issue #807).
+# Host-specific and reactive to local memory conditions, so it never belongs
+# in version control alongside the tracked benchmarks/*.csv result files.
+/benchmarks/.oom-record
--- a/scripts/bench_decode.sh
+++ b/scripts/bench_decode.sh
@@ -248,6 +248,8 @@ model_fits_in_memory() {
 #                 failed to allocate, cannot allocate memory, unable to allocate,
 #                 bad_alloc, "memory allocation of N bytes failed" (Rust abort)
 #   MLX Metal:    "greater than the maximum allowed buffer size", metal::malloc
+#   MLX CUDA:     CUDA_ERROR_OUT_OF_MEMORY, cudaErrorMemoryAllocation,
+#                 "cuMemAlloc failed" (issue #807, GB10 CUDA unified memory)
 #
 # Bare "oom" is intentionally excluded to avoid false positives (e.g. "zoom").
 is_oom_failure() {
@@ -264,7 +266,7 @@ is_oom_failure() {
   # Otherwise (exit 1 cxx exception, 134 bad_alloc abort, ...) it is OOM only
   # if the captured output names an allocator failure.
   printf '%s\n' "$err_text" | grep -qiE \
-    'out of memory|out-of-memory|insufficient memory|failed to allocate|cannot allocate memory|unable to allocate|bad_alloc|memory allocation of [0-9]|greater than the maximum allowed buffer size|metal::malloc'
+    'out of memory|out-of-memory|insufficient memory|failed to allocate|cannot allocate memory|unable to allocate|bad_alloc|memory allocation of [0-9]|greater than the maximum allowed buffer size|metal::malloc|CUDA_ERROR_OUT_OF_MEMORY|cudaErrorMemoryAllocation|cuMemAlloc failed'
 }
 
 # ---------------------------------------------------------------------------
--- a/scripts/bench_longprompt.sh
+++ b/scripts/bench_longprompt.sh
@@ -20,6 +20,57 @@
 # A model+length cell that OOMs (common at 32768 for large MoE models) is
 # recorded with the usual SKIP:oom / SKIP:oom_estimate classification and the
 # sweep continues.
+#
+# --- OOM avoidance (issue #807) ---------------------------------------------
+#
+# Near-OOM CUDA runs have repeatedly left the GB10 driver in a degraded state
+# (NVRM NV_ERR_NO_MEMORY accumulating for hours, eventually wedging the
+# kernel), so cells that are certain to OOM should never be launched at all.
+# Both layers below are purely empirical: they react to an observed SKIP:oom
+# cell, they never predict from a memory model. Only the SKIP:oom
+# classification (a true OOM at load/run time, see is_oom_failure() in
+# bench_decode.sh) feeds either layer; FAIL:bench, timeout (exit 124),
+# SKIP:oom_estimate, and cells for a missing ./models symlink or missing model
+# directory never do.
+#
+# 1. Monotonic backstop (within one sweep run). Once a (model, len) cell
+#    classifies as SKIP:oom, every remaining cell for that same model with
+#    length >= the OOM'd length is skipped before launch and logged as an
+#    explicit SKIP:oom_backstop CSV row (same 16-field layout as other rows)
+#    plus a stderr message.
+#
+# 2. Persistent OOM record (across runs). Each true SKIP:oom cell is appended
+#    to a gitignored record file at benchmarks/.oom-record, one line per
+#    event:
+#      model,prompt_target_len,date,commit,peak_bytes,source_csv
+#    `peak_bytes` is always blank today (the harness does not capture peak
+#    memory); `commit` is `git rev-parse --short HEAD` (or "unknown" outside a
+#    git checkout). On later runs, any cell whose model has a recorded OOM at
+#    a length <= the cell's length is pre-skipped and logged as a
+#    SKIP:oom_record row (the stderr message includes the matching record
+#    line). Set MLXCEL_BENCH_IGNORE_OOM_RECORD=1 to disable the pre-skip for
+#    one run; new SKIP:oom cells are still appended even when the pre-skip is
+#    disabled. A missing or empty record file is treated as "no records"; a
+#    malformed line (wrong field count, non-numeric length) is skipped with a
+#    stderr warning rather than aborting the sweep.
+#
+#    Staleness / pruning: records go stale whenever memory behavior improves.
+#    For example, the gemma-4-31b 32768-token OOM that originally motivated
+#    this design was fixed by the chunked-prefill change in PR #676, so a
+#    record line dated before that merge is no longer valid evidence. The
+#    per-line date and commit exist precisely so stale entries can be found
+#    and removed by hand, e.g.:
+#      grep -v '^gemma-4-31b-it-4bit,' benchmarks/.oom-record > /tmp/pruned \
+#        && mv /tmp/pruned benchmarks/.oom-record
+#    or by deleting lines whose date/commit predate a known fix. There is no
+#    automatic expiry; MLXCEL_BENCH_IGNORE_OOM_RECORD=1 is the low-effort way
+#    to re-test a single run without editing the file.
+#
+# 3. MLXCEL_BENCH_FAKE_OOM="model:len[,model:len...]" (testing hook). When the
+#    current cell's (model, len) pair is listed, the sweep skips invoking
+#    bench_decode.sh entirely and synthesizes a cell result containing a
+#    SKIP:oom row, so the backstop and persistent-record logic can be
+#    exercised deterministically without a real OOM.
 
 set -euo pipefail
 
@@ -30,6 +81,12 @@ BENCH_DECODE="${SCRIPT_DIR}/bench_decode.sh"
 MODELS_DIR="./models"
 BENCHMARKS_DIR="./benchmarks"
 DATE=$(date '+%Y-%m-%d')
+
+# Same 15-column header bench_decode.sh writes (its CSV_HEADER constant), so
+# aggregate rows -- whether they come from a real cell, a synthesized fake-OOM
+# cell, or a backstop/record skip row emitted directly by this script -- share
+# one schema. Trailing SKIP:*/FAIL:* tokens are an unnamed 16th field.
+CSV_HEADER="model,model_path,prompt_tokens,generated_tokens,prefill_ms,prefill_tok_s,decode_ms,decode_tok_s,date,hardware,mlx_version,build_type,max_tokens,prompt,prompt_target_len"
 
 # Representative subset: mix of dense (llama, qwen2.5), MoE (qwen3-a3b,
 # mixtral), and a large multimodal-capable text model (gemma-4).
@@ -44,6 +101,17 @@ WARMUP_TOKENS=4
 COOLDOWN=0
 BIG_COOLDOWN=0
 EXTRA_ARGS=()
+
+# OOM avoidance state (issue #807). See the header comment above for the
+# design; IGNORE_OOM_RECORD and FAKE_OOM_SET are read from the environment
+# only (no CLI flag) so a stray flag typo can never silently disable them.
+OOM_RECORD_FILE="${BENCHMARKS_DIR}/.oom-record"
+IGNORE_OOM_RECORD="${MLXCEL_BENCH_IGNORE_OOM_RECORD:-0}"
+declare -A OOM_MIN_LEN=()       # model_name -> smallest length observed SKIP:oom at
+declare -A FAKE_OOM_SET=()      # "model:len" -> 1, from MLXCEL_BENCH_FAKE_OOM
+OOM_RECORD_MODEL=()             # parallel arrays loaded from OOM_RECORD_FILE
+OOM_RECORD_LEN=()
+OOM_RECORD_LINE=()
 
 usage() {
   cat <<'EOF'
@@ -65,6 +133,37 @@ Options:
 
 All cells share one aggregate CSV with the 15-column bench_decode.sh schema
 (prompt_target_len appended). Cell order is model-major, length-minor.
+
+OOM avoidance (issue #807):
+  A cell that classifies as SKIP:oom (a true OOM at load/run time; see
+  is_oom_failure() in bench_decode.sh) triggers two purely empirical,
+  reactive-only skip layers -- neither one predicts from a memory model:
+
+  1. Monotonic backstop: every remaining cell for that model at length >= the
+     OOM'd length is skipped for the rest of this run, logged as
+     SKIP:oom_backstop.
+  2. Persistent record: the OOM is appended to the gitignored
+     benchmarks/.oom-record file (model,prompt_target_len,date,commit,
+     peak_bytes,source_csv). On later runs, any cell whose model has a
+     recorded OOM at a length <= the cell's length is pre-skipped and logged
+     as SKIP:oom_record.
+
+  Only SKIP:oom feeds either layer. FAIL:bench, a timeout (exit 124),
+  SKIP:oom_estimate, a missing ./models symlink, and a missing model
+  directory never do.
+
+Environment variables:
+  MLXCEL_BENCH_IGNORE_OOM_RECORD=1
+      Disable the persistent-record pre-skip for this run (new SKIP:oom cells
+      are still appended to benchmarks/.oom-record). Use this to re-test
+      cells after a memory fix lands; see the staleness/pruning guidance in
+      this script's header comment for removing the corresponding record
+      lines once the fix is confirmed.
+  MLXCEL_BENCH_FAKE_OOM="model:len[,model:len...]"
+      Testing hook: for each listed (model, len) pair, skip invoking
+      bench_decode.sh and synthesize a cell result containing a SKIP:oom row,
+      so the backstop and persistent-record logic can be validated without a
+      real OOM.
 EOF
 }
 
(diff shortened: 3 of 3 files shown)


inureyes:

### Intent
> Add two purely empirical, reactive-only OOM-avoidance layers to `scripts/bench_longprompt.sh` (within-sweep monotonic backstop emitting `SKIP:oom_backstop`; persistent gitignored `benchmarks/.oom-record` pre-skip emitting `SKIP:oom_record`, with `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` override and a `MLXCEL_BENCH_FAKE_OOM` test hook), a `.gitignore` entry, and CUDA OOM patterns in `bench_decode.sh` `is_oom_failure`.

### Findings Addressed
None. No CRITICAL/HIGH/MEDIUM findings; no code changes were required.

### Verification (all empirically exercised on this branch, shell-only, no cargo)
- [x] All stated requirements implemented (all 8 acceptance criteria met)
- [x] Backstop: a `SKIP:oom` cell skips every same-model cell at `len >=` the OOM length as `SKIP:oom_backstop` before launch, with a stderr message (verified via `MLXCEL_BENCH_FAKE_OOM`)
- [x] Record: true OOM appended to gitignored `benchmarks/.oom-record` as `model,len,date,short-commit,,(blank peak),source_csv`; `git check-ignore` confirms the ignore rule
- [x] Later runs pre-skip recorded model at `len >=` recorded length as `SKIP:oom_record` (matching line echoed to stderr)
- [x] `MLXCEL_BENCH_IGNORE_OOM_RECORD=1` bypasses the pre-skip yet still appends new `SKIP:oom` cells; documented in header + `usage()`
- [x] Only exact `SKIP:oom` feeds either layer (strict `==` equality); `FAIL:bench`, exit-124 timeout, `SKIP:oom_estimate`, missing `./models` symlink, and missing model dir (`[miss]` path precedes the gate) never do
- [x] Successful-row CSV schema unchanged: `CSV_HEADER` is byte-identical to `bench_decode.sh`; skip rows use the same 16-field layout with a trailing classification token
- [x] `cell_classification` parses the last comma-separated field, so a quoted prompt containing commas (e.g. "Hello, how are you today?") is classified correctly; short-prompt successful rows (empty trailing `prompt_target_len`) and long-prompt rows (numeric trailing field) are correctly treated as non-classified
- [x] Bash correctness under `set -euo pipefail`: associative-array access guarded with `:-` defaults, `check_oom_record` return-1 consumed inside an `if` condition, empty-array expansion safe on bash 5.2.21; malformed record lines are warned and skipped (not fatal); `bash -n` and `shellcheck` clean (only a pre-existing unrelated SC2002 in `bench_decode.sh`)
- [x] Documentation complete: header comment and `usage()` cover the record format, pre-skip semantics, override var, and staleness/pruning guidance (PR #676 precedent)
- [x] Integrated into the actual model-major/length-minor sweep loop, not orphaned helpers
- [x] No unintended structural changes; existing runner invocation preserved

### Remaining Items
None.

inureyes:
## Follow-up: supplementary review findings

A supplementary automated review pass surfaced three additional items. Disposition below; two are fixed in `4bfd870`, one is reported for maintainer judgment.

### Fixed (pushed in 4bfd870)
- [x] **HIGH: bash 4 requirement broke stock macOS.** `declare -A OOM_MIN_LEN` / `FAKE_OOM_SET` are bash 4+ only, but the script explicitly supports Darwin/Metal (it detects M1 Ultra and M5 Max) and stock macOS ships bash 3.2 as `/bin/bash`, where `declare -A` aborts the whole script at startup, including `--help`. This is the first bash-4 dependency in `scripts/` (no other bench script uses `declare -A`). Added an early, bash-3-compatible guard that re-execs under a newer bash if one is on PATH (Homebrew `/opt/homebrew/bin` or `/usr/local/bin`) and otherwise exits with an actionable message instead of the cryptic `declare: -A: invalid option`. The guard only re-execs into a bash whose major version is >= 4, so it cannot loop.
- [x] **MEDIUM: truncated OOM-record lines were accepted.** `load_oom_record` validated only the first two fields, so a truncated/hand-edited line like `model,32768` was accepted and pre-skipped indefinitely, contradicting the header comment that promises wrong-field-count lines are skipped. Now validates the full field count (>= 6, the documented `model,prompt_target_len,date,commit,peak_bytes,source_csv` format); a source path containing a comma only inflates the count and stays accepted.

### Reported, not changed
- [ ] **MEDIUM: `cuMemAlloc failed` may match non-OOM CUDA errors.** `cuMemAlloc` can also fail with `CUDA_ERROR_INVALID_VALUE`, `CUDA_ERROR_DEINITIALIZED`, etc., so matching the call name alone could classify a non-OOM defect as `SKIP:oom` and then persist it. Left as-is deliberately: MLX emits these strings from its vendored C++ runtime (not in this tree), so the exact OOM-path wording cannot be confirmed here, and tightening the pattern risks a false negative, which for this gate is the more dangerous direction. The other two CUDA markers (`CUDA_ERROR_OUT_OF_MEMORY`, `cudaErrorMemoryAllocation`) are unambiguous. Flagging for a maintainer who can check MLX's actual CUDA OOM output; if it always includes an OOM reason string, narrowing `cuMemAlloc failed` to co-require that reason would remove the false-positive surface.

Re-verified after the fixes with `bash -n` and `shellcheck` (clean), plus the fake-OOM backstop, persistent-record pre-skip, and malformed/truncated-line paths. Working tree left clean; the pre-existing untracked `benchmarks/*.csv` files were untouched.

Issue #686, opened 2026-07-07
bug(gemma4): dense gemma-4 text modeling is catastrophically degraded (ppl ~12k vs gemma-3's 15 on the same corpus); likely root of the temp-0 pad-collapse


inureyes:
## Summary
The #683 quality gate (teacher-forced perplexity via the new `examples/perplexity.rs`, merged in PR #685) exposed that the gemma-4 family's text modeling in mlxcel is severely degraded while every other family is healthy on the identical corpus and harness. This is a correctness bug in the gemma-4 forward path (or its checkpoint loading), not a quantization artifact, and it is almost certainly the same root cause as the known temp-0 greedy pad-collapse on gemma-4-12b-it noted during PR #682 validation.

## Measured (GB10, docs-corpus ~62k tokens, chunked teacher-forced NLL, fresh caches per chunk)
| model | perplexity | verdict |
|---|---|---|
| qwen3.5-9b-4bit | 7.0 | healthy |
| gemma-3-4b-it-4bit | 15.3 | healthy (gemma-3 port is fine) |
| llama-3.1-8b-4bit | 24.8 | healthy |
| gemma-4-e4b-it-4bit | 50.0 | degraded (mild; per-layer-input sub-arch) |
| gemma-4-31b-it-4bit | 317 | severely degraded |
| gemma-4-12b-it-4bit | ~12,000 to 29,000 | catastrophic (uniform-random over the 262k vocab would be ppl ~262k / nll 12.48; measured nll 9.4 to 10.3) |

Context-length scan on gemma-4-12b: nll is flat-bad from the shortest chunks (32 tokens: nll 9.44; 64: 9.69; 128-512: ~10.3), while the gemma-3 control shows the normal improving-with-context pattern (64: 34.2 -> 512: 15.2). A 32-token chunk exercises no sliding window, no rotating cache, no long-range RoPE, so the defect is in the per-token forward itself: candidate suspects are the embed `sqrt(hidden)` scale, sandwich-norm wiring (pre/post attention and MLP order), QK-norm epsilon or placement, `v_norm` (RMSNormNoScale on values), `layer_scalar` application, final norm or lm_head wiring, logit softcapping, and local/global RoPE theta assignment. Severity ordering (e4b 50 / 31b 317 / 12b 12k) may indicate the bug scales with depth or differs across the dense vs per-layer-input sub-paths.

## Why this went unnoticed
[...]

PR #811, opened 2026-07-17
feat(server): derive thinking_mode chat-template kwarg from template

### LINKED ISSUE #775: feat(server): derive the thinking_mode chat-template kwarg when the template supports it
## Summary
Some chat templates gate reasoning on a `thinking_mode` string variable rather than the boolean `enable_thinking`. mlxcel already derives an `enable_thinking` default from the template (`src/server/chat_template.rs:147-169`) and has generic template-kwarg plumbing (`src/server/chat_template_kwargs.rs`), but it never sets `thinking_mode`, so such templates silently fall back to the non-thinking branch even when thinking is requested.

## Semantics
When thinking is enabled for a request (explicitly or by default) AND the raw chat template text references the identifier `thinking_mode`, inject `thinking_mode: "enabled"` into the template kwargs, unless the client already supplied a `thinking_mode` kwarg. Inject nothing when thinking is disabled (templates default the variable themselves).

## Implementation plan
1. Template-reference check: a substring scan of the template source for `thinking_mode`, in the same style as the existing enable_thinking detection in `chat_template.rs`.
2. Inject during kwarg resolution in `chat_template_kwargs.rs`, with client-provided values winning.
3. Tests: a template gated on `thinking_mode` renders the thinking branch when enabled; a client override is respected; templates without the identifier are untouched.

## Acceptance criteria
- Unit tests pass; rendering for existing templates (qwen3 family, gemma family) is unchanged.

Effort: low.

inureyes:
## Summary

Some HuggingFace chat templates gate reasoning on a thinking_mode string variable rather than the boolean enable_thinking mlxcel forwards by default, so those templates silently took their non-thinking branch even when thinking was requested. This adds detection of the thinking_mode identifier in the template source and injects thinking_mode: "enabled" into the template kwargs whenever thinking is enabled and the template references it, unless the client already supplied its own thinking_mode value.

## What changed

- `src/server/chat_template.rs`: added `ChatTemplateProcessor::wants_thinking_mode_alias`, a substring scan for the `thinking_mode` identifier, mirroring the existing `wants_bare_thinking_alias` detection used for the DeepSeek-V3.2 bare-thinking idiom.
- `src/server/chat_template.rs`: `build_template_context` now injects `thinking_mode: "enabled"` when the template references the identifier and the resolved `enable_thinking` value is true, and does nothing when thinking is disabled (affected templates default the variable themselves). An explicit client-supplied `thinking_mode` kwarg always wins.
- `src/server/chat_template.rs`: added unit tests for identifier detection across template shapes, injection when thinking is enabled, no injection when thinking is disabled, honoring the tokenizer-derived default when no request kwarg is present, client override of the derived alias, parity across `apply_with_kwargs` and `apply_raw_with_kwargs`, and a no-op guard for templates that never reference the identifier.

## Test plan

- [x] `cargo test --release --features cuda --lib -- server::chat_template server::chat_template_kwargs` (89 passed, 0 failed, 1 ignored)
- [x] Existing qwen3 and gemma4 template rendering tests pass unchanged

Closes #775

### DIFF
--- a/docs/supported-models.md
+++ b/docs/supported-models.md
@@ -244,6 +244,10 @@ chat template that already reads `enable_thinking` directly and is
 unaffected. An explicit `thinking` entry in `chat_template_kwargs` always
 overrides the derived alias.
 
+### `thinking_mode` alias for chat templates gating on a string variable
+
+Some chat templates check a `thinking_mode` string (for example `thinking_mode == "enabled"`) instead of the `enable_thinking` boolean, so mlxcel also detects that identifier in the loaded template and injects `thinking_mode: "enabled"` whenever the resolved `enable_thinking` value is true. Nothing is injected when thinking is off, and an explicit `thinking_mode` entry in `chat_template_kwargs` always overrides the derived value.
+
 ### Loop-detection default-on for the Gemma 4 family
 
 Gemma 4 (31B Dense and 26B-A4B MoE, including the QAT 4-bit checkpoints) has an upstream, weights-level token-repetition collapse documented in [google-deepmind/gemma#622](https://github.com/google-deepmind/gemma/issues/622) and reproduced across other engines, including [vllm-project/vllm#40080](https://github.com/vllm-project/vllm/issues/40080). Generation can degenerate into a single repeated token (or short fragment) that fills the token budget, most often inside the thought channel. Tool declarations and `json_schema` structured-output requests amplify it, but it is not limited to those cases. Sampling penalties do not reliably recover it, because once the logits collapse the top-k candidates are themselves garbage.
--- a/src/server/chat_template.rs
+++ b/src/server/chat_template.rs
@@ -257,6 +257,26 @@ impl ChatTemplateProcessor {
         self.template.contains("not thinking is defined")
     }
 
+    /// Detect a chat template that gates reasoning on a `thinking_mode`
+    /// string variable rather than the conventional boolean `enable_thinking`
+    /// kwarg mlxcel forwards by default (issue #775).
+    ///
+    /// Some HuggingFace chat templates check `thinking_mode == "enabled"` (or
+    /// similar string comparisons) instead of the `enable_thinking` boolean.
+    /// mlxcel never populates `thinking_mode`, so such templates silently take
+    /// their non-thinking branch regardless of what the caller asked for.
+    ///
+    /// The check is a conservative substring scan for the literal identifier
+    /// `thinking_mode` anywhere in the template source, mirroring the style of
+    /// [`wants_bare_thinking_alias`] and the other template-introspection
+    /// heuristics in this file. It may false-positive on a template that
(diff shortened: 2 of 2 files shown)


inureyes:

### Intent
> Inject `thinking_mode: "enabled"` into chat-template kwargs when the template references the `thinking_mode` identifier and thinking is effectively enabled, unless the client already supplied a `thinking_mode` kwarg (issue #775).

### Findings Addressed
None. Independent review plus a supplementary automated review pass found no actionable correctness, security, or maintainability issues.

### Remaining Items
None.

### Verification
- [x] All stated requirements implemented: identifier detection (`wants_thinking_mode_alias`, chat_template.rs:276), injection gated on `effective_enable_thinking`, client-wins precedence, injects the spec-mandated `"enabled"` string, injects nothing when thinking is disabled.
- [x] No placeholder/mock code remaining: single-file diff, all real logic.
- [x] Integrated into project code flow: threaded through both `apply_with_kwargs` (:455) and `apply_raw_with_kwargs` (:520), both of which are the production render entry points called from `chat_request.rs:199,213`.
- [x] Project conventions followed: mirrors the existing `wants_bare_thinking_alias` heuristic and the `thinking` alias injection block; `effective_enable_thinking` hoisted and shared so both aliases resolve identically.
- [x] Existing modules reused where applicable: reuses the existing template-kwarg plumbing, `default_enable_thinking` resolution, and the bare-thinking precedent rather than reimplementing.
- [x] No unintended structural changes: no renames/moves; the only refactor is hoisting `effective_enable_thinking` out of the bare-thinking block (behavior-preserving).
- [x] Tests pass: 7 new tests cover detection, enabled/disabled injection, tokenizer-default derivation, client override, both apply paths, and a no-op guard for non-matching templates; PR reports 89 passed / 0 failed on the consolidated run.

### Reviewer notes (non-blocking)
- Precedence: `owned` is seeded only with reserved/default keys plus the client kwargs loop; `thinking_mode` is never a default key, so `!owned.contains_key("thinking_mode")` is true exactly when the client did not supply it. Client `thinking_mode` (including `"disabled"`) always wins. Correct.
- Prompt-cache key: injection is internal to `build_template_context` and does not mutate `merged_kwargs`, so `template_sig` hashes the pre-injection kwargs. Identical requests hash identically; thinking on/off already differs via `enable_thinking`. No key instability.
- Substring heuristic (`template.contains("thinking_mode")`) is looser than the `wants_bare_thinking_alias` phrase match, so it can match a longer identifier or a comment. Impact is nil: a spurious `thinking_mode` context entry is inert unless the template reads the exact variable, and the injected value only matters when the template compares against `"enabled"`. The docstring documents this trade-off. Consistent with the established alias heuristics.
- gemma-4 templates use `enable_thinking`, not `thinking_mode`, so the alias never fires and their rendering path is unchanged (byte-identical), preserving the #686 `wants_thinking_default_off` behavior.

Issue #818, opened 2026-07-19
fix(server): speculative serving aborts the whole process via MLX CUDA graph-cache "Cache thrashing" throw (lifetime miss counter vs default capacity 400)


inureyes:
## Summary

A `mlxcel-server` running batched speculative decoding (dflash drafter) dies with an uncaught C++ exception once MLX's CUDA graph cache has accumulated more than 2x its capacity in lifetime misses:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cache thrashing is happening, please set the environment variable MLX_CUDA_GRAPH_CACHE_SIZE to a larger value than 400 to fix degraded performance.
```

The whole process aborts: no request-level error, no recovery, all in-flight requests dropped. Reproduced on GB10 with both v0.4.1 and HEAD (3d738cf), so this is not a #816 regression; #816 release testing surfaced it.

## Repro

```
mlxcel-server -m models/qwen3.5-4b-4bit --draft-model models/qwen3.5-4b-dflash --port 18324
# fire 6 concurrent chat completions (varying prompts, max_tokens 48), then repeat
```

HEAD died after the first burst plus one follow-up request; the v0.4.1 binary died on the second burst of the identical scenario. With `MLX_CUDA_GRAPH_CACHE_SIZE=2000` the same scenario completes 13/13 requests across two bursts plus a single request, with dflash engaged (42 rounds, 630 proposed / 107 accepted tokens, 17% acceptance).

## Root cause (vendored MLX 0.32.1)

- `mlx/backend/cuda/lru_cache.h` line 91: `if (env_name_ && ++cache_misses_ > 2 * capacity_) throw std::runtime_error(...)`. `cache_misses_` (line 131) is a lifetime counter: never reset, not on hits, not on trim. Any process that ever inserts more than `2 * capacity` distinct graph keys throws, regardless of the hit rate at that moment.
- `mlx/backend/cuda/device.cpp` line 210: `graph_cache_("MLX_CUDA_GRAPH_CACHE_SIZE", /* default_capacity */ 400)`, so the throw threshold is 800 lifetime misses.
- `mlx/backend/cuda/device.cpp` line 490: the cache key is `graph_nodes_key_ + ":" + graph_deps_key_`, so every distinct captured graph shape is a distinct key.

[...]

PR #825, opened 2026-07-19
fix(server): survive MLX decode-eval throws instead of aborting the worker

### LINKED ISSUE #822: fix(server): catch MLX C++ aborts at the decode FFI boundary so a backend throw fails the request, not the process
## Summary

When MLX's C++ side throws (for example the graph-cache "Cache thrashing" `runtime_error` in #818, but also allocation failures and other internal errors), the exception propagates through an infallible cxx boundary and the whole `mlxcel-server` process dies with `terminate called after throwing an instance of 'std::runtime_error'`. Every in-flight request is dropped, with no request-level error and no operator-actionable log beyond the raw C++ `what()`. #818 removed the one specific trigger that we hit in practice (the graph-cache default), but the crash *class* remains: any MLX throw reached from the batch decode loop still aborts the process. This issue is the defense-in-depth follow-up (#818 proposal 2): route the decode-loop evaluation through the fallible boundary so an MLX failure fails the affected request/batch instead of the process.

## Root cause

The batch scheduler decode path evaluates on-device tensors through the infallible `mlxcel_core::eval` / `mlxcel_core::async_eval` wrappers, declared in `src/lib/mlxcel-core/src/lib.rs` as plain cxx functions (`fn eval(arr: &MlxArray);` at line 202, `fn async_eval(arr: &MlxArray);` at 1782). cxx generates a `noexcept` shim for a function that does not return `Result`, so a C++ exception crossing that boundary calls `std::terminate`. The scheduler uses the infallible `eval` at 10 sites in `src/server/batch/scheduler.rs` (4777, 4923, 5067, 5083, 5254, 5327, 6059, 6238, 6577) plus `async_eval` at 5959.

The fallible boundary already exists and is already used elsewhere: `fn try_eval(arr: &MlxArray) -> Result<()>;` (`lib.rs:212`) makes cxx catch the C++ exception and return it as a Rust `Err`. The audio and whisper workers deliberately go through it precisely so an MLX failure surfaces as an error rather than a process abort (`src/server/audio_worker.rs:399,493`, `src/models/whisper/mod.rs:149`, `src/models/whisper/decoding.rs:88`). The batch decode loop, which is the highest-throughput and longest-lived MLX consumer, is the one place that does not.

## Proposed change

1. Add a fallible `try_async_eval(arr: &MlxArray) -> Result<()>` (and, if the lookahead pipeline needs it, `try_async_eval_pair`) to the cxx bridge, mirroring `try_eval`. `async_eval` has no fallible variant today.
2. Route the scheduler decode-loop evaluations (the 10 `eval` sites and the `async_eval` site above) through the fallible variants. On `Err`, translate it into a request-level failure for the affected sequence(s)/batch: return a 500 `internal_error` to those requests with the MLX `what()` string in the operator log, and continue serving other sequences. This mirrors the audio/whisper pattern already in the tree.
3. Add a bounded-failure policy so a genuinely unrecoverable backend does not spin: after N consecutive decode-eval failures (or a failure whose message indicates allocator/device corruption), drain in-flight requests with an error and shut the worker down cleanly rather than looping.

## Risks and scope notes

- Not every MLX throw is safely recoverable. The graph-cache thrashing throw happens at graph *capture*, before any state mutation, so failing just that batch is clean. But an out-of-memory or an internal-invariant throw may leave the allocator or device in an undefined state; for those, continuing to serve other requests is unsafe. The bounded-failure policy (point 3) is why this is not simply "catch and keep going": the goal is a *graceful, logged, request-scoped failure or clean shutdown*, not pretend-recovery.
- Performance: `try_eval` returns `Result<()>` but is otherwise the same MLX call; the added cost is a branch on the hot path, negligible against a decode step. Confirm no measurable decode-throughput regression on a GB10 sweep before/after.
- This is defense in depth, not a fix for a currently-reproducing crash (that was #818). Priority is medium: it hardens the process against the next MLX throw we have not yet hit.

## Refs

Follow-up to #818 (proposal 2). Sibling follow-ups: the upstream MLX lifetime-miss report (proposal 3) and the paged-state sync warning spam.

inureyes:
## Summary

The batch scheduler decode loop force-evaluated on-device tensors through the infallible `mlxcel_core::eval` / `async_eval` wrappers, which cxx compiles as noexcept shims, so any MLX C++ throw reached from decode (the graph-cache abort from #818, an allocation failure, an internal invariant) crossed the FFI boundary uncaught and called `std::terminate`, killing the whole server and dropping every in-flight request with no request-level error. This routes the decode/prefill evaluation through a fallible boundary so a backend throw fails just the affected request(s), not the process.

## What changed

- `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.{h,cpp}` + `src/lib/mlxcel-core/src/lib.rs`: add `try_async_eval` (body identical to `async_eval`, declared `-> Result<()>` so cxx catches any thrown MLX exception at the FFI boundary), mirroring the existing `try_eval`.
- `src/server/batch/scheduler.rs`: route every decode and chunked-prefill evaluation site through the fallible `try_eval` / `try_async_eval` variants and translate an `Err` into the existing `abort_sequence` path, so the worker keeps serving the other sequences. A bounded consecutive-failure guard (`MAX_CONSECUTIVE_EVAL_FAILURES = 8`, reset on any success) treats a persistently-failing backend as unrecoverable: past the threshold it logs a fatal line, aborts the remaining in-flight sequences, and stops the scheduler loop cleanly instead of spinning.

## Scope and risk

Not every MLX throw is safely recoverable: the graph-cache throw happens at graph capture before any state mutation, so failing just that batch is clean, but an out-of-memory or internal-invariant throw can leave the allocator or device undefined. That is why the guard exists; the goal is a graceful request-scoped failure or a clean shutdown, not pretend-recovery. The success path is unchanged (`try_*` is the same MLX call plus a branch on `Result`).

## Test plan

- `try_async_eval_ok_matches_async_eval` exercises the fallible boundary's success path (schedules a valid graph, returns `Ok`, reads back the expected value).
- Scheduler failure-counter logic covered by pure-function unit tests: reset on success, increment and saturate on failure, threshold inclusivity, guard tripping after consecutive failures, single-success reset.
- Full CUDA build (`make release-cuda`) passes; `cargo check --lib --features cuda` clean (no Rust warnings). Server decode path runs normally end to end (3 concurrent chat completions complete correctly through the fallible-eval loop, no regression).

Closes #822


inureyes:

### Intent
> Route every batch-scheduler decode/prefill eval through the fallible `try_eval` / `try_async_eval` cxx boundary so an MLX C++ throw fails just the affected request(s) instead of `std::terminate`-ing the worker, guarded by a bounded consecutive-failure shutdown.

### Findings Addressed
- [x] Lookahead prime leaked a speculative KV position on a caught `try_async_eval` throw (HIGH) — fixed in `eb3e8ff`.

`prime_lookahead_with_input` returned `None` on the new caught-throw path without unwinding the speculative KV position `lookahead_forward` had already appended per sequence. The caller set `decode_lookahead = None` (and the steady path committed step n while dropping step n+1), so the next synchronous decode ran on a cache whose KV tail was one position longer than `generated_tokens` accounted for — a desync that corrupts the sequence(s) or forces a batch-wide re-throw, defeating the request-scoped recovery this change is meant to provide. It is a newly reachable path: the previous infallible `async_eval` could only hit the caches-vanished `None` (before any append). Fixed by trimming the one speculative position via the existing `apply_lookahead_trim(seq_ids, lookahead_teardown_positions(false))` teardown before bailing, mirroring the stale/bootstrap fallbacks; the caches-vanished `?` path is untouched (it returns before any append). All prime paths (bootstrap, stale, finishing re-prime, steady step n+1) funnel through this function, so one fix covers them.

### Remaining Items (non-blocking, informational)
- [ ] Batched-fused decode readback (`batched_fused_sample`) and the primed-token readback (`lookahead_tokens_to_host` → `array_evaluated_bytes`) still materialize through infallible host-copy paths (LOW) — outside the enumerated `eval`/`async_eval` scope of #822 (the #818 throw fires at graph capture, caught by the new `try_*` sites); the fused case is already documented in-code as a follow-up.

### Verification
- [x] All stated requirements implemented — all 10 enumerated eval sites converted (9 `try_eval` + 1 `try_async_eval`); no infallible `eval`/`async_eval` remains in the scheduler.
- [x] No placeholder/mock code remaining.
- [x] Integrated into project code flow — conversions live in the real decode/prefill functions reached via `run()` → `decide_action()`; `shutdown_requested` is honored at the loop head; guard drains both `active_batch` and `chunked_prefill_seq` without duplicate error events.
- [x] Project conventions followed — `try_*(&x).map_err(...)` matches the whisper/audio fallible-eval pattern; reuses the existing `abort_sequence` / `abort_sequence_with_error` / `apply_lookahead_trim` helpers.
- [x] Existing modules reused where applicable.
- [x] No unintended structural changes — deferred-init `prefill_eval` borrow handling is sound; success path is unchanged (same MLX call plus a `Result` branch).
- [x] Consecutive-failure guard correct — reset on success, saturating increment on failure, `count >= 8` threshold; no off-by-one (unit-tested).
- [x] Compiles clean — `cargo check --lib --features cuda` (no Rust warnings) after the fix.

PR #891, opened 2026-07-23
fix(gemma4): size chunked-prefill masks to returned keys per family

### LINKED ISSUE #885: bug: Gemma 4 sliding-window server output collapses to reserved <unused> tokens under chunked prefill
# mlxcel-server emits reserved `<unused…>` garbage tokens on Gemma 4 sliding-window models: chunked-prefill continuation drops the sliding-window attention mask

## Summary

On `mlxcel-server`, a Gemma 4 sliding-window model can collapse into a short repeating cycle of reserved `<unusedNNNN>` vocabulary tokens instead of a real answer. Observed on `gemma-4-e4b-it-qat-4bit` (a **dense** Gemma 4, `enable_moe_block=false`, `sliding_window=512`) with a thinking budget active. The server logs a matching warning at the moment it happens:

```
WARN chunked_prefill_continue{seq_id=seq-6 offset=2048 total=2201}:
  mlxcel::models::gemma4: trim_mask_to_keys: mask shorter than key length,
  discarding caller mask mask_len=160 key_len=671
```

Sample corrupted output:

```
<unused2115><unused2102><unused2120><unused2115><unused2120><unused2102><unused2102>...
```

(`<unused2115>` etc. are reserved high-vocab tokens, e.g. vocab ids 258027 / 258014 / 258032; the model should never emit them.)

## Root cause

During server-side **chunked prefill continuation** (long prompt split into `--prefill-chunk-size` chunks, default 512, so decode can be interleaved for other active sequences), the scheduler passes a chunk-local causal prefill mask into the model forward. That mask is sized to the current continuation chunk (padded/tile-aligned, e.g. `mask_len=160` for a ~153-token chunk), built via `create_padded_prefill_mask` (`src/server/batch/scheduler.rs:4863` / `5024` / `5186` / `5370`).

Gemma 4's **sliding-attention** layers, however, read a `RotatingKVCache` whose returned key length is the sliding window plus the new chunk tokens (here `key_len=671`, roughly `sliding_window(512) - 1 + chunk`). The per-layer guard `trim_mask_to_keys` (`src/models/gemma4.rs:2364-2387`) can only crop a mask that is **longer** than the keys; when the caller mask is **shorter** than the keys it cannot be fixed, so it is discarded and `None` is returned:

```rust
} else {
    // mask is smaller than key length ...
    tracing::warn!(mask_len, key_len, "trim_mask_to_keys: mask shorter than key length, discarding caller mask");
    None
}
```

With the mask dropped, the sliding-attention layers run for a **multi-token** continuation chunk (`l > 1`) with no window/causal mask. The kernel's internal fallback cannot reconstruct the correct causal + windowed alignment for queries at absolute positions `[offset, total)` against a rotated key buffer, so attention is computed incorrectly. The corrupted hidden state drives the LM head into a degenerate region and the model emits reserved `<unused…>` tokens in a short repeating cycle.

Why the visible failure is intermittent/hard to hit: it requires (1) a prompt long enough to cross a `prefill_chunk_size` boundary, and (2) concurrent load so the scheduler actually chunks-and-interleaves (the reporter's `seq_id=seq-6` indicates several concurrent sequences). A single idle request often prefills whole and does not trigger it. A thinking budget makes it more likely by extending sequences and keeping the batch busy.

## Reproduction

- Model: `gemma-4-e4b-it-qat-4bit` (dense, `sliding_window=512`).
- Serve with default chunking (`--prefill-chunk-size 512`) and `--max-batch-size > 1`.
- Drive concurrent requests, at least one with a prompt longer than 2048 tokens (crossing multiple chunk boundaries) while other sequences decode, optionally with a thinking budget (`thinking_budget_tokens`).
- Watch the server log for `trim_mask_to_keys: mask shorter than key length, discarding caller mask`; responses that coincide with it can contain `<unused…>` token cycles.

Note: the reporter reproduced this on a production server; a deterministic minimal repro tied to the exact concurrent chunked-continuation + rotating-cache state is still being reduced. The warning above is direct evidence of the mask-drop.

## Expected

The sliding-attention layers must always receive a mask whose key axis matches the `RotatingKVCache`'s returned key length for the current chunk (correct causal + window alignment for queries at `[offset, total)`), so `trim_mask_to_keys` never has to discard it. No reserved `<unused…>` tokens should ever appear in output.

## Proposed fix direction

- For chunked-prefill continuation, build the per-layer sliding mask from the same live-window length the rotating cache returns (mirror the non-chunked `l > 1` branch that uses `first_cache_live_len` / `create_sliding_window_prefill_mask`, `src/models/gemma4.rs:3139-3158`), instead of relying on a chunk-local caller mask that is shorter than the sliding key buffer.
- Make the too-short case in `trim_mask_to_keys` a hard error (or a correct rebuild) rather than a silent discard, at least for multi-token (`l > 1`) chunks where dropping the mask corrupts attention. Silently discarding is only safe for single-token decode where the kernel derives its own mask.
- Add a regression test: chunked prefill of a > `prefill_chunk_size` prompt on a sliding-window Gemma 4 model, asserting no `<unused…>` tokens and output parity against unchunked (`--prefill-chunk-size 0`) prefill.

## Impact

- Affects Gemma 4 sliding-window models served with chunked prefill under concurrent load. `gemma-4-e4b-it-qat-4bit` confirmed (dense, window 512); other Gemma 4 sliding-window checkpoints (window 1024) are likely susceptible under the same chunked-continuation condition.
- User-visible: garbage `<unused…>` responses.

## Notes / related

- This is independent of the separate Gemma 4 26B A4B (MoE) multi-turn corruption, which is the fused single-token-decode MoE kernel (`MLXCEL_FUSED_MOE=0` avoids that one); `gemma-4-e4b-it-qat-4bit` here is dense, so that path is not involved.
- Related but distinct from #884 (Gemma 4 `<|channel>thought` reasoning-channel leak in CLI output). This issue is the `<unused…>` garbage collapse from the chunked-prefill mask drop; #884 is the reasoning-channel text leak. Both are Gemma 4 output-quality issues but have different root causes.
- Environment: Linux GB10 (CUDA, sm_121), HEAD of `main`. `trim_mask_to_keys` at `src/models/gemma4.rs:2364`; scheduler prefill masks at `src/server/batch/scheduler.rs`.

inureyes:
## Summary

Gemma 4 sliding-window models served with chunked prefill under concurrent load could collapse into a repeating cycle of reserved `<unused...>` tokens instead of a real answer. The root cause is a chunk-local prefill mask that is sized to the wrong key axis and then silently discarded, dropping causal / sliding-window alignment during a chunked-prefill continuation. This makes each family's prefill mask match the keys its cache actually returns, so the discard never happens and chunked prefill matches unchunked prefill.

## Root cause (confirmed against code)

During a chunked-prefill CONTINUATION the scheduler passes the server's chunk-local `create_padded_prefill_mask` into the model forward. Gemma 4 is a model-owned-cache batching model (`supports_batching() == true`, `make_caches()` returns an empty `Vec`), so the cache pool hands the scheduler dummy placeholder caches and `continue_chunked_prefill` computes `kv_offset = caches.first().map_or(0, |c| c.offset) == 0`. The caller mask is therefore built with offset 0 (key axis 160 in the reporter's `mask_len=160` log). Inside the model, `forward_with_sequence_id` resolves the REAL per-sequence caches (offset 2048), so the full-attention `KVCache` returns `global_live_len + l` keys (2208) and the sliding `RotatingKVCache` returns `min(sliding_live_len, window - 1) + l` keys (671, the reporter's `key_len=671`). The `else if let Some(mask) = mask` branch of `forward_with_speculative_sinks` (`src/models/gemma4.rs:3058`) reused that single short mask for BOTH families, so `trim_mask_to_keys` (`src/models/gemma4.rs:2364`) took its too-short branch and discarded the mask. Every layer then fell through to the maskless `causal_attention` path, which cannot reconstruct the correct causal + windowed alignment for queries at `[offset, total)` against a rotated key buffer, corrupting attention and driving the LM head into the reserved `<unused...>` region.

The full-attention (global) layers were affected by the same discard: their offset-0 caller mask (key axis 160) is also shorter than the full cache's returned 2208 keys, so their mask was discarded too and they relied on the maskless causal fallback. The fix now gives them an explicit mask sized to the live window as well.

## What changed

- `src/models/gemma4.rs` (`forward_with_speculative_sinks`, `else if let Some(mask) = mask` branch): when `l > 1` and the caller mask's key axis is shorter than the keys either family returns, rebuild BOTH masks from the caches' live windows exactly as the proven non-caller `l > 1` branch does: `create_causal_mask(l, global_live_len)` for full attention and `create_sliding_window_prefill_mask(l, sliding_live_len, window)` for the sliding family. A correctly-sized caller mask (cold batched prefill, or a full prefill that adopted a prompt-cache prefix) is preserved unchanged and cropped per family by `trim_mask_to_keys`.
- `src/models/gemma4.rs` (`trim_mask_to_keys` / `Attention::attend`): the function now takes `query_len`. A too-short mask on a MULTI-token chunk (`query_len > 1`) fails loudly (panic referencing issue #885) instead of silently discarding, since silent discard is the corruption. Single-token decode (`query_len == 1`) keeps the safe discard-and-derive-own-mask behaviour. The mask-longer-than-keys crop path is unchanged.
- `src/models/gemma4_tests.rs` and the `gemma4_unified_mask_tests` module: deterministic regression coverage (below).

## Test plan

- [x] `cargo test --lib gemma4_unified_mask_tests` (9 passed) including new: `chunked_continuation_sliding_mask_matches_returned_keys` and `chunked_continuation_global_mask_matches_returned_keys` (rebuilt mask key axis equals the rotating/standard cache's returned keys, while the offset-0 caller mask is shorter), `trim_mask_to_keys_crops_when_mask_longer_than_keys`, `trim_mask_to_keys_panics_on_too_short_multi_token` (`#[should_panic]`), `trim_mask_to_keys_discards_too_short_single_token_decode`.
- [x] `cargo test --lib models::gemma4_tests::cache_isolation` (4 passed, 3 ignored) including new PRIMARY parity test `gemma4_chunked_prefill_with_server_caller_mask_matches_single_pass`: on a tiny synthetic sliding-window Gemma 4 model, chunked prefill that feeds the server's offset-0 `create_padded_prefill_mask` on every continuation chunk produces logits within 1e-3 of single-pass prefill for sliding (window 4 and 8, chunk 8, over-window) and full attention. The rebuild is load-bearing: without it the sliding window-4 continuation chunk hits the new `trim_mask_to_keys` panic.
- [x] `cargo test --lib models::gemma4` (88 passed, 8 ignored) confirms the MTP / ragged / divergent-verify and existing chunked-parity paths still pass.
- [x] `cargo check --lib` and `cargo clippy --lib` clean (only pre-existing mlxcel-core C++ build notes).

Full CUDA build plus server E2E on the GB10 host (`gemma-4-e4b-it-qat-4bit`, concurrent chunked prefill) is left to the orchestrator, since the MLX C++ link and a 26B/GPU run exceed the sandbox watchdog.

Closes #885

### DIFF
--- a/src/models/gemma4.rs
+++ b/src/models/gemma4.rs
@@ -1885,7 +1885,8 @@ impl Attention {
         values: &MlxArray,
         mask: Option<&MlxArray>,
     ) -> UniquePtr<MlxArray> {
-        let local_mask = trim_mask_to_keys(mask, keys);
+        let query_len = mlxcel_core::array_shape(queries)[2];
+        let local_mask = trim_mask_to_keys(mask, keys, query_len);
 
         // When mask was discarded (undersized) or originally None,
         // use causal attention if possible.
@@ -2361,7 +2362,11 @@ impl Attention {
     }
 }
 
-fn trim_mask_to_keys(mask: Option<&MlxArray>, keys: &MlxArray) -> Option<UniquePtr<MlxArray>> {
+fn trim_mask_to_keys(
+    mask: Option<&MlxArray>,
+    keys: &MlxArray,
+    query_len: i32,
+) -> Option<UniquePtr<MlxArray>> {
     let mask = mask?;
     let mask_shape = mlxcel_core::array_shape(mask);
     let key_shape = mlxcel_core::array_shape(keys);
@@ -2372,16 +2377,34 @@ fn trim_mask_to_keys(mask: Option<&MlxArray>, keys: &MlxArray) -> Option<UniqueP
     } else if mask_len > key_len {
         let start = mask_len - key_len;
         Some(slice_axis(mask, -1, start, mask_len))
+    } else if query_len > 1 {
+        // Mask SHORTER than the keys on a MULTI-token chunk (prefill /
+        // continuation / verify). Silently discarding it (the pre-#885
+        // behaviour) drops the causal + sliding-window alignment for every
+        // query row and routes the layer through the maskless
+        // `causal_attention` fallback, whose offset handling cannot
+        // reconstruct the correct window for a rotated key buffer. On a
+        // sliding-window Gemma 4 model that corrupts attention and collapses
+        // the output into reserved `<unused...>` tokens. The forward is
+        // responsible for building a mask whose key axis matches the keys this
+        // layer's cache returns (see `forward_with_speculative_sinks`), so a
+        // too-short mask here is an invariant violation: fail loudly instead
+        // of serving garbage.
+        panic!(
+            "trim_mask_to_keys: caller mask key axis ({mask_len}) is shorter than the returned \
+             key length ({key_len}) for a multi-token chunk (query_len={query_len}); the prefill \
+             mask must be sized to the cache's returned keys (issue #885)"
+        );
     } else {
-        // mask is smaller than key length — this happens when an external
-        // mask was created with stale offset (e.g. during chunked prefill on
-        // non-batching models).  Discard the undersized caller mask and let
-        // the attention kernel fall back to its internal causal handling by
-        // returning None, which will cause the caller to pass a null mask.
+        // Single-token decode (`query_len == 1`): a caller mask built with a
+        // stale offset can be shorter than the keys, but the decode kernel
+        // derives its own causal / sliding-window mask, so discarding the
+        // undersized mask and returning `None` is safe here.
         tracing::warn!(
             mask_len,
             key_len,
-            "trim_mask_to_keys: mask shorter than key length, discarding caller mask"
+            "trim_mask_to_keys: mask shorter than key length on single-token decode, \
+             discarding caller mask"
         );
         None
     }
@@ -3056,7 +3079,43 @@ impl Gemma4TextModel {
                 )),
             )
         } else if let Some(mask) = mask {
-            (Some(mlxcel_core::copy(mask)), Some(mlxcel_core::copy(mask)))
+            // A caller-supplied prefill mask (the server's
+            // `create_padded_prefill_mask`) is chunk-local. For a model-owned-
+            // cache batching model like Gemma 4 the scheduler builds it against
+            // its dummy (offset-0) cache view, so during a chunked-prefill
+            // CONTINUATION the mask's key axis is far shorter than the per-family
+            // caches actually return for this chunk: the full-attention
+            // `KVCache` returns `global_live_len + l` keys and the sliding
+            // `RotatingKVCache` returns `min(sliding_live_len, window - 1) + l`.
+            // Reusing that single short mask for BOTH families makes
+            // `trim_mask_to_keys` discard it (mask shorter than keys), dropping
+            // the causal / sliding-window alignment and collapsing sliding-
+            // window Gemma 4 output into reserved `<unused...>` tokens (issue
+            // #885). When the caller mask is too short for the returned keys,
+            // rebuild BOTH masks from the caches' live windows exactly as the
+            // non-caller `l > 1` branch below, so each family's mask key axis
+            // matches its returned keys and chunked prefill is byte-identical to
+            // unchunked prefill. A correctly-sized caller mask (cold batched
+            // prefill, or a full prefill that adopted a prompt-cache prefix) is
+            // preserved unchanged and cropped per family by `trim_mask_to_keys`.
+            let mask_key_len = *mlxcel_core::array_shape(mask).last().unwrap_or(&0);
+            let global_live_len = first_cache_live_len(caches, "full_attention");
+            let sliding_live_len = first_cache_live_len(caches, "sliding_attention");
+            let window = self.config.sliding_window as i32;
+            let full_returned = global_live_len + l;
+            let sliding_returned = sliding_live_len.min((window - 1).max(0)) + l;
+            if l > 1 && mask_key_len < full_returned.max(sliding_returned) {
+                (
+                    Some(create_causal_mask(l, global_live_len)),
+                    Some(create_sliding_window_prefill_mask(
+                        l,
+                        sliding_live_len,
+                        window,
+                    )),
+                )
+            } else {
+                (Some(mlxcel_core::copy(mask)), Some(mlxcel_core::copy(mask)))
+            }
         } else if has_padding {
             // Resident prompt padding present (ragged burst). Build per-row
             // left-padding-aware masks for ANY query width, including `l == 1`.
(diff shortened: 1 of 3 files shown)


inureyes:
## Security & performance review: one CRITICAL finding (fixed on-branch)

### CRITICAL: the `trim_mask_to_keys` panic is a reachable server-crash (DoS), via the MTP verify path

The prior reviewers concluded the new `panic!` is unreachable. That is correct **only for the chunked-prefill path**: the caller-mask-branch rebuild (`forward_with_speculative_sinks`, the `else if let Some(mask) = mask` arm) sizes both attention-family masks to the keys the caches return, so `trim_mask_to_keys` never sees a too-short multi-token mask there. Confirmed.

But `trim_mask_to_keys` is shared, and a **different** caller reaches it with a too-short `l>1` mask: the Gemma 4 MTP (speculative) verify path. A B=1 (or equal-length batched) verify round calls `forward_with_speculative_sinks` with `mask=None`, `left_padding=None`, uniform `per_row_valid_end`, so it takes the plain `else if l > 1` branch, which builds `create_sliding_window_prefill_mask(l, sliding_live_len, window)` (key axis capped at `min(sliding_live_len, window-1) + l`). In MTP the sliding cache is a **buffered** `RotatingKVCache` (`enable_speculative_buffer`), whose `update_and_fetch` returns `sliding_live_len + l` uncompacted keys. Once `sliding_live_len >= window` the mask is shorter than the returned keys, so the new panic fires and crashes the worker. Pre-#885 this same case silently discarded the mask and fell back to `causal_attention(..., window_size)`, which is correct for the chronological buffered layout (no crash).

Empirically reproduced with a real `Gemma4MtpTargetAdapter` on the synthetic `sliding_window=8` fixture: an 8-token prompt then one `K=2` verify block panics with `caller mask key axis (9) is shorter than the returned key length (10) ... (issue #885)`. Reverting only the panic makes the same call return normally. So the panic is a regression introduced by this PR, not a pre-existing crash. This is reachable by any normal MTP request that generates past the sliding window.

**Fix applied on this branch** (confined to `src/models/gemma4.rs` and `src/models/gemma4_mtp_target_tests.rs`): `trim_mask_to_keys` now discards a too-short mask with a warning for **any** `query_len` (the pre-#885 behavior), instead of panicking on `query_len > 1`. This removes the DoS while keeping the real #885 fix (the caller-mask-branch rebuild) fully intact: the rebuild already guarantees the chunked-prefill path never delivers a too-short multi-token mask to `trim_mask_to_keys`, so nothing on that path relies on the panic. The `#[should_panic]` unit test was converted to assert the discard, and a permanent MTP regression test (`mtp_multi_token_verify_at_window_does_not_panic`) was added.

Validation (narrow selectors, CPU sandbox): `cargo check --lib`, `cargo clippy --lib -- -D warnings` clean; `models::gemma4` suite 90 passed / 0 failed / 8 ignored, including the chunked-prefill parity test (`gemma4_chunked_prefill_with_server_caller_mask_matches_single_pass`) and the new MTP regression test. Full CUDA build + real-model MTP greedy-parity on the GB10 host is left to the orchestrator.

### Performance: no regression

The per-chunk mask rebuild is built **once per forward** (before the layer loop); the loop only selects a reference per layer, so it is not per-layer work. Both rebuilt masks use the same `create_causal_mask` / `create_sliding_window_prefill_mask` builders as the proven non-chunked `l > 1` branch, so cost is `O(l * keys)`, not quadratic-per-chunk. The common paths are untouched: single-token decode (`l == 1`) keeps the cheap `(None, None)` fallback, and the rebuild only fires on a genuine chunked continuation (`mask.is_some() && l > 1 && caller mask shorter than the returned keys`). A correctly-sized caller mask (cold batched prefill, prompt-cache-prefix adoption) still takes the cheap copy path.

### Shared-branch other users: no further regressions

`has_padding` (ragged MTP) builds full-axis windowed masks (`mask >= keys`, crop-safe, never discarded); `divergent_verify` and `per_row_valid_end` require `mask.is_none()` or coexist only on prefill where `per_row_valid_end` is `None`, and use full-axis / gap-masked builders. The only correctness-as-safety regression in the shared branch was the MTP verify panic, now fixed.

### Verdict on the panic-DoS reachability question

Reachable. CRITICAL. Not via chunked prefill (the reviewers' scope, where the rebuild makes it unreachable) but via the MTP speculative verify path, which shares `trim_mask_to_keys`. Fixed on-branch.

inureyes:
## PR Finalization Complete

Verified the branch at HEAD e230d8b (`fix/issue-885-gemma4-chunked-prefill-sliding-mask`) against the finalization checklist. No further changes were needed.

### Formatting / Lint
- `cargo fmt --all -- --check`: clean across the whole workspace.
- `cargo clippy --lib -- -D warnings`: clean (only pre-existing mlxcel-core C++ build notes from the vendored MLX headers, unrelated to this change).
- `cargo check --lib`: clean.

### Tests
- `cargo test --lib models::gemma4`: 90 passed, 8 ignored (serial-MLX-execution tests, expected), 0 failed. This includes the chunked-prefill parity test (`gemma4_chunked_prefill_with_server_caller_mask_matches_single_pass`), the mask-axis unit tests in `gemma4_unified_mask_tests` (including the two issue-#885 regression tests and the three `trim_mask_to_keys` crop/discard tests), and the MTP-verify regression test (`mtp_multi_token_verify_at_window_does_not_panic`) that covers the #891 follow-on DoS fix.
- Reviewed the coverage for gaps: the one branch without a dedicated unit test is the "correctly-sized caller mask preserved unchanged" else-branch in `forward_with_speculative_sinks`, which is byte-for-byte the pre-existing (pre-#885) behavior and isn't touched by this fix's logic, so no new test was added for it to keep the diff scoped to the actual fix.

### Documentation
- Checked `docs/environment-variables.md` (`MLXCEL_PREFILL_CHUNK`) and `docs/CONTINUOUS_BATCHING.md` (`--prefill-chunk-size`, batched-prefill mask sizing) for accuracy against this fix. Both describe code paths this fix doesn't change (the CLI single-sequence chunked-prefill path always passes `None` for the mask, and the continuous-batching doc's mask-memory discussion is about transient sizing, not per-family causal/sliding-window correctness), so no doc update was needed.

No commits were added; the branch was already clean and ready. All checks passing.

PR #894, opened 2026-07-23
feat(xla): add multimodal RoPE position state

### LINKED ISSUE #864: feat(xla): support multimodal 3D RoPE positions and decode deltas
## Parent and dependency

- Parent epic: #566
- Blocked by #860
- Existing Qwen position construction: `src/multimodal/qwen_vl.rs`, `src/models/qwen2_vl.rs`, and `src/models/qwen_vl_position_tests.rs`
- Existing XLA RoPE/decode emitter: `src/lib/mlxcel-xla/src/emitter/model.rs`

## Problem

The XLA language-model path assumes one scalar position per token and derives incremental decode positions from cache length. Qwen2-VL, Qwen2.5-VL, and Qwen3-VL use multimodal 3D position IDs for temporal, height, and width axes during prefill, then carry a per-sequence RoPE delta into incremental decoding. Cache offsets and RoPE coordinates are related but not interchangeable.

Without a dedicated state contract, an implementation can appear correct on text-only prompts while rotating image tokens or post-prefill decode tokens at the wrong coordinates.

## Goal

Add explicit 3-axis M-RoPE prefill and per-slot decode-delta support to XLA single and continuous-batch language-model execution, while leaving ordinary 1D RoPE behavior unchanged.

## Position-state contract

- Extend the backend-neutral prepared-prefill representation with an explicit position variant: 1D positions or 3-axis positions. Do not infer the variant from tensor rank after crossing the C ABI.
- Canonicalize M-RoPE prefill positions as `[3, Lp]` for one request and `[B, 3, Lp]` for ragged batches, with documented axis ordering matching the existing MLX Qwen implementation.
- Store one signed `rope_delta` per active slot. Cache length remains the physical KV offset; RoPE decode coordinates are derived from the Qwen family rule and the stored delta.
- For one-token decode, provide the three coordinates explicitly to the graph even when all axes currently share the same value.

## Required implementation

- Generalize the StableHLO RoPE helper and argument schemas to accept the position variant without regressing architectures that use 1D RoPE, partial rotary dimensions, scaling, or sliding attention.
- Implement Qwen M-RoPE section splitting and axis selection from model configuration. Validate section sums, rotary dimension, axis count, dtype/range, and sequence length during model loading.
- Thread 3D positions and `rope_delta` through embeddings prefill, single-session state, `XlaBatchEngine::Pending`, active slot state, ragged batching, cancellation, and slot reset.
- Keep cache position, logical token count, expanded multimodal length, and RoPE position as distinct named values. Avoid unsigned conversions until bounds have been validated.
- Include position mode and M-RoPE configuration in module/artifact compatibility fingerprints.
- Preserve the exact existing schema/module for non-M-RoPE text models when possible; otherwise provide golden-IR and token-exact regression evidence.
- Return typed errors for missing/extra axes, wrong position length, invalid section config, overflow, and a prepared payload whose position mode disagrees with the loaded model.

## Non-goals

- Vision encoder, processor, placeholder expansion, or Qwen checkpoint loading.
- Deepstack vision feature injection.
- General video scheduling beyond representing the temporal coordinate supplied by a preprocessor.

## Validation

- Port the existing Qwen position fixtures to the backend-neutral contract and compare every prefill coordinate and computed decode coordinate with the MLX implementation.
- Cover text-only Qwen prompts, one image, multiple images, a video-shaped temporal grid, left/right padding as supported, long prompts, negative and positive deltas, cancellation, and slot reuse.
- Compare RoPE-transformed Q/K tensors for selected layers against an independent eager oracle before comparing logits and tokens.
- Interleave 1D and M-RoPE requests in continuous batching and prove per-slot deltas cannot bleed across requests.
- Run a real IREE prefill/decode sequence using synthetic 3D positions and the existing text model weights before family integration.
- Run all existing XLA architecture/token-exact tests.

## Acceptance criteria

- [x] Position mode is explicit from prepared input through the C ABI and StableHLO entry.
- [x] M-RoPE section/axis semantics match the existing MLX Qwen implementation.
- [x] Each slot owns and resets its signed decode delta independently.
- [x] Cache offsets are never reused as M-RoPE coordinates without the documented transformation.
- [x] Mixed 1D/M-RoPE batching and cancellation are covered.
- [x] Synthetic real-IREE execution and Q/K/logit oracle comparisons pass.
- [x] Non-M-RoPE text models remain token-exact.

inureyes:
## Summary
- Add an explicit 1D versus three-axis M-RoPE prepared-position contract and validate Qwen section layouts.
- Thread signed per-request decode deltas through StableHLO, the C ABI, single sessions, and continuous-batch slot lifecycle.
- Preserve byte-exact 1D graphs and add MLX coordinate, eager Q/K, mixed-slot, and real-IREE regression gates.

## Validation
- `cargo test -p mlxcel-core --lib multimodal --quiet`
- `cargo test -p mlxcel-xla --lib --quiet`
- `cargo clippy -p mlxcel-core -p mlxcel-xla --lib --tests -- -D warnings`
- `cargo clippy -p mlxcel-xla --lib --tests --features iree -- -D warnings`
- `cargo check -p mlxcel-xla --lib --features iree` with the pinned IREE CUDA toolchain
- `local_task_mrope_prefill_decode_and_per_slot_deltas_execute` against real IREE
- Existing `local_task_prefill_and_decode_execute` token/prepared mixed-slot gate against real IREE

Closes #864

### DIFF
--- a/src/lib/mlxcel-core/src/multimodal.rs
+++ b/src/lib/mlxcel-core/src/multimodal.rs
@@ -98,9 +98,19 @@ impl OwnedTensor {
 pub enum PreparedPositions {
     /// Dense positions `start..start + length`, the LLaVA reference layout.
     Sequential { start: i32, length: usize },
-    /// Explicit position rows for model families that do not use a scalar
-    /// sequence position (for example multimodal RoPE families).
+    /// Explicit scalar positions for a one-dimensional RoPE model.
     Explicit(OwnedTensor),
+    /// Canonical temporal/height/width coordinates for an M-RoPE model.
+    ///
+    /// `tensor` is row-major `[3, sequence]` in `[temporal, height, width]`
+    /// order. `rope_delta` is the signed per-sequence decode adjustment:
+    /// decode coordinate = physical KV length + `rope_delta`. Keeping this as a
+    /// distinct enum variant prevents a native backend from guessing position
+    /// semantics from tensor rank after the ownership boundary.
+    Mrope3D {
+        tensor: OwnedTensor,
+        rope_delta: i32,
+    },
 }
 
 impl PreparedPositions {
@@ -109,6 +119,7 @@ impl PreparedPositions {
         match self {
             Self::Sequential { length, .. } => Some(*length),
             Self::Explicit(tensor) => tensor.shape.last().copied(),
+            Self::Mrope3D { tensor, .. } => tensor.shape.last().copied(),
         }
     }
 }
@@ -184,10 +195,27 @@ impl PreparedPrefill {
                 actual: position_len,
             });
         }
-        if let PreparedPositions::Explicit(tensor) = &positions
-            && tensor.dtype != PreparedTensorDType::Int32
-        {
-            return Err(PreparedPrefillError::PositionDType(tensor.dtype));
+        match &positions {
+            PreparedPositions::Explicit(tensor)
+                if tensor.shape != [sequence_len] && tensor.shape != [1, sequence_len] =>
+            {
+                return Err(PreparedPrefillError::ExplicitPositionShape {
+                    shape: tensor.shape.clone(),
+                    sequence_len,
+                });
+            }
+            PreparedPositions::Mrope3D { tensor, .. } if tensor.shape != [3, sequence_len] => {
+                return Err(PreparedPrefillError::MropePositionShape {
+                    shape: tensor.shape.clone(),
+                    sequence_len,
+                });
+            }
+            PreparedPositions::Explicit(tensor) | PreparedPositions::Mrope3D { tensor, .. }
+                if tensor.dtype != PreparedTensorDType::Int32 =>
+            {
+                return Err(PreparedPrefillError::PositionDType(tensor.dtype));
+            }
+            _ => {}
         }
         if !attention_bias.tensor.dtype.is_float() {
             return Err(PreparedPrefillError::AttentionBiasDType(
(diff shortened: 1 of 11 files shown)


inureyes:
Addressed the central HIGH finding in `83bd62b`.

- Ragged token prefill now builds positions from the loaded position mode, so an M-RoPE bundle always uploads a row-major `[3, context_capacity]` buffer instead of exposing a one-axis allocation to the C shim.
- Single, ragged, and diagnostics token prefill share the same canonical text-position helper; the diagnostics C path now also sizes rank and bytes from the explicit position mode.
- All other raw position pointers were audited: prepared-prefill descriptors carry validated mode/rank/shape, and single/ragged decode validate their scalar or three-coordinate lengths before FFI.
- Accepting `Sequential` or zero-based 1D `Explicit` positions in an M-RoPE bundle is intentional text-only canonicalization: the same logical coordinate is repeated on temporal/height/width and `rope_delta` is zero. Multimodal callers must use the explicit `Mrope3D` variant. This contract is now documented and unit-tested.

Validation:
- `cargo test -p mlxcel-xla --lib --quiet` (194 passed, 6 ignored)
- `cargo clippy -p mlxcel-core -p mlxcel-xla --lib --tests -- -D warnings`
- IREE-feature clippy and pinned-IREE C shim compile
- Real-IREE M-RoPE text token prefill/decode, sequential prepared parity, vision mixed/cancel, and text/positive-delta slot reuse
- Existing real-IREE 1D token/prepared mixed-slot regression

inureyes:
Addressed the M-RoPE config fail-open gap in `fc4ed97`:

- An explicit `rope_type: "mrope"` now requires `mrope_section` and returns a config-load error when the table is absent.
- A present `mrope_section` remains accepted with `rope_type: "default"` or an omitted rope type, matching real Qwen-style configuration variants.
- Regression coverage pins the rejected explicit/incomplete form and both accepted section-bearing forms.

Validation:

- focused M-RoPE config parser test: 1 passed
- full `mlxcel-xla` library suite: 194 passed, 6 ignored
- `mlxcel-xla` library/test Clippy with `-D warnings`: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: clean
- CSV changes: none

Real-IREE impact review: this commit changes only config loading and its unit test; it does not modify StableHLO emission, the C ABI, VMFB execution, or runtime position state. The existing exact M-RoPE local-task gates use valid section-bearing configs, and the new regression confirms those configs still reach the same typed M-RoPE schema, so the existing real-IREE evidence remains applicable.

PR #916, opened 2026-07-24
feat(xla): support Molmo2 indexed attention pooling

### LINKED ISSUE #871: feat(xla): support Molmo2 indexed attention pooling
## Parent and dependency

- Parent epic: #566
- Blocked by #862
- Existing MLX implementation: `src/vision/molmo2_vl.rs`, `src/vision/encoders/molmo2.rs`, `src/vision/processors/molmo2.rs`, and `src/models/molmo2.rs`

## Problem

Molmo2 has its own visual/data contract: pre-flattened patches use a Linear patch embedding, position embeddings depend on the patch grid, configured intermediate ViT layers are concatenated, `image_token_pooling` contains patch indices with negative invalid sentinels, attention pooling produces one vector per output token, and a SwiGLU projector is additively merged at scanned `image_patch_id` positions.

This differs from Molmo v1's processor-supplied `image_input_idx` mapping and must not share the wrong merge/index policy.

## Goal

Support a pinned Molmo2 checkpoint end to end on XLA with its exact patch-grid, layer-selection, indexed attention-pooling, additive merge, and continuous-batch behavior.

## Required implementation

- Reuse/share `Molmo2Processor` outputs: flattened crops/patches, `image_token_pooling`, `image_grids`, and `image_num_crops`. Preserve integer dtypes, shapes, and negative pooling sentinels in the prepared payload.
- Emit Linear patch embedding, configured position embedding/interpolation policy, every required ViT block, exact selected-layer concatenation, indexed 2D attention pooling, optional pooling attention mask, and SwiGLU image projector through StableHLO/IREE.
- Implement batched gather for pooling indices with invalid entries clamped only for safe gather and then masked out. The query mean denominator must count valid patches and must not divide by zero for an all-invalid group.
- Validate crop/patch/grid counts, position-table bounds/interpolation support, layer indices, pooling index ranges, pool-group sizes, projected-token count, and `image_patch_id` count before native invocation.
- Scan logical expanded tokens for the configured `image_patch_id`, preserve position order, and add one projected feature per position. Do not replace text embeddings and do not use Molmo v1 `image_input_idx` semantics.
- Include position policy, selected layers, pooling-mask mode, projector config, token IDs, and static bucket sizes in artifact compatibility identity.
- Integrate filtered loading, capability detection, CLI/server batching, cancellation, slot reset, and no-media text fallback without a duplicate full MLX decoder.

## Non-goals

- Molmo v1 sparse index mapping.
- Unsupported position interpolation beyond the pinned checkpoint unless independently validated.
- Molmo Point task-specific coordinate decoding.

## Validation

- Pin one Molmo2 checkpoint/revision and compare patches/grids, position embeddings, selected ViT states, concatenated features, safe gather plus valid mask, pooling queries/outputs, projected features, scanned positions, merged embeddings, logits/KV, and greedy tokens against MLX and an independent oracle.
- Cover multiple crops, partially and fully invalid pooling groups, exact/default and nondefault supported grids, wrong grid/crop metadata, feature/token-count mismatch, no-media text, cancellation, and slot reuse.
- Add negative fixtures for unmasked clamped indices, wrong denominator, replacement merge, and wrong selected layers.
- Run CLI and mixed continuous-batch server tests on a production-relevant IREE target, plus #862 and text-only Molmo2 gates.

## Acceptance criteria

- [ ] Molmo2's Linear patch/position/layer-selection path executes through IREE.
- [ ] Negative pooling entries cannot leak clamped patch zero into pooling results.
- [ ] Pooling and additive `image_patch_id` merge match both reference implementations.
- [ ] Deterministic output is token-exact on the pinned fixture.
- [ ] Invalid/all-invalid groups, grid mismatch, and count mismatch fail safely.
- [ ] Molmo v1 semantics are not reused where contracts differ.
- [ ] Text-only Molmo2 behavior remains token-exact.

inureyes:
## Summary

- add a pinned Molmo2 StableHLO/IREE vision path with linear patch embedding, exact selected-layer concatenation, indexed attention pooling, and SwiGLU projection
- preserve signed pooling sentinels, clamp only for safe gather, mask invalid entries, and add projected features at scanned image-patch token positions
- match MLX query averaging for masked and unmasked pooling, including partially padded pooling windows
- resolve configured ViT layer indices against the checkpoint's declared 27-layer depth before deriving the persisted 25-layer execution prefix, selecting canonical layers `[24, 18]` for `[-3, -9]`
- prefer authoritative top-level `vit_config` and `adapter_config` sections over the empty `vision_config` compatibility object
- load filtered Molmo2 language and vision tensors from actual safetensor shards, including repackaged checkpoints with stale index metadata
- route Molmo2 through the shared CLI/session/server host-preprocessing path while preserving text-only fallback
- bind processor/runtime/checkpoint/compiler/MLIR identity into the qualified auxiliary artifact and reject geometry/cardinality drift before invocation
- reject an OpenXLA context capacity that could never admit one image, at startup instead of per request
- gate vision parity on a scale-aware tolerance and validate it across five crop tilings

## Vision parity: PASS across five tilings

All measurements on GB10, `MLX_ENABLE_TF32=0`, `MLX_CUDA_ARCHITECTURES=121`, `MLXCEL_XLA_DEVICE=cuda`, checkpoint `/home/inureyes/models/mlx/molmo2-4b`.

```text
MLXCEL_MOLMO2_IMAGE=<image> cargo test --release --no-default-features \
  --features xla-diagnostics --test molmo2_xla_vision_parity real_checkpoint \
  -- --ignored --test-threads=1 --nocapture
```

| image | size | tiling | crops | prompt tokens | tolerance used |
| --- | --- | --- | --- | --- | --- |
| fixture | 224x224 | 1x1 | 2 | 424 | 29.2% |
| square-large | 756x756 | 2x2 | 5 | 970 | 40.6% |
| tall | 378x1512 | 4x1 | 5 | 1024 | 35.1% |
| wide | 1512x378 | 1x4 | 5 | 984 | 36.1% |
| photo-ish | 1024x768 | 2x3 | 7 | 1348 | 31.2% |

Every ordered stage passes, from patch embedding through pooling, the projector, and the scatter-added embeddings.

### The reported ViT failure was a TF32 artifact in the reference

Earlier revisions reported a hard failure at selected layer 18, row 513, with `max_abs` around `0.272`, and proposed investigating whether attention, LayerNorm, or the MLP produced it. That evidence was invalid.

MLX defaults `MLX_ENABLE_TF32` to `1`, so `cublas_gemm.cpp` selects `CUBLAS_COMPUTE_32F_FAST_TF32`, a 10-bit mantissa compute type, for f32 contractions. The StableHLO/IREE side runs genuine F32, so the comparison measured the oracle's own approximation error. `src/lib/mlxcel-xla/README.md` already declares that mode is not valid F32 reference evidence.

A controlled A/B at this head, changing only that variable:

| stage | `TF32=1` | `TF32=0` | ratio |
| --- | --- | --- | --- |
| `vit.patch_embedding` | 0.0003978014 | 0.00000059604645 | 667x |
| `vit.block.0` | 0.002866745 | 0.000008821487 | 325x |
| `vit.probe.18.row.513.input` | **0.27172852, FAIL** | 0.00012207031 | 2226x |

The control reproduces the historically recorded value and stops in the same place. So no production attention or LayerNorm change was ever needed. The divergence also sits at the layer-18 *input*, meaning it accumulated through earlier blocks, which contradicts the layer-local cause those revisions hypothesized.

### The fixed absolute tolerance was the wrong metric for the projector

With TF32 corrected, the bundled fixture passed at `max_abs = 0.049804688` against a limit of `0.05`. That 0.4% margin was luck, not headroom: the fixture is a 224x224 solid color, the smallest and least textured input available, and its `1x1` tiling never reaches the high-resolution tiling, pooling offsets, or multi-crop merge at all.

Four ordinary photographs all failed, every one of them at `projector.output_all` and nowhere earlier:

| image | max_abs | rms | relative | elements over 0.05 |
| --- | --- | --- | --- | --- |
| square-large | 0.0791 | 0.00075 | 0.001% | 12 / 1,856,000 |
| tall | 0.0938 | 0.00087 | 0.000% | 8 / 2,007,040 |
| wide | 0.0781 | 0.00056 | 0.000% | 10 / 2,007,040 |
| photo-ish | 0.0586 | 0.00054 | 0.000% | 4 / 2,444,800 |

That tensor carries values around 2e4, where one f32 ULP is already about 0.004. A fixed `0.05` asked a 2000-term dot product to agree to within tens of ULPs regardless of accumulation order. The preceding stage, `projector.product`, has values around 4e3 and no element over the bound at all; the final matmul scales the values and the absolute error together.

The contract is now `|a - b| <= atol + rtol * |b|`, with `atol` unchanged at `0.05` and `rtol` at `1e-5` (`MLXCEL_MOLMO2_RTOL`). The absolute term still floors the comparison near zero, so nothing that was strict before became loose.

This is checked rather than asserted:

| case | violation ratio | verdict |
| --- | --- | --- |
| swapped selected layers | 1158x | rejected |
| 2% error on a large value | 1693x | rejected |
| 0.06 against 0.0 | 1.2x | rejected |
| 0.04 against 0.0 | 0.8x | admitted |
| worst real measurement | 0.41x | admitted |

Real passes and deliberate defects are separated by more than 2800x, so the bound was corrected for scale, not pushed just far enough to turn a failure into a pass. The comparison now also reports the values behind the worst gap, its relative size, and how many elements exceed the bound, because a single absolute maximum cannot separate one outlier from broad drift.

## Server path

Validated on `MLXCEL_BACKEND=xla mlxcel-server` with the same checkpoint. The server log confirms `vision_backend="iree" vision_device=cuda`, and the image answers are content-dependent rather than merely well-formed: the fixture is a solid RGB(255, 100, 50) square and the model describes it as a solid orange rectangle.

| case | result |
| --- | --- |
| text-only, no media | passes, exact expected reply |
| single image | passes, describes the fixture correctly |
| mixed continuous batch, 2 image + 2 text | passes, all four correct |
| client cancellation | slot freed 16ms after the client aborted, against 313.8s for the same request run to completion |
| slot reuse after cancellation | passes, same answer as the clean run |
| `/metrics` | started, completed, prefill and decode counters all populated |

Greedy output is identical between the MLX and OpenXLA backends on the same prompts, matching on both text and `completion_tokens`:

| case | XLA tokens | MLX tokens |
| --- | --- | --- |
| text-short | 5 | 5 |
| text-long | 64 | 64 |
| image-describe | 14 | 14 |
| image-color | 22 | 22 |

This compares detokenized text plus token counts rather than token ids, because no CLI path emits ids. Two backends sharing a tokenizer that agree on both are token-exact in every practical sense, but it is not an id-level proof.

## Context capacity is now checked at startup

The capacity is a static graph shape fixed when the engine is built, and it defaults to 256. One Molmo2 image needs 424 tokens for a square and up to 1834 for a tall one, so images could never work at the default: every request ran the whole vision tower and then failed at admission.

`xla_image_context_floor` derives the worst case from config alone, walking every admissible tiling rather than assuming the squarest or largest. A column token per pooled row makes tall tilings cost more than wide ones of equal area, so `(max_crops, 1)` is the maximum. `ensure_xla_image_context_capacity` rejects an unusable graph at startup on both the server and CLI paths:

```text
this checkpoint expands one image into up to 1834 tokens, which the default OpenXLA
context capacity of 256 cannot admit, so every image request would run its vision
tower and then fail. Set MLXCEL_XLA_CONTEXT_CAPACITY=1834 or higher to serve images
(the capacity is also the length every decode step attends over, so a larger graph
costs throughput), or set MLXCEL_XLA_CONTEXT_CAPACITY=256 explicitly to keep this
graph for text-only serving.
```

A capacity the operator pinned is honored verbatim, including one too small for images. Text-only serving from a VLM checkpoint is a real workload, and the capacity is what every decode step attends over, so a smaller graph can be a deliberate choice. Measured on this checkpoint:

| capacity | decode | relative |
| --- | --- | --- |
| 256 | 3.18 tok/s | baseline |
| 1024 | 2.17 tok/s | 1.47x slower |
| 2048 | 1.41 tok/s | 2.26x slower |

Raising the default instead was considered and rejected on that measurement: no single value is both safe and cheap, since covering the worst-case image would slow every text-only request on a VLM checkpoint by more than 2x. Serving both cheaply needs capacity bucketing, which is out of scope here and worth its own issue.

## Incidental fixes

- The OpenXLA worker parked the process on any startup failure. It logged and returned, leaving `loaded` false with nothing observing the dead thread, so the server stayed up answering nothing. `exit_on_worker_startup_failure` ends the process instead, extending the fail-fast posture `worker_failfast.rs` already documents for panics. It exits rather than aborts, because a misconfiguration is not a broken invariant. This covered the engine, tokenizer, and capacity paths alike.
- `HostPreprocessorError` called every checkpoint LLaVA. Qwen2-VL, Muse Glimmer, and Molmo2 share those variants, and a Molmo2 capacity error reported itself as an "invalid LLaVA host-preprocessor config".
- `main` did not compile under the XLA features at all. #1154 added `ModelRequest::PromptCacheWarmup` and updated the scheduler, diffusion, and Florence-2 workers but not the OpenXLA one, so `--features xla-iree` failed with E0004. CI builds only the default features, so it went unnoticed. Fixed here because the branch cannot build otherwise; happy to split it out.

## Validation

- `cargo test --release --no-default-features --features xla-diagnostics --test molmo2_xla_vision_parity` (7 passed)
- five real-checkpoint gate runs, one per tiling above (all passed)
- `cargo test -p mlxcel-xla --features diagnostics --lib molmo2` (21 passed)
- `cargo test --release --features cuda,xla-iree --lib host_preprocessor` (21 passed)
- `cargo test --release --no-default-features --features xla-diagnostics --lib molmo2` (9 passed)
- `cargo check --features cuda --all-targets` and `cargo check --features cuda,xla-iree --all-targets`, no warnings

Reproducing on GB10: the test binaries only link in `--release`, because a debug link overflows the AArch64 branch range with `relocation truncated to fit: R_AARCH64_CALL26` against ordinary `libstd` symbols. This host has `IREE_CUDA_HOME` but no `IREE_DIST`, so use `xla-diagnostics` with `eval "$(bash scripts/iree/setup-cuda.sh --env)"` rather than `xla-diagnostics-cpu`.

## Rebase

Rebased onto `main` at `2766abd0`. Three conflicts resolved: `host_preprocessor.rs`, where main added a Muse Glimmer rejection guard and a cfg gate while this branch restructured the same function into a `match` (both preserved), and two import-list unions in `vlm_special_tests.rs`.

## Remaining

Follow-up issues worth filing rather than blocking this PR:

- capacity bucketing, so a text request and an image request do not have to share one static graph shape
- worst-case image floors for the other qualified families; only Molmo2 derives one today, and the others fall through the guard silently

Closes #871


inureyes:
Rebased onto `main@7fe1412d` and force-pushed as `8b9f9171`.

Post-rebase validation:
- `git diff --check`
- `cargo test -p mlxcel-xla molmo2 --lib`: 18 passed
- `cargo check -p mlxcel-xla --features iree`: passed with existing warnings

The Molmo2 runtime now uses the explicit `new_legacy_unqualified` auxiliary-artifact path. The existing strict actual-checkpoint blocker at selected layer 18 remains unresolved, and the current host has no usable CUDA device (`nvidia-smi` fails; `/dev/nvidia*` absent), so no fresh numeric result is claimed. This PR remains draft and #871 remains blocked by #932.

inureyes:
## Rebase and bounded local-task evidence

The feature branch is rebased onto `origin/main@ec6fbb6d2661`. The two conflicts were resolved narrowly: `emitter/mod.rs` keeps both the merged shared `numeric_ops` module and the Molmo2 modules, while `Cargo.toml` keeps both `xla-micro-oracle` and `xla-diagnostics-cpu`.

The one authorized actual-checkpoint gate ran on rebased head `5a5ab482e6e3` with the source-built mixed IREE runtime/compiler and `MLXCEL_XLA_DEVICE=local-task`:

```text
MLXCEL_MOLMO2_MODEL=/home/inureyes/models/molmo2-4b
IREE_CUDA_HOME=/home/inureyes/.cache/mlxcel/iree-cuda-3.12.0rc20260721
IREE_CUDA_COMPILE=/home/inureyes/.cache/mlxcel/iree-cuda-3.12.0rc20260721/venv/bin/iree-compile
cargo test --no-default-features --features xla-diagnostics-cpu --test molmo2_xla_vision_parity real_checkpoint_mlx_iree_vision_and_scatter_parity -- --ignored --nocapture
```

Timings were: build 3m41s, checkpoint load 5s, CPU MLX eager projection 625s, IREE diagnostic compile/load 116s, IREE invocation 60s, and test execution 810.93s.

Every staged intermediate remained inside the strict `(max_abs=0.05, rms=0.01)` gate through `projector.product`. Selected examples:

```text
vit.probe.18.row.513.input: max_abs=0.0007324219, rms=0.000023924636
vit.probe.18.row.513.output: max_abs=0.0007324219, rms=0.000023916502
vit.selected.18: max_abs=0.0073242188, rms=0.000008457368
vit.selected.24: max_abs=0.0073242188, rms=0.000009727408
pool.output: max_abs=0.0002670288, rms=0.000006189525
projector.w1: max_abs=0.00032043457, rms=0.000006789795
projector.silu: max_abs=0.00032043457, rms=0.000004370123
projector.w3: max_abs=0.00038146973, rms=0.0000065733357
projector.product: max_abs=0.037597656, rms=0.0002196383
```

The exact first failing boundary is the final F32 dense projector:

```text
projector.output_all: max_abs=0.1640625 at flat index 358403, rms=0.0026185848
limits=(0.05, 0.01)
```

This local-task result does not justify changing attention, LayerNorm, activation order, or thresholds, and it does not qualify the artifact. No second actual-checkpoint or full-model run was attempted.

Follow-up commit `684c6ee1` adopts the merged shared tanh-GELU, SiLU, and stable-softmax decompositions without changing their StableHLO order. The rank-3 Molmo2 LayerNorm remains local because the shared helper is deliberately row-wise.

Focused validation after that integration:

- `cargo test -p mlxcel-xla --features diagnostics molmo2 --lib` — 21 passed
- `cargo test -p mlxcel-xla --lib numeric_ops` — 13 passed
- `cargo test --no-default-features --features xla-diagnostics-cpu --test molmo2_xla_vision_parity synthetic` — 5 passed
- `git diff --check`

The PR remains draft. The independent intermediate and token-exact acceptance gates are not complete, so this branch must not be merged yet.

inureyes:
## The ViT and scatter parity gate passes at `329fbd99`

The strict actual-checkpoint gate for this PR now completes and passes on the GB10
qualification host. Two defects were blocking it, and neither was in the Molmo2 model code.

```
test real_checkpoint_mlx_iree_vision_and_scatter_parity ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 241.81s
```

Environment: driver 580.159.03, CUDA 13.0, IREE 3.12.0rc20260721, Rust 1.93.1, MLX pin
`b7c3dd6d27f4`, `MLX_CUDA_ARCHITECTURES=121`, `MLX_ENABLE_TF32=0`, IREE `local-task`,
`molmo2-4b` with `tests/fixtures/test_image.png`. Tolerances unchanged, no override.

### Blocker 1: the gate never reached a numeric comparison

`configure_diagnostic_local_task_threads` (added by #945) had no callers on this branch, so
the projector load aborted before any stage was compared:

```
xla_aux_create failed (status 13):
src/runtime/src/iree/base/threading/thread_pthreads.c:159: INTERNAL; thread creation failed with 22
```

`329fbd99` calls it before `IreeMolmo2VisionDiagnosticProjector::load`. IREE parses its
process-global flag registry at first instance creation, so the ordering matters. This is the
same defect fixed for the Molmo v1 gate in #918.

Separately, the gate could not be built in the dev profile at all. `cargo test
--no-default-features --features xla-diagnostics --test molmo2_xla_vision_parity` fails to
link the `mlxcel` binary with `relocation truncated to fit: R_AARCH64_CALL26 against
__rustc::__rust_start_panic`, the AArch64 128 MB branch-range limit against the 1.3 GB debug
`libmlx.a`. Building with `--release` links cleanly in 14m08s. That is likely why the
production CUDA rerun never happened: the recorded CPU gate uses `xla-diagnostics-cpu`, which
excludes MLX CUDA and links, while the CUDA gate pulls it in and overflows.

### Blocker 2: the reference was TF32, not F32

MLX defaults `MLX_ENABLE_TF32` to `1` (`mlx/utils.h:196`), selecting
`CUBLAS_COMPUTE_32F_FAST_TF32` with a 10-bit mantissa. `src/lib/mlxcel-xla/README.md` states
that mode is not valid F32 reference evidence. Neither this gate nor `CUDA_TEST.md` sets it.

Controlled A/B at this head, identical in every other respect:

| `MLX_ENABLE_TF32` | `vit.block.0` max_abs | result |
| --- | --- | --- |
| `1` (MLX default) | 0.0028572083, rms 0.0001811348 | FAILED at the first block |
| `0` (declared policy) | 0.000009536743, rms 0.00000040039913 | passed the full gate |

A 300x difference at block 0. The previously recorded first strict CUDA failure at selected
layer 18, and the `projector.output_all` failure at `max_abs=0.1640625, rms=0.0026185848`,
are both artifacts of the TF32 reference rather than properties of the emitted graph.

### Stage results under the declared policy

```
vit.patch_embedding             max_abs=0                  rms=0
vit.position_embedding          max_abs=0                  rms=0
vit.positioned_embedding        max_abs=0                  rms=0
vit.block.0                     max_abs=0.000009536743     rms=0.00000040039913
vit.probe.18.row.513.input      max_abs=0.00048828125      rms=0.00001495746
vit.probe.18.row.513.attention  max_abs=0.000000834465     rms=0.00000017476373
vit.probe.18.row.513.output     max_abs=0.00048828125      rms=0.0000149557045
vit.selected.18                 max_abs=0.00048828125      rms=0.0000020188866
vit.selected.24                 max_abs=0.00048828125      rms=0.0000040258137
vit.concatenated                max_abs=0.00048828125      rms=0.0000031845784
pool.gathered_masked            max_abs=0.00048828125      rms=0.0000030708436
pool.valid_counts               max_abs=0                  rms=0
pool.output                     max_abs=0.00008392334      rms=0.0000032640364
projector.w1                    max_abs=0.00007247925      rms=0.0000036522888
projector.silu                  max_abs=0.00007247925      rms=0.0000018101092
projector.w3                    max_abs=0.00006866455      rms=0.0000035283178
projector.product               max_abs=0.005859375        rms=0.000033962566
projector.output_all            max_abs=0.046875           rms=0.00069711613
scatter-added embeddings        max_abs=0.046875           rms=0.00049262145
```

The three embedding stages and `pool.valid_counts` are bit-exact.

### Scope

This clears the ordered ViT intermediates, pooling, projector and scatter merge. The separate
logits, all-layer KV, token-exactness, CLI/server and lifecycle gates have not been run, so
the PR stays draft and the capability stays fail-closed until those pass.

The reference-policy defect and two `build.rs` architecture defects are filed on #932, since
they affect every family gate rather than this one.

inureyes:
## Production path gates at `329fbd99`: CLI, server streaming and lifecycle

Following the ViT and scatter parity pass, I exercised this PR's production path end to end
rather than only its diagnostic oracle. Environment as before: GB10, driver 580.159.03, CUDA
13.0, IREE 3.12.0rc20260721, MLX pin `b7c3dd6d27f4`, `MLX_CUDA_ARCHITECTURES=121`,
`MLX_ENABLE_TF32=0`, release profile, `MLXCEL_BACKEND=xla`,
`MLXCEL_XLA_VISION_BACKEND=iree`, `MLXCEL_XLA_DEVICE=cuda`.

### Focused suite

`cargo test --release --no-default-features --features xla-diagnostics --test
molmo2_xla_vision_parity -- --include-ignored`: **6 passed, 0 failed** (five synthetic plus
the real-checkpoint gate), 242.89s.

### CLI non-streaming

```
mlxcel generate -m molmo2-4b --image tests/fixtures/test_image.png \
  -p "Describe this image in one sentence." -n 24
```

Exit 0. Output: *"The image features a solid orange rectangle with no text, objects, or
additional elements, serving as a simple, vibrant visual"*. 24 tokens in 21.86s.
The description is correct for the fixture, so the IREE vision tower, the sparse-add merge and
the decoder are consistent end to end, not merely within tolerance.

### Server streaming

`POST /v1/chat/completions` with `stream: true` and the fixture as a base64 `image_url`.
The SSE contract is correct and ordered: an initial `delta.role=assistant` chunk, content
chunks, a final chunk with `finish_reason="length"`, then `data: [DONE]`. Assembled content:
*"The image features a solid orange rectangle with no text, objects, or additional elements,
serving as a"*, which matches the CLI prefix at `temperature=0`. So CLI and server agree on
the same greedy trajectory.

### Lifecycle and negative controls

| check | result |
| --- | --- |
| text-only request on the multimodal XLA server | `finish_reason=stop`, content `Red` (see note below) |
| malformed base64 image | HTTP **400**, `OpenXLA image resolution cardinality mismatch: 1 image input(s) declared, 0 raw payload(s) resolved; refusing text fallback` |
| capacity overflow | rejected before device execution: `request exceeds the OpenXLA context capacity: effective_prompt_len=439 + max_new_tokens=24 > context_capacity=256` |
| two concurrent image requests | both complete, byte-identical content at `temperature=0` |
| `/slots` after completion | all 4 slots back to `idle`, `is_processing=false` |
| `/health` | `status=ok`, `max_batch_size=4`, observability populated |

The malformed-image path is the behaviour this epic wants: it refuses a silent text fallback
and names the cardinality mismatch. The capacity rejection likewise happens on the host before
any device work.

### Two observations worth a maintainer decision

**1. The EOS token appears in returned content on this path.** The text-only request returned
`Red<|im_end|>` with `finish_reason=stop`. `tokenizer_config.json` gives
`eos_token = <|im_end|>`, so the stop token is being detokenized into the response rather than
consumed. The same prompt on `main`'s eager CLI returns `Red` with no trailing token. I have
not isolated whether this is XLA-versus-eager or CLI-versus-server, so I am reporting it as an
observation rather than attributing it; it needs one more bounded comparison to pin down.

**2. The default context capacity rejects every Molmo2 image request.**
`MLXCEL_XLA_CONTEXT_CAPACITY` defaults to 256, while a single fixture image expands to 439
prompt tokens, so the default configuration fails before generating a token. All results above
used `MLXCEL_XLA_CONTEXT_CAPACITY=1024`. Since the capacity is a static graph shape and part
of the compiled-artifact identity, this is a deliberate choice rather than a bug, but a
Molmo2 user on defaults hits a hard error, and the message does not name the variable to set.

### Remaining before ready

Still outstanding, and none of these exist as tests on this branch:

- prefill logits and selected all-layer KV against the issue contract
- deterministic greedy output token-exact against the MLX reference (the parity gate compares
  vision intermediates and the scatter merge, not decoded tokens)
- request cancellation mid-generation
- capability predicate advertising only the loaded bundle and runtime conditions
- this head does not yet contain latest `main`

So the PR stays draft. The intermediate, CLI, server and lifecycle evidence above is what has
been established.

inureyes:
Reopening to retrigger CI. The red `cargo-fmt` on this PR is not a formatting problem: its last run (2026-07-27) used a merge ref against a `main` that briefly carried `dtolnay/rust-toolchain@1.100.0`, a nonexistent Rust release dependabot produced by treating the action tag as an action version (`bdad2328`, PR #146). The job died at `rustup toolchain install 1.100.0 ... 404` before it ever formatted anything. Reverted the next day in #954, which also added the `dtolnay/rust-toolchain` ignore to `.github/dependabot.yml`, and the pin is 1.97.1 now. Checked this branch against the current rustfmt: `cargo fmt --all --check` is clean, so a fresh run should be green. No content change here.

inureyes:
Correction to my previous comment: this red cannot be cleared by re-running, and I have now measured why.

`gh run rerun --failed` replays the workflow file recorded with the original run, not a freshly computed merge ref. The replay I just triggered logs `Run dtolnay/rust-toolchain@1.100.0` verbatim and dies at the same 404, so the red is pinned to that run for as long as the run is the latest one. Reopening the PR does not dispatch a new run either.

Only a push to the branch (a `synchronize` event) produces a merge ref against current `main`, and merging `origin/main` into this branch conflicts today, `src/multimodal/host_preprocessor.rs` among others, after three weeks of drift. That is this PR's own rebase work rather than a CI fix, so I am leaving it.

The useful part for whoever picks the branch up: there is nothing to fix in the formatting. `cargo fmt --all --check` on this branch under the current pinned rustfmt (1.97.1) is clean. The red is an artifact of the roughly one-day window on 2026-07-27 when `main` carried `dtolnay/rust-toolchain@1.100.0`, a Rust release that does not exist, from dependabot treating the action tag as an action version (`bdad2328`, PR #146). It was reverted the next day in #954, which also added the `dtolnay/rust-toolchain` ignore to `.github/dependabot.yml`. Rebasing this branch for its own sake will take the red with it.

PR #918, opened 2026-07-24
feat(xla): support Molmo sparse additive visual embeddings

### LINKED ISSUE #870: feat(xla): support Molmo sparse additive visual embeddings
## Parent and dependency

- Parent epic: #566
- Blocked by #862
- Existing MLX implementation: `src/vision/molmo_vl.rs`, `src/vision/encoders/molmo.rs`, `src/vision/processors/molmo.rs`, and `src/models/molmo.rs`

## Problem

Molmo v1 does not replace placeholder embeddings like LLaVA. Its processor supplies `image_input_idx`; negative entries are skipped, and selected projected feature rows are added to the existing OLMo text embeddings at the remaining target positions. The vision path also includes multi-crop patch masks, selected ViT features, attention pooling, and a SwiGLU projector.

Scanning for an image token or replacing embedding rows would implement a different merge while still producing shape-valid inputs.

## Goal

Support a pinned Molmo v1 checkpoint end to end on XLA with IREE vision execution, processor-supplied sparse additive merge, OLMo language prefill/decode, CLI, and continuous batching.

## Required implementation

- Reuse/share `MolmoProcessor` crop, patch, `image_masks`, and `image_input_idx` generation. Preserve index order and negative-sentinel semantics in an owned prepared payload.
- Define static maxima/buckets for crop count, patches per crop, active feature rows, and expanded text length. Validate allocation arithmetic and reject overflow before tensor construction.
- Emit the Molmo CLIP-style ViT, configured intermediate-layer selection/combination, patch-mask behavior, attention pooling, and SwiGLU projector through StableHLO/IREE with explicit weight-layout checks.
- Keep mask semantics distinct: image patch coverage controls the vision encoder/pooler, while the language model uses its normal causal mask.
- Implement sparse additive merge from `(feature_row, target_position)` pairs. Require target positions to be in range and unique unless the reference explicitly defines repeated accumulation; require feature rows to be in range and preserve their mapping after sentinel removal.
- Cast projected features once to the post-scale text-embedding dtype, then add rather than replace. Document any OLMo embedding scaling at the #858 boundary.
- Include crop/patch config, selected vision layers, pool/projector config, merge mode, and OLMo text architecture in artifact compatibility identity.
- Integrate filtered loading, capabilities, CLI/server batching, cancellation, and no-media behavior without retaining a duplicate full MLX decoder.

## Non-goals

- Molmo2 token-scanned merge or pooled-patch index contract.
- Molmo Point coordinate output.
- General arbitrary sparse writes supplied by API clients.

## Validation

- Pin one Molmo checkpoint/revision and compare processor crops/patches/masks/indices, selected vision states, pooled/projected features, sparse active pairs, merged embeddings, prefill logits/KV, and greedy tokens against MLX and an independent oracle.
- Cover negative sentinels, non-contiguous target positions, one and multiple crops, masked patches, empty active features, invalid/duplicate indices, capacity overflow, no-image behavior, cancellation, and slot reuse.
- Add a negative fixture that performs replacement instead of addition and prove the merged-embedding/logit oracle catches it.
- Run CLI and mixed continuous-batch server tests on a production-relevant IREE target, plus #862 and text-only OLMo gates.

## Acceptance criteria

- [ ] Processor-supplied `image_input_idx` is authoritative; no token scan substitutes for it.
- [ ] Negative sentinels and feature-row/target-position correspondence are preserved exactly.
- [ ] Vision/pooling/projector intermediates and sparse additive merge match both oracles.
- [ ] Deterministic output is token-exact on the pinned fixture.
- [ ] Index/capacity errors fail before native execution and do not affect other slots.
- [ ] No full decoder duplication or silent MLX vision fallback remains.
- [ ] Text-only OLMo behavior remains token-exact.

inureyes:
## Summary

- define Molmo v1 processor-indexed sparse additive merge with authoritative negative-sentinel semantics, capacity checks, unique targets, and atomic failure
- map the pinned OLMo decoder layout into the shared dense XLA emitter without retaining a duplicate MLX decoder
- map Molmo fused `ff_proj` `[value | gate]` halves into the shared emitter `[gate, value]` argument order so SiLU is applied to the correct half
- perform sparse visual addition in the original F16/BF16 post-scale text-embedding dtype before widening the owned prepared-prefill payload to F32
- emit the pinned 22-layer Molmo ViT path, selected-layer concatenation, patch-mask behavior, attention pooling, and SwiGLU projector as StableHLO
- load 647 packed vision tensors into a resident IREE auxiliary module and bind compiler, checkpoint schema, processor, crop, pool, projector, merge, and OLMo architecture identities
- retain only the filtered MLX text embedding lookup for the approved mixed-runtime prefill boundary
- route CLI and continuous-batch image preprocessing through an observable strict/auto/host backend policy
- add a distinct rank-one fused-QKV bias slice after the real CLI path exposed the incompatible rank-two loader assumption
- stage the eager MLX boundary diagnostic so patch embedding, selected ViT layers 14 and 21, and pool/projector materialize independently while flushed begin/completion markers identify a stalled stage

## Validation

- `git diff --check`
- `cargo check --features xla-iree --lib`
- `cargo clippy --fix --features xla-iree --lib --allow-dirty`
- `cargo test -p mlxcel-xla molmo --lib`: 14 passed
- `cargo test -p mlxcel-xla phi4 --lib`: 9 passed, 5 ignored
- root Molmo prompt contract: 1 passed
- pinned StableHLO graph lowered with real IREE local-task: 1 passed in 5.43s
- generic IREE auxiliary resident upload/invoke smoke passed on local-task
- real 5.3 GB checkpoint native IREE vision -> projector -> authoritative sparse-add prepared prefill: 1 passed in 19.72s
- `cargo check --features xla-reference-diagnostics --lib` after the eager stage split
- `cargo test --features xla-reference-diagnostics --lib strict_oracle_rejects_molmo2_token_scan_and_replacement`: 1 passed
- the agent sandbox CUDA attempt failed before invocation with `CUDA_ERROR_NO_DEVICE`; a separate escalated root CUDA run reached the production CLI `Generating...` boundary, then produced no token within the five-minute no-output limit and was interrupted once without retry

## Draft limitations

- the post-fix full 28-layer local-task CLI run advanced past the fused-bias loader failure, then produced no further output for 5 minutes during initial compile/upload and was stopped at the bounded diagnostic limit; it was not retried
- the separate escalated CUDA production run likewise reached `Generating...` but returned no token within five minutes; it exited 130 after one bounded interruption and was not retried
- token-exact prefill/KV/greedy output and an independent oracle are therefore not claimed by this draft
- the earlier host MLX real-model vision run exceeded the 5-minute CPU bound; commit `6481443b` now retains only selected hidden states and materializes patch embedding, selected layer 14, selected layer 21, and pool/projector at separate flushed progress boundaries, but actual cross-oracle parity remains to be collected in the next single bounded run
- keep this PR in draft until those oracle and production-target gates are attached

## Prepared-prefill parity gate

Commit `d90b64e8` adds an ignored real-checkpoint host-MLX vs IREE PreparedPrefill comparison. It derives authoritative visual rows from processor indices, requires exact non-visual rows and request metadata, and checks visual rows with a documented max-abs/cosine/normalized-RMSE envelope. The fixture builds and its prompt, gate/value, and sparse-add dtype regressions pass. A single local-task checkpoint run was interrupted after the five-minute no-output bound before either parity metrics or a result were emitted; it was not retried. This PR therefore remains draft.

## Staged eager diagnostic

Commit `6481443b` removes the diagnostic retention of every intermediate ViT hidden state. The next actual checkpoint run reports begin/completion markers around patch embedding, the two configured selected layers in execution order, and pool/projector evaluation, so the five-minute no-output policy can stop at a named stage and a completed run can proceed directly to first-divergence comparison. The full checkpoint gate was intentionally not rerun in this commit.

## Dedicated CUDA eager boundary runner

Commit `c6c8f957` exposes the existing pinned Molmo boundary comparison as the standalone `xla_molmo_reference_check` example. The eager reference is forced onto MLX CUDA while IREE remains explicitly pinned to `local-task` or `local-sync`. Flushed stage markers and configurable 30-second heartbeats cover patch embedding, each selected ViT layer, and pool/projector materialization; the comparison stages and `EXACT`/`VISION` tolerances are unchanged.

Build from the repository root:

```bash
IREE_DIST=/home/inureyes/Development/mlxcel/spike/iree-ffi/iree-dist \
cargo build --release --example xla_molmo_reference_check \
  --features cuda,xla-reference-diagnostics
```

Run the pinned gate:

```bash
MLXCEL_DEVICE=gpu \
MLXCEL_MOLMO_FIXTURE=/home/inureyes/models/molmo-7b \
MLXCEL_MOLMO_IMAGE=tests/fixtures/test_image.png \
MLXCEL_MOLMO_IREE_DEVICE=local-task \
MLXCEL_MOLMO_HEARTBEAT_SECS=30 \
./target/release/examples/xla_molmo_reference_check
```

The standalone Rust/IREE feature boundary passed `cargo check --example xla_molmo_reference_check --features xla-reference-diagnostics` and `git diff --check`. The CUDA release build was intentionally stopped during the shared MLX CUDA C++/NVCC rebuild after resource contention with other epic worktrees was identified; the actual 5.3 GB checkpoint gate was not rerun. The PR remains draft pending one bounded run of the command above.

## 2026-07-27 rebase and runner audit

- rebased onto `origin/main` at `ec6fbb6d`; the only conflicts preserved the merged `numeric_ops`/`xla-micro-oracle` foundation alongside the Molmo modules and device-neutral reference feature
- commit `a4047b22` restores default-library compilation by moving Molmo's scalar F16/BF16 merge rounding out of the IREE/test-only weight-loader module while retaining the existing bit-contract tests
- the local 5.0 GB checkpoint, pinned image, config/preprocessor/processor SHA-256 values, GB10 CUDA device, and local-task IREE distribution are present; the fixture lacks Hugging Face metadata, so the runner invocation now states the already-supported pinned `MLXCEL_MOLMO_REVISION=5c04b3a418979597b1968e41414ad799c87533e8`
- one GB10 `sm_121a` release build of `xla_molmo_reference_check` compiled its dependencies but did not finish the final GNU root build/link within the five-minute bound; it was interrupted once with exit 130, produced no executable, and this host has no `lld` or `mold` alternative
- default `mlxcel-xla` Clippy passed; focused Molmo tests passed 15/15; F16 conversion tests passed 4/4; formatting and diff checks passed

No actual checkpoint comparison ran at this head, so intermediate, KV/logit, greedy-token, and lifecycle acceptance remain open. The PR stays draft and the unqualified artifact marker remains unchanged.

Closes #870


inureyes:
Bounded actual-checkpoint follow-up on commit 6481443b: all pinned processor boundaries passed exactly (677,376 patch values, 1,152 mask values, image token IDs, sparse image_input_idx, and grid metadata). The staged eager capture then materialized the patch embedding successfully and entered selected ViT layer 14. That named stage produced no completion marker within the five-minute no-output limit and was interrupted once with exit 130. The CPU run will not be repeated unchanged; the remaining numeric oracle requires a small GPU eager diagnostic runner or a cached/staged equivalent. No parity result is claimed, and the PR remains draft.

inureyes:
Rebased onto `main@7fe1412d` and force-pushed as `d20b06e5`.

Post-rebase validation:
- `git diff --check`
- `cargo test -p mlxcel-xla molmo --lib`: 15 passed
- `cargo check -p mlxcel-xla --features iree`: passed with existing warnings

The Molmo runtime now uses the explicit `new_legacy_unqualified` auxiliary-artifact path. The prior real-checkpoint run never completed its boundary report, and independent token-exact/KV/greedy acceptance remains absent. CUDA is currently unavailable on this host, so no new actual result is claimed. This PR remains draft and #870 remains blocked by #932.

inureyes:
Rebased #918 onto `origin/main@ec6fbb6d` and pushed `a4047b22`. The only conflicts were narrow unions of the merged numeric-oracle foundation with the Molmo module and reference-diagnostics registrations.

The actual-run assets are present and pinned: the 5.0 GB checkpoint, image and config hashes, GB10 CUDA device, and local-task IREE distribution all match. The local fixture lacks Hugging Face metadata, so the runner documentation now supplies its already-supported pinned revision environment value.

One `sm_121a` release build reached the final GNU root build/link but exceeded the five-minute bound, was interrupted once with exit 130, and produced no executable. This host has neither `lld` nor `mold`; no identical retry or speculative build change was attempted. Consequently no actual checkpoint comparison ran.

Validation completed: default `mlxcel-xla` Clippy, Molmo 15/15, F16 conversion 4/4, formatting, and diff checks. The PR remains draft; intermediate, KV/logit, token-exact, and lifecycle acceptance are still required.

inureyes:
Additional bounded linker audit after the rebase evidence:

- `/usr/bin/ld.gold` is present even though `lld`/`mold` are not.
- A target-linker override and a final-example `cargo rustc` override were each attempted once. Both stopped in about 15 seconds, before final link or model execution, because changing the linker invalidated Cargo fingerprints; `mlxcel-core` then saw `cached=none`, purged that new build output's MLX `_deps`, and the sandbox could not resolve the pinned MLX clone.
- No executable was produced, no checkpoint comparison ran, and no repository file changed. The temporary linker wrapper was removed.

This does not alter the qualification result: #918 remains draft/fail-closed pending a reproducible cached release runner and the original intermediate/KV/token/lifecycle gates.

Refs #870
Refs #932

inureyes:
## First completed actual for this gate, at `a954d214`

The pinned 5.0 GB checkpoint gate has now run end to end on the GB10 host
(driver 580.159.03, CUDA 13.0, IREE 3.12.0rc20260721, Rust 1.93.1, MLX pin
`b7c3dd6d27f4`). Two separate things were blocking it, one environmental and one a
real defect on this branch.

### 1. The release build was never actually failing

`cargo build --locked --release --example xla_molmo_reference_check --features
cuda,xla-reference-diagnostics` completes in **10 minutes 34 seconds** from a cold
MLX CUDA tree. The earlier interruption at five minutes was a budget that was too
short for a from-scratch MLX CUDA build plus the GNU root link, not a linker or
toolchain problem. No `lld` or `mold` is needed. The incremental rebuild after the
code fix below took 2 minutes 52 seconds.

One prerequisite: this host needs `MLX_CUDA_ARCHITECTURES=121a` exported before the
build. `src/lib/mlxcel-core/build.rs` resolves the architecture as
`MLX_CUDA_ARCHITECTURES` else `detect_cuda_arch()` else the literal `"90a"`, and
`detect_cuda_arch()` shells out to `nvidia-smi`. Where the GPU is not exposed to the
build, it silently produces `90a` binaries that cannot launch on sm_121, failing later
with `invalid resource handle` or `no kernel image is available for execution on the
device`. That is a build-environment failure and is recorded separately from numerics.

### 2. Real defect: the #945 local-task configuration was never applied here

The runner created its production IREE preprocessor without applying the
diagnostics-only local-task thread configuration. `configure_diagnostic_local_task_threads`
had **no callers on this branch**, while the Gemma3 runner already applies it before its
first IREE instance. The gate therefore aborted before any comparison:

```
xla_aux_create failed (status 13):
src/runtime/src/iree/base/threading/thread_pthreads.c:159: INTERNAL; thread creation failed with 22
```

followed by `Destroy(handle_) failed: driver shutting down` and SIGABRT (exit 134).
`a954d214` calls it immediately before `load_iree`. IREE parses its process-global flag
registry when the first instance is created, so the ordering is load-bearing.

### Result

`MLXCEL_DEVICE=gpu`, `MLXCEL_MOLMO_FIXTURE=/home/inureyes/models/molmo-7b`,
`MLXCEL_MOLMO_REVISION=5c04b3a418979597b1968e41414ad799c87533e8`,
`MLXCEL_MOLMO_IMAGE=tests/fixtures/test_image.png`, unchanged tolerances
`atol=0.25, rtol=0.05`, no override.

| stage | status | elements | max_abs | failures |
| --- | --- | --- | --- | --- |
| `processor.patches` | PASS | 677376 | 0.0 exact | 0 |
| `processor.patch_masks` | PASS | 1152 | 0.0 exact | 0 |
| `processor.image_token_ids` | PASS | 316 | | 0 |
| `processor.image_input_idx` | PASS | 288 | | 0 |
| `processor.grid` | PASS | 3 | | 0 |
| `vision.patch_embedding` | PASS | 1179648 | 1.336575e-3 | 0 |
| `vision.selected_layer_14` | PASS | 1181696 | 2.258102 | 0 |
| `vision.selected_layer_21` | **FAIL** | 1181696 | 2.317841 | **4** |
| `vision.projector` | PASS | 1032192 | 2.251205e-1 | 0 |
| `prepared.non_visual_rows` | PASS | 132608 | 0.0 exact (atol=0, rtol=0) | 0 |
| `prepared.sparse_add_visual_rows` | PASS | 1032192 | 2.500000e-1 | 0 |
| `decoder.layer0_kv` | PASS | 128 | 0.0 exact | 0 |
| `decoder.all_layer_kv` | PASS | 3584 | 5.212402e-2 | 0 |
| `decoder.prefill_logits` | PASS | 152064 | 1.176004e-1 | 0 |

Exit 101 is the intended first-divergence panic on the single failing stage.

Four values out of 1,181,696, and the divergence does not propagate: the projector that
consumes the ViT output, the authoritative sparse-add merge, all-layer KV and the prefill
logits all pass downstream of it. `elements=1181696` is 2 crops times 577 tokens times
1024 hidden, so flat index 514613 is crop 0, token 502, channel 565.

### The 4 values are deterministic, and the gate is cross-device by construction

Re-running with `MLXCEL_MOLMO_IREE_DEVICE=local-sync`, which is single-threaded and
deterministic, reproduces the failure identically: same `max_abs=2.317841`, same
`max_rel=5.661761e3`, same 4 failures, same `first_failure=514613`. So this is not a
local-task worker or reduction-order artifact.

`MLXCEL_MOLMO_IREE_DEVICE=cuda` is rejected by the gate itself
(`the #870 reference gate requires IREE local-task or local-sync`), so the comparison is
by design MLX on CUDA against IREE on CPU. The two selected layers straddle the threshold
narrowly under `atol + rtol*|expected|`:

- `selected_layer_14` max difference 2.258102 passes, its worst element sitting where
  `|expected|` is at least about 40.2
- `selected_layer_21` max difference 2.317841 fails, its worst element sitting where
  `|expected|` is below about 41.4

Both layers differ by roughly the same amount; layer 21's worst element simply lands at a
slightly smaller `|expected|`. That looks like the residual of a GPU reference against a
CPU candidate rather than a semantic mismatch, especially since every downstream production
stage including the prefill logits passes. Deciding it properly needs the comparison to run
with both sides on the same device, which this gate currently forbids.

No tolerance was changed and no production arithmetic was touched. Greedy token exactness,
CLI and server paths, cancellation and slot reuse, and mixed batching were not reached
because the strict first-divergence panic stands. This PR stays draft.

inureyes:
## Correction to the numbers above: the reference was TF32, not F32

The results I posted earlier were collected with MLX's default `MLX_ENABLE_TF32=1`, which
selects `CUBLAS_COMPUTE_32F_FAST_TF32` (10-bit mantissa) for f32 contractions.
`src/lib/mlxcel-xla/README.md` declares that mode is not valid F32 reference evidence. The
gate does not set the variable and neither did I, so those numbers compared a TF32 MLX
reference against a true-F32 IREE candidate. Re-running at `a954d214` with
`MLX_ENABLE_TF32=0`, everything else identical:

| stage | TF32 default | `MLX_ENABLE_TF32=0` |
| --- | --- | --- |
| `vision.patch_embedding` | 1.336575e-3, 0 fail | 1.336575e-3, 0 fail |
| `vision.selected_layer_14` | 2.258102, 0 fail | 1.991226, 0 fail |
| `vision.selected_layer_21` | 2.317841, **4 fail** | 2.046043, **4 fail** |
| `vision.projector` | 2.251205e-1, 0 fail | 1.643753e-1, 0 fail |
| `prepared.sparse_add_visual_rows` | 2.500000e-1, 0 fail | 1.875000e-1, 0 fail |
| `decoder.all_layer_kv` | 5.212402e-2, 0 fail | 5.070114e-2, 0 fail |
| `decoder.prefill_logits` | 1.176004e-1, 0 fail | 1.173029e-1, 0 fail |

True F32 removes about 12 percent of the drift at both selected ViT layers and 25 to 27
percent at the projector and the sparse-add merge. The patch embedding is unchanged, which
matches `mlx/backend/cuda/conv.cpp:139` gating the convolution path on `enable_tf32()`
separately.

It does not close the gate. The same four elements at flat index 514613 still exceed
`atol=0.25, rtol=0.05`, and every other stage still passes. So the residual at
`vision.selected_layer_21` is now attributable to the graph rather than to the reference
policy, which is a cleaner statement of the remaining blocker than the earlier numbers
supported.

Two further observations on that residual, both from runs at this head:

- It is device independent. `local-task`, `local-sync` and `cuda` all produce 4 failures at
  the same flat index 514613 with max_abs `2.317841`, `2.317841` and `2.316116` respectively
  under the TF32 default. IREE agrees with itself across CPU and GPU, so the variation is not
  an IREE backend or reduction-order artifact.
- It is not a precision-policy mismatch of the kind seen in the Qwen families. This
  checkpoint has `torch_dtype: float32` with an unquantized vision tower
  (`skip_vision_non_divisible: true`), and the Molmo emitter applies no `Precision` demotion,
  so both sides are plain f32 by construction.

`MLXCEL_MOLMO_IREE_DEVICE=cuda` currently trips the gate's own assertion at line 122. I
tested it by widening that assertion locally and reverted the change, since it fixes nothing;
the result is recorded above only as evidence that the residual is device independent.

Filed the reference-policy issue and the two `build.rs` architecture defects on #932.
This PR stays draft.

inureyes:
Reopening to retrigger CI. The red `cargo-fmt` on this PR is not a formatting problem: its last run (2026-07-27) used a merge ref against a `main` that briefly carried `dtolnay/rust-toolchain@1.100.0`, a nonexistent Rust release dependabot produced by treating the action tag as an action version (`bdad2328`, PR #146). The job died at `rustup toolchain install 1.100.0 ... 404` before it ever formatted anything. Reverted the next day in #954, which also added the `dtolnay/rust-toolchain` ignore to `.github/dependabot.yml`, and the pin is 1.97.1 now. Checked this branch against the current rustfmt: `cargo fmt --all --check` is clean, so a fresh run should be green. No content change here.

inureyes:
Correction to my previous comment: this red cannot be cleared by re-running, and I have now measured why.

`gh run rerun --failed` replays the workflow file recorded with the original run, not a freshly computed merge ref. The replay I just triggered logs `Run dtolnay/rust-toolchain@1.100.0` verbatim and dies at the same 404, so the red is pinned to that run for as long as the run is the latest one. Reopening the PR does not dispatch a new run either.

Only a push to the branch (a `synchronize` event) produces a merge ref against current `main`, and merging `origin/main` into this branch conflicts today, `src/multimodal/host_preprocessor.rs` among others, after three weeks of drift. That is this PR's own rebase work rather than a CI fix, so I am leaving it.

The useful part for whoever picks the branch up: there is nothing to fix in the formatting. `cargo fmt --all --check` on this branch under the current pinned rustfmt (1.97.1) is clean. The red is an artifact of the roughly one-day window on 2026-07-27 when `main` carried `dtolnay/rust-toolchain@1.100.0`, a Rust release that does not exist, from dependabot treating the action tag as an action version (`bdad2328`, PR #146). It was reverted the next day in #954, which also added the `dtolnay/rust-toolchain` ignore to `.github/dependabot.yml`. Rebasing this branch for its own sake will take the red with it.

Issue #566, opened 2026-07-01
epic: OpenXLA multimodal / VLM architecture support (Qwen2-VL, Gemma3n, Phi4MM, Molmo, and more)


inureyes:
## Problem / current state

Parent architecture epic: #493. Scoping design: #503 and `spike/openxla/MULTIMODAL_VLM_DESIGN.md`.

The current OpenXLA/IREE path is still text-only at the request boundary, but the implementation has evolved beyond the original 2025 design assumptions:

- `src/lib/mlxcel-xla/src/emitter/model.rs` has a mature shared text layer stack, but token prefill still gathers `embed[tokens]`; there is no embeddings-prefill entry.
- `src/lib/mlxcel-xla/src/emitter/model.rs` and `src/lib/mlxcel-xla/src/iree.rs` hard-code a 256-token context/prefill shape. A normal LLaVA image alone contributes 576 or 729 patch tokens at common resolutions, before text and generation headroom.
- `src/lib/mlxcel-xla/src/iree.rs` and `src/lib/mlxcel-xla/csrc/xla_iree.c` load only token-prefill/decode modules, and `XlaBatchEngine::Pending` owns only token IDs.
- `src/backend/session.rs` has an XLA path that can ignore supplied input embeddings and run token-only generation. Multimodal support must fail closed until the real path exists.
- `src/server/batch/xla_worker.rs` rejects images, audio, and video. `spawn_xla_model_worker` owns no VLM/audio preprocessor.
- The mature MLX stack returns `MlxArray`-backed `InputEmbeddings`; its VLM loaders commonly construct a full text decoder. XLA needs an owned backend-neutral payload and filtered processor/encoder/embedding loading, not a duplicate decoder.
- Qwen M-RoPE, Qwen3 DeepStack, Gemma3 additive masks, Molmo sparse/indexed pooling, and Youtu windowed vision have different contracts and cannot be represented by one generic “merge image tokens” task.
- Phi4MM and Gemma3n audio are not currently executable reference paths in mlxcel. Phi4MM explicitly rejects audio placeholders and skips audio/speech-LoRA weights; Gemma3n loads text/vision but no audio tower. Their MLX reference implementations must be qualified before XLA ports.

## Goal

Accept supported multimodal requests on the OpenXLA/IREE backend and produce reference-correct output through CLI and continuous-batch OpenAI-compatible serving, while preserving text-only behavior and using explicit, bounds-checked contracts for:

1. compiled context capacity and admission;
[...]

PR #922, opened 2026-07-24
feat(xla): add Gemma3n audio execution

### LINKED ISSUE #878: feat(xla): support Gemma3n audio with dense PLE prefill
## Parent and dependencies

- Parent epic: #566
- Blocked by #873, #875, and #876
- Qualified MLX/oracle contract: #875
- Required XLA language/PLE contract: #876

## Problem

Gemma3n audio combines a specialized mel/Conformer encoder with the already-specialized Gemma3n AltUp/Laurel/dense-PLE language path. The audio frame mask drives convolutional subsampling and chunked local attention; the resulting projected tokens and per-layer inputs must align with expanded prompt positions before #876 seeds KV caches.

A host-only encoder fallback would leave a substantial second compute graph/runtime and would not complete the epic's OpenXLA parity target.

## Goal

Execute the qualified Gemma3n audio encoder and multimodal projection through StableHLO/IREE, feed exact merged embeddings plus dense PLE to #876, and serve bounded audio requests through CLI and continuous batching.

## Required implementation

- Keep waveform decode/resample and mel extraction in the bounded #873 host stage unless a measured, independently validated IREE front end is proposed. The `audio.main` graph accepts canonical mel features and mask/actual-frame metadata from #875.
- Define static frame buckets up to the pinned checkpoint's duration limit. Include mel bins, frame policy, convolution shapes/strides/padding, hidden size, Conformer depth, chunk/context sizes, projection/pooling config, and dtype in artifact identity.
- Emit both subsample convolution blocks with manual padding, cumulative group normalization, ReLU, mask/length propagation, and output projection exactly.
- Emit all Conformer components from #875: feed-forward halves, chunked local attention, relative position projection/shift, per-dimension query scaling, logit softcap, light causal/depthwise convolution, residuals, clipping behavior relevant to inference, and final normalization/pooling/projection.
- Carry masks through every stage; padded mel frames and padded output tokens must not contribute to cumulative statistics, attention, pooling, placeholder count, or PLE generation.
- Produce projected audio tokens plus explicit per-clip output lengths. Apply the exact #875 placeholder expansion/replacement and construct the post-scale merged embeddings and dense per-layer-input tensor required by `prefill_embeddings_ple.main`.
- Preserve multiple-clip order and request-scoped PLE. Submit effective expanded length to #857 admission, then release mel/encoder/projected/PLE buffers after successful prefill or on cancellation/error.
- Load audio weights conditionally. Capability is true only when the checkpoint contains compatible audio config/weights, `audio.main`, and the #876 language bundle. Text/image-only checkpoints must not pay the audio load/memory cost.
- Integrate CLI/server audio, mixed image/audio if qualified by #875, cancellation, metrics, and slot reset. No audio/PLE state may survive slot reuse.
- Do not retain or invoke a full MLX Gemma3n decoder/audio tower on the production XLA path.

## Non-goals

- Gemma4 audio or Phi4MM adapters.
- Speech output/streaming audio input.
- Claiming arbitrary Gemma3n checkpoint compatibility without the pinned processor/config oracle.

## Validation

- Compare every stage selected by #875: mel/mask input, both subsample blocks/cumulative stats, relative-position and chunk masks, selected Conformer blocks, pooled/projected audio tokens, placeholder merge, dense PLE, selected Gemma3n language layers, KV/logits, and greedy tokens.
- Cover duration/frame bucket boundaries, padded batches, one/multiple clips, malformed masks, all-padded rejection, context overflow after audio expansion, audio-only, text/image-only, qualified mixed media, cancellation, and slot reuse.
- Interleave requests with different clip lengths and modalities; prove masks, effective lengths, and dense PLE remain slot-local.
- Run deterministic ASR through CLI and streaming server on a production-relevant IREE target with token-exact output.
- Run #875/#876 and all existing Gemma3n/text/image/XLA gates.

## Acceptance criteria

- [ ] Gemma3n audio encoder/projection executes entirely through IREE math after host mel extraction.
- [ ] Cumulative normalization, local attention, relative shift, masks, pooling, and projected lengths match #875 intermediates.
- [ ] Merged embeddings and dense PLE match #875 and seed #876 KV/logits exactly.
- [ ] Frame/context bounds and all mask/length/cardinality invariants fail safely before native misuse.
- [ ] Audio capability and conditional weight loading reflect the actual loaded bundle.
- [ ] CLI/server deterministic output is token-exact and per-slot state is isolated.
- [ ] Existing Gemma3n text/image behavior remains token-exact.

inureyes:
## Summary

- add the split resident IREE Gemma3n audio encoder, projection, prepared-prefill, dense-PLE, CLI, and continuous-batch server path
- validate the pinned audio weight schema, cache identity, request ownership, cancellation, and slot reuse without a silent MLX fallback
- add an audio-only MLX-to-IREE intermediate diagnostic that avoids loading the 30 GB language bundle
- match the pinned MLX CUDA sequential BF16 affine dequantization contract and the explicit BF16 mel boundary before the first SSCP convolution
- keep BF16 SSCP operands with explicit F32 StableHLO convolution results followed by the existing BF16 output rounding
- match the MLX CUDA cumulative-normalization association order and expose the cumulative norm arithmetic boundaries
- replace the released dense-PLE width-256 RMS reduction with the explicit MLX CUDA lane/XOR schedule
- retain a diagnostics-only conv1 BF16-result candidate as negative evidence; production MLIR remains unchanged

## Latest CUDA evidence

The actual MLX CUDA versus IREE CUDA diagnostic proves the following boundaries bit-exact:

- `sscp_conv_0_convolution`
- every exposed conv0 cumulative-normalization arithmetic stage and final norm
- `sscp_conv_0` activation, which is the BF16 input consumed by conv1

The first divergence remains `sscp_conv_1_convolution`. The controlled result-type experiment is now complete:

| Conv1 form | max_abs | RMS | values over one BF16 ULP |
| --- | ---: | ---: | ---: |
| BF16 operands -> F32 StableHLO result -> BF16 round (production) | 0.25 | 6.284481559e-4 | 6 |
| BF16 operands -> BF16 StableHLO result -> F32 carrier (diagnostic candidate) | 12 | 0.3255866858 | 419,814 |

The BF16-result candidate is decisively worse. Result typing is therefore not the fix: the production F32 accumulation/result followed by BF16 rounding is the correct representation and remains unchanged.

## Backend audit

MLX CUDA's vendored convolution path constructs a cuDNN Frontend graph with BF16 I/O and F32 compute for BF16 operands. It requests Heuristic Mode A execution plans, checks support, builds a candidate, and caches the resulting graph by dtype, shapes, stride, padding, dilation, groups, and buffer alignments. It excludes plans marked `DOWN_CONVERT_INPUTS`; its TF32-specific tensor-core exclusion is only applied when the convolution I/O dtype itself is F32, not for this BF16-input convolution.

The selected cuDNN engine is backend state, not StableHLO state. The current MLX wrapper does not record the selected plan name/engine configuration or numerical notes. The vendored cuDNN Frontend can expose a selected plan name and plan metadata, and cuDNN API logging can expose backend calls, but obtaining that identity requires instrumenting the MLX/cuDNN layer or running an exact-shape cuDNN probe. Either approach can identify the MLX plan; neither creates an equivalent StableHLO/IREE selector.

A compiler-only exact-shape probe of the production conv1 operation was lowered with the pinned IREE CUDA compiler through `executable-configurations`. IREE emitted its own `linalg.generic` reduction with BF16-to-F32 extension, F32 multiply/add, and an IREE GPU reduction/workgroup tiling configuration. It did not emit a cuDNN execution plan. Repeating the same probe with StableHLO `precision_config = HIGHEST` produced the same IREE lowering configuration and arithmetic, so `precision_config` is not a usable diagnostics candidate for selecting the MLX/cuDNN reduction tree.

StableHLO convolution exposes semantic shape/layout/precision configuration, but no cuDNN engine global index, knob map, numerical-note filter, or exact reduction-tree attribute. Consequently, a tiny cuDNN-vs-StableHLO probe could reproduce and characterize the backend difference, but it cannot force IREE to execute the same cuDNN plan. The existing actual-checkpoint diagnostic already supplies the stronger comparison because it uses the exact conv1 input and weights.

The precise blocker is therefore backend-limited: matching the final six sparse BF16 residuals would require one of the following new backend capabilities, none of which is safe to guess in this PR:

1. an IREE CUDA external/cuDNN convolution path that can import and pin the same engine configuration and knobs;
2. an explicitly implemented custom CUDA reduction schedule proven to match the selected MLX cuDNN engine; or
3. an upstream compiler contract that represents convolution reduction/engine identity.

No production change, tolerance relaxation, heavyweight model rerun, or speculative candidate was added.

## Validation

- actual CUDA result-type comparison recorded above
- `cargo test -p mlxcel-xla emitter::gemma3n_audio_emit::tests::` — 5 passed, 5 environment-dependent tests ignored
- focused result-type identity test pins identical operands, padding, layout, stride, and shapes
- version-matched IREE `3.11.0rc20260316` CUDA compile of the released conv1-shape result-type pair passed
- exact-shape IREE `executable-configurations` audit: custom F32 reduction lowering confirmed; `DEFAULT` and `HIGHEST` select the same configuration
- `cargo check -p mlxcel-xla --features diagnostics --tests`
- `cargo clippy -p mlxcel-xla --tests --no-deps` — existing unrelated warnings remain
- `git diff --check`
- no generated CSV or model artifact is included

## Remaining blocker

This PR remains draft and `status:blocked` because the unchanged MLX-to-IREE audio thresholds have not passed. The result-type hypothesis is closed; the remaining mismatch is an unrepresentable backend reduction/engine-identity difference.

Closes #878

Parent epic: #566

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -349,3 +349,9 @@ name = "xla_phi4_audio_check"
 [[example]]
 name = "xla_gemma3n_audio_check"
 required-features = ["xla-iree"]
+
+# Audio-only MLX↔IREE first-divergence gate for #878. This intentionally does
+# not construct the 30 GB language bundle.
+[[example]]
+name = "xla_gemma3n_audio_reference_check"
+required-features = ["xla-diagnostics"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -343,4 +343,9 @@ required-features = ["xla-micro-oracle"]
 # Pinned #874 MLX-vs-IREE Phi4MM Conformer and speech/vision projection parity.
 [[example]]
 name = "xla_phi4_audio_check"
+
+# Deterministic Gemma3n audio split-runtime + dense-PLE language gate (#878).
+# Uses the pinned #875 WAV/checkpoint fixture selected by environment variables.
+[[example]]
+name = "xla_gemma3n_audio_check"
 required-features = ["xla-iree"]
--- a/examples/xla_gemma3n_audio_check.rs
+++ b/examples/xla_gemma3n_audio_check.rs
@@ -1,0 +1,185 @@
+// Copyright 2025-2026 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Real-IREE #875 fixture gate for the split Gemma3n audio runtime.
+
+use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicBool;
+
+use mlxcel::audio::gemma3n::Gemma3nAudioFeatureExtractor;
+use mlxcel::audio::{AudioFamilyPolicy, preprocess_wav_file};
+use mlxcel::tokenizer::load_tokenizer;
+use mlxcel_xla::{
+    GEMMA3N_AUDIO_MEL_BINS, GEMMA3N_AUDIO_SOFT_TOKENS, Gemma3nAudioInput, XlaInferenceSession,
+    select_gemma3n_audio_frame_bucket,
+};
+
+const PROMPT: &str = "Transcribe the following speech segment in English:";
+const AUDIO_TOKEN_ID: i32 = 262_273;
+const BOA_TOKEN_ID: i32 = 256_000;
+const EOA_TOKEN_ID: i32 = 262_272;
+const CONTEXT_CAPACITY: usize = 256;
+const EXPECTED_TRANSCRIPT: &str = "Chapter 1. Mrs. Rachel Lind is surprised.\n\nMrs. Rachel Lind";
+
+fn path_env(name: &str, default: &str) -> PathBuf {
+    std::env::var_os(name)
+        .map(PathBuf::from)
+        .unwrap_or_else(|| PathBuf::from(default))
+}
+
+fn language_layer_count(model_dir: &Path) -> Result<usize, String> {
+    let path = model_dir.join("config.json");
+    let bytes =
+        std::fs::read(&path).map_err(|error| format!("read {}: {error}", path.display()))?;
+    let config: serde_json::Value = serde_json::from_slice(&bytes)
+        .map_err(|error| format!("parse {}: {error}", path.display()))?;
+    let layers = config
+        .get("text_config")
+        .and_then(|text| text.get("num_hidden_layers"))
+        .and_then(serde_json::Value::as_u64)
+        .ok_or_else(|| format!("{} has no text_config.num_hidden_layers", path.display()))?;
+    usize::try_from(layers)
+        .map_err(|_| format!("Gemma3n language layer count {layers} does not fit usize"))
+}
+
+fn expanded_prompt(
+    model_dir: &Path,
+) -> Result<(mlxcel::tokenizer::MlxcelTokenizer, Vec<i32>), String> {
+    let tokenizer = load_tokenizer(model_dir).map_err(|error| error.to_string())?;
+    let rendered = format!(
+        "<bos><start_of_turn>user\n{PROMPT}<audio_soft_token><end_of_turn>\n\
+         <start_of_turn>model\n"
+    );
+    let mut tokens = tokenizer
+        .encode(&rendered, false)
+        .map_err(|error| error.to_string())?
+        .into_iter()
+        .map(|token| token as i32)
+        .collect::<Vec<_>>();
+    let wrapper = tokenizer
+        .encode("\n\n", false)
+        .map_err(|error| error.to_string())?
+        .into_iter()
+        .map(|token| token as i32)
+        .collect::<Vec<_>>();
+    mlxcel::vlm_runtime::expand_gemma3n_audio_tokens(
+        &mut tokens,
+        AUDIO_TOKEN_ID,
+        BOA_TOKEN_ID,
+        EOA_TOKEN_ID,
+        1,
+        GEMMA3N_AUDIO_SOFT_TOKENS,
+        &wrapper,
+        None,
+    )
+    .map_err(|error| error.to_string())?;
+    if tokens.len() != 210 {
+        return Err(format!(
+            "#875 expanded prompt has {} tokens, expected 210",
+            tokens.len()
+        ));
+    }
+    Ok((tokenizer, tokens))
+}
+
+fn audio_input(audio_path: &Path) -> Result<Gemma3nAudioInput, String> {
+    let waveforms = preprocess_wav_file(
+        audio_path,
+        AudioFamilyPolicy::gemma3n(),
+        &AtomicBool::new(false),
+    )
+    .map_err(|error| error.to_string())?;
+    let clips = waveforms
+        .clips
+        .into_iter()
+        .map(|clip| clip.samples)
+        .collect::<Vec<_>>();
+    let features = Gemma3nAudioFeatureExtractor::new().extract_batch(&clips)?;
+    if features.frames != 1_406 {
+        return Err(format!(
+            "#875 mel frontend emitted {} frames, expected 1406",
+            features.frames
+        ));
+    }
+    let bucket =
+        select_gemma3n_audio_frame_bucket(features.frames).map_err(|error| error.to_string())?;
+    if bucket != 2_048 {
+        return Err(format!(
+            "#875 selected frame bucket {bucket}, expected 2048"
+        ));
+    }
+    let mut mel = vec![0.0; bucket * GEMMA3N_AUDIO_MEL_BINS];
+    for frame in 0..features.frames {
+        let start = frame * GEMMA3N_AUDIO_MEL_BINS;
+        mel[start..start + GEMMA3N_AUDIO_MEL_BINS]
+            .copy_from_slice(&features.features[start..start + GEMMA3N_AUDIO_MEL_BINS]);
+    }
+    let mut valid_mask = vec![0; bucket];
+    for (output, valid) in valid_mask
+        .iter_mut()
+        .zip(features.valid_mask.iter().copied())
+    {
+        *output = u8::from(valid);
+    }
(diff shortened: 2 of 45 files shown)


inureyes:
Rebased onto `main@7fe1412d` and force-pushed as `876bcc3c`.

Rebase integration:
- resolved the `serde` dependency spelling conflict in favor of current main
- combined #934 numeric-contract test imports with this branch’s typed auxiliary weight storage
- adapted the newer Qwen2-VL loader to `AuxiliaryWeightStorage::Bytes`
- kept both Gemma3n audio artifacts explicitly `new_legacy_unqualified`

Post-rebase validation:
- `git diff --check`
- `cargo test -p mlxcel-xla emitter::gemma3n_audio_emit::tests::`: 5 passed, 5 environment-dependent ignored
- `cargo check -p mlxcel-xla --features diagnostics`: passed with existing warnings
- `cargo check -p mlxcel-xla --features diagnostics --tests`: reaches only the existing mainline `gemma3n_qmv` test-only `Builder` cfg/import defect; the branch-specific Qwen auxiliary-storage mismatch was fixed

The unchanged MLX-to-IREE audio threshold blocker remains, and CUDA is unavailable on the current host. No new numeric pass is claimed. This PR remains draft/blocked and #878 remains blocked by #932.

inureyes:
Foundation refresh complete at head 2b6d9491:

- rebased the full draft cleanly onto main@15a37117 (#943)
- replaced the family-local one-block cumulative-scan implementation with the shared mlx_cuda_prefix_sum_2d helper
- added a production-wrapper structural regression pinning the same 17-add / 3-concatenate CUDA hierarchy
- focused Gemma3n audio tests pass (6 runnable, 5 environment-dependent ignored), the shared schedule compiles through local-task CPU IREE, and diagnostics test compilation passes

The draft remains fail-closed and status:blocked. This refresh does not change the recorded first unresolved boundary: sscp_conv_1_convolution still requires a representable cuDNN/IREE engine or reduction-plan contract.

inureyes:
Rebased onto current `main` (`ec6fbb6d` foundation; new head `0a5aa90d`) and revalidated the Gemma3n audio branch.

Validation:
- `cargo check -p mlxcel-xla --features diagnostics --lib` — passed
- `cargo test -p mlxcel-xla gemma3n_audio --lib` — 17 passed, 6 ignored (environment-gated real IREE/checkpoint probes)
- `cargo fmt --all --check` — passed
- `git diff --check` — passed

The PR intentionally remains draft/fail-closed. The existing production-oracle blocker is unchanged: the first SSCP convolution still has 6 BF16-ULP mismatches (`max_abs = 0.25`), so the original #878 intermediate/token/lifecycle gates are not satisfied. No tolerance was relaxed, and this branch is not merge-qualified yet.

Refs #878
Refs #932

inureyes:
## Current-head CUDA actual at `0a5aa90d`: the first SSCP convolution now passes

Run on the GB10 qualification host (driver 580.159.03, CUDA 13.0, IREE 3.12.0rc20260721,
Rust 1.93.1, MLX pin `b7c3dd6d27f4`, `MLX_CUDA_ARCHITECTURES=121`, release profile).
Working tree clean at `0a5aa90d`. Tolerances unchanged, no override.

### Fast pre-gate

`cargo test --locked -p mlxcel-xla gemma3n_audio --lib`: 17 passed, 6 ignored.

### Small SSCP convolution characterization probe

`./target/release/examples/xla_gemma3n_sscp_conv_oracle` exits 0 and reports PASS:

```
channel=0 mlx=0.000000000e0 iree=0.000000000e0 bf16_ulp=0     schedule_outcomes=[0.0, 1.0]
channel=1 mlx=1.000000000e0 iree=1.000000000e0 bf16_ulp=0     schedule_outcomes=[0.0, 1.0]
channel=2 mlx=0.000000000e0 iree=0.000000000e0 bf16_ulp=0     schedule_outcomes=[0.0, 1.0]
channel=3 mlx=1.000000000e0 iree=0.000000000e0 bf16_ulp=16256 schedule_outcomes=[0.0, 1.0]
PASS (backend reduction schedules differ within the enumerated F32 accumulation envelope)
```

Channel 3 is worth recording explicitly: the probe enumerates the achievable F32 accumulation
orders and finds that both `0.0` and `1.0` are legitimate outcomes for the same sum. MLX
lands on one and IREE on the other. That is an exact-tie cancellation whose result is decided
by the reduction tree, so no precision policy can reconcile it, which supports the existing
characterization of this as a backend reduction/engine-identity difference rather than a
correctable semantic mismatch.

### Full audio intermediate actual, `MLXCEL_XLA_DEVICE=cuda`

The recorded blocker for this PR was 6 BF16 ULP outliers with `max_abs=0.25` on the first
SSCP convolution. At this head the first convolution no longer produces any BF16 ULP outlier,
and the first divergence has moved downstream:

```
first divergent stage=sscp_conv_0_norm_cumulative_sum
max_abs=1.562500000e-2  rms=2.432233746e-3  max_index=373
max_bf16_ulp=0  non_bf16_values=1024  over_one_bf16_ulp=0  captured_bf16_outliers=0
```

`max_abs` is about 16x smaller than the recorded blocker, and `max_bf16_ulp`,
`over_one_bf16_ulp` and `captured_bf16_outliers` are all zero at the failing stage. Mel
features, valid mask, frame lengths and bucket all pass ahead of it.

The gate still fails, now at the cumulative-sum normalization rather than the convolution, so
the PR stays draft and the capability stays fail-closed.

### The reference policy was checked and is not responsible

I filed on #932 that MLX defaults `MLX_ENABLE_TF32=1` and that
`src/lib/mlxcel-xla/README.md` declares that mode is not valid F32 reference evidence. It
decided the Molmo2 gate in #916, and `mlx/backend/cuda/conv.cpp:139` gates the convolution
path on `enable_tf32()` independently of the matmul path, so it was a plausible explanation
for a convolution blocker and had to be tested.

It is not the explanation here. A controlled A/B at this head, identical in every other
respect, is byte-identical:

| `MLX_ENABLE_TF32` | first divergent stage |
| --- | --- |
| `1` (MLX default) | `sscp_conv_0_norm_cumulative_sum` max_abs `1.562500000e-2` rms `2.432233746e-3` index 373 |
| `0` (declared policy) | `sscp_conv_0_norm_cumulative_sum` max_abs `1.562500000e-2` rms `2.432233746e-3` index 373 |

So the improvement relative to the recorded blocker is attributable to this head rather than
to the reference policy, and the remaining divergence is a property of the emitted graph.
Recording the A/B because the negative result is what makes the attribution safe.

Remaining, unchanged: cumulative normalization and shared prefix scan, local attention,
relative shift, pooling, projection lengths, projected audio embeddings and dense PLE, then
logits/KV/greedy ASR token exactness and the CLI/server, isolation, cancellation and slot
reuse gates.

Issue #637, opened 2026-07-02
perf(cuda/quant): sm_120/sm_121-tuned quantized GEMM for Blackwell prefill


inureyes:
Part of #623. Depends on #625 (upstream JIT qmm rework changes the starting point) and should follow #629 (MoE prefill fix determines which GEMM path matters most).

## Context

The only architecture-specialized quantized GEMM in the CUDA path is `qmm_sm90`, compiled solely when `90a` (Hopper) is in `MLX_CUDA_ARCHITECTURES` (`MLX_CUDA_SM90A_ENABLED` gate: `src/lib/mlxcel-core/build.rs:265-271`; docs `docs/installation.md:101-105`). A GB10 build passes `121`, so `supports_qmm_sm90` is false and quantized prefill matmuls fall through to `qmm_sm80` (Ampere-class tiles) or `qmm_naive` (`src/lib/mlx-cpp/patches/mlx/backend/cuda/quantized/quantized.cpp:53-113`). Blackwell (sm_120/121 consumer-class: no wgmma/TMA from sm_90a, but 5th-gen tensor cores incl. fp4/fp8 paths and larger shared memory) gets a two-generations-old kernel for the compute-bound phase.

Decode is not the target here: it is GEMV-shaped and bandwidth-bound (`qmv`/`fp_qmv`, plus upstream qmv_wide after #625). The payoff is prefill and batched prefill on long prompts, plus MoE grouped GEMM shapes.

Upstream is also moving here (QQMM both-sides-quantized work targets compute capability 10.0+; Hopper QMM heuristics; NAX kernels). This issue is research-then-implement: prefer upstream adoption over local kernels.

## Scope

Quantify the gap and close what is real. Explicitly allowed outcome: a findings doc concluding the sm_80 path is within noise of achievable on sm_121, with profile evidence, plus an upstream issue if a kernel gap is confirmed. Otherwise: an sm_120-tuned qmm in the overlay.

## Implementation plan

1. Quantify (needs #624 long-prompt bench): prefill tok/s at 2048/8192/32768 on llama-3.1-8b-4bit and qwen2.5-7b-4bit; compute achieved TFLOPS and compare against (a) the same model's bf16 prefill (cuBLAS/cuDNN-backed, the practical ceiling) and (b) ncu-measured tensor-core utilization of the current qmm_sm80 path.
[...]

PR #923, opened 2026-07-26
fix(cuda): drain cp.async before qmm_sm80 epilogue smem store (#910)

### LINKED ISSUE #910: bug: Gemma 4 26B A4B temp-0 decode is non-deterministic run-to-run and corrupts near-tied tokens independently of the fused MoE kernel

inureyes:
## Summary

Fixes the run-to-run bitwise non-determinism (and occasional garbage tokens) of temp-0 decode on all 4-bit CUDA models by closing a shared-memory write-write race in the vendored MLX qmm_sm80 kernel: the epilogue staged its output tile into shared memory that aliases the A/B cp.async pipeline slots (SharedStorage is a union) without first retiring in-flight cp.async groups, so a late-landing async copy could overwrite freshly staged C values before they were copied to global memory.

## Root cause (refines the issue's working hypothesis)

The issue's Layer A hypothesis pointed at the async-allocator/command-encoder machinery; evidence-driven bisection landed one level deeper, inside the kernel itself. Repeated identical in-process forwards flip 23/23 at prefill length 8 and 0/23 at length 4, exactly the `M * B < 8 ? qmv : qmm_sm80` dispatch boundary in quantized.cpp, and `compute-sanitizer --tool racecheck` deterministically reports a write-write hazard in `qmm_sm80_kernel` (214 errors pre-fix, 0 post-fix). Even `CUDA_LAUNCH_BLOCKING=1` still flipped 1/23 in a tight in-process loop, proving the race is intra-kernel rather than a host-side missing stream dependency; launch-blocking (and graphs-off, and graph-cache-off) merely narrow the DMA timing window, which is why they looked like discriminators. In upstream MLX b7c3dd6d `mlx/backend/cuda/device/qmm_sm80.cuh`, the mainloop only ever waits down to `K_PIPE_MAX-2` outstanding cp.async groups (lines 235, 249) and its final iteration issues a redundant `fetch_gmem` of the last K tile (line 257), yet the epilogue (lines 265-272) writes the aliasing C tile with no `cp_async_wait<0>()` and no leading `__syncthreads()`. bf16 dense models were never affected because they use cuBLASLt and do not reach this kernel; the issue's Layer B (MoE router expert-set flips) is downstream of the same defect via the batched gather_qmm dispatch of qmm_sm80 and resolves with this fix (no separate Layer B change needed).

## What changed

- `src/lib/mlx-cpp/patches/mlx/backend/cuda/device/qmm_sm80.cuh` (new overlay): adds `cp_async_wait<0>()` + `__syncthreads()` at the top of the qmm_sm80_mainloop epilogue, retiring all outstanding copy groups and re-converging the CTA before any thread overwrites the union (also closes the cross-thread write-after-read hazard on the final pipeline-slot reads); otherwise byte-identical to upstream b7c3dd6d.
- `src/lib/mlx-cpp/patches/mlx/backend/cuda/quantized/qmm/qmm_sm80.cu` (existing overlay): JIT module name bumped to `qmm_sm80_r2` because the on-disk PTX cache is keyed by module name only, so without the bump a stale cached PTX would keep serving the racy kernel after an upgrade; top-of-file overlay header now documents both the #637 tile change and this fix.
- `tests/cuda_qmm_determinism.rs` (new): CUDA-gated regression test that repeats an identical prefill (64 tokens, forcing the qmm_sm80 path) + fixed-token decode sequence with fresh KV caches and asserts byte-identical logits at every step across 10 iterations; failed 23/23 iterations pre-fix, skips gracefully when the model checkpoint is absent, model dir overridable via `MLXCEL_TEST_QMM_MODEL_DIR`.
- `docs/upstream/mlx-cuda-quantized-matmul-nondeterminism.md` (new): upstream MLX bug report draft with the evidence tables, exact file/line root cause, minimal fix diff, and standalone repro (value-level loop plus deterministic racecheck detection), ready for manual submission to ml-explore/mlx.

## Test plan

- [x] `MLX_CUDA_ARCHITECTURES=121 cargo test --release --features cuda --test cuda_qmm_determinism` passes (GB10)
- [x] 20/20 temp-0 CLI runs of 128 tokens byte-identical on llama-3.2-1b-4bit, matching the `CUDA_LAUNCH_BLOCKING=1` reference output (pre-fix: 3+ distinct outputs in 4 runs)
- [x] 6/6 byte-identical on qwen3-4b-4bit, 3/3 on qwen3-30b-a3b-4bit (MoE, Layer B path)
- [x] 6/6 byte-identical with `MLX_USE_CUDA_GRAPHS=0` (pre-fix ~1/10 flips)
- [x] `compute-sanitizer --tool racecheck`: 0 hazards (pre-fix 214 errors in qmm_sm80_kernel)
- [x] Decode throughput unchanged (256 vs 259 tok/s, run-to-run noise); orchestrator to run the full 6-model determinism matrix including the gemma4 26B 5-paragraph repro

Closes #910

### DIFF
--- a/docs/upstream/mlx-cuda-quantized-matmul-nondeterminism.md
+++ b/docs/upstream/mlx-cuda-quantized-matmul-nondeterminism.md
@@ -1,0 +1,153 @@
+# MLX CUDA: qmm_sm80 shared-memory epilogue race makes quantized matmul non-deterministic
+
+Upstream bug report draft for ml-explore/mlx. Found and fixed in mlxcel (lablup/mlxcel#910); the fix is carried as a source overlay on the pinned MLX commit until it lands upstream.
+
+## Summary
+
+`qmm_sm80_kernel` (mlx/backend/cuda/device/qmm_sm80.cuh) has a shared-memory write-write race between the epilogue's C-tile staging and still-in-flight `cp.async` copies from the mainloop. `SharedStorage` is a union, so the epilogue C tile aliases the A/B pipeline slots, and the epilogue writes it without first retiring outstanding `cp.async` groups or re-converging the CTA. A late-landing async copy of A/B data overwrites freshly staged C values before the smem-to-gmem copy reads them, so the kernel publishes corrupted output on a timing-dependent subset of launches.
+
+Observable effect: every quantized (tested with affine 4-bit) model decodes non-deterministically at temperature 0 on CUDA. Two runs of the same binary with the same prompt diverge bitwise, usually within the first few generated tokens, and the corrupted logits occasionally surface as garbage tokens (foreign-script injections, nonsense words) at positions where clean runs are confident by 14+ logits. Any `quantized_matmul` / `gather_qmm` dispatch with M >= 8 is affected (the `qmv` path used for M * B < 8 is clean); prefill is therefore always affected, and in MoE models the corrupted `gather_qmm` outputs also jitter the router scores so the selected expert set flips run to run.
+
+Environment: GB10 (DGX Spark, sm_121), CUDA 13, Linux 6.17, MLX commit b7c3dd6d. The race is not architecture-specific in construction; the timing window will vary by device.
+
+## Evidence
+
+### bf16 vs 4-bit (same binary, temp-0 CLI decode, byte-diff of token streams)
+
+| Model | Quant | Result |
+|---|---|---|
+| llama-3.1-8b-bf16 | none | deterministic (4/4 byte-identical) |
+| llama-3.1-8b-4bit | affine 4-bit | non-deterministic |
+| llama-3.2-1b-4bit | affine 4-bit | non-deterministic (3 distinct outputs in 4 runs of 128 tokens) |
+| qwen3-4b-4bit | affine 4-bit | non-deterministic |
+| gemma MoE 26b-a4b-4bit | affine 4-bit MoE | non-deterministic, occasional garbage tokens |
+| qwen3-30b-a3b-4bit | affine 4-bit MoE | non-deterministic |
+
+bf16 dense models never touch the qmm kernels (cuBLASLt path), which is why they are clean.
+
+### Dispatch-boundary bisection (in-process, repeated identical forward, fresh KV caches, logits byte-hashed)
+
+| Prefill length | Dispatch (QuantizedMatmul::eval_gpu) | Result over 24 iterations |
+|---|---|---|
+| 1 | qmv (M * B < 8) | deterministic (0/23 diverge) |
+| 4 | qmv | deterministic (0/23) |
+| 8 | qmm_sm80, CTA tile (16,128,64) | non-deterministic (23/23) |
+| 32 | qmm_sm80 | non-deterministic (23/23) |
+
+The flip from 0/23 to 23/23 exactly at M = 8 matches the `M * B < 8 ? qmv : qmm_sm80` dispatch boundary in quantized.cpp.
+
+### Environment-variable discriminators (and why they misled)
+
+| Config | Result | Interpretation |
+|---|---|---|
+| default | non-deterministic, high rate | race window wide under normal launch pressure |
+| `MLX_USE_CUDA_GRAPHS=0` | still non-deterministic (~1/10 runs of 128 tokens) | not a graph-machinery bug; slower launch path narrows the window |
+| graph-exec cache disabled (`MLX_CUDA_GRAPH_CACHE_SIZE=1` + thrash check off) | 0 divergence in 16 runs | per-commit graph instantiation slows the host enough to mask the race almost completely |
+| `CUDA_LAUNCH_BLOCKING=1` | 0 divergence in CLI runs, but 1/23 iterations still diverged in a tight in-process loop | serialized launches keep the SM idle so the async copy nearly always lands before the epilogue store; suppression, not correctness |
+
+The launch-blocking residual (1/23 in-process) is the decisive clue that this is an intra-kernel race rather than a host-side missing stream dependency: with fully serialized launches there is no host/device or cross-stream asynchrony left, yet the kernel still occasionally flips.
+
+### compute-sanitizer racecheck (direct detection)
+
+Running any M >= 8 quantized forward under `compute-sanitizer --tool racecheck` reports the hazard deterministically, pre-fix:
+
+```
+Error: Race reported between Write access at
+  mlx::core::cu::qmm_sm80_kernel<(int)64, cutlass::half_t,
+  cutlass::integer_subbyte<(int)4, (bool)0>, cutlass::half_t,
+  cute::tuple<cute::C<(int)16>, cute::C<(int)128>, cute::C<(int)64>>>+0x2db0
+and Write access at ...qmm_sm80_kernel<...>+0x5020 [32768 hazards]
+RACECHECK SUMMARY: 214 errors
+```
+
+Post-fix: `RACECHECK SUMMARY: 0 hazards displayed (0 errors, 0 warnings)`.
+
+## Root cause
+
+File: `mlx/backend/cuda/device/qmm_sm80.cuh` at commit b7c3dd6d.
+
+1. `SharedStorage` (line 18) is a union: the epilogue's `C` staging tile occupies the same shared memory bytes as the mainloop's `A`/`B` cp.async pipeline slots.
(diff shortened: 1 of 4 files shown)


inureyes:

### Intent
> Close the intra-kernel shared-memory write-write race in the vendored MLX `qmm_sm80` epilogue that made temp-0 decode bitwise non-deterministic (and occasionally garbage) on all 4-bit CUDA models (issue #910).

### Verdict on the two highest-priority questions

**Sync-fix correctness: CORRECT.** `cp_async_wait<0>()` + `__syncthreads()` is inserted at the epilogue top (right after `// Epilogue.`, before the register-only accumulator conversion and before the first union smem write `copy(r2s_copy_c, r2s_tCrC, r2s_tCsC)`). `cp_async_wait<0>()` retires all outstanding cp.async groups landing in sA/sB, and the barrier re-converges the CTA and makes those writes visible before any thread overwrites the aliasing union with C. The two lines are unconditional (no `if constexpr` guard), so they apply to every template instantiation and tile shape the kernel is compiled for. No epilogue path writes the union before the barrier, and the barrier also closes the cross-thread write-after-read on the final pipeline-slot reads. Both statements are required: `__syncthreads()` alone does not wait for outstanding cp.async DMAs. The fix reuses the exact `cp_async_wait<N>(); __syncthreads();` idiom the mainloop already uses at its two wait points.

**JIT-name-bump completeness: COMPLETE.** `module_name` is constructed once (`qmm_sm80_r2_tn_...`) and consumed only by `get_jit_module`, which keys both the in-memory registry (`unordered_map<string, JitModule>` by name) and the on-disk PTX cache. Confirmed in the fetched `jit_module.cpp`: `get_ptx_path()` returns `cache_dir / (module_name + ".ptx")` with no source hash, so the rationale holds exactly. Without the bump a stale pre-fix `qmm_sm80_tn_*.ptx` would keep serving the racy kernel after an upgrade. The old name `qmm_sm80_tn_` is absent from all tracked source. `jit_module.cpp` still `#include`s the source *file* `qmm_sm80.cuh` (a path, correctly unchanged), and NVRTC's `module_name + ".cu"` program name is cosmetic. Build and lookup agree.

### Verification
- [x] All stated requirements implemented (kernel fix + cache-name bump + regression test + upstream report)
- [x] No placeholder/mock code (real cute/CUDA fix)
- [x] Integrated into project code flow: the new `.cuh` overlay is picked up automatically by `file(GLOB_RECURSE patches/mlx/backend/cuda/*)` + `configure_file(COPYONLY)`; verified the fix is present in the current release build's fetched MLX copy
- [x] Project conventions followed (overlay doc header, Apache license header on the test, cxx-bridge API usage matching existing tests)
- [x] Existing modules reused (cute `cp_async_wait`/`__syncthreads` idiom; `mlxcel_core` cxx APIs; existing `LanguageModel` trait)
- [x] No unintended structural changes: `.cuh` overlay is byte-identical to pristine upstream b7c3dd6d except the header/epilogue doc comments and the two functional lines (diffed against the raw source at the pinned commit); `.cu` diff is exactly the pre-existing #637 tile overlay plus the #910 name bump
- [x] No perf regression risk: the drain+barrier runs once per output-tile epilogue (after the K loop), not in the inner K/block loop; PR reports 256 vs 259 tok/s (noise)
- [x] Test quality: `#![cfg(feature = "cuda")]`, PREFILL_LEN=64 so M*B=64 >= 8 hits `qmm_sm80` (confirmed against the `M*B < 8 ? qmv : qmm_sm80` dispatch in `quantized.cpp`), asserts byte-identical logit hashes across 10 iterations, skips gracefully when the model is absent (`MLXCEL_TEST_QMM_MODEL_DIR` overridable). All referenced APIs exist with matching signatures.
- [ ] Tests executed here: not run. The CUDA determinism matrix (cold `--features cuda` build) is left to the orchestrator per the review constraints; test compilation was validated statically by resolving every API used.

### Notes (informational, no action needed)
- The automated regression test uses a dense model (`llama-3.2-1b-4bit`), so it exercises `qmm_sm80` via the M=64 prefill but not the batched `GatherQMM` (MoE) dispatch. Both share `qmm_sm80_kernel`, so the fix is covered; the PR verified the MoE path manually (qwen3-30b-a3b-4bit, qwen3-4b-4bit).
- The test compares 64-bit SipHash digests of the logit bytes rather than raw bytes; collision risk is negligible.
- Fix scope is correctly confined to `qmm_sm80`: it is the kernel racecheck named and the only M*B>=8 quantized-matmul kernel reachable on GB10 (sm_121, cc major 12); `qmm_sm90` requires cc major 9 and is not used there.

### Remaining Items
- None. No findings required auto-fix.

PR #924, opened 2026-07-26
feat(models): add GPT-2 (gpt2) text model support

### LINKED ISSUE #833: feat(models): add GPT-2 (gpt2) text model support
## Summary

GPT-2 is the classic OpenAI decoder-only transformer and a foundational architecture that many derived checkpoints and educational/regression checkpoints still ship in. mlx-lm supports it, but mlxcel does not yet load the `gpt2` architecture. Porting it gives mlxcel a small, cheap-to-run reference model that is useful for smoke-testing the text path and unlocks the family of GPT-2 derivatives. The port is low effort because the block structure is close to the existing Llama-style skeleton once RMSNorm is swapped for LayerNorm and the Conv1D weight layout is handled.

## Upstream reference

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gpt2.py

## Public checkpoint

`mlx-community/gpt2` (and larger GPT-2 variants such as gpt2-medium / gpt2-large on the same org). Verify mlx-community checkpoint availability before validation.

## Architecture notes

Classic GPT-2 decoder with these distinctive traits:

- Learned absolute position embeddings (`wpe`) added to token embeddings (`wte`). No RoPE anywhere.
- `nn.LayerNorm` (with bias), NOT RMSNorm.
- Fused `c_attn` QKV projection and `c_proj` output projection stored in the HuggingFace Conv1D layout, which is transposed relative to `nn.Linear`. ~~`sanitize` must transpose the `c_attn`, `c_proj`, and MLP `c_fc` weights so they load as standard linear layers.~~ **Correction (verified against the merged `src/models/gpt2.rs`, PR #924, `c7e9b6e0d`): three details of that sentence are wrong. First, the transpose covers four projections per block, not three: the attention `c_attn` and `c_proj` plus the MLP `c_fc` and the MLP `c_proj` all load through `Gpt2Layout::conv1d_linear`. Second, it applies to the `.weight` tensors only, never to bias vectors: `c_attn.bias` (`[3 * n_embd]`), `c_proj.bias` and `c_fc.bias` are 1-D and are copied through untouched. Third, there is no `sanitize` in the shipped port. `Gpt2Layout::detect` probes the shape of `h.0.attn.c_attn.weight` once (`[n_embd, 3 * n_embd]` means the Conv1D layout and needs the transpose, `[3 * n_embd, n_embd]` means an already-transposed MLX conversion and does not), and the transpose is then applied per projection at load, with every other projection shape-checked against that one decision. Deciding by shape rather than by key prefix is what keeps `h.N.attn.c_proj.weight` correct, since it is square (`[n_embd, n_embd]`) and carries no layout signal of its own.**
- Tied output embedding (lm_head shares `wte`).
- GELU MLP.

## Implementation plan

- Base the decoder block on the existing Llama-style block skeleton in `src/models/llama3.rs`, swapping RMSNorm for LayerNorm. LayerNorm already exists in mlxcel (used in the VLM encoders), so reuse that rather than adding a new norm.
- Add a learned position-embedding table added to the token embeddings at the input boundary.
- ~~Add the Conv1D weight transpose (`c_attn`, `c_proj`, `c_fc`) in `sanitize`.~~ **Correction: applied at load rather than in a `sanitize` pass, and over four projections rather than three, per the architecture-note correction above. Detect the layout once from `h.0.attn.c_attn.weight`, then transpose the attention `c_attn` / `c_proj` and the MLP `c_fc` / `c_proj` weights, leaving the biases alone.**

Effort: LOW.

**Delivered and closed.** Implemented by PR #924, merged as `c7e9b6e0d`; `mlxcel arch` now reports the family. Test coverage: 25 unit tests in `src/models/gpt2_tests.rs` plus a detection test. Real-checkpoint validation: 40 greedy tokens from `models/gpt2` reproduce the mlx-lm reference token id sequence exactly. Three real key layouts were additionally validated end to end: the bare raw layout, `distilbert/distilgpt2` under a `transformer.` prefix, and `mlx-community/gpt2-base-mlx` under a `model.` prefix with pre-transposed weights. The unticked boxes below are the original pre-implementation plan and were not maintained during the work; treat the merged PR and its review thread as the record of what shipped.

- [ ] Config struct + serde parse for the `gpt2` config.
- [ ] `from_weights` constructor.
- [ ] Weight-key remap, including the Conv1D transpose of the attention `c_attn` / `c_proj` and the MLP `c_fc` / `c_proj` weights (weights only, decided once from a shape probe at load rather than in a `sanitize` pass; see the correction above).
- [ ] Add the `gpt2` arch-string arm to `src/models/detection.rs`.
- [ ] Register the model in `src/model_metadata.rs` (`for_each_model_registration!`).
- [ ] Add the TP/distributed inference arch-string if applicable.
- [ ] `_tests.rs` unit tests beside the implementation.
- [ ] Update `docs/supported-models.md`.

## Effort

LOW.

inureyes:
## Summary

Adds GPT-2 (`gpt2`) as a fully integrated text model: config parse, loader, decoder graph, detection, registration, `LoadedModel` dispatch, docs, and unit tests. GPT-2 is the first family in this tree to use learned absolute position embeddings, so the KV-cache offset now reaches a position lookup rather than a rotary rotation. A follow-up review pass added a diagnostic for position-table overrun and corrected a docs claim, and a security pass closed one CRITICAL and three HIGH findings plus one MEDIUM in the loader's handling of untrusted `config.json` input.

## What changed

- `src/models/gpt2.rs` (new): `ModelArgs` for the original OpenAI config field names (`n_embd` / `n_head` / `n_layer` / `n_positions` / `layer_norm_epsilon`), the decoder graph, `Gpt2Layout` checkpoint-layout detection, `strip_causal_mask_buffers`, the pure `position_ids` helper, and the `LanguageModel` impl.
- Learned absolute positions: `Gpt2Model::forward` reads `caches[0].offset` before any layer runs and adds `wpe(arange(seq_len) + offset)` to the `wte` token embeddings at the input boundary. There is no RoPE anywhere, so this is the only place the cache offset enters the graph. `position_ids` clamps at the last row of the table, because GPT-2's 1024-token context is reachable from the CLI and an out-of-range `wpe` lookup would otherwise index off the end. `exceeds_position_table` reports when that clamp actually engages, and `Gpt2Model::forward` emits one `eprintln!` per process the first time it does, since a prompt or `-n N` that overruns the context previously degraded output silently on both the CLI and the server.
- Block shape, following `src/models/phi.rs`: LayerNorm with bias (not RMSNorm), one fused `c_attn` QKV projection split three ways on the last axis, multi-head attention (`n_kv_heads == n_head`), a tanh-approximate GELU MLP at four times the model width, and a tied output head through `wte.as_linear` (the checkpoint ships no `lm_head` tensor).
- Conv1D trap: raw HuggingFace exports store `c_attn` / `c_proj` / `c_fc` in the `[in, out]` `Conv1D` layout that `mlxcel_core::layers::Linear` cannot use. `Gpt2Layout::conv1d_linear` transposes those weights to `[out, in]` and copies the 1-D bias vectors through untouched. The transpose decision comes from the `c_attn` weight shape rather than the key prefix, because `attn.c_proj.weight` is square and carries no layout signal of its own; a quantized or otherwise already-sanitized MLX conversion therefore loads unchanged. Every projection in every layer is checked against the shape that decision implies, rather than trusting the rest of the checkpoint to agree with layer 0: a mismatched layer or a degenerate rank previously reached `matmul` unrejected, and an MLX C++ exception crossing the cxx bridge is an uncatchable `std::terminate`.
- Causal-mask-buffer trap: `h.N.attn.bias` is HuggingFace's `[1, 1, n_ctx, n_ctx]` registered mask buffer, not a projection bias. The graph only ever asks for the `.c_attn` and `.c_proj` sub-prefixes so it can never be mis-loaded as a Linear bias, and `strip_causal_mask_buffers` drops it during `load()` in a single `retain` pass keyed on the weight-map key shape, so the cost is proportional to the checkpoint rather than to `n_layer` read out of `config.json`.
- Untrusted-config hardening: `config.json` reaches the loader as third-party input (`mlxcel generate -m <org>/<repo>` downloads and loads a HuggingFace repo in the same command, and nothing between the download and the loader parses the file). `ModelArgs::validate` now rejects `n_head == 0`, an indivisible head split, and out-of-range `n_embd` / `n_layer` / `n_positions` at load, before `n_layer` sizes the `Vec::with_capacity` in `from_weights` and before any derived `as i32` cast can truncate. `validate_position_table` checks the loaded `wpe` table's row count against `n_positions` and rejects a config that overstates it, since `position_ids` clamps to a config-derived bound and MLX's embedding gather range-checks a negative index but not a positive one. `last_position_row` saturates instead of truncating, since a truncating cast can turn the clamp bound negative and panic `clamp` (a process abort on a server worker thread).
- `src/models/mod.rs`: module declaration, `Gpt2Model` re-export, and `ModelType::Gpt2` in all four places that must agree (the enum, `ALL_MODEL_TYPES`, the `metadata()` display table, and the `all_variants!` exhaustiveness list in the unit tests).
- `src/models/detection.rs`: `"gpt2" => Ok(ModelType::Gpt2)`.
- `src/model_metadata.rs`: standard config-backed registration (`dir_loader`, `args`, `weight_builder`, `wrap`), so both the directory route and the owned-weights route go through the shared registry.
- `src/loaded_model.rs`: `LoadedModel::Gpt2` variant and its `delegate_language_model!` arm.
- `src/distributed/tensor_parallel/inference.rs`: `"gpt2"` fallback architecture string. GPT-2's fused `c_attn` and Conv1D key names have no shard rules, so the planner rejects the string; the entry exists to keep the exhaustive dispatch table total.
- `src/models/gpt2_tests.rs` (new, 25 tests) and one new detection test in `src/models/detection_tests.rs`.
- `docs/supported-models.md`: GPT-2 entry describing the position-embedding path, both layout traps, the load-time position-table and cross-layer shape checks, and the three checkpoint layouts validated end to end. The entry also corrects an earlier draft that claimed CLI generation needs `--no-chat-template`; the flag is optional because a checkpoint that ships no chat template already renders the prompt verbatim.

## Test plan

- [x] `cargo clippy --release --lib --tests --features metal,accelerate` (no new findings; the two warnings in `src/multimodal/host_preprocessor*.rs` are pre-existing on `main` from #915 and are untouched here)
- [x] `cargo test --release --lib --features metal,accelerate models::gpt2` (25 tests)
- [x] `cargo test --release --lib --features metal,accelerate models::detection_tests::gpt2`
- Unit coverage: the real `openai-community/gpt2` config field set and the bare-config defaults; scalar and list `eos_token_id`; config-magnitude validation against seven hostile scalar cases; `position_ids` at prefill, at a decode step past the prefill, on a multi-token continuation, clamped at the end of the table, and saturating instead of panicking on an absurd table size; position-table-overrun reporting at the exact boundary; layout detection for the bare, `transformer.`-prefixed, and `model.`-prefixed key shapes plus both rejection paths; the weight-only Conv1D transpose across all four projections with bias shapes asserted; the mask-buffer strip matching key shape rather than the config layer count, and leaving every real projection bias in place; rejection of a layer that disagrees with layer 0 about the Conv1D layout, a degenerate projection shape, a `wpe` table that is shorter than `n_positions` claims, and a `wpe` table of the wrong width; and a tiny single-layer end-to-end prefill-then-decode that asserts the logits shape and the advancing cache offset.
- Real-checkpoint validation: token-exact against the mlx-lm reference on `models/gpt2`; `openai-community/gpt2` (bare keys), `distilbert/distilgpt2` (`transformer.`-prefixed) and `mlx-community/gpt2-base-mlx` (`model.`-prefixed, already transposed) all load and generate the same text for the same prompt; the security fixes were confirmed not to change generated token ids on `models/gpt2` and `mlx-community/gpt2-base-mlx`.

Closes #833


inureyes:

\`mlxcel arch\` reports the family: \`GPT-2 (learned absolute positions, Conv1D weights)\`.

Generation against the raw HuggingFace \`models/gpt2\` checkpoint (F32, 160 tensors, tied head, no chat template) is **token-exact against the mlx-lm reference** for 40 greedy tokens. Command: \`mlxcel generate -m models/gpt2 -p "The capital of France is" -n 40 --no-chat-template\`, throughput 156.4 tok/s.

Reference (`mlx_lm.stream_generate`, `temp=0.0`, same raw prompt) produced the identical continuation and the identical token id sequence: \`262 3139 286 262 4141 2066 11 290 262 3139 286 262 4141 2066 318 262 3139 286 262 4141 2066 13 198 198 464 4141 2066 318 262 3139 286 262 4141 2066 13 198 198 464 4141 2066\`.

This matters specifically for the learned absolute position embeddings, which are new to the text path in this tree: a wrong prefill/decode offset still yields fluent English, so the token-exact diff rather than an eyeball check is what establishes the offset is threaded correctly.

inureyes:

### Intent
Add GPT-2 (`gpt2`) as a fully integrated text model: config, loader, decoder graph, detection, registration, dispatch, docs and tests (issue #833).

### Verification performed beyond the existing token-exact gate

The bare-key `openai-community/gpt2` path was already proven token-exact against mlx-lm. This review covered what that did not:

- **`transformer.`-prefixed layout, real checkpoint.** `distilbert/distilgpt2` loads and generates. To prove the prefix handling is layout-equivalent rather than merely plausible, a copy with the `transformer.` prefix mechanically stripped produced byte-identical output.
- **`model.`-prefixed layout, real checkpoint.** `mlx-community/gpt2-base-mlx` (the only GPT-2 in that org; `mlx-community/gpt2` does not exist) ships `model.`-prefixed, already-transposed `[2304, 768]` `c_attn` and no mask buffers. It loads unchanged and generates text identical to the raw HuggingFace checkpoint, confirming both the prefix arm and the shape-based Conv1D short-circuit.
- **Server path.** `mlxcel-server` serves `/v1/models` and `/v1/completions` for `gpt2`, including two concurrent requests through the batch scheduler.
- **Prefix-cache offset threading.** A warm-cache server request whose prompt extends an earlier cached prefix produced output identical to a cold CLI run of the same prompt, confirming the `caches[0].offset` to `wpe` lookup is correct under prefix-cache adoption, not just on a cold cache.
- **Registration completeness.** Every surface a peer dense text model touches (`ModelType`, `ALL_MODEL_TYPES`, `metadata()`, the `all_variants!` list, `detection.rs`, `model_metadata.rs`, `loaded_model.rs`, `tensor_parallel/inference.rs`) is covered; the file sets for `ModelType::Gpt2` and `LoadedModel::Gpt2` match `Olmo3`/`StarCoder2` exactly. `mlxcel arch` reports the family.

### Findings Addressed
- [x] Over-running the 1024-row `wpe` table degraded output silently on both CLI and server (MEDIUM). A 1600-token prompt returns `the jumps the jumps the jumps` with no diagnostic. Added `exceeds_position_table()` and a once-per-process message from `Gpt2Model::forward`. `eprintln!` rather than `tracing::warn!` because only `mlxcel-server` installs a tracing subscriber and the overrun is reachable from the CLI, where a `tracing` event is a no-op.
- [x] `docs/supported-models.md` claimed CLI generation needs `--no-chat-template` (MEDIUM, factually wrong). `ChatTemplateProcessor::from_model_path` returns `Ok(None)` for a checkpoint with no template and `resolve_cli_prompt` then passes the prompt through verbatim; both `gpt2` and `distilgpt2` generate correctly without the flag. Entry rewritten with the three validated layouts and the position-table bound.

### Reviewed and deliberately left as-is
- **Clamping instead of erroring past `n_positions`** is the right call. mlxcel already has the convention this needs: `context_window_from_config` reads `n_positions`, so the default `-n -1` resolves to `n_positions - prompt_len` (measured: exactly 1020 tokens from a 4-token prompt). No new convention was invented, and a hard error inside `forward` would fail an in-flight server request where the rest of the tree degrades gracefully. Only the missing diagnostic needed fixing.
- **Four accepted key prefixes** is not scope creep. Three map to real in-the-wild checkpoints, all now validated end to end. Only `model.transformer.` is unexercised, and it costs one string comparison.
- **`strip_causal_mask_buffers` on the directory route only.** The claim is accurate. The graph only ever requests the `.c_attn` / `.c_proj` sub-prefixes, so `h.N.attn.bias` can never be mis-loaded as a projection bias, and a unit test pins that. The weight route is reachable only through `load_model_with_adapter`, where the cost is the retained buffers (about 48 MB for a 12-layer, 1024-context checkpoint), not a correctness problem.
- **623-line `src/models/gpt2.rs`.** Consistent with every peer model file.

### Follow-ups worth a separate issue (not fixed here)
- **LoRA fusion is not Conv1D-aware.** `fuse_lora_weights_into` does `add(base, delta)` with no shape check. On a raw HuggingFace GPT-2 the base weights are still `[in, out]` at fusion time (the transpose happens later, inside `Gpt2Layout::conv1d_linear`), so a matching adapter key would broadcast-fail on `c_attn` / `c_fc` and, worse, silently add a transposed delta to the square `c_proj`. In practice mlx-lm trains adapters against the sanitized tree, so real adapter keys are `model.`-prefixed and only resolve against an already-transposed checkpoint where fusion is correct, and PEFT adapters use a `lora_A.weight` suffix that never matches. Narrow, but unguarded.
- **`n_positions` for the clamp comes from `config.json`, not from the loaded `wpe` tensor.** A config that overstates the table size would clamp to a row that does not exist. No real checkpoint does this, and deriving the bound from `wpe.weight`'s axis 0 (valid for both plain and packed-quantized layouts) would make the guard self-consistent, matching this PR's own choice to take the Conv1D decision from the weight shape rather than the config.
- **`n_ctx` and `intermediate_size()` are only read by tests.** Harmless and mirrors mlx-lm's `ModelArgs`; noted so a future reader does not assume they gate anything.

### Verification
- [x] Integrated into project code flow (CLI, server, batch scheduler, prefix cache)
- [x] Tests pass (`models::gpt2` 18, `models::detection_tests` 22, `model_metadata` 8, `models::metadata_tests` 4; `cargo fmt --check` clean; clippy adds no new findings, the two `host_preprocessor*` warnings are pre-existing on `main`)

inureyes:

Correctness was already established, so this pass covered only the security and performance surface. Six findings, all fixed in ea6a002e4. Two performance observations are reported without a change.

### Trust boundary

`config.json` is untrusted input at this loader. `mlxcel generate -m <org>/<repo>` resolves an arbitrary HuggingFace repo id, downloads it, and loads it in the same command; the download layer hardens repo ids, filenames and transport but never parses the file, and nothing between it and `Gpt2Model::load` range-checks a single field. Every `ModelArgs` field arrives exactly as the checkpoint author wrote it.

### CRITICAL: out-of-bounds read of the learned position table

`src/models/gpt2.rs` `position_ids` / `Gpt2Model::forward`

`position_ids` clamps to `n_positions - 1`, and `n_positions` comes from `config.json`, not from `wpe.weight`. Nothing else bounds the lookup.

I probed the gather behind `UnifiedEmbedding::forward` (`mlx::core::take`) against an 8-row table. It wraps a negative index correctly (`-1` returned row 7) but performs no range check on a positive one: index 64 returned a value belonging to neither row 0 nor row 4 nor row 7, and indices 8 and 100000 returned zeros. No fault, no error, no clamp. The reads land wherever the offset points.

Failure scenario: a checkpoint ships `wpe.weight` with 2 rows and a `config.json` declaring `n_positions: 16`. The model builds without complaint, and an ordinary 8-token prompt gathers positions 2 through 7 from past the end of the buffer. I ran exactly this: the prefill completed and returned logits, so the out-of-bounds values flow into the output. `context_window_from_config` also reads `n_positions`, so an inflated value widens the token budget and lets a long generation walk further past the table.

Fix: `validate_position_table` reads the actual row count and width of the loaded table and rejects a config that overstates either. `Gpt2Model::n_positions` is now private, so the invariant established at construction cannot be broken afterwards by a caller.

### HIGH: unbounded strip loop hangs the load path

`src/models/gpt2.rs` `strip_causal_mask_buffers`

The function probed `<prefix>h.<layer>.attn.bias` for four prefixes and every `layer` in `0..n_layer`, formatting a key per iteration, with no early exit. It runs in `load()` before `from_weights`, so no weight lookup could reject the config first, and `n_layer` had no ceiling.

Failure scenario: `"n_layer": 1000000000000` produces four trillion `format!` plus hash lookups. The allocation footprint stays flat, so there is no OOM to kill it; the load simply never returns. In the server this hangs startup.

Fix: a single `retain` pass over the weight map that matches the key shape. The cost is now proportional to the checkpoint, not to a config field. `is_causal_mask_buffer_key` requires `h.` to start a path segment, so `blah.0.attn.bias` is not mistaken for layer 0, and the real `.c_attn.bias` / `.c_proj.bias` / `.c_fc.bias` biases never match.

### HIGH: panic from a truncating cast in the position helpers

`src/models/gpt2.rs` `position_ids`, `exceeds_position_table`

`n_positions.saturating_sub(1) as i32` truncates instead of saturating. `1usize << 32` becomes `-1` and `(1usize << 40) - 1` becomes `-1`, and `i32::clamp` panics when `min > max`.

Failure scenario: `"n_positions": 4294967296` panics on the first forward. Release builds unwind, and every core inference worker re-imposes fail-fast with `catch_unwind` plus `process::abort` (ADR 0003), so this is a process abort.

Fix: `last_position_row` saturates into `i32`. Both helpers are public, so neither now depends on its caller having validated the input.

### HIGH: no magnitude validation on the architecture scalars

`src/models/gpt2.rs` `ModelArgs`

`n_layer` reached `Vec::with_capacity(args.n_layer)` unchecked, and `n_embd` / `n_head` fed `as i32` casts that truncate (`n_head: 4294967308` casts to 12 while the divisibility check saw the untruncated value; `3 * n_embd` overflows for a large `n_embd`).

Fix: `ModelArgs::validate` following the `SparseAttentionConfig::validate` precedent in `src/models/minimax_m3_config.rs`. It rejects `n_head == 0` before the divisibility check (`0.is_multiple_of(0)` is true, so the old combined check relied on the short-circuit), an indivisible head split, and out-of-range `n_embd` / `n_layer` / `n_positions`. It runs on both routes: after the config parse in `load()`, before the checkpoint is read, and at the top of `from_weights` for the owned-weights route. The ceilings (1024 layers, `n_embd` 65536, 4M positions) sit orders of magnitude above GPT-2 XL and above every dense decoder in this tree.

### MEDIUM: cross-layer weight-shape trust

`src/models/gpt2.rs` `Gpt2Layout::conv1d_linear`

The Conv1D layout is probed once from `h.0.attn.c_attn.weight` and then applied to every projection in every layer, with no check that the rest of the checkpoint agrees.

Failure scenario: `h.0` is Conv1D `[in, out]` and `h.5` is already `[out, in]`. Layer 5 gets transposed a second time and the mismatch surfaces at `matmul`. An MLX C++ exception crossing the cxx bridge is `std::terminate`, not a recoverable error. I hit this incidentally during the review and the test binary died with SIGABRT and no Rust-level diagnostic. The same applies to a degenerate rank (`mx::transpose` with no axes reverses any rank without complaining) and to a mis-sized projection bias.

Fix: `conv1d_linear` takes the expected input and output widths and validates the weight shape, rank, and bias length against the detected layout, returning a load error that names the tensor. Quantized projections are still handed to `UnifiedLinear` unchecked, since a packed shape matches neither float layout.

### Performance: assessed, no change

The per-decode-step host `Vec<i32>` plus `from_slice_i32` for the position ids is a single-element allocation and a small host-to-device copy against 12 layers of attention and MLP work. It is not measurable at this model size and is the clearest expression of the offset threading, so it stays.

The Conv1D transpose happens once at load, not per forward: it builds one lazy node per projection, MLX evaluates it on the first forward, and `eval` detaches the node from its input afterwards. Worth noting for a future large-variant port: until that first eval, the transposed output and the original weight are both live, so peak memory during load through first token is roughly double the projection weights. That is immaterial for GPT-2 124M and would be worth an `eval_all` plus `detach_all` at load (the pattern already used in `src/models/sanitize.rs`) if GPT-2 XL becomes a target.

`wte.as_linear` computes logits over the whole prefill sequence. That matches every other decoder in this tree, so it is not a finding here.

### Validation

- `cargo check --release --lib --tests --features metal,accelerate`
- `cargo clippy --release --lib --tests --features metal,accelerate` (no new findings; the two `src/multimodal/host_preprocessor*.rs` warnings are pre-existing on `main`)
- `cargo test --release --lib --features metal,accelerate models::gpt2` (25 passed, 8 new)
- `cargo test --release --lib --features metal,accelerate models::detection` (22 passed)
- Real checkpoints: `models/gpt2` and `mlx-community/gpt2-base-mlx` both load after the change and produce identical greedy token ids for the same prompt, so the added validation does not reject a genuine checkpoint and the numerics are untouched.

inureyes:
## PR Finalization Complete

### Tests
Verified the 25 unit tests in `src/models/gpt2_tests.rs` (17 from the initial port, 8 from the security pass) plus the detection test in `src/models/detection_tests.rs` are coherent rather than overlapping: each test file section maps to a distinct behavior and nothing was left duplicated across the three commits. For every CRITICAL/HIGH/MEDIUM fix from the security pass, I temporarily reverted just that fix in isolation and re-ran the suite to confirm the paired regression test actually fails without it: the position-table row-count check (`from_weights_rejects_a_config_that_overstates_the_position_table`, `..._of_the_wrong_width`), the config magnitude ceilings (`config_validation_rejects_hostile_scalars`), the cross-layer Conv1D shape check (`from_weights_rejects_a_layer_that_disagrees_with_layer_zero_about_conv1d`, `..._a_degenerate_projection_shape`), and the saturating cast (`position_ids_saturate_instead_of_panicking_on_an_absurd_table_size`, which genuinely panics with `assertion failed: min <= max` under the old truncating cast) all failed exactly as expected, and nothing else did. The unbounded-strip-loop fix has no literal timing test (impractical in a unit suite) but is structurally enforced: `strip_causal_mask_buffers` no longer takes `n_layer` at all, so no caller can regress to the old per-layer probe. No new tests were added; existing coverage already isolates each fix.

### Documentation
`docs/supported-models.md`'s GPT-2 entry was accurate for the runtime clamp/warn behavior but predated the security commit, so it did not mention that a checkpoint whose `wpe` row count or per-layer projection shapes disagree with `config.json` is now rejected at load. Added one clause documenting that load-time consistency check. No Korean translation exists for this file in the repository, so none was added.

### Lint/Format
`cargo fmt --all -- --check` and `cargo clippy --release --lib --tests --features metal,accelerate -p mlxcel` are both clean; the only warnings present are the two pre-existing `host_preprocessor*.rs` warnings from #915, untouched by this PR.

### PR body
Updated to describe what actually shipped across all three commits (the position-table-overrun diagnostic, the docs correction, and the five security fixes), updated the test count to 25, and kept `Closes #833`.

PR #926, opened 2026-07-26
feat(models): add GPT-BigCode (gpt_bigcode) text model support

### LINKED ISSUE #834: feat(models): add GPT-BigCode (gpt_bigcode) text model support
## Summary

GPT-BigCode is the architecture behind the BigCode StarCoder / SantaCoder code models. It is a GPT-2 derivative whose defining change is Multi-Query Attention, which makes it cheap to run at long context. mlx-lm supports it; mlxcel does not yet load the `gpt_bigcode` architecture. Porting it adds code-model coverage and is straightforward once the GPT-2 skeleton is in place, since MQA is just GQA with a single KV head, which mlxcel already supports everywhere.

## Upstream reference

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gpt_bigcode.py

## Public checkpoint

`bigcode/gpt_bigcode-santacoder` or a StarCoderBase checkpoint. Verify mlx-community checkpoint availability (confirm an existing MLX conversion) before validation.

## Architecture notes

GPT-BigCode with these distinctive traits:

- Multi-Query Attention: with `multi_query=True`, a single KV head is shared across all query heads.
- Learned absolute position embeddings (added to token embeddings). No RoPE.
- `nn.LayerNorm` (with bias), not RMSNorm.
- Fused `c_attn` projection producing `[dim + 2 * kv_dim]` (query block plus a single shared key and value block). ~~stored in the Conv1D layout like GPT-2~~ **Correction (verified against `bigcode/gpt_bigcode-santacoder`): this is NOT the Conv1D layout.** HuggingFace `GPTBigCode` uses `nn.Linear`, so `c_attn.weight` is `[2304, 2048]` = `[out, in]` already, and upstream `gpt_bigcode.py` has no `sanitize` method at all. Adding a transpose would silently corrupt every projection.

## Implementation plan

- Reuse the gpt2 skeleton (learned position embeddings + LayerNorm). No Conv1D transpose, per the correction above.
- MQA maps directly onto mlxcel GQA with `n_kv_heads = 1`, which is already supported across the attention path; no new attention code should be needed.
- Split the fused `c_attn` output into the query block and the single shared KV block, respecting the `[dim + 2 * kv_dim]` layout.

Dependency note: land this after or alongside the gpt2 port (lablup/mlxcel#833), since it reuses the gpt2 learned position embeddings, LayerNorm wiring, ~~and Conv1D transpose~~ and shared validation helpers. **Correction (verified against the merged `src/models/gpt_bigcode.rs`, PR #926, `4ed5e3b78`): the Conv1D transpose is struck from this list because there is none to reuse, per the correction in the architecture notes above, where it was already established that GPT-BigCode is `nn.Linear` and not the Conv1D layout. What the merged port actually imports from `src/models/gpt2.rs` is `position_ids` and `exceeds_position_table` (the learned position-embedding helpers), `layer_norm_from_weights` (the biased LayerNorm loader), and `dim_eq` plus `validate_embedding_table` (the shared load-time validation helpers). Nothing layout-related crosses between the two modules.**

Effort: LOW-MEDIUM.

**Delivered and closed.** Implemented by PR #926, merged as `4ed5e3b78`; `mlxcel arch` now reports the family. Test coverage: 20 unit tests in `src/models/gpt_bigcode_tests.rs`. Real-checkpoint validation: 60 greedy tokens from `models/gpt_bigcode-santacoder` reproduce the mlx-lm reference exactly, exercising the MQA path end to end. GPT-2 was re-verified byte-identical after this PR's shared-helper refactor. The unticked boxes below are the original pre-implementation plan and were not maintained during the work; treat the merged PR and its review thread as the record of what shipped.

- [x] Config struct + serde parse for the `gpt_bigcode` config (including `multi_query`).
- [x] `from_weights` constructor.
- [x] Fused `c_attn` `[dim + 2 * kv_dim]` split at offsets `[dim, dim + kv_dim]`. No `sanitize` and no Conv1D transpose: the checkpoint is already `[out, in]`, and a transposed weight is rejected by name instead.
- [x] Add the `gpt_bigcode` arch-string arm to `src/models/detection.rs`.
- [x] Register the model in `src/model_metadata.rs` (`for_each_model_registration!`), plus `src/models/mod.rs` (enum, `ALL_MODEL_TYPES`, `metadata()` table, `all_variants!`) and `src/loaded_model.rs`.
- [x] Add the TP/distributed inference arch-string if applicable (added to keep the dispatch table total; the planner rejects it, since the fused `c_attn` has no shard rules and a single KV head cannot be split).
- [x] `_tests.rs` unit tests beside the implementation (20 tests).
- [x] Update `docs/supported-models.md`.

## Effort

LOW-MEDIUM.

inureyes:
## Summary

Adds GPT-BigCode (`gpt_bigcode`), the architecture behind the BigCode StarCoder and SantaCoder code models, as a fully integrated text model: config parse, loader, decoder graph, detection, registration, `LoadedModel` dispatch, docs and unit tests. It is a GPT-2 derivative, so it builds on the block skeleton landed in #924 rather than restating it.

## Correction to the issue body

The issue says `c_attn` / `c_proj` are "stored in the Conv1D layout like GPT-2". **That is wrong, and following it would have silently corrupted every projection.** HuggingFace `GPTBigCode` builds its projections with `nn.Linear`, not the `Conv1D` that GPT-2 uses, so the weights are already `[out, in]`, which is the layout `mlxcel_core::layers::Linear` wants. Shapes read directly from the `bigcode/gpt_bigcode-santacoder` SafeTensors header:

```
transformer.h.0.attn.c_attn.weight   F16 [2304, 2048]   = [out, in]
transformer.h.0.attn.c_proj.weight   F16 [2048, 2048]
transformer.h.0.mlp.c_fc.weight      F16 [8192, 2048]   = [out, in]
transformer.h.0.mlp.c_proj.weight    F16 [2048, 8192]
```

Compare GPT-2, whose `c_attn.weight` is `[768, 2304]` = `[in, out]` and does need the transpose. Upstream `gpt_bigcode.py` correspondingly has no `sanitize` method at all. There is therefore no transpose in this port, and `load_linear` rejects a weight that arrives in the transposed orientation by name with an error that explains why GPT-BigCode is not Conv1D, so the claim cannot be reintroduced silently later. `a_conv1d_oriented_weight_is_rejected_and_named` pins that behavior.

## What changed

- `src/models/gpt_bigcode.rs` (new): `ModelArgs` for the GPT-2-style config field names plus `multi_query` and `n_inner`, the decoder graph, `detect_prefix`, `load_linear`, and the `LanguageModel` impl.
- **Multi-Query Attention**, the defining feature. With `multi_query: true` exactly one KV head is shared by every query head, so the fused `c_attn` projection produces `n_embd + 2 * kv_dim` features rather than `3 * n_embd` and is split at `[n_embd, n_embd + kv_dim]`, which is *not* an even three-way split. On santacoder that is 2048 + 2 * 128 = 2304 for 16 query heads of width 128 over a single KV head. This maps onto the existing GQA attention path with `n_kv_heads = 1`; MLX broadcasts the shared KV head across the query heads exactly as it does for the GQA families, so no new attention code is involved. `ModelArgs::qkv_split_offsets` exposes the split as a pure function so the uneven offsets are unit-tested without a checkpoint, and the tiny forward tests assert the cached KV head count directly (1 under MQA, `n_head` when `multi_query` is false).
- **MLP width from `n_inner`** (8192 on santacoder), not a hardcoded `4 * n_embd`. An absent or `null` `n_inner` falls back to `4 * n_embd`, matching HuggingFace.
- **No causal-mask buffer to strip.** HuggingFace registers GPT-BigCode's mask with `persistent=False`, so unlike GPT-2 no `h.N.attn.bias` tensor reaches the checkpoint; the santacoder export carries 292 tensors, none matching that shape. GPT-2's `strip_causal_mask_buffers` pass is deliberately not copied.
- **Single key prefix.** Both HuggingFace and upstream mlx-lm nest the decoder under a `transformer` submodule, and upstream ships no `sanitize` that could rename it, so GPT-2's multi-prefix probe is not copied either; `detect_prefix` accepts `transformer.` and rejects anything else with a message naming the key it looked for.
- **Untrusted-config hardening built in from the start**, following what #924's security pass arrived at. `ModelArgs::validate` rejects `n_head == 0`, an indivisible head split, and out-of-range `n_embd` / `n_layer` / `n_positions` / `vocab_size` / `n_inner` at load, before `n_layer` sizes the `Vec::with_capacity` in `from_weights` and before any derived `as i32` cast can truncate. Every gather is bounded by the real tensor shape rather than a config field: both the `wpe` and the `wte` table are checked against `n_positions` and `vocab_size` at load, because MLX wraps a negative gather index but does not range-check a positive one, so a config that overstates a table turns an ordinary prompt into an out-of-bounds read whose result reaches the logits. Every projection, bias and norm is checked against the shape the config implies in *every* layer, not just layer 0, since an MLX C++ exception crossing the cxx bridge is an uncatchable `std::terminate` rather than a load error. There is no unbounded probe loop over `0..n_layer` on the load path.
- **Output head.** Every published checkpoint ties the head to `wte` and ships no `lm_head` tensor, which is why `tie_word_embeddings` defaults to true. An untied config must actually carry the tensor; silently falling back to the tied path would produce logits from the wrong matrix. GPT-BigCode has no family-wide `<|endoftext|>` id (santacoder uses 49152, StarCoder uses 0), so `eos_token_ids` falls back from `eos_token_id` to `bos_token_id` and then to empty rather than guessing an id that could truncate at an ordinary token.
- `src/models/gpt2.rs`: three helpers are made shared rather than duplicated, each carrying the `// Used by:` comment the project convention asks for. `dim_eq` and `layer_norm_from_weights` become `pub(crate)`, with the latter now taking the model width and epsilon directly and validating the norm shape; `validate_position_table` is generalized to `validate_embedding_table` so it covers the token table as well as the position table. GPT-2 behavior is unchanged apart from the added norm-shape rejection.
- `src/models/mod.rs`: module declaration, `GptBigCodeModel` re-export, and `ModelType::GptBigCode` in all four places that must agree (the enum, `ALL_MODEL_TYPES`, the `metadata()` display table, and the `all_variants!` exhaustiveness list).
- `src/models/detection.rs`: `"gpt_bigcode" => Ok(ModelType::GptBigCode)`, plus a detection test asserting it does not fall through to the `gpt2` arm.
- `src/model_metadata.rs`: standard config-backed registration, so both the directory route and the owned-weights route go through the shared registry. `src/loaded_model.rs`: variant and `delegate_language_model!` arm. `src/distributed/tensor_parallel/inference.rs`: `"gpt_bigcode"` fallback architecture string; the fused `c_attn` key names have no shard rules and the multi-query KV block cannot be split across ranks anyway. `generate_shard_plan` has no `gpt_bigcode` arm and still produces a generic transformer plan through its fallback, but `validate_supported_runtime`'s `runtime_kind_for` match has no arm for `ModelType::GptBigCode` either, so TP is refused with an unsupported-architecture error before any load is attempted; the entry here exists to keep the dispatch table total.
- `src/models/gpt_bigcode_tests.rs` (new, 20 tests) and `docs/supported-models.md`.

## Follow-up fix: quantized projection output width (`165f8199c`)

`load_linear` returned early for any projection carrying a `.scales` key, so a packed weight reached the graph with neither its output width nor its bias shape checked. That carve-out is wrong for this family specifically, because `Attention::forward` slices the fused `c_attn` output at offsets derived from `config.json` (`[n_embd, n_embd + kv_dim]`), not from the tensor itself. A packed `c_attn` wider than the config claims leaves every slice in bounds, so the model loads and decodes fluent-looking output while K and V are silently read from arbitrary interior channels; the reachable case is a quantized conversion whose `config.json` omits `multi_query` (which then defaults to `true`) against a weight packed for full multi-head attention. A packed weight narrower than the config claims takes the other branch instead, where MLX's `slice` clamps the stop index and the following `reshape` then throws, crossing the cxx bridge as an uncatchable `std::terminate`.

Quantization packs the input axis only, so the row count is untouched by packing and is now checked on the quantized path the same way it already was on the float path, along with the 1-D bias; the packed input width stays unchecked, since no float layout applies to it and `UnifiedLinear` reconciles that half of the contract itself. `a_quantized_projection_skips_the_packed_input_width_but_not_the_output_width` (renamed and extended from the earlier, now-inaccurate `load_linear_delegates_a_quantized_projection_unchecked`) covers a wider packed weight, a narrower one, and a mis-sized packed bias.

## Finalization polish

Corrected two doc comments this chain had left inconsistent: `src/models/gpt2.rs` described the shared `mlxcel_core::utils::gelu_approx` helper as the tanh approximation, which contradicts both the C++ implementation in `mlx_cxx_bridge.cpp` (erf-based) and this port's own, correct description of the same helper in `src/models/gpt_bigcode.rs`; and the `src/distributed/tensor_parallel/inference.rs` comment for `ModelType::GptBigCode` named the wrong mechanism for the TP refusal (corrected wording is in the `What changed` bullet above). Also extended the `docs/supported-models.md` GPT-BigCode entry, written before `165f8199c` landed, to describe the quantized output-width check that commit added.

## Test plan

- [x] `cargo clippy --release --lib --tests --features metal,accelerate` (no new findings; the two warnings in `src/multimodal/host_preprocessor*.rs` are pre-existing on `main` and untouched here)
- [x] `cargo test --release --lib --features metal,accelerate models::gpt_bigcode` (20 passed)
- [x] `cargo test --release --lib --features metal,accelerate models::gpt2` (25 passed, unchanged by the shared-helper refactor)
- [x] `cargo test --release --lib --features metal,accelerate detection_tests` (23 passed)
- [x] `cargo test --release --lib --features metal,accelerate models::metadata_tests` (4 passed: the registration and `ALL_MODEL_TYPES` exhaustiveness gates)
- [x] `cargo test --release --lib --features metal,accelerate distributed::tensor_parallel` (338 passed, 4 ignored; unaffected by the comment-only inference.rs change)
- [x] Reverted only the `src/models/gpt_bigcode.rs` half of `165f8199c` while keeping its paired test, confirming `a_quantized_projection_skips_the_packed_input_width_but_not_the_output_width` is the sole failure in the `models::gpt_bigcode` suite, then restored the fix.
- Unit coverage: the real santacoder config field set and the bare-config defaults; scalar, list and empty-list token-id fields with the bos fallback; config-magnitude validation against nine hostile scalar cases; the MQA head count, `kv_dim`, `c_attn` output width and the uneven split offsets, asserted to differ from both `3 * n_embd` and the even thirds; the `multi_query: false` degeneration; `n_inner` versus the `4 * n_embd` fallback including explicit `null`; prefix detection and its rejection path; every projection loading with the stored `[out, in]` shape unchanged; rejection of a Conv1D-oriented weight, a rank-3 weight, a mis-sized bias, a missing projection, a later layer that disagrees with the config, a `wpe` or `wte` table shorter than the config claims, a table or norm of the wrong width, and an untied config with no `lm_head`; a quantized projection that skips only the packed input-width check while its output row count (both wider and narrower than the config implies) and its bias are still validated and rejected by name; and two tiny end-to-end prefill-then-decode runs asserting the logits shape, the advancing cache offset and the cached KV head count under both `multi_query` settings.
- Not run here: the release build and real-checkpoint generation against `models/gpt_bigcode-santacoder`, which the maintainer is running separately.

Closes #834


inureyes:

**GPT-BigCode is token-exact against the mlx-lm reference.** `mlxcel generate -m models/gpt_bigcode-santacoder -p "def fibonacci(n):" -n 60 --no-chat-template` reproduces the reference continuation exactly (correct recursive Fibonacci plus the trailing `print(fibonacci(10))`), at 48.9 tok/s. This exercises the MQA path end to end: one shared kv head across all 16 query heads, with the fused `c_attn` split at `[dim, dim + kv_dim]` rather than a 3-way even split.

`mlxcel arch` reports `GPT-BigCode (StarCoder / SantaCoder, multi-query attention)`.

**No GPT-2 regression from the shared-helper refactor.** This PR lifts `dim_eq`, `layer_norm_from_weights`, and the generalized `validate_embedding_table` out of the merged `src/models/gpt2.rs`, so I re-ran the GPT-2 gate: generation from `models/gpt2` is byte-identical to the output on merged `main` before the refactor.

**On the GELU question.** The implementation matches upstream mlx-lm, which applies the erf-based `nn.gelu` even though this config declares `gelu_pytorch_tanh`. I checked whether that distinction is observable here by monkey-patching the mlx-lm reference to use `nn.gelu_approx` (the tanh form) and regenerating: both variants emit **the identical 60 token ids**. So the sub-1e-3 activation difference flips no argmax on this checkpoint, and matching upstream is not masking a defect. Worth keeping in mind only if a future GPT-BigCode checkpoint shows divergence from HuggingFace output.

inureyes:

### Intent

Add GPT-BigCode (`gpt_bigcode`, StarCoder / SantaCoder) as a fully integrated text model, reusing the GPT-2 block skeleton from #924 and mapping Multi-Query Attention onto the existing GQA path with `n_kv_heads = 1`.

### Findings Addressed

None. No CRITICAL or HIGH finding was identified, so no code was changed on this branch.

### Verified Rather Than Accepted

- **Shared-helper generalization is not a weakening.** `validate_position_table` to `validate_embedding_table` (`src/models/gpt2.rs:834`) is logic-identical to the merged form; only `n_positions` / `n_embd` became parameters. GPT-2's call site (`gpt2.rs:777`) passes the original values, `src/models/gpt2_tests.rs` is untouched by this PR, and both regression tests still assert on the original substrings (`n_positions`, `n_embd`). 25 GPT-2 tests pass unchanged.
- **The Conv1D rejection cannot fire on a legitimate checkpoint.** `load_linear` (`src/models/gpt_bigcode.rs:406`) reaches the rejection only when `is_expected` is false. For a square projection `[in, out]` and `[out, in]` are the same predicate, so `is_expected` is true and the branch is unreachable. The transposed hint is text on an already-failing path.
- **MQA split matches upstream exactly.** `mlx_lm/models/gpt_bigcode.py` splits at `[dim, dim + kv_dim]` giving Q, then K, then V. `gpt_bigcode.rs:465-467` produces the same three ranges in the same order; K and V are not swapped. `cache.update_and_fetch(k, v)` argument order is correct.
- **No mask buffer to strip, confirmed against the checkpoint.** The `models/gpt_bigcode-santacoder` SafeTensors header carries 292 tensors, zero keys ending `.attn.bias`, and zero tensors of rank greater than 2. `c_attn.weight` is `[2304, 2048]`, `wte` is `[49280, 2048]`, `wpe` is `[2048, 2048]`, no `lm_head.weight`, all keys `transformer.`-prefixed. Every claim in the PR body holds.
- **EOS fallback has no falsy-value bug.** `eos_token_ids` (`gpt_bigcode.rs:336`) selects on `!ids.is_empty()`, not on the id value, so StarCoder's `eos_token_id: 0` yields `[0]` correctly. The empty-list outcome is reachable only for a config declaring neither `eos_token_id` nor `bos_token_id`; `-n -1` still resolves through `context_window_from_config`, which reads `n_positions`, so generation remains bounded. Declining to invent a family-wide id is the right call here.
- **Config-trust hardening is real.** `validate()` runs before `Vec::with_capacity` on both routes (`load` and `from_weights`), `n_head <= n_embd <= 65536` is implied by the divisibility check so every `as i32` cast is in range, there is no probe loop over `0..n_layer`, and both `wte` and `wpe` are bounded against real tensor rows. The `wte` check is new relative to GPT-2.
- **Registration surfaces are complete.** `ModelType`, `ALL_MODEL_TYPES`, `metadata()`, `all_variants!`, `detection.rs`, `model_metadata.rs`, `loaded_model.rs`, `distributed/tensor_parallel/inference.rs` and `docs/supported-models.md` all carry the variant. `display_name()` and `family()` derive from `metadata()`, and family `Specialized` is already in `FAMILY_ORDER`. Paged decode, TurboQuant symmetric KV and the hybrid-SSM prompt-cache deny list are all opt-in or deny-by-default, so no entry is required. Tensor parallel is refused for `tp_size > 1` via `runtime_kind_for`'s `_ => None`.

### Remaining Items

- (MEDIUM) `src/execution/kv_arch.rs:244` and `:163`, `src/execution/memory_estimate.rs:510` and `:721`, `src/execution/quant_advisor.rs:61` alias `num_hidden_layers` / `n_layers` / `num_layers` and `hidden_size` / `d_model` / `dim`, but never the OpenAI-era `n_layer` / `n_embd` / `n_inner`. `mlxcel inspect` therefore reports KV cache 0 bytes and activation 0 bytes for this family, `--recommend-kv-cache` prints nothing, and `mlxcel serve` logs a KV-geometry-unavailable warning because `--kv-cache-budget` defaults to `auto`. Inert today (the family does not opt into paged decode) and shared verbatim with `gpt2` from #924, so it is a follow-up rather than a blocker. Trap for whoever fixes it: `kv_arch.rs:150` has no branch for the boolean `multi_query` flag, only ChatGLM's `multi_query_group_num`, so adding the aliases alone would produce a 16x KV over-estimate for SantaCoder.
- (MEDIUM) `src/models/gpt_bigcode.rs:396` short-circuits a quantized projection past the float shape contract, so a malformed quantized `c_attn` would reach `slice_last_dim` and abort through the cxx bridge rather than returning a load error. Deliberate, documented, unit-tested, and identical to `gpt2.rs:323`. No public quantized `gpt_bigcode` checkpoint exists.
- (LOW) `src/distributed/tensor_parallel/inference.rs:221-224` states that "the planner's supported-architecture validation rejects this string". `generate_shard_plan` actually accepts it through the `_ =>` generic-transformer fallback at `plan_generator.rs:157`. The real refusal is `runtime_kind_for`'s `_ => None` at `llama_runtime.rs:2225`, surfaced by `validate_supported_runtime`. The outcome is correct, only the mechanism is misdescribed. Copied from the GPT-2 comment.
- (LOW) `src/models/gpt2.rs:29-30` claims `gelu_approx` is the tanh approximation. It is erf-based (`mlx_cxx_bridge.cpp:2653`, "Use erf-based exact GELU instead of the original tanh approximation"). `gpt_bigcode.rs:549-552` states this correctly, so two sibling files now contradict each other. Pre-existing from #924.
- (LOW) `intermediate_size()` maps `n_inner: 0` to the `4 * n_embd` fallback instead of rejecting it. HuggingFace would yield 0. Harmless: the c_fc shape check then fails loudly against the real tensor.

### Suggestions Reviewed and Declined

- Adding a per-file mlx-lm derivation notice: declined. Zero of the 157 files under `src/models/` carry one, including `gpt2.rs`. The convention there is `//! Reference: <GitHub URL>`, and `NOTICE` lines 28-48 already carry the mlx-lm MIT attribution for "mlxcel's model architectures". The per-file convention is applied to `src/lib/mlxcel-core/src/cache/turbo/` and `src/audio/qwen3_omni_moe/`, which is what the guideline's example targets.
- Requiring an exact `wte` row count for tied heads: declined. It would reject legitimate checkpoints that pad the embedding table for alignment while declaring the true `vocab_size`, and slicing the tied logits would diverge from upstream mlx-lm, whose `Model.__call__` applies `wte.as_linear` over every row unsliced. That would break the token-exactness gate this PR was validated against.

### Verification

- [x] Tests pass

`cargo check --release --lib --tests --features metal,accelerate` clean (only the two pre-existing `host_preprocessor` warnings). `cargo test --release --lib --features metal,accelerate models::gpt` 46 passed, 0 failed. `cargo fmt --all -- --check` clean. `scripts/ci/check_cross_repo_refs.py` clean.

inureyes:

Scope: changed files, with the config surface and the MQA arithmetic traced end to end. No CRITICAL or HIGH findings. One MEDIUM fixed in 165f8199c, three reported below.

### Verified clean

- **Untrusted-config hardening is complete for this family's own surface.** Every scalar that sizes an allocation, indexes a table, divides, or feeds an `as i32` cast is bounded in `ModelArgs::validate` before it is used: the zero check on `n_head` precedes the divisibility check (`0.is_multiple_of(0)` is true, so the order matters), and `n_embd` being a nonzero multiple of `n_head` transitively bounds `n_head <= n_embd <= 65536`. That makes every derived quantity fit `i32` with room to spare: `kv_dim <= 65536` and `c_attn_out_features() = n_embd + 2 * kv_dim <= 196608`, so the MQA arithmetic cannot overflow or truncate. `n_inner` is checked through `intermediate_size()`, which is the value that actually sizes the MLP projections. There is no unbounded probe loop on the load path, and `Vec::with_capacity(n_layer)` is capped at 1024.
- **The CRITICAL from #924 is correctly carried over and extended.** `n_positions` is only stored on the model after `validate_embedding_table` has proved `n_positions <= wpe.weight` rows, so the clamp bound in `position_ids` is bounded by the real tensor, not by a config field. The same check is newly applied to `wte` against `vocab_size`.
- **Shared helpers were not weakened.** `validate_embedding_table` is `validate_position_table` with the row bound and the field name parameterized; the rejection condition (`claimed_rows > rows`), the zero-row guard and the quantized-width carve-out are unchanged. `layer_norm_from_weights` is strictly stronger than before: it now rejects a norm weight or bias that is not 1-D of width `n_embd`, which previously reached MLX. `dim_eq` is unchanged. Nothing here loosens the merged GPT-2 path.
- **MQA maps onto the existing GQA path with no materialization.** `n_kv_heads = 1` reaches `mlx::core::fast::scaled_dot_product_attention` unchanged; MLX broadcasts the shared KV head internally and there is no `repeat_kv` anywhere on the route, so K and V are stored and read once rather than 16 times. `KVCache` holds one head per layer, which is the whole point of the architecture.
- **Per-decode-step position-id cost is negligible and identical to GPT-2's.** A one-element `Vec<i32>` and a 4-byte host-to-device array per step, with no forced evaluation. The cache offset is read once before the layer loop, so `update_and_fetch` advancing it mid-loop cannot corrupt the lookup.
- Split offsets, `n_kv_heads`, `scale`, the erf GELU, the tied head and the top-level `lm_head` key all match https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gpt_bigcode.py line for line.

### MEDIUM (fixed in 165f8199c)

**A quantized projection reached the graph with its output width unchecked.** `src/models/gpt_bigcode.rs` `load_linear` returned early for any prefix carrying a `.scales` tensor, skipping both the shape contract and the bias check. That carve-out is inherited from GPT-2, but it is more dangerous here because `Attention::forward` slices the fused `c_attn` output at offsets derived from `config.json` rather than from the tensor.

Failure scenario: a 4-bit conversion whose `config.json` omits `multi_query` (it defaults to `true`, so the loader computes `2048 + 2 * 128 = 2304`) while the packed `c_attn` is the multi-head `3 * 2048 = 6144` width. All three slices stay in bounds, so Q is correct while K and V are read from arbitrary interior channels of the projection. The model loads, decodes and emits fluent-looking output with no error anywhere, which is exactly the outcome `load_linear` exists to prevent for the transposed float case. The mirror case (packed weight narrower than the config claims) takes the other branch: MLX's `slice` clamps the stop index rather than erroring, and the following `reshape` throws an `std::invalid_argument` that crosses the cxx bridge as an uncatchable `std::terminate`, so the process aborts with no diagnostic instead of returning a load error naming the tensor.

Affine and block-float packing both compress the input axis only, so a quantized weight is still `[out_features, packed_in]`. The input width stays unchecked and is left to `UnifiedLinear` to reconcile; the row count and the 1-D `.bias` are now checked on both paths. No legitimate export is rejected, since `mx.quantize` on an `[out, in]` weight always yields `[out, in * bits / 32]`. Covered by the extended `a_quantized_projection_skips_the_packed_input_width_but_not_the_output_width`.

### MEDIUM (reported, not fixed: the fix belongs in shared estimation code)

**GPT-BigCode is invisible to the architecture-aware KV estimator, so the server's default `--kv-cache-budget auto` leaves the paged KV pool unbounded.** `src/execution/kv_arch.rs:244` requires `num_hidden_layers` / `n_layers` / `num_layers`, and `attn_dims` at `kv_arch.rs:163` requires `hidden_size` / `d_model` / `dim` / `model_dim`. GPT-BigCode writes `n_layer` and `n_embd`, so `classify` returns `None`, `estimate_kv_arch` yields nothing, `kv_cache_bytes_per_token` returns 0, `paged_block_bytes` returns `None`, and `resolve_paged_block_budget` returns `None`. `src/server/model_worker.rs:729` then logs "the model's KV geometry is unavailable; leaving the paged pool unbounded".

Failure scenario: serving a StarCoder-class checkpoint with the default flags, a burst of long-context completion requests grows the paged pool with no admission or eviction ceiling until the OS reclaims the process, instead of being block-budgeted as it would be for a Llama or Qwen3 model. The pre-load `mlxcel inspect` estimate also reports KV as 0 bytes.

The condition is inherited: GPT-2 (merged in #924) writes the same key names and is in the same state on `main`, which is why this is a follow-up rather than a change inside a model-port PR. Anyone fixing it should land all three pieces together, because the first two alone introduce a new error: adding `n_layer` and `n_embd` to the key lists is correct for GPT-2 (plain MHA, `num_kv_heads` correctly defaults to `n_head`), but for GPT-BigCode `attn_dims` has no `multi_query` handling, so `num_kv_heads` would default to `n_head` and over-reserve MQA KV by `n_head` times: 16x on santacoder, 48x on StarCoder. An over-reservation of that size is the opposite outage, admitting roughly 1/48 of the sequences the model can actually hold.

**A quantized embedding table's dequantized width is still unchecked.** `validate_embedding_table` skips the `cols == n_embd` comparison when the table is quantized, because the packed width is a function of the bit depth rather than the model width. A quantized `wte` or `wpe` whose real width is not `n_embd` therefore reaches `ln_1.forward` and the residual `add` with a mismatched last axis, which is the `std::terminate` path again rather than a load error. Closing this needs the dequantized `in_features` from `UnifiedEmbedding` rather than anything derivable from the row count, and the helper is shared with GPT-2, so it is left as a follow-up.

### LOW

`intermediate_size()` treats `n_inner: 0` as absent and falls back to `4 * n_embd`, where HuggingFace would take the 0 literally. The consequence is only that `validate`'s `intermediate == 0` branch is unreachable (`n_embd >= 1` is already established, so the fallback is at least 4). No security or correctness impact on any real config; noting it so the dead branch is not mistaken for live coverage.

### Verification

`cargo fmt --all -- --check`, `cargo clippy --release --lib --tests --features metal,accelerate` (only the two pre-existing `src/multimodal/host_preprocessor*.rs` warnings, unchanged from `main`), and `cargo test --release --lib --features metal,accelerate` for `models::gpt_bigcode` (20 passed), `models::gpt2` (25 passed, confirming the shared-helper refactor plus the new norm-shape rejection leaves GPT-2 unchanged) and `detection_tests` (23 passed).

PR #928, opened 2026-07-26
feat(models): add GPT-NeoX (gpt_neox) text model support

### LINKED ISSUE #836: feat(models): add GPT-NeoX (gpt_neox) text model support
## Summary

GPT-NeoX is the EleutherAI decoder architecture behind the Pythia model suite and other NeoX-derived checkpoints. It introduces partial RoPE and an optional parallel-residual block layout. mlx-lm supports it; mlxcel does not yet load the `gpt_neox` architecture. Porting it adds the Pythia family and gives mlxcel a well-documented set of small-to-mid checkpoints. Effort is contained because the distinctive pieces (partial RoPE, parallel residual, LayerNorm) already have precedents in the mlxcel tree.

## Upstream reference

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gpt_neox.py

## Public checkpoint

EleutherAI Pythia family (for example a `pythia-*` checkpoint). Verify mlx-community checkpoint availability (confirm an existing MLX conversion) before validation.

## Architecture notes

GPT-NeoX with these distinctive traits:

- Partial RoPE: only `rotary_pct` of `head_dim` is rotated; the remaining channels pass through unrotated.
- Optional parallel residual: with `use_parallel_residual=True`, the attention sub-layer and the MLP sub-layer both read the same pre-norm input and their outputs are summed into the residual (rather than the MLP reading the attention output).
- `nn.LayerNorm` (with bias), not RMSNorm.
- Separate `input_layernorm` and `post_attention_layernorm`.

## Implementation plan

- Reuse the existing partial-RoPE handling used by gemma4 (`partial_rotary_factor`) in `src/models/gemma4.rs` and by phi in `src/models/phi.rs` to rotate only the `rotary_pct` fraction of `head_dim`.
- Reuse the parallel-residual block pattern for the `use_parallel_residual=True` path from ~~`src/models/gpt_oss.rs`~~ **`src/models/stablelm.rs`. Correction (verified against the tree at the merge parent of PR #928, `f7a877e76`): `src/models/gpt_oss.rs` has no parallel-residual path at all, and never had one; its block is the ordinary chained form (`h = x + attn(input_layernorm(x))`, then `h + mlp(post_attention_layernorm(h))`). The in-tree precedent is `src/models/stablelm.rs`, which branches on `use_parallel_residual` in `TransformerBlock::forward`. One difference from that precedent matters here: StableLM's parallel branch feeds both sub-layers from a single `input_layernorm`, whereas GPT-NeoX keeps two separate norms and computes `x + attention(input_layernorm(x)) + mlp(post_attention_layernorm(x))`, which is the form that shipped.**
- Reuse LayerNorm from the gpt2 work (lablup/mlxcel#833).

Effort: LOW-MEDIUM.

**Delivered and closed.** Implemented by PR #928, merged as `f7a877e76`; `mlxcel arch` now reports the family. Test coverage: 32 unit tests in `src/models/gpt_neox_tests.rs`. Real-checkpoint validation: 40 greedy tokens from `models/pythia-1b` reproduce the mlx-lm reference exactly, which exercises the interleaved per-head QKV split, partial RoPE over 64 of 256 head channels, and the parallel-residual block together. A mutation run substituting the naive flat split failed exactly three tests and passed 26, confirming the tests catch that specific bug. The unticked boxes below are the original pre-implementation plan and were not maintained during the work; treat the merged PR and its review thread as the record of what shipped.

- [ ] Config struct + serde parse for the `gpt_neox` config (including `rotary_pct` / `use_parallel_residual`).
- [ ] `from_weights` constructor.
- [ ] `sanitize` weight remap.
- [ ] Add the `gpt_neox` arch-string arm to `src/models/detection.rs`.
- [ ] Register the model in `src/model_metadata.rs` (`for_each_model_registration!`).
- [ ] Add the TP/distributed inference arch-string if applicable.
- [ ] `_tests.rs` unit tests beside the implementation, covering both the parallel-residual and the sequential paths.
- [ ] Update `docs/supported-models.md`.

## Effort

LOW-MEDIUM.

inureyes:
## Summary

Adds GPT-NeoX (`gpt_neox`), the EleutherAI decoder behind the Pythia suite and the NeoX-derived checkpoints, as a fully integrated text model: config parse, loader, decoder graph, detection, registration, `LoadedModel` dispatch, docs and unit tests. It shares `LayerNorm` with bias and a fused QKV projection with the GPT-2 lineage landed in #924 and #926, so the shared helpers there are reused rather than duplicated, but the three things that define the family are its own.

## The trap: interleaved per-head QKV

`query_key_value.weight` is `[6144, 2048]` on `EleutherAI/pythia-1b`, plain `nn.Linear` orientation, no transpose. But the split is **not** a three-way split of the output axis. Upstream reshapes to `(..., num_heads, 3 * head_dim)` and splits *that*, so the projection output is head-major: `[q_i | k_i | v_i]` per head, each of the 8 heads contributing 768 contiguous channels for `head_dim` 256.

A flat `split(qkv, 3, axis=-1)`, which is exactly what `src/models/gpt2.rs` and `src/models/gpt_bigcode.rs` do to their own fused projections, produces Q, K and V of byte-identical shape assembled from the wrong channels. Nothing throws, the cached KV shape is unchanged, the logits shape is unchanged, and the model generates fluent English out of a scrambled attention. No shape assertion anywhere can catch it.

The split is therefore factored into `split_interleaved_qkv` so the layout can be pinned by channel value rather than only observed through a whole forward pass, and it is covered three ways:

- `interleaved_qkv_split_pins_the_head_major_layout` feeds a 24-channel projection whose values are their own channel indices and asserts Q is `[0,1,2,3, 12,13,14,15]`, K is `[4,5,6,7, 16,17,18,19]`, V is `[8,9,10,11, 20,21,22,23]`.
- `a_flat_three_way_split_would_take_different_channels` asserts the same tensors differ from the literal flat-split runs `0..7` / `8..15` / `16..23`.
- `attention_forward_returns_the_interleaved_v_block` is an end-to-end pin that does not touch the helper: with a single token, attention softmaxes over exactly one key, so with a zeroed `query_key_value` weight, a counting bias and an identity `dense`, the block output *is* the V channels. It must be `[8,9,10,11, 20,21,22,23]`, and is asserted not to be the flat split's `[16..23]`.

**Verified by mutation, not just by passing.** `split_interleaved_qkv` was temporarily replaced with the naive flat form and the suite re-run: exactly those three tests failed and all 26 others passed, including the full prefill-then-decode, both residual tests, and every layout and loading test. `attention_forward_returns_the_interleaved_v_block` reported `[16.0, 17.0, 18.0, 19.0, 20.0, 21.0, 22.0, 23.0]`, the flat split's V, confirming the test observes the exact failure mode it was written for. The mutation was then reverted and the suite re-run green.

## What changed

- `src/models/gpt_neox.rs` (new): `ModelArgs`, `GptNeoxLayout`, `strip_registered_buffers`, `load_linear`, `split_interleaved_qkv`, the decoder graph and the `LanguageModel` impl.
- **Partial RoPE.** Only `int(head_dim * rotary_pct)` channels of each head rotate and the rest pass through untouched (64 of 256 for Pythia's `rotary_pct` 0.25), expressed with the `dims` argument of `fast_rope` with `traditional = false`, following the `partial_rotary_factor` precedent in `src/models/phi.rs`. `partial_rope_leaves_the_unrotated_channels_untouched` asserts the tail channels are bit-unchanged at a non-zero offset and that a partial rotation differs from rotating the whole head.
- **Both residual layouts.** `use_parallel_residual: true` (every Pythia checkpoint) feeds attention and the MLP the *same* pre-norm input and sums both into the residual; `false` chains them so `post_attention_layernorm` reads the post-attention residual. Both produce identically shaped output from identical weights, so running the wrong one is a silent quality regression rather than a crash. The two tests recompute each form from the block's own sub-modules, which pins the wiring rather than the arithmetic, and a third asserts the two layouts do not produce the same logits.
- **Three registered buffers dropped at load.** `attention.bias` (a `[1, 1, 2048, 2048]` causal mask whose name collides with the projection-bias namespace exactly as GPT-2's `h.N.attn.bias` does), `attention.masked_bias` and `attention.rotary_emb.inv_freq`. Removed in a single `retain` pass over the weight map, not a probe of `<prefix>.<i>.attention.bias` for every `i` in `0..num_hidden_layers`: `num_hidden_layers` is attacker-controlled and unchecked at that point on the `load()` path, and the probing form has no early exit. That is the same replacement the previous chain PR had to make after review. A test asserts the real `query_key_value.bias`, `dense.bias` and `input_layernorm.bias` all survive the pass.
- **Two key layouts.** The raw HuggingFace export (`gpt_neox.layers.N`, top-level `embed_out`) and the layout upstream mlx-lm's `sanitize` produces (`model.h.N`, `model.embed_out`). The candidate list is fixed at two entries probed by the embedding key alone, with no shape-derived branch, because unlike GPT-2 both layouts store every projection in the same `nn.Linear` `[out, in]` orientation and there is nothing to transpose. `both_layouts_build_the_same_model` asserts they produce identical logits. Only the raw layout is checkpoint-validated; the second is derived directly from upstream's `sanitize`, which is the only thing that can produce an MLX conversion of this family.
- **Untrusted-config hardening built in from the start**, following what the security passes on #924 and #926 arrived at rather than waiting to be told again. `ModelArgs::validate` rejects `num_attention_heads == 0` *before* any divisibility check (`0.is_multiple_of(0)` is true, so the divisibility check alone lets a zero-zero config reach `head_dim()`), an indivisible head split, and out-of-range `hidden_size` / `num_hidden_layers` / `max_position_embeddings` / `vocab_size` / `intermediate_size`, all before `num_hidden_layers` sizes the `Vec::with_capacity` in `from_weights` and before any derived `as i32` cast can truncate to a negative number.
- **`rotary_pct` is the first float this lineage lets a config author control**, so it gets its own validation. `mlx::core::fast::rope` does check `dims`: at the pinned MLX commit `mlx/fast.cpp` throws on a non-positive `dims`, on an odd `dims`, and on a `dims` larger than the input's last axis. None of that is recoverable here, because `fast_rope` crosses the cxx bridge as `UniquePtr<MlxArray>` rather than a `Result`, so the C++ `std::invalid_argument` becomes an uncatchable `std::terminate` (SIGABRT), and it fires at the first forward pass rather than at load. Every value MLX would throw on is therefore rejected in Rust, at load: zero, negative, odd, greater-than-`head_dim`, NaN and both infinities (the `as i32` cast saturates, turning NaN into 0 and infinity into `i32::MAX`, but the non-finite case is rejected explicitly so the message names the real problem). `rotary_emb_base` is rejected when non-positive or non-finite, since RoPE takes its logarithm and a zero or negative base makes every rotated channel NaN without anything throwing. `config_validation_rejects_out_of_range_rope_parameters` covers all of these plus both accepted boundaries (the floor is two rotary dimensions, not one, since one is odd), and `every_accepted_rotary_pct_survives_a_real_rope_call` sweeps `rotary_pct` at both head widths and puts every accepted value through a real `fast_rope` call, so the guard drifting away from MLX's contract aborts the test binary rather than production.
- **`layer_norm_eps` and the `quantization` block are validated at load too**, closing the same failure shape for two more config-controlled floats that `validate_rope` did not cover. `layer_norm_eps` is the third float this family lets a config author control; `fast::layer_norm` never looks at it (`x * rsqrt(mean(x^2) + eps)`), so a non-finite, negative or zero value makes every hidden state NaN without anything throwing, and the checkpoint loads cleanly and then generates uniform garbage instead of failing to load. `quantization.group_size` and `quantization.bits` reach `quantized_matmul` and `dequantize` unreconciled (`reconcile_quantization_layout` treats a non-positive pair as insufficient shape info and passes it through unchanged), and MLX's own `packed_in * 32 / bits` check divides by zero at `bits == 0` and collapses to zero above 32, throwing the same uncatchable `std::terminate` the `rotary_pct` guard exists to prevent; the bound is a range (`1..=32` for `bits`, positive for `group_size`) rather than an allowlist, since mlxcel deliberately re-derives an effective bit width from the tensor shapes when the declared one disagrees, and an allowlist would reject the mixed-precision exports that behavior serves. `config_validation_rejects_a_layer_norm_eps_that_would_nan_every_hidden_state` and `config_validation_rejects_a_quantization_block_that_would_abort_an_mlx_kernel` cover both guards, including the values that must stay accepted.
- **Every gather and slice bound by the real tensor shape.** The `embed_in` table is checked against `vocab_size` at load, because MLX wraps a negative gather index but does not range-check a positive one, so a config that overstates the table turns an ordinary prompt into an out-of-bounds read whose result reaches the logits. Every projection, bias and norm is checked against the config-implied shape in *every* layer, not just layer 0.
- **The quantized branch does not skip the shape contract.** Packing compresses the input axis only, so the row count is untouched and is checked on the quantized path exactly as on the float path, along with the 1-D bias; only the packed input width is left to `UnifiedLinear`. That check is load-bearing here specifically because `Attention::forward` reshapes the fused projection to `[batch, seq, num_heads, 3 * head_dim]` using config-derived widths, so a packed `query_key_value` of a different real width makes the reshape throw, and an MLX C++ exception crossing the cxx bridge is an uncatchable `std::terminate` rather than a load error. This is the same carve-out #926 had to close after review, closed here before it was opened.
- **Which GELU, stated explicitly.** `mlxcel_core::utils::gelu_approx` is the erf-based *exact* GELU in this tree despite its name (the tanh form was deliberately replaced in `mlx_cxx_bridge.cpp` because `power(x, 3)` produced NaN for negative bf16 inputs). That matches Pythia's declared `hidden_act` of `"gelu"`, which HuggingFace resolves to the exact erf GELU, and differs from the tanh approximation upstream mlx-lm applies by under 5e-4 per element (MLX's own documented bound over `[-6, 6]`). The reasoning is written out at the call site. A config declaring a non-GELU `hidden_act` is reported via `eprintln!` rather than rejected, since the activation cannot corrupt memory and `tracing::warn!` is a no-op on the CLI path this is reachable from.
- **No grouped-query attention.** Upstream's `num_key_value_heads` is vestigial: its `__post_init__` defaults it to `num_attention_heads` but `Attention` never reads it, and HuggingFace `GPTNeoXAttention` has no such concept. The fused projection is `3 * hidden_size` wide and reshaped to `(num_heads, 3 * head_dim)`, so Q, K and V always carry the same head count; the field is deliberately absent from this config struct and the forward test asserts the cached KV head count is `num_attention_heads`.
- `src/models/gpt2.rs`: the shared `validate_embedding_table` gains a `width_field` parameter so its rejection message names the config field the reader will actually find in their `config.json` (`hidden_size` here, `n_embd` in the GPT-2 lineage). The three shared helpers' `// Used by:` comments are updated. GPT-2 and GPT-BigCode behavior is otherwise unchanged; `src/models/gpt_bigcode.rs` is touched only for the two updated call sites.
- `src/models/mod.rs`: module declaration, `GptNeoxModel` re-export, and `ModelType::GptNeoX` in all four places that must agree (the enum, `ALL_MODEL_TYPES`, the `metadata()` display table, and the `all_variants!` exhaustiveness list). `src/models/detection.rs`: `"gpt_neox" => Ok(ModelType::GptNeoX)`, plus a detection test asserting it does not fall through to either GPT-2-lineage arm. `src/model_metadata.rs`: standard config-backed registration. `src/loaded_model.rs`: variant and `delegate_language_model!` arm. `src/distributed/tensor_parallel/inference.rs`: `"gpt_neox"` fallback architecture string, keeping the dispatch table total; TP is refused by `validate_supported_runtime`, whose `runtime_kind_for` match has no arm for this variant. Sharding would need its own rules regardless, since a row split at any offset a generic plan would pick cuts through the middle of a head's Q, K and V rather than between whole heads.
- `src/models/gpt_neox_tests.rs` (32 tests) and `docs/supported-models.md`.
- **Finalization pass.** Two `format!` error strings in `validate_norm_eps` and `validate_quantization` had literal multi-space gaps left over from wrapping the prose without converting it to the backslash-continued style the rest of the file uses; rewritten with no behavior change. The `docs/supported-models.md` entry now also names the `layer_norm_eps` and `quantization` rejections above. Both fix commits were verified by reverting each in isolation: without the odd-`dims` rejection, `config_validation_rejects_out_of_range_rope_parameters` fails cleanly and `every_accepted_rotary_pct_survives_a_real_rope_call` then feeds an odd `dims` to a real `fast_rope` call and aborts the whole test binary with SIGABRT (`[rope] dims must be even but got 1.`), rather than failing cleanly; without the `layer_norm_eps` / `quantization` guards, the two tests named above fail cleanly (no MLX call is reached, so no abort). All three pass again with the fixes restored.

## Test plan

- [x] `cargo clippy --release --lib --tests --features metal,accelerate` (no new findings; the two warnings in `src/multimodal/host_preprocessor*.rs` are pre-existing on `main` and untouched here)
- [x] `cargo test --release --lib --features metal,accelerate models::gpt_neox` (32 passed)
- [x] `cargo test --release --lib --features metal,accelerate models::gpt2` (25 passed, unchanged by the `validate_embedding_table` signature change)
- [x] `cargo test --release --lib --features metal,accelerate models::gpt_bigcode` (20 passed, likewise)
- [x] `cargo test --release --lib --features metal,accelerate detection_tests` (24 passed)
- [x] `cargo test --release --lib --features metal,accelerate models::metadata_tests` (4 passed: the registration and `ALL_MODEL_TYPES` exhaustiveness gates)
- [x] `cargo test --release --lib --features metal,accelerate distributed::tensor_parallel` (338 passed, 4 ignored)
- [x] Mutation check on the interleaved split, described in full above: the naive flat form fails exactly the three value-level tests and passes the other 26.
- Unit coverage beyond the split: the real pythia-1b config field set and the bare-config defaults; scalar, list and empty-list token-id fields with the bos fallback; partial-RoPE dimensioning including `int()` truncation and the full-rotation case; `intermediate_size` versus the `4 * hidden_size` fallback including explicit `null`; the non-GELU activation predicate; eight hostile integer scalars plus `max_position_embeddings` and the `4 * hidden_size` overflow boundary; nine out-of-range RoPE parameters plus both accepted boundaries; detection of both key layouts, their rejection path, and their producing identical logits; the registered-buffer strip including that the real biases survive and that a second pass removes nothing; the model building identically with and without the buffers present; every projection loading with its stored `[out, in]` shape unchanged; rejection of a transposed weight, a rank-3 weight, a mis-sized bias, a missing projection, a later layer that disagrees with the config, an `embed_in` table shorter than `vocab_size` claims, a table or either norm of the wrong width, and an untied config with no `embed_out`; a quantized projection that skips only the packed input-width check while its output row count (wider and narrower) and its bias are still rejected by name; both residual layouts recomputed from their own sub-modules and asserted non-interchangeable; and prefill-then-decode under both residual settings asserting the logits shape, the advancing cache offset and the cached KV head count.
- Not run here: the release build and the token-exact real-checkpoint validation against `models/pythia-1b`, which the maintainer is running separately.

Closes #836


inureyes:

**Token-exact against the mlx-lm reference** for 40 greedy tokens on `models/pythia-1b`: `mlxcel generate -m models/pythia-1b -p "The capital of France is" -n 40 --no-chat-template` reproduces the reference continuation exactly, starting `Paris, and the capital of France is Paris.`, at 60.3 tok/s. `mlxcel arch` reports `GPT-NeoX (EleutherAI Pythia, partial RoPE, parallel residual)`.

This is the strongest gate of the three ports in this chain so far, because it exercises all three of the distinctive mechanisms at once and any one of them being wrong changes the output: the interleaved per-head QKV split, partial RoPE over 64 of 256 head channels, and the parallel-residual block. In particular the naive flat 3-way split that GPT-2 uses would still produce correctly-shaped tensors and fluent English here, so matching the reference token for token is what rules it out.

**On the GELU divergence, which turned out not to matter.** The implementation uses the erf form on the grounds that pythia declares `hidden_act: "gelu"`, which HuggingFace resolves to exact erf GELU, while upstream mlx-lm applies `nn.gelu_approx` (genuinely the tanh form in MLX Python). So this run compared an erf-based mlxcel against a tanh-based reference and still got identical token ids, which both confirms the sub-5e-4 activation difference flips no argmax on this checkpoint and means the gate above is not accidentally insensitive. The same pattern held for the GPT-2 port earlier in this chain.

inureyes:

### Intent

Add GPT-NeoX (`gpt_neox`) as a fully integrated text model: interleaved per-head QKV, partial RoPE, both residual layouts, registration across every dispatch surface (issue #836).

### Findings addressed

- [x] **HIGH: an odd rotary dimension count was accepted and aborts the process at the first forward pass** (`src/models/gpt_neox.rs:339`, `validate_rope`). Fixed in `d677ea6f1`.

  The guard accepted any `rope_dims` in `1..=head_dim` on the stated premise that `mlx::core::fast::rope` does no range checking and that an odd value is "asymmetric but not out of bounds". Both halves are wrong at the pinned MLX commit `b7c3dd6d`. `mlx/fast.cpp:409-424` throws on a non-positive `dims`, on an **odd** `dims`, and on a `dims` larger than the input's last axis.

  That is not recoverable: `fast_rope` is declared across the cxx bridge as returning `UniquePtr<MlxArray>` rather than a `Result`, so the C++ `std::invalid_argument` becomes an uncatchable `std::terminate`. Verified empirically rather than inferred, by calling `fast_rope` with `dims = 3` on a 4-wide head inside this suite:

  ```
  libc++abi: terminating due to uncaught exception of type std::invalid_argument: [rope] dims must be even but got 3.
  process didn't exit successfully: ... (signal: 6, SIGABRT: process abort signal)
  ```

  The whole test binary went down, taking 22 unrelated tests with it. The throw fires at the **first forward pass**, not at load, so a `config.json` declaring `rotary_pct` 0.01171875 against Pythia's 256-wide head passes validation, loads cleanly, and then aborts the process on the first request. This is exactly the class the PR set out to pre-empt, and the PR's own test asserted `rope_dims() == 1` was accepted.

  The fix rejects an odd count alongside the existing cases, moving the floor from 1 to 2. No published GPT-NeoX is affected: every `head_dim` and `rotary_pct` pair in the wild multiplies out even. New test `every_accepted_rotary_pct_survives_a_real_rope_call` sweeps `rotary_pct` at both the 4-wide and the 256-wide head and puts every accepted value through a real `fast_rope` call, so future drift away from MLX's contract aborts the test binary rather than production. The doc comments and the `docs/supported-models.md` entry that stated the incorrect premise are corrected.

### The five self-flagged items

1. **`model.h.N` alternate layout: keep.** I did not take the derivation on assertion. I read upstream's `sanitize` and confirmed the mapping character by character: prefix `model.`, then `.gpt_neox.layers.` to `.h.`, then `.gpt_neox.` to `.`, which sends `gpt_neox.embed_in.weight` to `model.embed_in.weight`, `gpt_neox.layers.N` to `model.h.N`, and top-level `embed_out` to `model.embed_out`. That is precisely the second entry of `GPT_NEOX_LAYOUTS`, and it is the only layout an `mlx_lm.convert` of this family can emit, so it is the layout a user quantizing a Pythia today would actually produce. The probe is two fixed key lookups with no attacker-controlled bound, and `both_layouts_build_the_same_model` pins logit equality. The difference from PR #924 cuts the other way: there the candidates differed in weight orientation and a wrong guess silently transposed, here both layouts are identical `nn.Linear` `[out, in]` and the only thing unvalidated is a rename. Cheap, verified against the authoritative producer, worth keeping.

2. **`intermediate_size` deviation: correct, and the fallback cannot disagree silently.** `MLP::from_weights` loads `dense_h_to_4h` as `[intermediate, hidden]` and `dense_4h_to_h` as `[hidden, intermediate]`, and `load_linear` checks both axes on the float path, so a config value that disagrees with the tensor is rejected by tensor name before any matmul. On the quantized path `dense_h_to_4h`'s row count still pins `intermediate`. Reading the field with a `4 * hidden_size` fallback is the more faithful of the two references (HF `GPTNeoXMLP` reads it, mlx-lm hardcodes), and no overflow is reachable: `hidden_size` is capped at 65536, so the fallback tops out at 262144, well inside `MAX_INTERMEDIATE_SIZE`. Left as is.

3. **`hidden_act` warn-and-continue: right call, and the diagnostic does reach the CLI.** Confirmed that `tracing_subscriber` is installed only in `src/server/startup.rs:1017,1022`, so `tracing::warn!` is a no-op in the `mlxcel` binary and `eprintln!` is the only thing that surfaces. It also matches the precedent `src/models/gpt_bigcode.rs:697` established in #926, with the same reasoning comment. Warning rather than rejecting is right here: the activation cannot corrupt memory, and refusing an otherwise loadable checkpoint over it would be worse.

4. **RoPE `dims` evenness: the asymmetry claim was wrong.** See the HIGH above. The `1..=head_dim` guard was still worth having, but for the opposite reason than stated: MLX validates by throwing, and a throw here is SIGABRT, so the value has to be rejected in Rust at load. Fixed.

5. **Sequential residual: genuinely correct.** Checked against upstream, not just present. Upstream's `else` branch is `attn_output = attention(input_layernorm(x)); x = x + attn_output; ffn_output = mlp(post_attention_layernorm(x)); x = x + ffn_output`, which is `h = x + attn(ln1(x)); out = h + mlp(ln2(h))`, exactly what `TransformerBlock::forward` computes, and it agrees with HF `GPTNeoXLayer`. `sequential_residual_feeds_the_mlp_the_post_attention_residual` recomputes it from the block's own sub-modules so it pins the wiring rather than the arithmetic, and `the_two_residual_layouts_are_not_interchangeable` proves the two are separable by value alone.

### Also verified

- **Registered buffers.** One `retain` pass over the weight map keyed on the three suffixes, no `0..n_layer` probe loop, so `num_hidden_layers` cannot drive a load-path spin. The real `.attention.query_key_value.bias` and `.attention.dense.bias` do not match any suffix and survive.
- **Previously merged files not weakened.** `src/models/gpt2.rs` gains a `width_field` parameter to `validate_embedding_table` that is used only inside a rejection message; the row check, the quantized carve-out and the return value are byte-identical, and both GPT-2 lineage call sites pass `"n_embd"`, so their messages are unchanged too. `src/models/gpt_bigcode.rs` changes only two call sites. Suites confirm: gpt2 25 passed, gpt_bigcode 20 passed.
- **Config-trust class.** The zero check on `num_attention_heads` precedes the divisibility check, so the `0.is_multiple_of(0)` hole is closed; the divisibility check then transitively bounds `num_attention_heads` by `hidden_size`, so every `as i32` cast (`head_dim`, `3 * hidden_size`, `num_heads * head_dim`) stays inside `i32`. `num_hidden_layers` is capped before it sizes `Vec::with_capacity`. `embed_in` is checked against `vocab_size` by real row count. The quantized branch checks the output row count and the 1-D bias, leaving only the packed input width to `UnifiedLinear`, which is the same accepted carve-out as #926. `rotary_pct` negative, zero, NaN, both infinities and greater-than-one were all already rejected; odd is the case that was missing.
- **Exhaustive registration.** `ModelType` enum, `ALL_MODEL_TYPES`, `metadata()` display table, `all_variants!`, `detection.rs`, `model_metadata.rs`, `loaded_model.rs` (variant plus `delegate_language_model!`), `distributed/tensor_parallel/inference.rs`, `docs/supported-models.md`. The TP refusal claim is accurate: `runtime_kind_for` in `llama_runtime.rs:2195` has no `GptNeoX` arm and falls to `_ => None`. The arch renderer gates pass (3 passed), so the family is not silently dropped from `mlxcel arch`.
- **Convention conformance.** `causal_attention(q, k, v, scale, 0.0, 0)` with the `attention_from_ptr` fallback matches `gpt2.rs:512`, `gpt_bigcode.rs:499` and `phi.rs:186` exactly. `gelu_approx` is confirmed erf-based in `mlx_cxx_bridge.cpp:1295`, so the call-site reasoning is accurate. `strip_registered_buffers` is called from `load()` only, the same placement as `strip_causal_mask_buffers` in `gpt2.rs:761`.

### Deliberately left

- `intermediate_size: 0` falls back to `4 * hidden_size` rather than erroring. Harmless, since `load_linear` then rejects any disagreement with the real tensor by name. LOW.
- `quantization.group_size` and `quantization.bits` reach MLX unvalidated. This is identical in `gpt2.rs`, `gpt_bigcode.rs` and every other family in the tree, so closing it belongs in its own cross-family change rather than a single model port. LOW.
- `layer_norm_eps` is not checked for NaN or a negative value. It produces NaN output rather than memory unsafety, and matches the rest of the lineage. LOW.
- `ModelArgs::interleaved_qkv_channel_offsets` is exercised only by a test and the docs. It is a deliberate spec function that states the family's trap as a pure function of the config, which is worth the public surface. LOW.

### Verification

- [x] Tests pass

`cargo fmt --all -- --check` clean. `cargo check --release --lib --tests --features metal,accelerate` clean. `cargo clippy --release --lib --tests --features metal,accelerate` reports only the two pre-existing `src/multimodal/host_preprocessor*.rs` warnings that are already on `main`. `models::gpt_neox` 30 passed, `models::gpt2` 25 passed, `models::gpt_bigcode` 20 passed, `detection_tests` 24 passed, `models::metadata_tests` 4 passed, `--bin mlxcel supported_models` 3 passed. `scripts/ci/check_cross_repo_refs.py` OK.

inureyes:

Scope: the changed files, with the MLX C++ sources at the pinned commit `b7c3dd6d27f45b5365b08a840310187dc503f1db` read directly rather than assumed. The hunt followed the defect class the previous HIGH exposed: a config-derived value that passes Rust-side validation but violates an undocumented precondition of an MLX entry point, which then terminates the process instead of returning an error.

No CRITICAL or HIGH findings. Two MEDIUM findings were fixed in `87f5f4afd`; the rest is reported below.

### Enumeration of MLX calls taking a config-derived argument

| Call site | Config-derived argument | MLX precondition at the pinned commit | Status |
|---|---|---|---|
| `fast_rope` (`gpt_neox.rs:739-740`) | `rope_dims`, `rope_base`, `offset` | `mlx/fast.cpp:409-424` throws on `dims <= 0`, odd `dims`, `dims > x.shape(-1)`; `x.ndim() >= 3`; base is unchecked but `std::log(base)` is taken in the fallback | Fully covered by `validate_rope`, plus the `every_accepted_rotary_pct_survives_a_real_rope_call` abort-oracle |
| `reshape` / `slice_last_dim` (`gpt_neox.rs:693-696`) | `num_heads`, `3 * head_dim` | element count must match | Covered: `load_linear` pins `query_key_value` at `3 * hidden_size` rows on both the float and the quantized path |
| `layer_norm` (`LayerNorm::forward`) | `layer_norm_eps` | `mlx/fast.cpp:190` validates weight and bias shapes only, never `eps` | **Was unchecked. Fixed.** |
| `quantized_matmul` / `dequantize` (`UnifiedLinear`, `UnifiedEmbedding`) | `quantization.group_size`, `quantization.bits` | `mlx/ops.cpp:114` computes `w.shape(-1) * 32 / bits` | **Was unchecked. Fixed.** |
| `take` (embedding gather) | `vocab_size` | wraps a negative index, does not range-check a positive one | Covered by `validate_embedding_table` on the float path; see MEDIUM-3 for the quantized carve-out |
| `causal_attention` / `attention_from_ptr` | `scale = 1/sqrt(head_dim)` | none reachable | Covered: `head_dim >= 1` after validation, and `causal_attention` derives its own `k_len - q_len` offset so a warm prefix cache stays correctly aligned |

### MEDIUM-1 (fixed): `layer_norm_eps` reached `fast_layer_norm` unvalidated

`src/models/gpt_neox.rs:116` and `:937`. A previous pass left this at LOW pending a determination of whether it crashes or produces garbage. It produces garbage, and silently.

`mlx/fast.cpp:190` validates only the shapes of the weight and bias; `eps` is never inspected. The computation is `x * rsqrt(mean(x^2) + eps)`, so a NaN `eps` makes every element of every hidden state NaN, a negative `eps` does the same as soon as `mean(x^2) + eps` goes below zero, and `eps == 0` gives `rsqrt(0)` on an all-zero row so `0 * inf` is NaN again. Nothing throws, so unlike the rope case this is not a process abort: the NaN propagates through all remaining layers into the logits and the sampler draws from a NaN distribution.

Concrete failure: `mlxcel generate -m <org>/<repo>` on a repo whose `config.json` declares `"layer_norm_eps": -1e-5` loads cleanly, reports no error, and emits garbage tokens for every prompt. Nothing in the output identifies the config field as the cause.

Fixed by `ModelArgs::validate_norm_eps`, rejecting non-finite, negative and zero. Tested both directions, including that every small positive value a checkpoint might plausibly declare stays accepted.

### MEDIUM-2 (fixed): `quantization.group_size` and `quantization.bits` reached MLX unvalidated

`src/models/gpt_neox.rs:174-178`, consumed at `:436-445`. This is a true instance of the target defect class, and it is not intercepted on the way down.

`Quantization { group_size: i32, bits: i32 }` is parsed with no bounds and threaded into `UnifiedLinear::from_weights` and `UnifiedEmbedding::from_weights`. It is not reconciled away first: `mlxcel_core::layers::reconcile_quantization_layout` (`src/lib/mlxcel-core/src/layers.rs:978`) treats `group_size <= 0 || bits <= 0` as insufficient shape information and deliberately returns the declared pair unchanged. MLX then reaches `validate_quantized_input` (`mlx/ops.cpp:114`):

```cpp
if (w.shape(-1) * 32 / bits != scales.shape(-1) * group_size) {
```

At `bits == 0` that is an integer division by zero. Above 32 it evaluates to zero, which can match no real scales tensor. Either way the `std::invalid_argument` that follows crosses the cxx bridge as `UniquePtr<MlxArray>` rather than a `Result`, so it is an uncatchable `std::terminate`, exactly the shape of failure `validate_rope` exists to prevent for rope's `dims`.

Concrete failure: a HuggingFace repo carrying real `.scales` tensors and `"quantization": {"group_size": 64, "bits": 0}` aborts the process. On `mlxcel-server` that takes down the server, not one request.

Fixed by `ModelArgs::validate_quantization`, bounding `bits` to `1..=32` and requiring a positive `group_size`. The bound is deliberately a range and not an allowlist of the widths MLX supports: mlxcel re-derives an effective bit width from the tensor shapes when the declared one disagrees, so an allowlist would reject the mixed-precision exports that behavior serves. Tested that `(32,4)`, `(64,4)`, `(128,4)`, `(64,8)`, `(64,6)` and `(16,4)` all stay accepted.

**This gap is tree-wide, not specific to this PR.** No model family in the tree validates its quantization block, and `gpt2.rs` and `gpt_bigcode.rs` have the identical exposure through the same helper. The fix here closes it for `gpt_neox` because this file's `validate` explicitly claims to be that gate; the durable fix belongs in `reconcile_quantization_layout`, whose "trust the caller" early return is where hostile values currently survive. Worth a separate issue.

### MEDIUM-3 (reported, not fixed): a quantized embedding skips the width check, so a width disagreement aborts at first inference

`validate_embedding_table` in `src/models/gpt2.rs:873` skips the column check when the table is quantized, because the packed width is a function of the bit depth rather than the model width. `gpt_neox.rs:1043` inherits that. For a quantized checkpoint the dequantized `embed_in` width is therefore never compared against `hidden_size`.

If they disagree, the first `input_layernorm` gets an `x` whose last axis does not match its validated `hidden_size`-wide weight, and `mlx/fast.cpp:212` throws "weight must have the same size as the last dimension of x". `layer_norm` also crosses the bridge as `UniquePtr<MlxArray>`, so that is another uncatchable `std::terminate` at the first forward pass rather than a load error. The same holds for a quantized `query_key_value` or `embed_out` whose packed input width disagrees, which `load_linear:619` explicitly leaves to `UnifiedLinear`.

Not fixed here: the check lives in a helper shared with GPT-2 and GPT-BigCode, closing it properly means computing the dequantized width from the reconciled bit depth, and doing that inside this PR would change two already-merged families. Impact is a hostile-checkpoint abort with no memory-safety consequence, which is why it is MEDIUM rather than HIGH.

### Items checked and found clean

- **`rotary_pct` across its whole float domain.** Negative, zero, NaN, both infinities and greater-than-one are all rejected, and the NaN bypass the checklist warns about does not apply: `validate_rope:324` tests `is_finite()` *before* any range comparison, so NaN never reaches a comparison whose falsity could admit it. The saturating `as i32` would fold NaN to 0 and infinity to `i32::MAX` anyway, and both land in the rejecting arm.
- **Interleaved QKV arithmetic and overflow.** `3 * head_dim`, `num_heads * head_dim` and `3 * hidden_size` are all bounded by `MAX_HIDDEN_SIZE = 65536` transitively (the divisibility check forces `num_attention_heads <= hidden_size`), so the largest i32 the split can compute is 196608. Overflow is unreachable. A `query_key_value` whose real output width disagrees with `3 * hidden_size` is rejected by `load_linear` on the float path *and* on the quantized path, since packing compresses the input axis only and leaves the row count intact.
- **Magnitude validation and ordering.** The zero check on `num_attention_heads` is first, ahead of the divisibility check, so the `0.is_multiple_of(0) == true` bypass is closed. Every ceiling matches the values `gpt_bigcode.rs` already uses. `num_hidden_layers` is bounded to 1024 before it sizes the `Vec::with_capacity` at `:1052`, and `validate()` runs before `load_text_weights` on the `load()` path and at the top of `from_weights` on the weight-builder path, so neither route can reach the loop unvalidated.
- **No unbounded probe loop.** `strip_registered_buffers:573` is a single `retain` pass over the weight map, not a per-index probe. The only `0..n` loop is the layer loop, which runs after validation.
- **`.attention.bias` never reaches a bias slot.** `load_linear` is only ever called with `{prefix}.attention.query_key_value` and `{prefix}.attention.dense`, so it probes `...query_key_value.bias` and `...dense.bias`. The registered causal-mask buffer at `...attention.bias` matches neither, which means the strip is a memory optimization rather than a correctness requirement, and skipping it on the weight-builder route (as `gpt2.rs` also does) is harmless.
- **Sampled token ids stay inside the embedding table.** For the untied path `load_linear` pins `embed_out` at exactly `vocab_size` rows and `validate_embedding_table` requires `vocab_size <= embed_in` rows, so a sampled id is always in bounds on the next step. For the tied path the logits width is the table's own row count. There is no way for a NaN-logits draw to produce an out-of-range gather.
- **`gpt2.rs` is message-only, as claimed.** The `width_field` parameter is threaded into one format string. No predicate, no ordering and no early return changed, and the `gpt_bigcode.rs` diff is the two updated call sites. `models::gpt2` (25) and `models::gpt_bigcode` (20) pass unchanged.

### Performance

No findings. Three things were checked concretely.

**Partial RoPE does not force a concat on the hot path.** `RoPE::use_fallback` (`mlx/backend/metal/rope.cpp:10`) returns true only for `Device::cpu`, so the GPU path never runs the fallback whose `dims < x.shape(-1)` branch concatenates the untouched tail. The Metal kernel handles a partial `dims` natively.

It does, however, always copy: `rope.cpp:46` takes `if (dims_ < D)` first and issues an unconditional `copy_gpu` of the whole tensor, which also makes the `head_seq_transpose` fast path further down unreachable for this family. That cost is inherent to MLX's partial-rope implementation rather than anything this PR chose, and it is small: for `pythia-1b` at a full 2048-token prefill it is 8 MB per tensor across Q and K over 16 layers, roughly 256 MB of copy traffic, well under a millisecond at Apple Silicon bandwidth and negligible against the projections themselves. At decode it is 4 KB per tensor. Using `fast_rope`'s `dims` argument remains the right call; slicing the rotary channels and concatenating them back would be strictly worse.

**The interleaved split costs the same bytes as a flat split.** Both forms produce three strided last-axis slices that MLX materializes; the interleaved form differs only in having a shorter contiguous run per head (256 elements versus 2048 on `pythia-1b`), which is a coalescing difference, not an extra copy. This matches upstream exactly.

**`eos_token_ids()` is not on the per-token path.** It clones its `Vec`, but `generate.rs` calls it once per generation (four call sites, all outside the decode loop), so a long `eos_token_id` list costs nothing per token.

### LOW

1. **`unsafe` block with no `// SAFETY:` comment**, `src/models/gpt_neox.rs:748`. The raw pointer is a plain `Option`-to-null conversion of a reference that outlives the call, so it is sound, but the project's Rust conventions ask for the justification in a comment. Copied verbatim from `gpt2.rs:515` and `gpt_bigcode.rs:502`, so fixing it should be one change across all three rather than only here.
2. **`head_dim()`, `rope_dims()` and `interleaved_qkv_channel_offsets()` are `pub` and divide by `num_attention_heads`.** A caller that skips `validate()` gets a divide-by-zero panic. That is a clean Rust panic rather than UB, no in-tree caller does it, and `gpt_bigcode.rs` has the identical shape.
3. **`GptNeoxModel::load` has no direct test coverage.** The suite exercises `from_weights` thoroughly but nothing calls the path-based loader, so the `config.json` read plus `strip_registered_buffers` sequencing is only covered by the real-checkpoint run.

### Verification

- `cargo fmt --all -- --check`, clean.
- `cargo clippy --release --lib --tests --features metal,accelerate`, no new findings. The two warnings in `src/multimodal/host_preprocessor*.rs` are pre-existing on `main`.
- `cargo test --release --lib --features metal,accelerate models::gpt_neox`, 32 passed (30 existing plus 2 new).
- `models::gpt2` 25, `models::gpt_bigcode` 20, `detection_tests` 24, `models::metadata_tests` 4, all unchanged.
- `python3 scripts/ci/check_cross_repo_refs.py`, OK.

PR #800, opened 2026-07-16
feat(vlm): add MiniMax-M3-VL multimodal support

inureyes:
## Summary

Add MiniMax-M3-VL (`model_type: "minimax_m3_vl"`), the vision-language variant on top of the merged MiniMax-M3 text backbone (#763). The only public checkpoint (`MiniMaxAI/MiniMax-M3`, 427B) is itself the VL model, with a top-level config nesting `text_config` plus `vision_config`.

## What changed

New files:

- `src/vision/encoders/minimax_m3_vl.rs`: the CLIP-style vision tower (hidden 1280, 16 heads, head_dim 80, 32 layers, patch 14, `pre_layrnorm`, LayerNorm + exact-GELU blocks with separate `q/k/v/out` projections), reusing the shared Qwen2-VL `cu_seqlens` variable-length attention and rotary-frequency-table helpers, driven by a genuine 3D `(t, h, w)` vision RoPE matching the reference `MiniMaxVLVisionTransformer`: the head dimension splits into three equal axis sections plus a trailing unrotated tail (`axis_dim = 2 * ((head_dim / 2) / 3 / 2)`, `rot_dim = 3 * axis_dim`; 26/78 for head_dim 80, 20/60 for head_dim 64), and the temporal axis is inert for images (`grid_t == 1`) but still reserves its slice of the head dim, so the split cannot collapse to the 2D `(h, w)` form. A two-stage projector then maps features to the text width (per-patch `multi_modal_projector` into `projection_dim` 6144, then a `patch_merge_mlp` folding `spatial_merge_size^2 = 4` adjacent patches into the text width). The whole tower runs in f32 (the vision weights are non-quantized in the checkpoint) to avoid mixed-dtype matmuls; the LLaVA-style merge casts features back to the text embedding dtype.
- `src/vision/processors/minimax_m3.rs`: a faithful port of the checkpoint `image_processor.py` (Qwen2-VL-style `smart_resize`, CLIP mean/std, temporal padding, the exact merge-grouped patchify) emitting `pixel_values [num_patches, 1176]` plus `image_grid_thw`.
[...]

PR #946, opened 2026-07-27
feat(models): add Ant Group Ling / Bailing MoE (bailing_moe) text model support

### LINKED ISSUE #838: feat(models): add Ant Group Ling / Bailing MoE (bailing_moe) text model support
## Summary

Add support for the Ant Group Ling / Bailing MoE text model family (`bailing_moe`) to the mlxcel Rust inference runtime, porting the architecture from mlx-lm. Bailing MoE is a DeepSeek-style mixture-of-experts model with routed experts, an always-on shared expert, and optional grouped top-k routing, so much of it maps onto mlxcel's existing DeepSeek-V2/V3 MoE machinery. The routing details differ from DeepSeek in several places where a blind reuse is silently wrong; those places are called out below.

This body was revised after a pre-implementation pass that read the real checkpoint at `models/ling-lite-1.5` and the upstream reference line by line. Three config fields the original text named do not exist in the checkpoint, one mapping instruction was a divergence from upstream rather than a port of it, and the shared-expert shape was described incorrectly.

## Upstream reference

- mlx-lm implementation: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/bailing_moe.py

## Validation checkpoint

`inclusionAI/Ling-lite-1.5`, present locally at `models/ling-lite-1.5`. No mlx-community conversion is required: the upstream weights are raw bf16 SafeTensors and load directly.

- 16.8B total / 2.75B activated, 28 layers, `hidden_size` 2048, `intermediate_size` 1408, `moe_intermediate_size` 1408, 64 routed experts, `num_experts_per_tok` 6, `num_shared_experts` 2, 16 query heads / 4 KV heads (`head_dim` 128), `vocab_size` 126464, `rope_theta` 600000, `rms_norm_eps` 1e-6, `tie_word_embeddings` false, `rope_scaling` null.
- Roughly 33.6 GB across 4 shards, entirely bf16. Validation needs a machine with headroom for that plus KV cache.
- Real weight keys per layer: `model.layers.N.attention.query_key_value.weight` (fused QKV, `[3072, 2048]` = `(16 + 2*4) * 128`), `model.layers.N.attention.dense.weight`, `model.layers.N.mlp.gate.weight` (router, `[64, 2048]`), `model.layers.N.mlp.experts.{0..63}.{gate_proj,up_proj,down_proj}.weight` (per-expert, not pre-stacked), `model.layers.N.mlp.shared_experts.{gate_proj,up_proj,down_proj}.weight`, plus `input_layernorm` and `post_attention_layernorm`. Top level: `model.word_embeddings.weight`, `model.norm.weight`, `lm_head.weight`. Note `word_embeddings`, not `embed_tokens`.
- `use_bias` and `use_qkv_bias` are both false, so no projection carries a bias tensor, and no `expert_bias` tensor exists in the checkpoint.

## Config fields the checkpoint does not contain

Three fields the original issue text named are absent from `models/ling-lite-1.5/config.json`: `n_group`, `moe_router_enable_expert_bias`, and `moe_router_enable_routed_scaling`. They must be optional with serde defaults matching upstream, or the config will not parse at all. The full set of upstream-optional routing fields:

| Field | Upstream default |
|---|---|
| `moe_router_enable_expert_bias` | `false` |
| `moe_router_enable_routed_scaling` | `true` |
| `routed_scaling_factor` | `1.0` |
| `score_function` | `"softmax"` |
| `n_group` | `1` |
| `topk_group` | `4` |
| `moe_shared_expert_intermediate_size` | `None` |
| `moe_router_enable_shared_expert` | `true` |

`score_function` defaults to `"softmax"`. DeepSeek-V3's router is sigmoid. Since this issue directs the implementer to reuse the DeepSeek-V2/V3 MoE machinery, this is exactly the place where reuse goes wrong: taking the DeepSeek gate as-is changes the routing distribution on every token while producing perfectly finite output. The score function must be read from config and both `"softmax"` and `"sigmoid"` supported.

Also absent from the checkpoint and therefore default-bearing: `use_qk_norm` (false), `partial_rotary_factor` (1.0), `rotary_dim` (None). With those defaults the rope dim is the full `head_dim` of 128 and there are no QK norm weights to load.

## What the validation checkpoint actually exercises

This matters for how much a green real-model run is worth:

- `n_group` defaults to 1, so the grouped top-k branch never runs.
- `moe_router_enable_expert_bias` defaults to false, so the bias add never runs and there is no `expert_bias` tensor to load.
- `first_k_dense_replace` is 0, so all 28 layers are MoE and there is no dense prefix to exercise for this checkpoint.

Ling-lite-1.5 therefore exercises only the plain softmax top-k path with a shared expert. The grouped-routing branch, the expert-bias branch, the sigmoid score function, and the dense-prefix path all still have to be implemented, and each needs unit-test coverage against a synthetic config, because no checkpoint available here reaches them. A token-exact real-model comparison will not validate those branches. Do not read a passing real-model run as full coverage.

## Router: selection uses biased scores, weights come from unbiased scores

`group_expert_select` computes `scores = softmax(gates.astype(float32))` (or `sigmoid` when `score_function == "sigmoid"`), saves `orig_scores = scores`, then adds the correction bias to `scores` only. The grouped-group masking, when `n_group > 1`, also applies to that biased copy. Top-k indices come from `argpartition` on the biased scores, and the returned weights are then gathered from `orig_scores` via `take_along_axis`. Applying the bias to the weights as well as to the selection is a silent correctness bug: output stays finite and plausible while every routed contribution is misweighted. This is the same selection-only-bias pattern DeepSeek-V3 and ERNIE-4.5 already implement in this tree; follow those.

Two further details from the same function:

- `norm_topk_prob` is conditional and has an epsilon. Upstream applies it only when `top_k > 1`, and the denominator is `scores.sum(-1, keepdims=True) + 1e-20`, not a bare sum.
- The score function runs in float32 and the result is cast back to the input dtype only at the very end, after normalization and scaling.

## `moe_router_enable_routed_scaling` is dead code upstream

`BailingMoeGate` stores `self.enable_routed_scaling = args.moe_router_enable_routed_scaling` and never reads it again. `__call__` passes `routed_scaling_factor` unconditionally, and `group_expert_select` ends with an unconditional `scores = scores * routed_scaling_factor`. The flag has no effect upstream.

The original instruction in this issue to "map `moe_router_enable_routed_scaling` onto routed scaling" would therefore be a deliberate divergence from the reference rather than a port of it. It is unobservable on Ling-lite-1.5, where `routed_scaling_factor` defaults to 1.0, but it is observable on any checkpoint that sets a non-unit factor together with the flag false.

Pick one behavior explicitly: either mirror upstream (always scale, flag parsed and ignored) or honor the flag (scale only when true). State which one was chosen and why in the PR body, and pin it with a unit test so a later reader cannot mistake it for an oversight.

## Shared experts are one wide MLP, not N separate experts

`shared_dim = moe_shared_expert_intermediate_size or moe_intermediate_size`, and the shared MLP is constructed with `intermediate_size = shared_dim * num_shared_experts`. For Ling-lite-1.5 that is a single MLP of width `1408 * 2 = 2816`, which the real tensors confirm: `mlp.shared_experts.gate_proj.weight` is `[2816, 2048]` and `mlp.shared_experts.down_proj.weight` is `[2048, 2816]`. There is no per-shared-expert axis anywhere in the checkpoint.

It is built only when `num_shared_experts > 0 and moe_router_enable_shared_expert`, it is added to the routed mixture at a fixed weight of 1.0 (`out = out + self.shared_experts(x)`), and it is never packed into the switch tensors. Do not stack it into `SwitchLinear` and do not give it a routing weight.

## Vendor config keys upstream partly ignores

`config.json` sets `norm_head: false` and `norm_softmax: false`.

- `norm_head` is live upstream, contrary to what a quick read suggests. mlx-lm's `ModelArgs` declares it, `Model` stores it, and `sanitize` L2-normalizes `lm_head.weight` along axis 0 in float32 with a `+1e-7` epsilon before casting back to the original dtype. The vendored `modeling_bailing_moe.py` does the same normalization at inference time, once, then clears the flag. Implement it in `sanitize` to match, and unit-test it: it is a few lines, and a checkpoint that sets it true would otherwise produce wrong logits with no error.
- `norm_softmax` is dead everywhere. mlx-lm declares it in `ModelArgs` and never reads it, and it does not appear anywhere in the vendored `modeling_bailing_moe.py`; `configuration_bailing_moe.py` does not even name it, so it survives only via `**kwargs`. A checkpoint setting it true is asking for behavior no released code defines. Reject that config at load rather than parsing it and silently ignoring it.

## Weight remap

Upstream `sanitize` does two things per layer at or after `first_k_dense_replace`:

1. Stacks `mlp.experts.{e}.{gate_proj,down_proj,up_proj}.{weight,scales,biases}` over `e in 0..num_experts` into `mlp.switch_mlp.{m}.{k}`. The `scales` and `biases` variants exist only for quantized checkpoints; the raw bf16 checkpoint has `weight` only.
2. Renames the router `mlp.gate.weight` to `mlp.gate.gate_proj.weight`, and `mlp.gate.bias` to `mlp.gate.gate_proj.bias` when present.

Watch the naming collision in step 2: `mlp.gate.*` is the router, while `gate_proj` is the SwiGLU gate of an expert MLP, and after the rename the router itself lives at `mlp.gate.gate_proj.weight`. A remap rule that matches on `gate_proj` without anchoring the prefix will swallow the router weight.

Upstream also carries a `quant_predicate` that quantizes `mlp.gate.gate_proj` at `group_size 64, bits 8` while leaving the rest at the default, and a `cast_predicate` that excludes `expert_bias` from dtype casting. Both matter only for quantized conversions, but note them if one is validated later.

## Architecture notes

- DeepSeek-style MoE block:
  - `num_experts` routed experts (64 here) plus a single shared MLP sized `moe_intermediate_size * num_shared_experts`, gated on `moe_router_enable_shared_expert`.
  - Optional `n_group` grouped top-k routing (experts partitioned into `n_group` groups, `topk_group` groups kept, the rest zeroed before top-k). Inactive by default and inactive on this checkpoint.
  - Optional router expert bias (`moe_router_enable_expert_bias`), the same role as DeepSeek's `e_score_correction_bias`, applied to selection only.
  - Routed scaling by `routed_scaling_factor`, unconditional upstream (see above).
  - `norm_topk_prob`, applied only when `top_k > 1`, with a `1e-20` epsilon in the denominator.
  - `first_k_dense_replace` layers before the MoE layers use a plain `BailingMoeMLP` at `intermediate_size`. Zero on this checkpoint.
- Standard RMSNorm + RoPE attention, no exotic attention variant, but the QKV projection is fused into one `query_key_value` tensor split at `[q_size, q_size + kv_size]`, and the output projection is named `dense`.
- Optional QK RMSNorm per head (`use_qk_norm`), applied after the head reshape and before RoPE. Off on this checkpoint.

## Implementation plan

Much of this maps onto mlxcel's existing DeepSeek-V2/V3 MoE path (`src/models/deepseek_v2.rs`, `src/models/deepseek_v3.rs`), which already implements grouped top-k routing, shared experts, the `e_score_correction_bias` router bias, and routed scaling. Reuse that machinery where the semantics actually match, and do not assume they match by default:

- Experts: reuse `src/models/switch_layers.rs` (`SwitchLinear` / fused experts) as the DeepSeek path already does.
- Routing: reuse the grouped top-k router and shared-expert combination, but parameterize the score function (`softmax` default, not DeepSeek's sigmoid), keep the bias selection-only, keep the `top_k > 1` guard and the `1e-20` epsilon on `norm_topk_prob`, and decide the `moe_router_enable_routed_scaling` question explicitly.
- Shared expert: one MLP at `shared_dim * num_shared_experts`, added at weight 1.0, outside the switch tensors.
- Attention: RMSNorm + RoPE with a fused `query_key_value` split and a `dense` output projection, reusing existing shared layers.
- Main new work is config field mapping (including the optional-with-defaults set above), the Bailing weight-key naming in `sanitize` (expert stacking plus the router rename), and the `norm_head` lm_head normalization.
- New model file under `src/models/` (for example `src/models/bailing_moe.rs`), following `docs/adding-models.md`.

## Validation

A token-exact comparison against the mlx-lm reference requires `--no-chat-template` on both sides. The reference harness feeds the raw prompt, while this checkpoint does ship a chat template (embedded in `tokenizer_config.json` as a `<role>HUMAN</role>` style template; there is no separate `chat_template.json`). Without the flag the two sides tokenize different strings and diverge on the first token for reasons unrelated to the port.

Captured reference for the raw prompt `What is the capital of France?`: generation stops at EOS after 11 token ids.

- text: `\n\n\nThe capital of France is Paris.`
- ids: `198 198 198 678 7706 300 11406 341 13997 13 126081`

The trailing `126081` is `<|endoftext|>`, which `config.json` sets as both `eos_token_id` and `pad_token_id`.

Then run a second time without `--no-chat-template`. That run is the real-usage sanity check; it is expected to produce different text, and it should be coherent and terminate.

## Defensive requirements

Every port already merged in this series (#833, #834, #836, #837) produced findings of the same class, so these are requirements rather than suggestions. The underlying hazard: a config-derived value that passes Rust-side validation but violates an undocumented precondition of an MLX C++ entry point terminates the process at the first inference rather than failing at load, because those bridge functions return `UniquePtr` rather than `Result` and a C++ throw crossing the cxx bridge is an uncatchable `std::terminate`. Rejecting at load is the only containment.

- Gather and slice bounds must be taken from real tensor shapes, not from config fields. MLX `take` wraps negative indices but does not range-check positive ones, so an out-of-range expert index or slice offset silently reads adjacent memory and the result reaches the logits as plausible numbers.
- Magnitude validation on the architecture scalars (`hidden_size`, `num_hidden_layers`, `num_attention_heads`, `num_key_value_heads`, `num_experts`, `num_experts_per_tok`, `num_shared_experts`, `intermediate_size`, `moe_intermediate_size`, `moe_shared_expert_intermediate_size`, `vocab_size`, `max_position_embeddings`, `n_group`, `topk_group`, `first_k_dense_replace`) before any `Vec::with_capacity`, truncating `as i32` cast, or use as a divisor. Zero checks go before divisibility checks: `0.is_multiple_of(0)` is true in Rust, so a divisibility check on its own lets a zero divisor through. `src/models/gpt_neox.rs` has the established shape of this (`MAX_*` ceilings plus a `validate()` returning `Result<(), String>`).
- Routing-specific bounds beyond magnitude: `num_experts_per_tok` at least 1 and no greater than `num_experts`; when `n_group > 1`, `num_experts` divisible by `n_group`, and `topk_group` between 1 and `n_group`. Upstream computes `k = n_group - topk_group` and calls `argpartition(kth=k - 1)`, which is out of range as soon as `topk_group >= n_group`.
- RoPE dims validated even, positive, and no larger than the last axis before reaching `fast_rope`. `partial_rotary_factor` is a float, so a hostile value can be NaN or infinite; the saturating `as i32` cast turns those into 0 and `i32::MAX` respectively, and both must be rejected with a message naming the actual field. `rope_theta` must be finite and positive, since RoPE takes its logarithm.
- `rms_norm_eps` validated finite and positive.
- Reject `norm_softmax: true` rather than ignoring it, per the section above.
- The tree-wide unvalidated `quantization` block is tracked separately as #929. This port only needs to avoid adding new exposure, which the per-model `validate_quantization` pattern already present in `src/models/gpt_neox.rs` achieves at negligible cost.

**Delivered and closed.** Implemented by PR #946, merged as `a3c823611`; `mlxcel arch` now reports the family. Test coverage: 42 unit tests in `src/models/bailing_moe_tests.rs`, mutation-checked by reverting five behaviors and observing exactly five paired failures. Real-checkpoint validation: `models/ling-lite-1.5` reproduces the captured reference ids `198 198 198 678 7706 300 11406 341 13997 13` and stops at EOS `126081`, run with `--no-chat-template` on both sides so the reference and mlxcel see the same raw prompt. Note the gate does NOT cover grouped routing or the expert-bias branch, because `n_group` defaults to 1 and `moe_router_enable_expert_bias` defaults false, so those two rest entirely on the unit tests. The unticked boxes below are the original pre-implementation plan and were not maintained during the work; treat the merged PR and its review thread as the record of what shipped.

- [ ] Config struct + serde parsing for the Bailing MoE config, with `n_group`, `topk_group`, `moe_router_enable_expert_bias`, `moe_router_enable_routed_scaling`, `routed_scaling_factor`, `score_function`, `moe_shared_expert_intermediate_size`, `moe_router_enable_shared_expert`, `use_qk_norm`, `partial_rotary_factor`, and `rotary_dim` all optional and defaulted to the upstream values in the table above. A parse test over the verbatim `models/ling-lite-1.5/config.json` proves none of them is required.
- [ ] `score_function` read from config and both `"softmax"` and `"sigmoid"` implemented, with a unit test that the default is softmax and that an unknown value is rejected rather than silently falling back.
- [ ] `from_weights` constructor wiring the MoE block (routed experts, single wide shared MLP added at weight 1.0, grouped top-k router) and RMSNorm/RoPE attention with the fused `query_key_value` split.
- [ ] Router implemented with selection on the biased scores and weights gathered from the unbiased scores, `norm_topk_prob` applied only when `top_k > 1` with the `1e-20` epsilon, and the float32 score computation cast back at the end. Unit tests pin each of these three separately.
- [ ] `moe_router_enable_routed_scaling` behavior chosen explicitly (mirror upstream or honor the flag), stated in the PR body, and pinned by a unit test.
- [ ] Grouped routing (`n_group > 1`) and the expert-bias path implemented and unit-tested against synthetic configs, since the validation checkpoint reaches neither.
- [ ] `first_k_dense_replace` dense-prefix path implemented and unit-tested against a synthetic config, since the validation checkpoint sets it to 0.
- [ ] `sanitize` stacking `mlp.experts.{e}.{gate_proj,down_proj,up_proj}.{weight,scales,biases}` into the fused `SwitchLinear` layout and renaming `mlp.gate.weight` / `mlp.gate.bias` to `mlp.gate.gate_proj.*`, with a test that the router weight is not captured by the expert `gate_proj` rule.
- [ ] `norm_head` implemented in `sanitize` (axis-0 L2 normalization of `lm_head.weight` in float32 with `+1e-7`, cast back) and unit-tested.
- [ ] `norm_softmax: true` rejected at load with a message explaining that no released implementation defines it.
- [ ] Config validation covering the magnitude, routing, RoPE, and norm-eps requirements in the Defensive requirements section, with a rejection test per rule.
- [ ] Arm added to `src/models/detection.rs` for the `bailing_moe` architecture / `model_type`.
- [ ] TP / distributed arch-string entry if applicable.
- [ ] Unit tests in the model's `_tests.rs` (config parse, sanitize expert remap and router rename, router selection, forward shape) in addition to the targeted tests above.
- [ ] `docs/supported-models.md` updated to list Bailing MoE / Ling.
- [ ] Real-checkpoint validation: `mlxcel arch` reports the architecture, and `./target/release/mlxcel generate` on `models/ling-lite-1.5` reproduces the captured reference token ids exactly with `--no-chat-template`.
- [ ] Second real-checkpoint run without `--no-chat-template` produces coherent output and terminates.

Done only when the model loads and generates from the real checkpoint (per `docs/adding-models.md`).

## Effort

MEDIUM. The expert machinery and the general MoE shape already exist on the DeepSeek-V2/V3 path, but the reuse is not free: the score function default differs from DeepSeek, the shared expert is one wide MLP rather than N experts, several config fields are absent from real checkpoints and must be defaulted, and two of the routing branches cannot be validated against any available checkpoint and need synthetic-config tests instead. Add the usual registration, detection, validation, test, and doc touchpoints.

inureyes:
## Summary

Ports the Ant Group Ling / Bailing MoE (`bailing_moe`) text architecture from mlx-lm. Bailing is a DeepSeek-shaped sparse decoder (RMSNorm before attention and before the FFN, grouped-query attention with RoPE, routed SwiGLU experts, one always-on shared expert), so the expert machinery is the shared `switch_layers` path unchanged. The work went into the places where reusing the DeepSeek-V2/V3 path as-is would be silently wrong, and into the config guards that have to reject at load because an MLX C++ exception crossing the cxx bridge is an uncatchable `std::terminate` at the first forward pass rather than a Rust error.

This PR is four commits: the port itself, a review-fix pass that closed a gap in the output-head and quantized-embedding validation and registered the new `Bailing` display family, a follow-up that also fixes two `FAMILY_ORDER` entries (`MiniMax VLM`, `Step VLM`) that were already missing on `main` before this branch touched that array, a security fix on an unbounded cast in the router config, and a finalization pass that brought `docs/supported-models.md` up to date with the review-fix commit and corrected an overstated doc comment. Details below.

## The three decisions this PR makes explicitly

**`moe_router_enable_routed_scaling` is dead code upstream, and this port mirrors upstream.** `BailingMoeGate` stores the flag and never reads it again: `__call__` passes `routed_scaling_factor` unconditionally and `group_expert_select` ends with an unconditional `scores = scores * routed_scaling_factor`. Honoring the flag would be a deliberate divergence from the reference rather than a port of it, so this loader multiplies unconditionally and a checkpoint decodes here exactly as it decodes under mlx-lm. The two readings coincide at the default factor of 1.0, which is the only value any published checkpoint uses; they diverge only for a non-unit factor with the flag false, and `ModelArgs::warn_on_ignored_routed_scaling_flag` prints a diagnostic naming both fields for exactly that combination, so the choice is not invisible. Pinned by `routed_scaling_is_applied_even_when_the_flag_is_false`.

**`norm_head` is implemented, `norm_softmax` is rejected.** `norm_head` is live upstream despite looking vestigial: mlx-lm's `sanitize` L2-normalizes `lm_head.weight` along axis 0 in float32 with a `1e-7` epsilon and casts back to the original dtype, and the vendored `modeling_bailing_moe.py` does the same once at inference. It is implemented here in `normalize_lm_head_weight`, and refused in combination with a quantized head, whose stored `.weight` is a packed bit field rather than the weight itself. `norm_softmax` is declared by upstream's `ModelArgs` and never read, does not appear anywhere in the vendored `modeling_bailing_moe.py`, and is not even a named parameter of `configuration_bailing_moe.py` (it survives only through `**kwargs`), so a checkpoint setting it true is asking for behavior no released implementation defines. That is rejected at load rather than parsed and silently ignored.

**Grouped routing and the router expert bias are implemented and unit-tested, but the real-checkpoint gate does not cover them.** `inclusionAI/Ling-lite-1.5` declares neither `n_group` nor `moe_router_enable_expert_bias`, and both default to off, so a token-exact real-model run says nothing about either branch. They are pinned against hand-computed values instead: `group_mask_scores_keeps_the_groups_with_the_largest_top_two_sum` uses values where the group total and the group max would each keep a different pair of groups than the top-two sum does, `grouped_routing_restricts_the_selection_to_the_surviving_groups` builds a case where the globally best expert lives in the losing group, and the two expert-bias tests cover both the tensor-present and tensor-absent paths (upstream initializes zeros when the flag is set but no tensor ships). `first_k_dense_replace` is 0 on Ling-lite too, so the dense-prefix branch gets the same synthetic-config treatment.

## What else the port had to get right

- The checkpoint uses GPT-NeoX naming, not Llama naming: the token table is `model.word_embeddings`, the output projection is `attention.dense`, and Q, K and V arrive fused in one `attention.query_key_value` matrix. On Ling-lite that matrix is `[3072, 2048]`, which is `(16 + 2 * 4) * 128` for 16 query heads and 4 KV heads of width 128, so the split offsets are `(2048, 2560)`, the uneven GQA form and not an even three-way split. MLX's `slice` clamps an out-of-range stop instead of throwing, so the even split would produce tensors of exactly the right shape assembled from the wrong channels.
- The raw checkpoint stores each routed expert separately at `mlp.experts.{e}.{gate,up,down}_proj`, 64 experts times 3 projections times 28 layers on Ling-lite. `SwitchLinear` joins those individually-keyed tensors into its stacked `[num_experts, out, in]` layout at load, and an mlx-lm conversion that already ships the pre-stacked `mlp.switch_mlp.{proj}` form loads unchanged on the same path.
- Upstream's `sanitize` renames the router from `mlp.gate.weight` to `mlp.gate.gate_proj.weight`, which ends in the same `gate_proj` segment every routed expert's SwiGLU gate uses. `router_prefix` probes two fully anchored keys rather than matching a suffix, so both the raw `inclusionAI` export and an mlx-lm conversion load and neither can read an expert projection as the router. `router_lookup_is_anchored_against_the_expert_gate_proj` puts both tensors in one map at deliberately different widths and asserts which one comes back.
- Selection uses the biased scores while the returned weights come from the unbiased ones. `orig_scores` is captured before the correction bias is added, the top-k indices come from the biased copy (and from the group-masked copy when `n_group > 1`), and the weights are gathered from `orig_scores`. Gathering from the biased copy leaves the output finite and plausible while misweighting every routed contribution, so the test constructs a bias large enough to change the selected set and then asserts both the weights it must produce and the value the wrong gather would have produced.
- `norm_topk_prob` is applied only when `num_experts_per_tok > 1` and divides by `sum + 1e-20`, not by a bare sum. The epsilon is pinned on an all-zero score row, which the softmax path cannot produce but the sigmoid path can.
- `score_function` defaults to `"softmax"`, not to DeepSeek-V3's sigmoid, and an unrecognized spelling is rejected rather than falling back. This is the single place where taking the DeepSeek gate as-is changes the routing distribution on every token while producing perfectly finite output.
- The shared expert is one wide MLP of `(moe_shared_expert_intermediate_size or moe_intermediate_size) * num_shared_experts`, which is `1408 * 2 = 2816` on Ling-lite, added to the routed mixture at a fixed weight of 1.0 and never packed into the switch tensors. There is no per-shared-expert axis anywhere in the checkpoint.
- Every routing knob is optional with the upstream default, because Ling-lite declares none of `n_group`, `topk_group`, `score_function`, `routed_scaling_factor`, `moe_router_enable_expert_bias`, `moe_router_enable_routed_scaling`, `moe_shared_expert_intermediate_size`, `moe_router_enable_shared_expert`, `use_qk_norm`, `partial_rotary_factor` or `rotary_dim`. Declaring any of them required would fail the parse of the real file outright.

## Config-trust guards

`config.json` arrives from a third-party HuggingFace repo in the `mlxcel generate -m <org>/<repo>` flow, and the download layer never parses it, so `ModelArgs::validate` and `validate_weights` reject before anything reaches MLX: architecture scalars that could size an allocation, divide, or truncate through an `as i32` cast, with the zero checks placed before the divisibility checks because `0.is_multiple_of(0)` is true in Rust; routing parameters that would put `argpartition(kth = n_group - topk_group - 1)` out of range or reshape the score row into unequal groups, neither of which upstream checks, including a `topk_group` that is carried into the gate with a saturating rather than a truncating cast so an absurd value above `i32::MAX` cannot become negative; a rotary width that is zero, odd or wider than the head, including the NaN and infinity a float `partial_rotary_factor` can express through a saturating cast; a non-positive or non-finite `rope_theta` or `rms_norm_eps`; a `rope_scaling` block other than absent, empty or `"default"`, since upstream threads it into `initialize_rope` while this loader always builds the plain rotation; a `quantization` block whose `bits` falls outside `1..=32` or whose `group_size` is non-positive; the output head's input width against `hidden_size`, checked exactly against upstream's own strict `nn.Linear(hidden_size, vocab_size)` load and skipped only when `tie_word_embeddings` ships no separate `lm_head` tensor; a quantized token table's own input width, reconstructed the same way; and any tensor whose real shape disagrees with the config, on both axes, per expert index rather than only the first (a checkpoint missing a late expert would otherwise register a short stack the router can index past), and on the quantized path, where packing compresses the input axis only so the input width is recovered as `scales.shape(-1) * group_size` the way MLX recovers it.

## Review, security and finalization fixes

Three defects surfaced in review, found by comparing the port against guards its sibling ports (helium, gpt_bigcode, gpt_neox) already carry: `validate_weights` never checked the output head, so a checkpoint whose `lm_head` disagreed with `hidden_size` reached `mx::matmul` (or `quantized_matmul`) at the first forward pass instead of failing at load; the same input-width reconstruction was missing on the quantized token table, whose width the shared `validate_embedding_table` deliberately leaves to the scales; and `ModelType::BailingMoe`'s `"Bailing"` family label was absent from `FAMILY_ORDER` in `src/main.rs`, which `family_order_is_exhaustive` catches. That same test was already red on `main` independently of this branch, because `"MiniMax VLM"` and `"Step VLM"` (added by #800 and #781) were never added to `FAMILY_ORDER` either; since this branch already touches that array, both entries are added alongside `"Bailing"` rather than leaving the surrounding test red. The review pass also updated the `Used by:` lists on the four shared helpers `bailing_moe` consumes (`gpt2::dim_eq`, `gpt2::validate_embedding_table`, `switch_layers::group_mask_scores`, `switch_layers::moe_weighted_sum`), per `docs/code-guidelines.md`.

A follow-up security fix addressed an unbounded cast: `BailingMoeGate::from_weights` carried `topk_group` into the gate with a plain `as i32`, and since `validate_routing` only bounds that field inside the `n_group > 1` branch while `n_group` defaults to 1 and is absent from every published Bailing config, the field reached the cast completely unbounded and a value above `i32::MAX` would truncate to a negative number. The stored value is unreachable today (`forward` reads `topk_group` only under `n_group > 1`, where the bound does apply), so this is defense in depth rather than a live bug, fixed by saturating instead of truncating, the same way `ModelArgs::rope_dims` already handles an absurd `rotary_dim`.

A finalization pass verified each of the fix commits above has a regression test that actually fails without it, by reverting each in isolation and observing the paired test fail before restoring, and brought two documentation gaps up to date: `docs/supported-models.md` was written for the first commit and did not mention the output-head/quantized-embedding validation or the per-expert-to-switch-layout join, and a doc comment on `validate_quantized_packing` claimed the check is "exactly as strict as the loader itself" without qualification, which is true for the plain-projection callers but not for `validate_experts`, since `SwitchLinear::from_weights` pins the quantization mode to `"affine"` unconditionally rather than inferring it from `.biases` presence the way `UnifiedLinear::from_weights` does. The comment is corrected; no behavior changed.

Not addressed here, all tracked separately: the tree-wide unvalidated `quantization` block (#929); a `group_mask_scores` missing-bounds-check finding affecting five other MoE families, filed as its own issue; a redundant per-expert `Copy` in `src/models/switch_layers.rs`, filed as its own issue; and the `topk_group == n_group` divergence from upstream, which is a deliberate tested decision, not a defect.

## What changed

- `src/models/bailing_moe.rs`: config and validation, the router, the MoE block, attention, the decoder layer and the model, plus `router_prefix`, `normalize_lm_head_weight`, the weight-shape guards, the output-head and quantized-embedding validation added in review, and the saturating `topk_group` cast.
- `src/models/bailing_moe_tests.rs`: 42 unit tests, all checkpoint-free.
- `src/models/mod.rs`, `src/models/detection.rs`, `src/model_metadata.rs`, `src/loaded_model.rs`: the module, `ModelType::BailingMoe`, the `bailing_moe` detection arm, the `for_each_model_registration!` entry and the `LoadedModel` dispatch arm.
- `src/distributed/tensor_parallel/inference.rs`: the arch-string arm that keeps the dispatch table total. Tensor parallelism is refused for this family by `validate_supported_runtime`, because the generic transformer shard plan names neither the fused `attention.query_key_value` nor the `mlp.experts.{idx}` / `mlp.switch_mlp` expert key space.
- `src/main.rs`: `FAMILY_ORDER` gains `"Bailing"`, `"MiniMax VLM"` and `"Step VLM"`, the last two independent of this port and already missing on `main`.
- `src/models/gpt2.rs`, `src/models/switch_layers.rs`: `Used by:` list updates for the shared helpers this port consumes.
- `docs/supported-models.md`: the Bailing MoE entry, kept current with every commit in this series.

## Test plan

- [x] `cargo check --release --lib --tests --features metal,accelerate` clean, except the two pre-existing `host_preprocessor` warnings.
- [x] `cargo clippy --release --lib --tests --features metal,accelerate` clean, same two pre-existing warnings and no new lints.
- [x] `cargo test --release --lib --features metal,accelerate models::bailing_moe`: 42 passed, 0 failed.
- [x] `cargo test --release --features metal,accelerate --bin mlxcel family_order`: 2 passed (`family_order_is_exhaustive`, `family_order_has_no_orphans`).
- [x] Mutation check, so the suite bites rather than merely coexisting with correct code: reverting the selection-only bias, the `top_k > 1` guard, the `1e-20` normalization epsilon, the unconditional routed-scaling multiply and the `1e-7` `norm_head` epsilon each fails its own test and nothing else.
- [x] Regression check on the review, security and doc-comment fixes: reverting the `lm_head` validation fails `validate_weights_rejects_an_output_head_that_disagrees_with_the_config` with "an output head of the wrong width"; reverting only the quantized `word_embeddings` check (with `lm_head` left intact) fails the same test later at "a token table packed for a different input width"; reverting the saturating `topk_group` cast fails `an_absurd_topk_group_saturates_rather_than_truncating_when_grouping_is_off` with `left: -1, right: 2147483647`; and removing the three added `FAMILY_ORDER` entries fails `family_order_is_exhaustive` with `missing: ["MiniMax VLM", "Bailing", "Step VLM"]`. Each revert leaves every other test in the suite passing.
- [x] Positive control: the verbatim `models/ling-lite-1.5/config.json` and the full 5603-tensor Ling-lite shape signature both pass every guard, so the guards cannot be tightened into rejecting the genuine checkpoint.
- [x] Real-checkpoint validation on `models/ling-lite-1.5` (33.6 GB): `mlxcel list` reports the architecture, `mlxcel generate --no-chat-template -p "What is the capital of France?"` reproduces the captured reference ids `198 198 198 678 7706 300 11406 341 13997 13 126081` exactly, and a second run without `--no-chat-template` produces coherent output and terminates.

Closes #838


inureyes:
Real-checkpoint validation on Apple M1 Ultra (Metal + Accelerate, release build), against `models/ling-lite-1.5` (raw upstream bf16, 5603 tensors across 4 shards, 31.30 GB resident after the Apple Silicon f16 conversion, 19.2 s to load).

**Token-exact against the mlx-lm reference.** With the raw prompt `What is the capital of France?` and `--no-chat-template` on both sides, mlxcel emits `\n\n\nThe capital of France is Paris.` and stops at EOS after 10 content tokens, matching the reference ids `198 198 198 678 7706 300 11406 341 13997 13` followed by `126081`. `mlxcel arch` reports the family under a `Bailing:` heading as `Ling / Bailing MoE (shared + routed experts)`.

The flag matters for the comparison and not for correctness: this checkpoint does ship a chat template, so `--no-chat-template` is what makes both sides see the same raw prompt. As a separate real-usage check, the default templated path answers cleanly: prompt `What is the capital of France? Answer in one sentence.` yields `The capital of France is Paris.` in 7 tokens.

**What this gate does and does not establish.** It exercises the plain softmax top-k router, the shared expert added at weight 1.0, the uneven fused QKV split, GQA with 16 query heads against 4 kv heads, and the join of 5376 per-expert tensors into the `SwitchLinear` layout across all 28 MoE layers, since `first_k_dense_replace` is 0 and there is no dense prefix. It does NOT cover grouped routing or the expert-bias branch, because `n_group` defaults to 1 and `moe_router_enable_expert_bias` defaults false, so neither runs on this checkpoint. Those two paths rest entirely on the unit tests, which is worth weighting in review.

Throughput is 24.0 tok/s raw and 18.2 tok/s templated for a 16.8B model with 2.75B active. The loader correctly warns that a quantized conversion would be substantially faster on Apple Silicon.

inureyes:

### Intent

Port the Ant Group Ling / Bailing MoE (`bailing_moe`) text architecture from mlx-lm, with the two routing branches no available checkpoint reaches (grouped top-k, router expert bias) covered by unit tests instead of the real-model gate.

The port is faithful. I re-derived the router against upstream `group_expert_select` line by line and found no arithmetic divergence: selection-only bias, weights gathered from `orig_scores`, the `top_k > 1` guard, the `1e-20` epsilon, the f32 computation cast back at the end, the unconditional routed-scaling multiply, the uneven GQA QKV split, the one-wide shared expert at weight 1.0. `group_mask_scores` reproduces upstream exactly: unflatten to `[n, n_group, epg]`, top-2 sum per group (via `argpartition` on the negated array, equivalent to `mx.topk(scores, 2)`), `argpartition(kth = k - 1, axis = -2)`, `[..., :k, :]`, `put_along_axis` zero, flatten. The diff is purely additive: two new files plus 22 lines across six registration and doc surfaces, zero deletions and zero edits to existing logic.

### Findings Addressed

- [x] **HIGH** `src/models/bailing_moe.rs:1209` `validate_weights` never checked `lm_head`. It is the one projection that skipped `validate_projection`, and the axis it skipped is the one that aborts. The row count only bounds an argmax, but the input width is the inner dimension of the matmul that produces the logits, and `mlx/ops.cpp:3329` throws `std::invalid_argument` on a mismatch. `matmul` and `quantized_matmul` cross the cxx bridge as `UniquePtr<MlxArray>` rather than `Result`, so that throw is an uncatchable `std::terminate` at the first forward pass, exactly the defect class every prior PR in this series produced. `helium.rs:714` and `gpt_bigcode` both check theirs. Fixed with an exact-row `validate_projection` call (upstream loads the tensor into `nn.Linear(hidden_size, vocab_size)` under a strict `load_weights`, so exact is the reference's own requirement), skipped under `tie_word_embeddings`.
- [x] **HIGH** `src/models/bailing_moe.rs:1218` The quantized token table's input width was also unchecked. The shared `gpt2::validate_embedding_table` deliberately checks the width only on the float path, so a table packed for a different `hidden_size` reached the first RMSNorm as the wrong shape. Closed with the file's own `validate_quantized_packing`, a no-op for an unquantized table.
- [x] **HIGH** `src/main.rs:1823` `ModelType::BailingMoe` introduced the brand-new family label `"Bailing"` (`models/mod.rs:757`) without adding it to `FAMILY_ORDER`. `family_order_is_exhaustive` in `main_tests.rs` failed with `missing: ["MiniMax VLM", "Bailing", "Step VLM"]`, and `mlxcel arch` renders an unlisted family appended alphabetically instead of next to the other vendor MoE families. Added after `"Hunyuan"`. The other two are absent on `main` as well and are left alone.
- [x] **MEDIUM** Four shared helpers this port consumes did not name it in their `Used by:` lists, which `docs/code-guidelines.md` makes the discovery mechanism for what a change to a shared function breaks: `gpt2::dim_eq`, `gpt2::validate_embedding_table`, `switch_layers::group_mask_scores`, `switch_layers::moe_weighted_sum`.

Added `validate_weights_rejects_an_output_head_that_disagrees_with_the_config`: a positive control, a wrong-width head, a head short of `vocab_size`, a missing head, the tied case, and both the honest and mis-packed quantized forms plus the same mis-packing on the token table. 41 tests pass, up from 40.

### Remaining Items

- [ ] **MEDIUM** `hidden_act` is not a config field at all, and `models/ling-lite-1.5/config.json` declares `"hidden_act": "silu"`. A checkpoint declaring anything else runs through `compiled_swiglu_activation` silently. Upstream hardcodes `swiglu` too, so rejecting would diverge, but `gpt_neox.rs:513 activation_is_gelu` sets the precedent of reporting rather than dropping. The right shape here is a `hidden_act: Option<String>` plus a warning next to the existing `warn_on_ignored_routed_scaling_flag`. Not fixed: MEDIUM, and it is a deliberate behavior addition rather than a correction.
- [ ] **MEDIUM** `validate_routing` rejects `topk_group == n_group`, which upstream accepts and handles correctly. `k = n_group - topk_group` is 0, so `kth = -1`, and `mlx/ops.cpp:2757` normalizes a negative `kth` (`kth_ = kth + a.shape(axis)`) to `n_group - 1`, which is in range; the `[..., :0, :]` slice is then empty and no group is zeroed, so all groups survive. That is the natural reading of "keep all `n_group` groups", and issue #838's own acceptance criterion says `topk_group` between 1 and `n_group`. Not fixed: relaxing the guard would newly exercise `put_along_axis` with an empty index array, which nothing here tests, and no published Bailing checkpoint sets `n_group > 1` at all.
- [ ] **MEDIUM** `ModelArgs::score_function()` (`bailing_moe.rs:446`) maps any unrecognized spelling to `Softmax` and relies on `validate()` to have rejected it. `BailingMoeGate::from_weights` is `pub` and does not revalidate, so a direct caller silently softmaxes. Only in-crate caller is the model, which validates first.
- [ ] **MEDIUM** `quant_mode` (`bailing_moe.rs:894`) is a third copy of the mode heuristic already inlined twice in `mlxcel-core/src/layers.rs` (lines 260 and 1121). The copies can drift silently. The fix is a shared `detect_quant_mode` in core used by all three, which is a core change and belongs in its own PR.
- [ ] **MEDIUM** Pre-existing, not this PR: `switch_layers::moe_weighted_sum`'s `Used by:` list names 21 models but 30 files call it (missing `cohere2_moe`, `ernie4_5_moe_vl`, `dots1`, `granitemoehybrid`, `lfm2`, `mellum`, `minimax_m3_moe`), and `group_mask_scores` omits `dots1` and `solar_open`. Worth a separate cleanup.
- [ ] **LOW** `validate_projection`'s `looks_transposed` hint fires whenever the actual rows equal the expected `in_features`, so a wrong-but-square weight gets told it is in `[in, out]` orientation when it is not a transpose of anything. Message-only. Tightening it to also require `dim_eq(shape[1], out_features)` would fix it.
- [ ] **LOW** The `mx::full` laziness the 5603-tensor positive control depends on is documented (`bailing_moe_tests.rs:1809`) but not asserted, so an eager `full_f32` would degrade the test into materializing tens of GB rather than failing loudly. `mlxcel-core` exposes no evaluation-status predicate today, so pinning it needs a new bridge function.
- [ ] **LOW** No `bailing_moe` arm in `detection_tests.rs`, unlike `helium` and `gpt_bigcode`. The `model_type` string is unambiguous, so this is consistency rather than a hole.

### Judgement on the other self-reported observations

The three implicit contracts hold up. `sigmoid(-200) == 0.0` is pinned by an explicit `assert_eq!` on the raw sigmoid before the epsilon assertion, so a change in MLX underflow fails with a precondition message. The f16 contract is pinned twice, once against `logit_dtype` and once against `FLOAT16` literally, so a `Linear::forward` that promoted would fail the second assertion. The `mx::full` one is not pinned (above).

The router-collision test does discriminate despite not being mutation-checked. It asserts the resolved prefix does not contain `.experts.`, and separately that the loaded projection emits `num_experts` (4) rather than `moe_intermediate_size` (16) channels. Any unanchored suffix match resolves into the expert key space and fails both, independently of what such a mutation would do to `validate_weights`.

### Verification

- [x] All stated requirements implemented (all 18 acceptance criteria in #838 are met)
- [x] No placeholder, mock, or stub code
- [x] Integrated into the code flow: `models/mod.rs` (module, re-export, `ModelType`, `ALL_MODEL_TYPES`, `metadata()`, `all_variants!`), `detection.rs:191`, `model_metadata.rs:192`, `loaded_model.rs:152` and its `delegate_language_model!` arm, `inference.rs:212`, `docs/supported-models.md:28`, and now `main.rs` `FAMILY_ORDER`. TP is refused because `runtime_kind_for` (`llama_runtime.rs:2195`) has no arm and falls through to `_ => None`, matching how `helium` and `gpt_bigcode` do it.
- [x] Existing modules reused: `switch_layers` `SwitchGLU` / `group_mask_scores` / `moe_weighted_sum`, `UnifiedLinear` / `UnifiedEmbedding` / `RMSNorm`, `gpt2::dim_eq` / `validate_embedding_table`
- [x] No unintended structural changes: additive only, no deletions
- [x] Tests pass: `models::bailing_moe` 41 passed / 0 failed, parallel and single-threaded; clippy clean apart from the two pre-existing `host_preprocessor` warnings; `cargo fmt --all -- --check` clean

Note on test stability: the very first execution after a fresh link aborted the binary with SIGABRT at 26 of 41 tests. It did not reproduce in eight subsequent runs (six parallel, one single-threaded, one with the new test skipped), so it reads as a cold Metal-initialization race rather than anything in this suite. Recording it in case it recurs in CI.

inureyes:

No CRITICAL and no HIGH findings. One LOW fixed in `dfdeacfe8`; the rest is reported below.

### The defect class this series keeps producing

Enumerated every MLX entry point this port feeds a config-derived argument and checked each precondition against the pinned checkout (`b7c3dd6d27f45b5365b08a840310187dc503f1db`): `fast::rope` (`fast.cpp:409-424`, dims positive / even / not exceeding the last axis, plus `ndim >= 3`), `argpartition` (`ops.cpp:2743-2769`), `scatter_axis` behind `put_along_axis` (`ops.cpp:3615-3675`), `take_along_axis`, `slice`, `matmul`, `quantized_matmul` / `gather_qmm` via `extract_quantized_matmul_dims`, `gather_mm`, `fast::rms_norm`, and the embedding gather. Every one of them is reached only with an argument that `ModelArgs::validate` or `validate_weights` has already bounded, and both run before anything touches a kernel (`from_weights` revalidates rather than trusting that `load` ran first, `bailing_moe.rs:1972-1974`).

Specifically confirmed:

- **`argpartition` pivot, plain branch.** `bailing_moe.rs:1492-1496` takes the expert count from the real score row, not from config, so `kth = n_experts - top_k` is in `0..n_experts` for any `top_k >= 1`. This is stricter than the shared DeepSeek-V3 gate, which passes the config-derived `self.top_k - 1` straight through (`deepseek_v3.rs:547`).
- **`argpartition` pivot, grouped branch.** `group_mask_scores` uses `kth = 1` on an axis of `experts_per_group` and `kth = n_group - topk_group - 1` on an axis of `n_group`. `validate_routing` (`bailing_moe.rs:610-644`) enforces `n_group >= 1`, `num_experts % n_group == 0`, `experts_per_group >= 2` and `1 <= topk_group <= n_group - 1`, which puts both pivots in range and makes the `reshape` into equal groups exact.
- **Expert-bias width.** `validate_projection` on the router requires `dim_eq(shape[0], num_experts)` exactly, and quantization packs the input axis only, so the router's real output width equals `num_experts` on both the float and the quantized path. The synthesized zero bias at `bailing_moe.rs:1537-1541` is therefore the real output width, not merely the config's, and a shipped `expert_bias` tensor is shape-checked at `bailing_moe.rs:1292-1303`.
- `num_experts_per_tok <= num_experts` (`:593`) and `first_k_dense_replace <= num_hidden_layers` (`:552`) are both enforced, zero checks precede divisibility checks throughout, and the only `Vec::with_capacity` is bounded by `MAX_NUM_HIDDEN_LAYERS`.
- No new `quantization` exposure beyond #929: `validate_quantization` bounds `bits` to `1..=32` and `group_size` to `>= 1` per this family, as #838 asked.

### Fixed

**LOW - truncating `as i32` on an unbounded `topk_group`** (`src/models/bailing_moe.rs:1552`). `validate_routing` bounds `topk_group` only inside the `n_group > 1` branch. `n_group` defaults to 1 and `models/ling-lite-1.5/config.json` declares neither field, so on every published Bailing config `topk_group` reached the cast unbounded and a value above `i32::MAX` truncated to a negative number stored in the gate. Unreachable today, since `forward` reads the field only under `n_group > 1`, but #838 lists `topk_group` among the scalars requiring magnitude validation before a truncating cast, and a negative value would become `slice_axis(group_idx, -2, 0, k)` with a negative `k` if a later change did reach it, where `slice_axis` reads `-1` as "to the end of the axis" rather than as an error and would zero every expert group instead of failing. Now saturates via `i32::try_from(...).unwrap_or(i32::MAX)`, matching how `rope_dims` already handles an absurd `rotary_dim`, so no config upstream accepts changes behavior. Pinned by `an_absurd_topk_group_saturates_rather_than_truncating_when_grouping_is_off`.

### Reported, not changed

**MEDIUM - `topk_group == n_group` is rejected here and accepted upstream.** `bailing_moe.rs:632` requires `topk_group <= n_group - 1`. Upstream computes `k = n_group - topk_group = 0` and `argpartition(kth = -1)`, which `ops.cpp:2757` normalizes to `n_group - 1`, then slices `[..., :0, :]` and scatters an empty index set, so all groups survive and the grouped mask is an identity. Two ways to reach the rejection: a config that sets `n_group` and `topk_group` equal, and (more plausibly) a config that sets `n_group: 4` while omitting `topk_group`, whose serde default is 4 both here and upstream. Both load under mlx-lm and fail here.

On the question that led the previous pass to leave it: relaxing would **not** have to reach `put_along_axis` with an empty index set at all, because `topk_group == n_group` means zero groups are masked, so gating the mask on `topk_group < n_group` is bit-exact with upstream and skips the call entirely. And even if it did reach it, the pinned MLX handles it: `broadcast_shapes` propagates a zero extent through a size-1 axis explicitly (`utils.cpp:152-154`, `out_shape[i] = a * b`), and `scatter_axis` over a zero-size index array is a no-op. So the rejection is safe to relax, and the only reason to keep it is that a config declaring grouping that does nothing is arguably worth surfacing. Left as-is because it is a deliberate, documented and tested decision rather than a defect; flipping it is three edits plus the `(4, 4)` case in `config_validation_rejects_grouped_routing_that_would_index_out_of_range`.

**MEDIUM (pre-existing, not this PR) - `group_mask_scores` has no internal bounds check, and Bailing is the only caller that validates its inputs.** `switch_layers.rs:698-700` computes `k = n_group - topk_group` and calls `slice_axis(group_idx, -2, 0, k)`. At `topk_group == n_group + 1`, `k = -1`, `argpartition(kth = -2)` normalizes to a valid pivot, and `slice_axis` reads the `-1` stop as "to the end of the axis", so **every** group is zeroed and every routed score becomes 0 with nothing throwing. `bailing_moe.rs` blocks this. DeepSeekV3, DeepSeekV32, GLM4Moe, GLM4MoeLite and ExaOneMoe all pass `args.topk_group as i32` through with no validation at all. Worth its own issue against those five.

**MEDIUM - `quant_mode()` diverges from the loader on the expert path.** `bailing_moe.rs:894-904` reproduces the mode heuristic inlined in `QuantizedEmbedding::from_weights` (`layers.rs:265-273`) and `UnifiedLinear::from_weights` (`layers.rs:1120-1134`), and matches both. It does **not** match `SwitchLinear::from_weights`, which pins `mode = "affine"` unconditionally (`switch_layers.rs:192`). So for a quantized expert tensor with no `biases`, `validate_quantized_packing` reconciles under `mxfp4`/`mxfp8`/`nvfp4` (bits fixed, group size re-derived) while the loader will use affine (group size trusted, bits re-derived), which makes the doc claim at `bailing_moe.rs:920-923` that the check is "exactly as strict as the loader itself and no stricter" false for `validate_experts`. Inert for now, since no public Bailing checkpoint is quantized and `SwitchLinear` would abort in `gather_qmm` on a biases-free tensor anyway (a pre-existing gap shared by every MoE family). This reinforces the earlier call that de-duplicating the heuristic belongs in its own PR; that PR should also give `SwitchLinear` the same auto-detect the other two loaders already have.

**LOW - `norm_head` builds an unevaluated float32 graph.** `normalize_lm_head_weight` (`:1379-1389`) holds `astype -> square -> sum -> sqrt -> add -> divide -> astype` lazily, so on a Ling-lite-sized head the first forward materializes roughly 3 GB of float32 intermediates from a 518 MB bf16 tensor. Dormant, since `norm_head` is false on every published checkpoint, but an explicit `eval` after the divide would bound it.

### Performance

**Expert join: once at load, no double buffer, confirmed.** `SwitchGLU::from_weights` runs once per layer inside `DecoderLayer::from_weights`, so the 5376 per-expert tensors become 84 stacked tensors exactly once (28 layers x 3 projections x 64 experts). Nothing on the decode path re-stacks. `mx::copy` and `mx::stack` are both lazy graph nodes, so `from_weights` itself allocates nothing; the 31.30 GB is the post-load weight map, whose bf16 -> f16 conversion is materialized one key at a time with the source entry replaced in place (`sanitize.rs:1891-1899`), which is why load does not double buffer. The stacked layout is materialized at the **first forward**, not at load, and MLX detaches each node's inputs as it evaluates, so the transient overshoot is bounded by roughly one projection's worth (64 x 1408 x 2048 x 2 = 369 MB) rather than by the model. Worth knowing that on the server path the first request pays that materialization.

**MEDIUM - 5376 redundant `Copy` primitives at first forward** (`switch_layers.rs:412-423`). `stack_individual_experts_with_count` calls `mlxcel_core::copy` on each per-expert tensor purely to obtain owned `UniquePtr`s for `stack_owned`, but `mx::stack` already allocates its own output buffer, so every one of those copies is a full extra allocate-and-write of the expert weights, about 30 GB of pointless memory traffic on Ling-lite. `ops::stack_owned` already lowers to `ffi::stack(&[*const MlxArray], axis)`, so the borrowed `&MlxArray` from the weight map can be passed straight through and the copies dropped. Shared code (Qwen2Moe, Mixtral and DeepSeek v1 also use it, Bailing by far the heaviest), so its own PR.

**LOW - per-decode-step router allocations.** `BailingMoeGate::forward` materializes two one-element MLX arrays per layer per token (the `1e-20` normalization epsilon and the routed-scaling factor) plus a full `copy(&orig_scores)` when there is no expert bias. Identical to the DeepSeek-V3 gate, so this is the tree's established shape rather than a regression. The one cheap win specific to Bailing is that `routed_scaling_factor` is 1.0 on every published checkpoint and validated finite, so skipping the multiply at exactly 1.0 is bit-exact and removes 28 scalar allocations and 28 elementwise multiplies per token.

### Verification

- `cargo check --release --lib --tests --features metal,accelerate`: clean apart from the two pre-existing `host_preprocessor` warnings.
- `cargo clippy --release --lib --tests --features metal,accelerate`: same two, no new lints.
- `cargo test --release --lib --features metal,accelerate models::bailing_moe`: 42 passed, 0 failed. No SIGABRT on any run.
- `models/ling-lite-1.5` was not loaded, per the review budget.

PR #950, opened 2026-07-28
fix: bounds-check topk_group inside group_mask_scores

### LINKED ISSUE #947: fix: bounds-check topk_group inside group_mask_scores
## Problem / Background

`group_mask_scores` (`src/models/switch_layers.rs:681`) is the shared grouped-expert routing mask behind every DeepSeek-V3-style MoE gate. It takes `n_group` and `topk_group` as plain `i32` and never checks either one against the other or against the score row:

```rust
let k = n_group - topk_group;
let group_idx = mlxcel_core::argpartition(&group_scores, k - 1, -2); // [n, n_group, 1]
let group_idx = slice_axis(&group_idx, -2, 0, k);                    // [n, k, 1]
let grouped = mlxcel_core::put_along_axis(&grouped, &group_idx, &zero, -2);
```

`slice_axis` (`src/lib/mlxcel-core/src/utils.rs:43`) implements Python slice semantics: `end == -1` means "to the end of the axis", and any other negative `end` is taken relative to the axis length. A negative `k` therefore does not raise, it selects a different set of groups.

### The silent case: `topk_group == n_group + 1`

`k` is `-1`. `argpartition(kth = k - 1 = -2, axis = -2)` is fine because MLX normalizes a negative `kth` (`mlx/ops.cpp:2757` at the pinned commit `b7c3dd6d`, `int kth_ = kth < 0 ? kth + a.shape(axis) : kth;`), giving `n_group - 2`. Then `slice_axis(group_idx, -2, 0, -1)` selects the **whole** group axis, so `group_idx` is a full permutation of the group indices and `put_along_axis` writes zero into every group of the row.

What the model emits is not zero output. Every caller masks a *selection* copy of the scores and gathers the combine weights from the untouched `orig_scores`, so the damage lands entirely on selection: with an all-zero (hence all-equal) selection row, `argpartition` returns a tie-break order that has nothing to do with the router, so every token routes to the same arbitrary expert set, and the combine weights are the genuine unbiased sigmoid/softmax scores at those arbitrary experts, renormalized and scaled as usual. The output is finite, in range, and plausible. The router has simply been switched off. Nothing throws and nothing logs.

### The neighboring regimes

- `n_group + 2 <= topk_group <= 2 * n_group - 1`: `k <= -2`, `slice_axis` resolves `end` to `n_group + k`, so the wrong *number* of groups is zeroed. Also silent, also wrong routing.
- `topk_group >= 2 * n_group`: the normalized `kth` goes below zero and MLX throws `std::invalid_argument`. `fn argpartition` is declared non-`Result` in the bridge (`src/lib/mlxcel-core/src/lib.rs:924`), so the exception is an uncatchable abort at the first forward pass rather than a load error.
- `topk_group == 0`: `k == n_group`, so the slice takes the whole group axis and every group is zeroed. Same silent-degenerate-routing shape as the `n_group + 1` case.
- `num_experts % n_group != 0` fails the same way one step earlier: `experts_per_group` is computed by truncating integer division, so `reshape(scores, [n, n_group, n_experts / n_group])` throws on the size mismatch and aborts.

### Caller inventory

Eight production call sites, not six. Only Bailing MoE validates:

| Call site | Passes | Validation |
|---|---|---|
| `src/models/bailing_moe.rs:1491` | `self.topk_group` | `validate_routing`, at load |
| `src/models/deepseek_v3.rs:542` | `self.topk_group` (stored as `args.topk_group as i32` at `:592`) | none |
| `src/models/deepseek_v32.rs:613` | `self.topk_group as i32` | none |
| `src/models/glm4_moe.rs:778` | `self.topk_group as i32` | none |
| `src/models/glm4_moe_lite.rs:640` | `self.topk_group` (`as i32` at `:720`) | none |
| `src/models/exaone_moe.rs:340` | `self.topk_group as i32` | none |
| `src/models/solar_open.rs:617` | `self.topk_group as i32` | none |
| `src/models/dots1.rs:379` | `self.topk_group as i32` | partial: the `:378` guard skips the mask when `n_group == topk_group`, no upper bound |

All eight already gate the call on `self.n_group > 1`, so the `n_group <= 1` regime is covered by every current caller. That is a call-site convention rather than a property of the function, which is the same shape of exposure the rest of this issue describes: it holds until a ninth caller forgets it.

`validate_routing` (`src/models/bailing_moe.rs:583`) is the model to generalize from. On the grouped branch it checks `num_experts % n_group == 0` (the reshape precondition), `experts_per_group >= 2` (group scoring sums each group's top two, so a one-expert group is meaningless), and `1 <= topk_group <= n_group - 1`, and it runs at load so a bad config is a load error with a message rather than wrong tokens.

The function's own `// Used by:` comment lists six callers ("BailingMoe, DeepSeekV3, DeepSeekV32, GLM4Moe, GLM4MoeLite, ExaOneMoe") and is missing SolarOpen and Dots1. `docs/code-guidelines.md` makes that comment the discovery mechanism for "what breaks if I change this", so the stale list is part of the problem here and is very likely why the caller count was under-counted in the first place.

Both the function and the seven unvalidated call sites predate #946 / #838. The Bailing MoE work is only where this surfaced; nothing in that PR introduced it.

## Proposed Solution

Put the bounds check inside `group_mask_scores` so all eight call sites are covered at once, instead of copying `validate_routing`'s `topk_group` clause into seven more model files (and having to remember it for the ninth caller).

The function has an unmasked identity available, so it can be made total: when the mask is not well defined, return the scores unmasked rather than computing a negative `k`. That covers every case above:

- `n_group <= 1`, or `topk_group >= n_group`: mask nothing. This is what the math already means (you cannot keep more groups than exist) and it matches the `n_group != topk_group` guard `dots1.rs:378` already carries.
- `topk_group == 0`: today `k == n_group`, which zeroes every group. Same silent-degenerate-routing shape, so it belongs in the same guard.
- `n_experts % n_group != 0` or `n_experts / n_group < 2`: mask nothing instead of letting `reshape` or `argpartition` throw across a non-`Result` bridge and abort the process.

So that an out-of-range config is not merely silently ignored, pair the guard with a one-time diagnostic (a warning naming the model, `n_group`, and `topk_group`) on the skip path. Model families that want a hard failure can keep or add a load-time check like `validate_routing`; the in-function guard is the floor that stops wrong output and aborts, not a replacement for load-time validation. Note that `tracing::warn!` is a no-op in the `mlxcel` CLI binary (see the implementation context below), so the diagnostic has to be `eprintln!` to be visible on the CLI path.

## Acceptance Criteria

- [x] `group_mask_scores` validates `n_group`, `topk_group`, and the score row width before computing `k`, and never reaches `slice_axis` or `argpartition` with a negative `k` or a negative normalized `kth`.
- [x] An out-of-range `topk_group` can no longer produce a mask that zeroes every group, and can no longer reach an MLX exception through the non-`Result` bridge.
- [x] Regression test that fails on `main`: call `group_mask_scores` directly with `n_group = 4`, `topk_group = 5` and assert the returned row still carries the input scores. On `main` that call returns an all-zero row.
- [x] Regression test for the abort regime: `n_group = 4`, `topk_group = 8` returns a value instead of terminating the process.
- [x] Regression test for `topk_group == 0`.
- [x] In-range behavior is bit-identical: for every valid `(n_group, topk_group)` pair the masked output matches `main` exactly, including the existing `group_mask_scores_keeps_the_groups_with_the_largest_top_two_sum` and the `n_group = 4, topk_group = 3` case in `src/models/bailing_moe_tests.rs`.
- [x] Existing Bailing MoE behavior is unchanged: `validate_routing` still rejects the same configs at load with the same messages, and `cargo test --release --lib models::bailing_moe` passes with no test modified.
- [ ] Real-checkpoint smoke run on at least one MoE family whose gate calls this function (for example `models/ling-lite-1.5` or `models/solar-open-100b-4bit`), token-exact against `main`. Deferred to the orchestrator: the only local checkpoint in these families is `models/ling-lite-1.5` at 33.6 GB, outside this unit's budget. Every published checkpoint in the eight families is in range, so the guard is a no-op for them by construction.
- [x] The `// Used by:` comment above `group_mask_scores` lists all eight callers.

## Technical Considerations

**Upstream divergence, offered as context and not as a required change.** Bailing MoE rejects `topk_group == n_group`, and its error text says `argpartition(kth = n_group - topk_group - 1)` "is out of range as soon as topk_group reaches n_group". That reasoning does not hold against the pinned MLX: `mlx/ops.cpp:2757` normalizes a negative `kth`, so `kth = -1` becomes `n_group - 1`, which is in range, and `k == 0` then makes the slice empty so no group is masked. Upstream mlx-lm accepts `topk_group == n_group` for exactly that reason. If the guard proposed above skips the mask for `topk_group >= n_group`, relaxing Bailing's check from `>= n_group` to `> n_group` would be bit-exact, because `topk_group == n_group` masks zero groups either way. Whoever picks this up should decide whether to relax it here or leave it to a separate change; either way the error message's stated rationale needs correcting.

**Hot path.** The guard runs once per MoE layer per forward, on scalars already held in the gate struct, so it is arithmetic on `i32`s with no array work and no measurable cost. Do not add an `eval` or a device round-trip to implement it.

**Scalar casts.** Several callers reach the function through `args.topk_group as i32` from a `usize` config field (`deepseek_v3.rs:592`, `glm4_moe_lite.rs:720`, and the inline casts in the table above). A `usize` above `i32::MAX` wraps to a negative `i32` there, before `group_mask_scores` ever sees it. That makes `k` larger than `n_group` rather than negative, which zeroes every group via the `end.min(dim_size)` clamp and then hands `argpartition` a `kth` at or above `n_group`, so it lands in the abort regime. `bailing_moe.rs:1568` already saturates with `i32::try_from(args.topk_group).unwrap_or(i32::MAX)`; the in-function guard should be written so a negative `topk_group` also reaches the mask-nothing path.

<details>
<summary><b>Implementation context (machine facts, validation method, and a recurring defect class)</b></summary>

**Most relevant to this issue:** the recurring defect class, and the note that `// Used by:` comments go stale. One of the unguarded regimes here reaches `argpartition` with a negative normalized `kth`, which is exactly the "config value violates an undocumented MLX precondition and aborts uncatchably at the first forward pass" pattern described below. And the stale `// Used by:` comment listing six callers where there are eight is the same one referenced below; it is why this issue's own caller count was initially under-counted.

Collected while porting five text-model families in #924, #926, #928, #930, and #946. This
is background for whoever picks this up, not additional scope.

### Build and test on the current development machine (Apple M1 Ultra, macOS)

- Every cargo invocation needs `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer`, or cmake fails with `xcrun: error: unable to find utility "metal"`. `xcode-select` still points at CommandLineTools.
- Only the RELEASE profile is warm. A debug `cargo check` triggers a COLD MLX C++ build that takes many minutes. Use `DEVELOPER_DIR=... cargo check --release --lib --tests --features metal,accelerate`.
- Narrow test selectors only: `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate models::<module>`. A bare `cargo test`, `cargo test --lib`, or `cargo clippy --workspace --all-targets` runs long enough to stall an agent with a stream-idle watchdog, and a cold `cargo build --release` can exceed it on the MLX C++ link step alone.
- `cargo fmt --all -- --check`, never a bare-path fmt check, which produces spurious edition-2024 diffs.
- CI runs no `cargo test` at all (only `changes`, `cargo-deny`, `cargo-fmt`, and the cross-repo-ref guard). Local runs plus a real checkpoint are the only gate. Two deterministically failing tests were found sitting on `main` during this series for exactly that reason: see #939, and the `FAMILY_ORDER` entries fixed in #946.
- `tracing::warn!` is a NO-OP in the `mlxcel` CLI binary; only `src/server/startup.rs` installs a subscriber. Use `eprintln!` for any CLI-facing diagnostic.
- A test that trips an MLX C++ throw aborts the whole test binary with SIGABRT rather than failing cleanly. That is expected, and the abort is stronger evidence than a clean assertion failure. One SIGABRT on the first run after a fresh link that does not reproduce on reruns is a cold Metal-initialization race, observed independently by two reviewers.

### Validating model behavior: a token-exact reference oracle

Shape tests do not catch the failures that matter here. A wrong prefill/decode offset, a flat instead of interleaved QKV split, the wrong RoPE convention, or a routing bias applied to the wrong copy all produce correctly-shaped tensors and fluent, plausible output. Only the token id sequence separates them.

What worked: create a scratch venv, `pip install mlx-lm`, and drive `mlx_lm.generate.stream_generate` with `make_sampler(temp=0.0)`, printing both text and token ids. Then compare against `mlxcel generate`. **Pass `--no-chat-template` to mlxcel even when the checkpoint ships a chat template**, so both sides see the same raw prompt, and run the templated path separately as a usability check.

Caveat: the oracle is blind wherever mlxcel and mlx-lm make the same choice. GELU was one such case, closed by monkey-patching the reference to the other variant and regenerating. Note `mlxcel_core::utils::gelu_approx` is erf-based despite its name, while MLX Python's `nn.gelu_approx` really is the tanh form.

### A recurring defect class worth checking for

A value from `config.json` passes every Rust-side check, violates an undocumented precondition of an MLX C++ entry point, and kills the process at the FIRST FORWARD PASS rather than failing at load. Most cxx bridge functions are declared returning `UniquePtr<MlxArray>` rather than `Result`, so the C++ throw is an uncatchable `std::terminate`, not something `catch_unwind` contains. The model loads cleanly and the server dies on its first request. This class produced one CRITICAL and five HIGH findings across the five ports.

Specifics established by reading the pinned MLX checkout rather than assuming:

- Gathers do NOT range-check positive indices. `take` wraps negative indices but an out-of-range positive index silently returns values belonging to no row, and that reaches the logits with no fault. Bound every gather by the real tensor shape, never by a config field.
- `slice` CLAMPS an out-of-range stop instead of throwing, so a too-wide split silently loses trailing channels.
- `fast_rope` requires `dims` even, positive, and no larger than the last axis.
- `rms_norm` and `layer_norm` never inspect `eps`. A NaN or negative eps yields NaN hidden states with no error at all, which is harder to diagnose than a crash.
- `matmul` throws on an inner-dimension mismatch, which is why an unchecked INPUT axis is fatal even where the row axis bounds only an argmax.
- `quantized_matmul` divides by `bits`, so `"bits": 0` is a division by zero. On AArch64 the divide returns 0 and `std::invalid_argument` fires; on x86-64 the hardware raises `SIGFPE` and kills the process before any exception exists.
- Reconstruct a quantized input width the way MLX does, as `scales.shape(-1) * group_size`, and check `.biases` shape equals `.scales` shape. A self-consistency check cannot catch a checkpoint honestly packed for the wrong `hidden_size`, because such a checkpoint IS self-consistent.
- Put zero checks BEFORE divisibility checks, because `0.is_multiple_of(0)` is true.
- Avoid unbounded probe loops over `0..n_layer` in the load path: a huge `n_layer` hangs with a flat allocation footprint, so no OOM kill rescues it.

`src/models/gpt_neox.rs` and `src/models/helium.rs` carry the current `ModelArgs::validate` / `validate_weights` shape to copy from.

### Conventions that bite

- The `// Used by:` comments above shared helpers are the designated discovery mechanism for "what breaks if I change this", and they can themselves be stale. One was found listing six callers where there were eight. Verify by grep rather than trusting the comment, and update it when you touch the helper.
- `mlxcel arch` is the architecture registry. `mlxcel list` lists downloaded models and will not tell you whether a family is supported.

</details>

inureyes:
## Summary

`group_mask_scores` (`src/models/switch_layers.rs`) computed `k = n_group - topk_group` and handed it to `slice_axis` and `argpartition` with no check of `topk_group` against `n_group` or against the score row. Seven of its eight callers pass the value straight from `config.json`. The bounds check now lives inside the function, so all eight call sites are covered at once instead of seven more model files each growing their own copy of it.

## Why the silent regime matters

`topk_group == n_group + 1` gives `k == -1`. `slice_axis` implements Python slice semantics, where `end == -1` means "to the end of the axis", so the bottom-k slice selected the whole group axis and `put_along_axis` zeroed every group. That did not produce zero output. Every caller applies the mask to a selection copy and gathers its combine weights from an untouched `orig_scores`, so an all-zero (hence all-equal) selection row hands routing to `argpartition`'s tie-break order, which has nothing to do with the router, while the combine weights remain the genuine unbiased scores at those arbitrary experts. Every token routes to the same arbitrary expert set, the output stays finite and in range, and nothing throws or logs. The router is simply off.

`n_group + 2 <= topk_group <= 2 * n_group - 1` resolves `end` to `n_group + k` and zeroes the wrong *number* of groups, equally silently, and `topk_group == 0` gives `k == n_group` and zeroes every group.

The remaining regimes abort instead. `topk_group >= 2 * n_group` and any negative `topk_group` drive `argpartition`'s normalized `kth` out of range; `num_experts % n_group != 0` and a group holding one expert fail one step earlier, in the reshape and in the top-2 `argpartition` respectively. `mlxcel_core::argpartition` is declared non-`Result` in the bridge, so an MLX C++ throw is an uncatchable `std::terminate` at the first forward pass rather than a load error: the model loads cleanly and the server dies on its first request. A negative `topk_group` is reachable without a negative config value, because several callers cast a `usize` config field through `args.topk_group as i32`, and a `usize` above `i32::MAX` wraps there before `group_mask_scores` ever sees it.

## What changed

- `src/models/switch_layers.rs`: new `group_mask_plan(shape, n_group, topk_group) -> Result<i32, GroupMaskSkip>` classifies the triple before any array work and returns `experts_per_group` when the mask is well defined. `group_mask_scores` returns the scores unmasked on every skip reason. The guard is pure `i32` arithmetic on values already held in the gate struct: no array work, no device round-trip, no `eval`, so the hot path is unaffected.
- `GroupMaskSkip` separates the two ordinary reasons to mask nothing (`n_group <= 1`, and `topk_group == n_group`, which keeps every group and is what upstream mlx-lm means by it) from the five broken-config reasons (`InvalidGroupCount`, `TopkGroupOutOfRange`, `UnevenGroups`, `GroupTooSmall`, `BadScoreShape`). Only the broken ones are reported.
- `report_group_mask_skipped` writes one `eprintln!` line per process naming the reason, `n_group`, `topk_group`, and the score-row shape. `tracing::warn!` is a no-op in the `mlxcel` CLI binary, where nothing installs a subscriber, so a `tracing` call would have been invisible on the CLI path. Every MoE layer of a model shares one `(n_group, topk_group)` pair, so one line is the right granularity; firing per layer per token would bury it. Verified: eight distinct bad-config calls in one process emit exactly one line.
- Check ordering puts `n_group <= 0` ahead of the `n_experts / n_group` division so a zero group count cannot divide by zero, and puts the group-width check after divisibility so an empty score row (which passes `0 % n_group == 0`) is still caught.
- The `// Used by:` comment listed six callers. A grep finds eight; it now names `Dots1` and `SolarOpen` as well, per `docs/code-guidelines.md`. The upstream reference was also converted from a bare `deepseek_v3.py` mention to the web-accessible GitHub URL form the repository guidelines require.
- Eleven new tests in the `models::switch_layers` module, including one pure test of `group_mask_plan` that exercises the whole classification table without touching MLX.

## Caller inventory, verified by grep

Eight production call sites, all of which currently gate the call on `self.n_group > 1`. The guard covers that regime too, so it no longer depends on every future caller repeating the convention.

| Call site | Passes | Load-time validation |
|---|---|---|
| `src/models/bailing_moe.rs:1491` | `self.topk_group` | `validate_routing` |
| `src/models/deepseek_v3.rs:542` | `self.topk_group` | none |
| `src/models/deepseek_v32.rs:613` | `self.topk_group as i32` | none |
| `src/models/glm4_moe.rs:778` | `self.topk_group as i32` | none |
| `src/models/glm4_moe_lite.rs:640` | `self.topk_group` | none |
| `src/models/exaone_moe.rs:340` | `self.topk_group as i32` | none |
| `src/models/solar_open.rs:617` | `self.topk_group as i32` | none |
| `src/models/dots1.rs:379` | `self.topk_group as i32` | partial, the call-site guard skips the mask when `n_group == topk_group` |

## Deliberately not changed

Bailing MoE's `validate_routing` still rejects `topk_group == n_group` where upstream mlx-lm accepts it, and its error text still explains that with an `argpartition` range claim that does not hold against the pinned MLX (a negative `kth` is normalized, so `kth == -1` becomes `n_group - 1`). The issue frames relaxing this as optional context, and the acceptance criteria require `validate_routing` to keep rejecting the same configs with the same messages, so it is untouched here. Correcting that rationale, and deciding whether to relax the bound, belongs in its own change.

`src/models/llada2_moe/mod.rs:480` carries a private near-duplicate, `group_limited_mask`, which fills non-kept groups with a large negative constant rather than zero and has the same unguarded `k = n_group - topk_group` arithmetic. It is not a caller of `group_mask_scores`, so it is out of scope for this issue, but it is the same defect and should get its own issue.

## Test plan

- [x] `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate models::switch_layers` passes 19/19.
- [x] Each new regression test was run individually against the unmodified function first. The silent regimes fail on an assertion, the loud ones kill the binary with SIGABRT: `topk_group = 5` returned `[0, 0, ... 0]` instead of the input row; `topk_group = 6` returned a row with the wrong two groups zeroed; `topk_group = 0` returned all zeros; `topk_group = 8` aborted with `[argpartition] Received invalid kth -5 along axis -2 for array with shape: (1,4,1)`; `topk_group = -1` aborted with `[argpartition] Received invalid kth 4`; `n_group = 3` over 16 experts aborted with `[reshape] Cannot reshape array of size 16 into shape (1,3,5)`; `n_group = 16` over 16 experts aborted with `[argpartition] Received invalid kth 1 along axis -1`; and `n_group = 0` panicked on integer division by zero.
- [x] In-range behavior is bit-identical. The `n_group = 4, topk_group = 2` and `topk_group = 3` cases, the `topk_group == n_group` identity, and a second `n_group = 2, topk_group = 1` case all produce the same rows before and after the guard.
- [x] `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate models::bailing_moe` passes 42/42 with no test modified.
- [x] The other caller families pass together, 61/61: `models::deepseek_v3`, `models::deepseek_v32`, `models::glm4_moe`, `models::exaone_moe`, `models::solar_open`, `models::dots1`, `models::llada2_moe`.
- [x] `DEVELOPER_DIR=... cargo clippy --release --lib --tests --features metal,accelerate` exits 0 with nothing new; the only two warnings are the pre-existing `src/multimodal/host_preprocessor.rs` and `host_preprocessor_export.rs` ones already on `main`.
- [ ] Real-checkpoint smoke run on an MoE family whose gate calls this function, token-exact against `main`. Not run here: the available local checkpoint is `models/ling-lite-1.5` at 33.6 GB, which is outside this unit's budget. Every published checkpoint in the eight families is in range, so the guard is a no-op for them by construction, and the in-range bit-identity tests above cover the path they take.

Closes #947

### DIFF
--- a/src/models/switch_layers.rs
+++ b/src/models/switch_layers.rs
@@ -21,6 +21,8 @@
 //! SwitchGLU: SwiGLU MLP routing through SwitchLinear
 //! group_mask_scores: group-based expert masking for MoE gates with n_group > 1
 
+use std::sync::atomic::{AtomicBool, Ordering};
+
 use mlxcel_core::utils::slice_axis;
 use mlxcel_core::weights::WeightMap;
 use mlxcel_core::{MlxArray, UniquePtr, dtype};
@@ -670,19 +672,163 @@ pub fn moe_weighted_sum(
     }
 }
 
+/// Why [`group_mask_scores`] cannot apply a grouped-routing mask to a given
+/// `(score row, n_group, topk_group)` triple.
+///
+/// The first two variants are ordinary configurations that happen to mask
+/// nothing. The rest are out-of-range values; [`group_mask_scores`] documents
+/// what each one used to do once it reached MLX.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum GroupMaskSkip {
+    /// `n_group <= 1`: the model declares no expert groups.
+    NotGrouped,
+    /// `topk_group == n_group`: every group is kept, so the mask is the
+    /// identity. Upstream mlx-lm accepts this and it is what `k == 0` already
+    /// produced through an empty bottom-k slice.
+    KeepsEveryGroup,
+    /// `n_group` is zero or negative.
+    InvalidGroupCount,
+    /// `topk_group` is outside `1..=n_group`.
+    TopkGroupOutOfRange,
+    /// The score row does not split into `n_group` equal groups.
+    UnevenGroups,
+    /// A group holds fewer than two experts.
+    GroupTooSmall,
+    /// The router scores are not a rank-2 `[tokens, n_experts]` array.
+    BadScoreShape,
+}
+
+impl GroupMaskSkip {
+    /// Whether this is a broken config worth reporting, as opposed to a
+    /// configuration that legitimately masks nothing.
+    fn is_config_error(self) -> bool {
+        !matches!(self, Self::NotGrouped | Self::KeepsEveryGroup)
+    }
+
+    fn reason(self) -> &'static str {
+        match self {
+            Self::NotGrouped => "the model declares no expert groups",
+            Self::KeepsEveryGroup => "topk_group keeps every group",
+            Self::InvalidGroupCount => "n_group must be at least 1",
+            Self::TopkGroupOutOfRange => "topk_group must be between 1 and n_group",
+            Self::UnevenGroups => "num_experts is not divisible by n_group",
+            Self::GroupTooSmall => {
+                "every group must hold at least 2 experts, because a group is scored by the sum \
+                 of its top two"
+            }
+            Self::BadScoreShape => "the router scores are not a rank-2 [tokens, num_experts] array",
+        }
+    }
+}
+
+/// Bounds check for [`group_mask_scores`], returning `experts_per_group` when
+/// the grouped-routing mask is well defined for `shape`.
+///
+/// Pure `i32` arithmetic over values already held in the gate struct: no array
+/// work, no device round-trip, no `eval`. It runs once per MoE layer per
+/// forward and costs nothing measurable.
+fn group_mask_plan(shape: &[i32], n_group: i32, topk_group: i32) -> Result<i32, GroupMaskSkip> {
+    if shape.len() != 2 {
+        return Err(GroupMaskSkip::BadScoreShape);
+    }
+    let n_experts = shape[1];
+    // Before the division below, and so `n_group == 0` cannot divide by zero.
+    if n_group <= 0 {
+        return Err(GroupMaskSkip::InvalidGroupCount);
+    }
+    if n_group == 1 {
+        return Err(GroupMaskSkip::NotGrouped);
+    }
+    if topk_group == n_group {
+        return Err(GroupMaskSkip::KeepsEveryGroup);
+    }
+    // Catches both `topk_group == 0` and the negative value a caller's
+    // `args.topk_group as i32` produces from a `usize` above `i32::MAX`.
+    if !(1..=n_group).contains(&topk_group) {
+        return Err(GroupMaskSkip::TopkGroupOutOfRange);
+    }
+    if n_experts % n_group != 0 {
+        return Err(GroupMaskSkip::UnevenGroups);
+    }
+    // Also rejects an empty score row: `0 % n_group == 0` passes the
+    // divisibility test above and only the group width catches it.
+    let experts_per_group = n_experts / n_group;
+    if experts_per_group < 2 {
+        return Err(GroupMaskSkip::GroupTooSmall);
+    }
+    Ok(experts_per_group)
+}
+
+/// Report an out-of-range grouped-routing config once per process.
+///
+/// `tracing::warn!` is a no-op in the `mlxcel` CLI binary, where nothing
+/// installs a subscriber, so this has to reach stderr directly. Every MoE layer
+/// of a model shares one `(n_group, topk_group)` pair, so one line is the right
+/// granularity: firing per layer per token would bury it.
+fn report_group_mask_skipped(skip: GroupMaskSkip, shape: &[i32], n_group: i32, topk_group: i32) {
+    static REPORTED: AtomicBool = AtomicBool::new(false);
+    if REPORTED.swap(true, Ordering::Relaxed) {
+        return;
+    }
+    eprintln!(
+        "warning: grouped MoE expert routing is disabled because {}. config n_group={n_group}, \
+         topk_group={topk_group}; router scores {shape:?}. Every token now routes as if the \
+         model declared no expert groups, which changes which experts it selects. Correct \
+         n_group / topk_group in the model's config.json.",
+        skip.reason()
+    );
+}
+
 /// Group-based expert masking for MoE gates with n_group > 1.
 ///
 /// Selects the top `topk_group` expert groups (by sum of top-2 scores per group)
 /// and zeros out scores for experts in non-selected groups.
 ///
-/// Used by: BailingMoe, DeepSeekV3, DeepSeekV32, GLM4Moe, GLM4MoeLite, ExaOneMoe
+/// Used by: BailingMoe, DeepSeekV3, DeepSeekV32, Dots1, ExaOneMoe, GLM4Moe,
+///          GLM4MoeLite, SolarOpen
 ///
-/// Reference: mlx-lm deepseek_v3.py group_expert_select()
+/// # Bounds
+///
+/// `n_group` and `topk_group` arrive straight from `config.json` on seven of the
+/// eight call sites, and this function is the floor that keeps a bad pair from
+/// reaching MLX. When the mask is not well defined it returns the scores
+/// unmasked and reports the config once, rather than computing a negative `k`:
+///
+/// - `topk_group == n_group + 1` gives `k == -1`, and [`slice_axis`] reads that
+///   `end` as "to the end of the axis", so the bottom-k slice became the whole
+///   group axis and every group was zeroed. Nothing threw. Callers gather their
+///   combine weights from an untouched copy of the scores, so an all-equal
+///   selection row silently handed routing to `argpartition`'s tie-break order
+///   while the weights stayed plausible: the router was off, the output was not.
+/// - `n_group + 2 <= topk_group <= 2 * n_group - 1` resolved `end` to
+///   `n_group + k` and zeroed the wrong *number* of groups. Also silent.
+/// - `topk_group >= 2 * n_group`, and any negative `topk_group`, drove
+///   `argpartition`'s `kth` out of range. `mlxcel_core::argpartition` is
+///   declared non-`Result`, so the MLX throw was an uncatchable abort at the
+///   first forward pass rather than a load error.
+/// - `topk_group == 0` gives `k == n_group`, zeroing every group.
+/// - `n_experts % n_group != 0` truncated `experts_per_group` and threw in the
+///   reshape; fewer than two experts per group threw in the top-2
+///   `argpartition`, since a group is scored by the sum of its top two.
+///
+/// This is not a substitute for load-time validation. A family that wants a bad
+/// config to be a load error keeps its own check, as Bailing MoE does in
+/// `ModelArgs::validate_routing`.
+///
+/// Reference: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/deepseek_v3.py (group_expert_select)
 pub fn group_mask_scores(scores: &MlxArray, n_group: i32, topk_group: i32) -> UniquePtr<MlxArray> {
     let shape = mlxcel_core::array_shape(scores);
+    let experts_per_group = match group_mask_plan(&shape, n_group, topk_group) {
+        Ok(experts_per_group) => experts_per_group,
+        Err(skip) => {
+            if skip.is_config_error() {
+                report_group_mask_skipped(skip, &shape, n_group, topk_group);
+            }
+            return mlxcel_core::copy(scores);
+        }
+    };
     let n = shape[0];
     let n_experts = shape[1];
-    let experts_per_group = n_experts / n_group;
 
     // Unflatten: [n, n_experts] -> [n, n_group, experts_per_group]
     let grouped = mlxcel_core::reshape(scores, &[n, n_group, experts_per_group]);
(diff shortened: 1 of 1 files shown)


inureyes:
The red `cargo-fmt` check is a pre-existing CI breakage, not a formatting problem in this branch. The job fails at the toolchain-install step, before `cargo fmt` ever runs:

```
info: syncing channel updates for 1.100.0-x86_64-unknown-linux-gnu
error: could not download nonexistent rust version `1.100.0-x86_64-unknown-linux-gnu`: ... 404
```

`bdad23287` bumped `dtolnay/rust-toolchain` from `1.93.1` to `1.100.0`. That action's tag is the Rust toolchain version, and Rust 1.100.0 does not exist, so the pin 404s. The same job failed on `main` in runs 30251560172 and 30253948313; the later green `main` run only passed because that commit touched no Rust and the `changes` filter skipped the job. `cargo-deny`, `Detect changes` and `cross-repo refs` all pass here.

Formatting was verified locally with `cargo fmt --all -- --check` (rustfmt 1.8.0-stable, Rust 1.93.1), which is clean. The CI pin needs its own fix.

PR #955, opened 2026-07-28
fix(lora): make adapter weight fusion Conv1D-layout aware

### LINKED ISSUE #925: fix(lora): make adapter weight fusion Conv1D-layout aware
## Problem / Background

`fuse_lora_weights_into` fuses a LoRA delta into a base tensor with a bare element-wise add and no shape check:

```rust
// src/lora/loader.rs:149
let fused = mlxcel_core::add(base_weight, &delta);
```

`compute_lora_delta` always produces a delta in `mlxcel_core` `Linear` layout, `[out_features, in_features]`. Fusion has no notion of how the base tensor is actually stored, and it never validates that the two operands agree. That held up while every supported family stored 2-D projections as `[out, in]`.

GPT-2 (#833, PR #924) breaks that assumption. It is the first family in this tree whose raw HuggingFace exports store `c_attn`, `c_proj`, and `c_fc` in the HF `Conv1D` layout `[in, out]`, transposed relative to what `Linear` expects. `Gpt2Layout::detect` and `Gpt2Layout::conv1d_linear` in `src/models/gpt2.rs` handle this with a load-time transpose, and they do it correctly.

The reachability comes from ordering. In `load_model_with_adapter` (`src/loading/mod.rs`), the raw safetensors map is loaded, `lora::apply_lora_adapters` runs on it, and only afterwards does `load_model_from_weights` construct the model, which is where `Gpt2Layout` performs the transpose. So at the moment fusion runs against a raw HF GPT-2 checkpoint, the base weight is still in `[in, out]` Conv1D layout while the delta is in `[out, in]`. GPT-2 is adapter-enabled: `src/model_metadata.rs` registers `Gpt2` with `adapter: None`, meaning there is no adapter-unsupported message, so nothing gates the path. The pipeline-parallel entry point `apply_stage_lora_adapter` funnels through the same `fuse_lora_weights_into` primitive and has the same exposure.

### Concrete outcomes on a raw HF GPT-2 checkpoint

For `n_embd = H` (H = 768 on GPT-2 small):

| Base weight | Raw HF (Conv1D) shape | Delta shape | Result of `add` |
|---|---|---|---|
| `h.N.attn.c_attn.weight` | `[H, 3H]` = `[768, 2304]` | `[3H, H]` = `[2304, 768]` | broadcast failure |
| `h.N.mlp.c_fc.weight` | `[H, 4H]` = `[768, 3072]` | `[4H, H]` = `[3072, 768]` | broadcast failure |
| `h.N.mlp.c_proj.weight` | `[4H, H]` = `[3072, 768]` | `[H, 4H]` = `[768, 3072]` | broadcast failure |
| `h.N.attn.c_proj.weight` | `[H, H]` = `[768, 768]` | `[H, H]` = `[768, 768]` | **succeeds, silently wrong** |

The square attention `c_proj` is the case that motivates this issue. Writing the true Linear weight as `W` (`[out, in]`), the checkpoint holds `Wᵀ`, so fusion yields `Wᵀ + D`. `conv1d_linear` then transposes at construction time, and the model ends up running `W + Dᵀ` instead of `W + D`. `Dᵀ` is a different rank-r update of the same shape, so there is no error, no warning, and no shape anomaly. The adapter appears to load and the model produces plausible but incorrect output.

The three non-square projections fail hard, but not cleanly. `add` is declared in the cxx bridge as `fn add(a: &MlxArray, b: &MlxArray) -> UniquePtr<MlxArray>` with no `Result`, and the C++ side (`src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:430`) calls `mlx::core::add` with no try/catch. The MLX broadcast error is a C++ exception escaping a cxx shim that is not declared fallible, so it terminates the process rather than surfacing as an `anyhow::Error` the caller can report.

### Why this is narrow in practice today

Reaching the bug requires an adapter whose keys match the bare GPT-2 projection names against a raw (untransposed) checkpoint. The two adapter formats in circulation do not:

- **mlx-lm adapters** carry the `model.` prefix, matching the MLX-converted GPT-2 layout. A `model.`-prefixed GPT-2 checkpoint in circulation is an MLX conversion and is already stored `[out, in]`, so those adapters resolve only against weights that need no transpose. Note this is a property of how the published conversions are packed, not something the loader enforces: `Gpt2Layout::detect` decides the layout purely from the shape of `h.0.attn.c_attn.weight` and never consults the prefix, so a `model.`-prefixed checkpoint that was still in Conv1D layout would be detected as Conv1D and would be exposed exactly like the bare form.
- **HuggingFace PEFT adapters** name their tensors `lora_A.weight` / `lora_B.weight`. The key scan in `fuse_lora_weights_into` matches on `.lora_a` / `.lora_b` suffixes, so PEFT keys never match at all and fusion is a no-op for them.

So the trigger is an adapter using mlxcel's own `.lora_a` / `.lora_b` naming on bare `h.N.attn.c_proj`-style keys. That is a real and producible artifact (it is exactly what `apply_stage_lora_adapter` and the local fine-tuning conventions describe), but it is not what today's published adapters look like. The defect is latent rather than actively breaking users, which is why the priority is low. It is worth fixing because the failure mode on the square projection is silent numerical corruption, and it becomes progressively more likely as GPT-2 adapter tooling appears.

### Re-verified on current `main` (`a3c823611`)

Every claim above was re-checked after #924, #926, #928, #930 and #946 merged. Nothing moved:

- `src/lora/loader.rs:149` is still the unguarded `mlxcel_core::add(base_weight, &delta)`. `compute_lora_delta` still normalizes both adapter conventions to `[out, in]` (transposing the `a @ b` product on the mlx-lm branch, computing `b @ a` on the PEFT branch), and `find_base_weight_name` still falls through to `{name}.weight` when no candidate matches.
- `apply_stage_lora_adapter` (`src/lora/loader.rs:298`) still calls the same primitive, at `src/lora/loader.rs:334`, so the pipeline-parallel path has identical exposure and no guard of its own.
- `add` is still declared without `Result` at `src/lib/mlxcel-core/src/lib.rs:219`, and the shim at `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:430` is still a bare `mlx::core::add` with no try/catch. Only 7 declarations in the whole bridge return `Result`, and `add` is not one of them.
- `src/model_metadata.rs:204` still registers `Gpt2` with `adapter: None`, so the `adapter_unsupported_message` gate in `load_model_from_weights` (`src/loading/mod.rs:656`) has nothing to reject on.
- GPT-2 is still the only Conv1D-layout family in the tree. `src/models/gpt_bigcode.rs:40-48` documents that GPT-BigCode builds its projections with `nn.Linear`, and its `load_linear` rejects a `[in, out]` weight by name with a message that points at the GPT-2 orientation. GPT-NeoX, Helium and Bailing MoE all store `[out, in]`. The guard is still worth having for the next family that lands transposed, but no family other than GPT-2 is exposed today.

## Proposed Solution

Make weight fusion layout-aware, and make it fail loudly rather than guess.

1. **Shape-check before `add`.** In `fuse_lora_weights_into`, compare `array_shape(base_weight)` against `array_shape(&delta)` before the add. On a mismatch, return an `anyhow::Error` naming the base weight key and both shapes. This alone converts three of the four GPT-2 cases from a process abort into a diagnosable error.
2. **Reject, do not auto-transpose.** If the base shape is the reverse of the delta shape, that is ambiguous evidence, not a repair instruction: for a square weight the two are indistinguishable, so silently transposing would fix `c_fc` while leaving `c_proj` just as wrong. Report the layout mismatch and refuse the fusion.
3. **Give fusion the layout it needs.** Options, in rough order of preference:
   - Normalize the checkpoint layout before fusion, so `load_model_with_adapter` hands `fuse_lora_weights_into` a weight map that is already in `[out, in]`. This is the cleanest fix because it removes the ordering hazard entirely rather than patching around it, and it keeps the fusion primitive family-agnostic.
   - Failing that, thread a per-key layout hint into fusion so it can transpose the delta (not the base) for known Conv1D keys.
4. **Extend the guard to the pipeline-parallel path.** `apply_stage_lora_adapter` shares the primitive, so it inherits the fix, but the stage-local path needs its own coverage since it never goes through `load_model_with_adapter`.

Whatever shape the fix takes, a square Conv1D-layout weight must not be able to accept a transposed delta without an error.

## Acceptance Criteria

- [x] `fuse_lora_weights_into` validates operand shapes before the element-wise add and returns a descriptive `anyhow::Error` (naming the base weight key, the base shape, and the delta shape) instead of letting MLX broadcast rules decide.
- [x] A base/delta shape mismatch never reaches `mlxcel_core::add`, so no LoRA fusion path can abort the process through an uncaught C++ exception.
- [x] A square base weight stored in Conv1D `[in, out]` layout cannot be fused with an `[out, in]` delta without an error. Silent transposed accumulation is eliminated, not merely made unlikely.
- [x] Fusion never silently transposes an operand to make shapes agree.
- [x] A real test fuses an adapter using `.lora_a` / `.lora_b` keys against an actual Conv1D-layout GPT-2 checkpoint (raw HuggingFace `gpt2`, bare `h.N.*` keys, no `model.` prefix) and asserts the outcome. Synthetic shape-only unit tests are acceptable as supplementary coverage but are not sufficient on their own, since the whole defect is that the shapes look fine.
- [x] The test explicitly covers the square `h.N.attn.c_proj.weight` case, which is the one that produces no error today.
- [x] The non-square projections (`c_attn`, `c_fc`, and the MLP `c_proj`) produce the new descriptive error rather than a crash.
- [ ] Whichever fix direction is chosen, the end state is verified end to end: loading a GPT-2 checkpoint with an adapter through `load_model_with_adapter` either produces numerically correct fused weights or a clear error, and generation runs against a real checkpoint to confirm it.
- [x] `apply_stage_lora_adapter` (pipeline-parallel stage-local fusion) is covered by the same guarantee and has its own test.
- [x] Existing LoRA fusion behavior for `[out, in]` families is unchanged, verified against the current adapter tests in `src/distributed/pipeline/partial_loading_adapter_tests.rs`.

## Technical Considerations

- Relevant code: `src/lora/loader.rs` (`fuse_lora_weights_into`, `compute_lora_delta`, `find_base_weight_name`, `apply_stage_lora_adapter`), `src/loading/mod.rs` (`load_model_with_adapter`), `src/models/gpt2.rs` (`Gpt2Layout`), `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp` (`add`).
- `find_base_weight_name` resolves `h.0.attn.c_proj` to `h.0.attn.c_proj.weight` on a raw checkpoint, and falls through to that same candidate even when no key matches. Any shape guard has to sit after resolution, on the tensor that was actually found.
- The two "base weight not found" and "incomplete LoRA pair" paths in `fuse_lora_weights_into` report through `tracing::warn!`, which is a no-op in the `mlxcel` CLI binary because only `src/server/startup.rs` installs a subscriber. Any new diagnostic on the CLI-reachable path has to be a returned error or an `eprintln!` to be visible at all.
- `compute_lora_delta` already infers the adapter convention (mlx-lm versus PEFT) from shapes. Adding a base-layout inference on top of that stacks one heuristic on another, which argues for normalizing the checkpoint layout ahead of fusion instead.
- Quantized GPT-2 checkpoints short-circuit the Conv1D probe in `Gpt2Layout::detect` (a packed weight shape carries no layout signal, and a quantized export can only come from an already-sanitized MLX conversion). Fusion against quantized base weights is a separate concern and is out of scope here.
- This is not GPT-2-specific in principle. Any future family that lands with a transposed on-disk projection layout hits the same primitive, so the guard is worth having independent of how the GPT-2 case is resolved.

<details>
<summary><b>Implementation context (machine facts, validation method, and a recurring defect class)</b></summary>

**Most relevant to this issue:** that MLX `slice` clamps an out-of-range stop instead of throwing (the `c_attn` three-way split rides on that behavior), that a C++ throw crossing the cxx bridge is an uncatchable `std::terminate` rather than a recoverable error (which is why the three non-square projections abort the process instead of returning an error), and the reminder to update the `// Used by:` comment when touching a shared helper, since `fuse_lora_weights_into` carries one.

Collected while porting five text-model families in #924, #926, #928, #930, and #946. This
is background for whoever picks this up, not additional scope.

### Build and test on the current development machine (Apple M1 Ultra, macOS)

- Every cargo invocation needs `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer`, or cmake fails with `xcrun: error: unable to find utility "metal"`. `xcode-select` still points at CommandLineTools.
- Only the RELEASE profile is warm. A debug `cargo check` triggers a COLD MLX C++ build that takes many minutes. Use `DEVELOPER_DIR=... cargo check --release --lib --tests --features metal,accelerate`.
- Narrow test selectors only: `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate models::<module>`. A bare `cargo test`, `cargo test --lib`, or `cargo clippy --workspace --all-targets` runs long enough to stall an agent with a stream-idle watchdog, and a cold `cargo build --release` can exceed it on the MLX C++ link step alone.
- `cargo fmt --all -- --check`, never a bare-path fmt check, which produces spurious edition-2024 diffs.
- CI runs no `cargo test` at all (only `changes`, `cargo-deny`, `cargo-fmt`, and the cross-repo-ref guard). Local runs plus a real checkpoint are the only gate. Two deterministically failing tests were found sitting on `main` during this series for exactly that reason: see #939, and the `FAMILY_ORDER` entries fixed in #946.
- `tracing::warn!` is a NO-OP in the `mlxcel` CLI binary; only `src/server/startup.rs` installs a subscriber. Use `eprintln!` for any CLI-facing diagnostic.
- A test that trips an MLX C++ throw aborts the whole test binary with SIGABRT rather than failing cleanly. That is expected, and the abort is stronger evidence than a clean assertion failure. One SIGABRT on the first run after a fresh link that does not reproduce on reruns is a cold Metal-initialization race, observed independently by two reviewers.

### Validating model behavior: a token-exact reference oracle

Shape tests do not catch the failures that matter here. A wrong prefill/decode offset, a flat instead of interleaved QKV split, the wrong RoPE convention, or a routing bias applied to the wrong copy all produce correctly-shaped tensors and fluent, plausible output. Only the token id sequence separates them.

What worked: create a scratch venv, `pip install mlx-lm`, and drive `mlx_lm.generate.stream_generate` with `make_sampler(temp=0.0)`, printing both text and token ids. Then compare against `mlxcel generate`. **Pass `--no-chat-template` to mlxcel even when the checkpoint ships a chat template**, so both sides see the same raw prompt, and run the templated path separately as a usability check.

Caveat: the oracle is blind wherever mlxcel and mlx-lm make the same choice. GELU was one such case, closed by monkey-patching the reference to the other variant and regenerating. Note `mlxcel_core::utils::gelu_approx` is erf-based despite its name, while MLX Python's `nn.gelu_approx` really is the tanh form.

### A recurring defect class worth checking for

A value from `config.json` passes every Rust-side check, violates an undocumented precondition of an MLX C++ entry point, and kills the process at the FIRST FORWARD PASS rather than failing at load. Most cxx bridge functions are declared returning `UniquePtr<MlxArray>` rather than `Result`, so the C++ throw is an uncatchable `std::terminate`, not something `catch_unwind` contains. The model loads cleanly and the server dies on its first request. This class produced one CRITICAL and five HIGH findings across the five ports.

Specifics established by reading the pinned MLX checkout rather than assuming:

- Gathers do NOT range-check positive indices. `take` wraps negative indices but an out-of-range positive index silently returns values belonging to no row, and that reaches the logits with no fault. Bound every gather by the real tensor shape, never by a config field.
- `slice` CLAMPS an out-of-range stop instead of throwing, so a too-wide split silently loses trailing channels.
- `fast_rope` requires `dims` even, positive, and no larger than the last axis.
- `rms_norm` and `layer_norm` never inspect `eps`. A NaN or negative eps yields NaN hidden states with no error at all, which is harder to diagnose than a crash.
- `matmul` throws on an inner-dimension mismatch, which is why an unchecked INPUT axis is fatal even where the row axis bounds only an argmax.
- `quantized_matmul` divides by `bits`, so `"bits": 0` is a division by zero. On AArch64 the divide returns 0 and `std::invalid_argument` fires; on x86-64 the hardware raises `SIGFPE` and kills the process before any exception exists.
- Reconstruct a quantized input width the way MLX does, as `scales.shape(-1) * group_size`, and check `.biases` shape equals `.scales` shape. A self-consistency check cannot catch a checkpoint honestly packed for the wrong `hidden_size`, because such a checkpoint IS self-consistent.
- Put zero checks BEFORE divisibility checks, because `0.is_multiple_of(0)` is true.
- Avoid unbounded probe loops over `0..n_layer` in the load path: a huge `n_layer` hangs with a flat allocation footprint, so no OOM kill rescues it.

`src/models/gpt_neox.rs` and `src/models/helium.rs` carry the current `ModelArgs::validate` / `validate_weights` shape to copy from.

### Conventions that bite

- The `// Used by:` comments above shared helpers are the designated discovery mechanism for "what breaks if I change this", and they can themselves be stale. One was found listing six callers where there were eight. Verify by grep rather than trusting the comment, and update it when you touch the helper.
- `mlxcel arch` is the architecture registry. `mlxcel list` lists downloaded models and will not tell you whether a family is supported.

</details>

inureyes:
## Summary

LoRA weight fusion added a delta to a base tensor with no shape check and no notion of how the base was stored on disk. Deltas are always `[out_features, in_features]`; raw HuggingFace GPT-2 exports store their projections in the `Conv1D` layout `[in_features, out_features]`, and `Gpt2Layout` only transposes them at model construction, which runs after fusion on the `load_model_with_adapter` path. Fusion is now layout-aware and shape-checked, and it reports rather than guesses.

## The two failure modes this closes

The square attention `c_proj` (`[768, 768]` against a `[768, 768]` delta) broadcast cleanly, so the model silently ran `W + Dᵀ` instead of `W + D`. That is a different rank-r update of the same shape: no error, no warning, no shape anomaly, just wrong numbers. This is the case the issue was filed for, and it is the reason a shape check alone is not a fix.

The three non-square projections (`c_attn`, `c_fc`, MLP `c_proj`) could not broadcast, and `mlxcel_core::add` is not declared fallible in the cxx bridge, so the MLX C++ exception was an uncatchable `std::terminate` that killed the process instead of returning an error.

## What changed

- `src/lora/loader.rs`: `reject_conv1d_layout_fusion` takes a whole-map layout verdict before the first `add`. The verdict comes from the fused QKV projection, which is `[n_embd, 3 * n_embd]` on disk and `[3 * n_embd, n_embd]` once transposed and can never be confused; that sibling is the only evidence available, since the square `c_proj` carries no layout signal of its own. Quantized projections are skipped on the same `.scales` probe `Gpt2Layout::detect` uses, and the reported evidence key is the lexicographically smallest match so the diagnostic does not depend on `HashMap` order.
- `src/lora/loader.rs`: a shape guard compares base and delta before every `add` and returns an `anyhow::Error` naming the adapter layer, the base weight key and both shapes, with an added note when the two are exact transposes. This covers any future family that lands with a transposed on-disk layout.
- Nothing is auto-repaired. Transposing the delta would fix the non-square projections and leave the square one exactly as wrong, because a square weight carries no signal about which orientation was intended. Rejection is the only honest outcome, so the error says what to use instead.
- Both guards live in the shared primitive, so they cover the single-process path through `apply_lora_adapters` and the pipeline-parallel `apply_stage_lora_adapter`, which never goes through `load_model_with_adapter` and would have missed a fix applied at the loading layer.
- Both diagnostics are returned errors, not `tracing::warn!`, which is a no-op in the `mlxcel` CLI binary because only `src/server/startup.rs` installs a subscriber.
- Rejection is per-target, not blanket: an adapter that lands on nothing transposed (an embedding-only adapter, say) still fuses on a Conv1D checkpoint.
- `src/lora/loader_tests.rs` (new): the inline `mod tests` moved out to a sibling `_tests.rs` per the project convention, keeping `loader.rs` under the file-size limit, plus the new coverage.
- `src/models/gpt2.rs`, `src/models/gpt2_tests.rs`: the checkpoint-layout fixtures are now `pub(crate)` so the LoRA tests assert against the same maps instead of re-deriving them.

## Test plan

The layout tests fuse an mlxcel-convention (`.lora_a` / `.lora_b`) adapter against the raw HuggingFace GPT-2 fixtures from `src/models/gpt2_tests.rs`, reused rather than reinvented: a locally invented fixture would assume the very thing under test, whereas `raw_hf_weights` is the map `Gpt2Layout::detect` is separately proven to classify as a Conv1D export. Every layout test re-asserts that classification before touching fusion.

- [x] `conv1d_checkpoint_rejects_fusion_into_the_square_attention_c_proj` asserts base and delta shapes are equal before asserting the rejection, which is what demonstrates that a shape check alone cannot catch this case, and asserts the base tensor is left byte-unchanged.
- [x] `conv1d_checkpoint_rejects_every_non_square_projection_without_aborting` covers `c_attn`, `c_fc` and MLP `c_proj`. Reaching the assertions at all is the evidence: on an unguarded build each of these aborts the whole test binary with SIGABRT.
- [x] `an_already_transposed_gpt2_checkpoint_still_fuses_correctly` and `conv1d_checkpoint_still_fuses_an_adapter_that_targets_nothing_transposed` are positive controls asserting exact fused sums.
- [x] `stage_local_fusion_rejects_a_conv1d_layout_stage_weight_map` and `stage_local_fusion_still_applies_to_an_out_in_stage_weight_map` cover `apply_stage_lora_adapter` end to end through a real on-disk adapter directory.
- [x] `a_transposed_base_weight_is_reported_instead_of_reaching_mlx_add` and `a_mismatch_that_is_not_a_transpose_omits_the_transpose_hint` cover the generic guard on a family with no Conv1D signal.
- [x] `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate lora` (20 passed)
- [x] `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate partial_loading_adapter` (14 passed, including the pre-existing `stage_filtered_fusion_matches_full_fusion_on_stage_subset` parity test, so `[out, in]` behavior is unchanged)
- [x] `DEVELOPER_DIR=... cargo test --release --lib --features metal,accelerate models::gpt2` (25 passed)
- [x] `DEVELOPER_DIR=... cargo clippy --release --lib --tests --features metal,accelerate` (clean; the two `host_preprocessor` warnings are pre-existing on `main`)

Closes #925

### DIFF
--- a/src/distributed/pipeline/partial_loading_adapter_tests.rs
+++ b/src/distributed/pipeline/partial_loading_adapter_tests.rs
@@ -458,3 +458,121 @@ fn stage_filtered_fusion_is_noop_on_empty_stage_adapter() {
     let val = mlxcel_core::item_f32(&sum);
     assert!((val - 10.0).abs() < 1e-5, "base drifted: {val}");
 }
+
+// ---------------------------------------------------------------------------
+// apply_stage_lora_adapter — the pipeline-parallel entry point
+//
+// This path never goes through `load_model_with_adapter`, so it needs its own
+// coverage of the base-weight layout guard rather than inheriting the non-PP
+// path's.
+// ---------------------------------------------------------------------------
+
+fn ones_tensor(rows: usize, cols: usize) -> OwnedTensor {
+    let mut data = Vec::with_capacity(rows * cols * 4);
+    for _ in 0..rows * cols {
+        data.extend_from_slice(&1.0f32.to_le_bytes());
+    }
+    OwnedTensor {
+        dtype: safetensors::tensor::Dtype::F32,
+        shape: vec![rows, cols],
+        data,
+    }
+}
+
+/// Write an mlx-lm-style adapter directory: `lora_a` is `[in, rank]` and
+/// `lora_b` is `[rank, out]`, both all ones, so every delta entry is
+/// `scale * rank`.
+fn write_adapter_dir(dir: &std::path::Path, layer: &str, in_f: usize, rank: usize, out_f: usize) {
+    use std::collections::HashMap;
+
+    std::fs::create_dir_all(dir).unwrap();
+    std::fs::write(
+        dir.join("adapter_config.json"),
+        format!(
+            r#"{{"fine_tune_type": "lora", "lora_parameters": {{"rank": {rank}, "scale": 1.0}}}}"#
+        ),
+    )
+    .unwrap();
+
+    let mut tensors: HashMap<String, OwnedTensor> = HashMap::new();
+    tensors.insert(format!("{layer}.lora_a"), ones_tensor(in_f, rank));
+    tensors.insert(format!("{layer}.lora_b"), ones_tensor(rank, out_f));
+    safetensors::serialize_to_file(&tensors, None, &dir.join("adapters.safetensors")).unwrap();
+}
+
+fn sum_of(weights: &mlxcel_core::weights::WeightMap, key: &str) -> f32 {
+    let w = weights.get(key).unwrap();
+    mlxcel_core::eval(w);
+    let sum = mlxcel_core::sum_all(w);
+    mlxcel_core::eval(&sum);
+    mlxcel_core::item_f32(&sum)
+}
+
+#[test]
+fn stage_local_fusion_rejects_a_conv1d_layout_stage_weight_map() {
+    // A pipeline stage holding a raw HuggingFace GPT-2 export: `c_proj` is
+    // square, so the delta broadcasts and nothing downstream would notice the
+    // transposed accumulation. The stage path must refuse it just like the
+    // single-process path does.
+    let args = crate::models::gpt2::tests::tiny_args();
+    let mut base = crate::models::gpt2::tests::raw_hf_weights(&args, "transformer.");
+    let h = args.n_embd;
+
+    let tmp_dir = temp_dir("conv1d");
+    write_adapter_dir(&tmp_dir, "transformer.h.0.attn.c_proj", h, 2, h);
+
+    let filter = LayerFilter {
+        layer_range: 0..1,
+        has_embedding: true,
+        has_lm_head: true,
+    };
+
+    let key = "transformer.h.0.attn.c_proj.weight";
+    let before = sum_of(&base, key);
+
+    let err = match crate::lora::apply_stage_lora_adapter(&mut base, &tmp_dir, &filter) {
+        Ok(()) => panic!("a Conv1D-layout stage weight map must not fuse"),
+        Err(err) => err.to_string(),
+    };
+    assert!(err.contains("Conv1D"), "{err}");
+    assert!(err.contains(key), "{err}");
+
+    let after = sum_of(&base, key);
+    assert!(
+        (before - after).abs() < 1e-6,
+        "stage weights must be left untouched: {before} -> {after}"
+    );
+
+    let _ = std::fs::remove_dir_all(&tmp_dir);
+}
+
+#[test]
+fn stage_local_fusion_still_applies_to_an_out_in_stage_weight_map() {
+    // Positive control: the layout guard must not disturb the layout every
+    // other family (and every MLX conversion) actually ships.
+    use mlxcel_core::weights::WeightMap;
+
+    let mut base: WeightMap = WeightMap::new();
+    base.insert(
+        "model.layers.0.self_attn.q_proj.weight".into(),
+        mlxcel_core::ones(&[4, 4], mlxcel_core::dtype::FLOAT32),
+    );
+
+    let tmp_dir = temp_dir("outin");
+    write_adapter_dir(&tmp_dir, "model.layers.0.self_attn.q_proj", 4, 2, 4);
+
+    let filter = LayerFilter {
+        layer_range: 0..1,
+        has_embedding: false,
+        has_lm_head: false,
+    };
+
+    crate::lora::apply_stage_lora_adapter(&mut base, &tmp_dir, &filter)
+        .expect("stage fusion succeeds on an [out, in] weight map");
+
+    // 16 entries of 1.0, plus a delta of scale (1.0) * rank (2) everywhere.
+    let fused = sum_of(&base, "model.layers.0.self_attn.q_proj.weight");
+    assert!((fused - 48.0).abs() < 1e-4, "unexpected fused sum: {fused}");
+
+    let _ = std::fs::remove_dir_all(&tmp_dir);
+}
--- a/src/distributed/pipeline/partial_loading_adapter_tests.rs
+++ b/src/distributed/pipeline/partial_loading_adapter_tests.rs
@@ -460,7 +460,7 @@ fn stage_filtered_fusion_is_noop_on_empty_stage_adapter() {
 }
 
 // ---------------------------------------------------------------------------
-// apply_stage_lora_adapter — the pipeline-parallel entry point
+// apply_stage_lora_adapter: the pipeline-parallel entry point
 //
 // This path never goes through `load_model_with_adapter`, so it needs its own
 // coverage of the base-weight layout guard rather than inheriting the non-PP
--- a/src/lora/loader.rs
+++ b/src/lora/loader.rs
@@ -94,7 +94,24 @@ pub fn fuse_lora_weights(
 /// the filter will produce fusion updates — `base_weights` entries that are
 /// not referenced by any `lora_a` / `lora_b` pair are left untouched.
 ///
-/// Used by: [`fuse_lora_weights`], `apply_lora_adapters_in_place`
+/// # Layout
+///
+/// Every delta this produces is in `mlxcel_core` `Linear` layout,
+/// `[out_features, in_features]`. The base weight has to be in the same layout
+/// for the element-wise add to mean anything, and nothing upstream guarantees
+/// that: `load_model_with_adapter` fuses into the raw safetensors map and only
+/// afterwards constructs the model, which is where a family such as GPT-2
+/// transposes its on-disk `Conv1D` projections. Both mismatch cases are checked
+/// here, before any operand reaches MLX. See
+/// [`reject_conv1d_layout_fusion`] and the shape guard in the fusion loop.
+///
+/// # Errors
+///
+/// Returns an error, leaving `base_weights` untouched, when the base weight map
+/// stores its projections transposed, or when a resolved base weight and its
+/// delta disagree on shape.
+///
+/// Used by: [`fuse_lora_weights`], [`apply_stage_lora_adapter`]
 pub fn fuse_lora_weights_into(
     base_weights: &mut WeightMap,
     adapter_weights: &WeightMap,
@@ -118,6 +135,15 @@ pub fn fuse_lora_weights_into(
             lora_pairs.entry(base_name).or_insert((None, None)).1 = Some(mlxcel_core::copy(weight));
         }
         // Ignore other weights (like scales for DoRA)
+    }
+
+    // Refuse up front if the base weight map still stores its projections
+    // transposed. This has to be a whole-map verdict taken before the first
+    // `add`, because the square `attn.c_proj` is precisely the case where the
+    // two shapes agree and the corruption would be silent.
+    if !lora_pairs.is_empty() {
+        let adapter_layers: Vec<&str> = lora_pairs.keys().map(String::as_str).collect();
+        reject_conv1d_layout_fusion(base_weights, &adapter_layers)?;
     }
 
     // Fuse each LoRA pair with the corresponding base weight
@@ -145,6 +171,34 @@ pub fn fuse_lora_weights_into(
         // Compute the LoRA delta: scale * (lora_b @ lora_a)
         let delta = compute_lora_delta(&lora_a, &lora_b, scale)?;
 
+        // Never hand mismatched operands to `mlxcel_core::add`. MLX broadcasts,
+        // so a mismatch is not reliably an error: a base weight that is one
+        // broadcastable step away from the delta fuses cleanly and produces a
+        // silently wrong tensor. And where broadcasting genuinely fails, `add`
+        // is not declared fallible in the cxx bridge, so the MLX C++ exception
+        // is `std::terminate` and takes the process down instead of returning
+        // an error the caller can report.
+        let base_shape = mlxcel_core::array_shape(base_weight);
+        let delta_shape = mlxcel_core::array_shape(&delta);
+        if base_shape != delta_shape {
+            let is_transpose = base_shape.len() == 2
+                && delta_shape.len() == 2
+                && base_shape[0] == delta_shape[1]
+                && base_shape[1] == delta_shape[0];
+            let hint = if is_transpose {
+                " The two are exact transposes of each other, so this base weight is stored \
+                 transposed relative to the [out_features, in_features] layout every LoRA delta \
+                 uses. Fusion does not transpose an operand to make the shapes agree; convert the \
+                 checkpoint to [out_features, in_features] instead."
+            } else {
+                ""
+            };
+            anyhow::bail!(
+                "LoRA shape mismatch fusing adapter layer {base_name} into base weight \
+                 {base_weight_name}: base is {base_shape:?}, delta is {delta_shape:?}.{hint}"
+            );
+        }
+
         // Fuse: W_fused = W_base + delta
         let fused = mlxcel_core::add(base_weight, &delta);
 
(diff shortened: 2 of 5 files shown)


inureyes:
One acceptance criterion is intentionally left unticked on the issue: the end-to-end run that loads a GPT-2 checkpoint with an adapter through `load_model_with_adapter` and generates against it. No GPT-2 checkpoint is present in this environment, and real-model validation is handled separately from this change.

What that run should show, given the guard: a raw HuggingFace `gpt2` checkpoint plus an adapter targeting any of `h.N.attn.c_attn`, `h.N.attn.c_proj`, `h.N.mlp.c_fc` or `h.N.mlp.c_proj` now fails at load with the "LoRA fusion refused" error naming `h.0.attn.c_attn.weight`, its `[n_embd, 3 * n_embd]` shape, and the affected targets, instead of aborting the process (non-square) or loading a silently corrupted `attn.c_proj` (square). An mlx-community GPT-2 conversion plus the same adapter fuses and generates exactly as before, since its projections are already `[out_features, in_features]` and the layout probe finds no Conv1D signal.

inureyes:
## CI blocker, pre-existing on `main`, unrelated to this PR

`Pipeline Parallel CI / 2-host logical` is red. The pipeline-parallel job itself passed; only the trailing `Clippy on PP modules` step failed, on two findings in `src/multimodal/`:

```
error: unused import: `export_qwen2_vl_prefill`
  --> src/multimodal/host_preprocessor.rs:47:51
error: function `export_qwen2_vl_prefill` is never used
  --> src/multimodal/host_preprocessor_export.rs:154:15
```

This PR touches five files, none of them under `src/multimodal/`:

```
src/distributed/pipeline/partial_loading_adapter_tests.rs
src/lora/loader.rs
src/lora/loader_tests.rs
src/models/gpt2.rs
src/models/gpt2_tests.rs
```

The cause is on `main`. `export_qwen2_vl_prefill` has exactly one non-test call site, `src/multimodal/host_preprocessor.rs:428`, and it sits inside a block gated on `#[cfg(feature = "xla-iree")]`. The CI step runs `cargo clippy -p mlxcel --lib --tests -- -D warnings` with no features, so that call site is compiled out, the import goes unused and the function goes dead, and `-D warnings` turns both into errors. It reproduces locally on the same command with no local changes applied.

It has gone unnoticed because this workflow is path-filtered to the PP modules and last ran on 2026-07-20, while the code arrived on 2026-07-24 in `edc0ebbb6` (`feat: add Qwen2-VL OpenXLA vision path`). This is simply the first PP-touching PR since then.

Left unfixed here rather than folded in, since it is an unrelated module and belongs in its own change. The fix is to gate the import and the function on `feature = "xla-iree"` the way the call site already is, and it will unblock every PP-touching PR, not just this one.

Issue #932, opened 2026-07-26
feat(xla): define operator-level MLX/IREE numeric contracts


inureyes:
## Parent and affected work

- Parent epic: #566
- Current default-branch baseline: `ec6fbb6d` (shared foundation through #945)
- Deterministic MLX CUDA Q4 baseline: #923
- Affected family issues and draft PRs: #866 / #920, #868 / #921, #869 / #917, #870 / #918, #871 / #916, #872 / #919, and #878 / #922

## Problem

The remaining OpenXLA multimodal family ports now fail at a shared boundary rather than at missing model topology: MLX CUDA and StableHLO/IREE can execute mathematically equivalent operators with different operand materialization, accumulator precision, reassociation, reduction trees, fused-kernel ordering, or backend-selected convolution plans. The current family implementations encode some of these choices locally, so each full-model oracle run is rediscovering the same backend contract one architecture at a time.

This is not a request to relax tolerances. The existing family gates must remain unchanged, and final greedy output remains token-exact. If an MLX kernel has no deterministic, representable execution contract, the affected XLA capability must remain fail-closed until a reproducible contract is implemented or the reference path itself is made deterministic.

The reference baseline also changed after the current seven branches were cut: #923 fixed a shared-memory race in MLX CUDA `qmm_sm80` and changed the kernel artifact identity. Any Q4 evidence collected before `f7a877e7` must therefore be treated as historical until it is reproduced on the deterministic kernel.

## Current evidence

| Family | Draft PR | First unresolved boundary |
| --- | --- | --- |
| Molmo2 | #916 at `684c6ee1` | Current-head actual local-task gate reaches `projector.output_all`; first failure is flat index 358403 with `max_abs=0.1640625`, `rms=0.0026185848`. The immediately preceding `projector.product` passes (`max_abs=0.037597656`, `rms=0.0002196383`), isolating final dense-projection accumulation/association. |
| Gemma3 VLM | #917 at `9ab13992` | Current-head actual run was bounded before model start by a one-time pinned-MLX cache rebuild. The last valid actual evidence remains `siglip.hidden.block1.output` at `002bb77e`: flat index 658517, 8 failures, `max_abs=0.6672516`; current head passes the exact example check and focused Gemma3 tests but makes no new parity claim. |
[...]

PR #968, opened 2026-07-30
fix(xla): stop emitting the terminating EOS token as output

### LINKED ISSUE #963: fix(xla): terminating EOS token leaks into generated output on every XLA path
## Problem

Every OpenXLA generation path returns the terminating EOS id as if it were generated
output. The eager MLX paths drop it. The result is a visible control token in user
content and a completion-token count that is one too high.

Reproduced on GB10 (driver 580.159.03, CUDA 13.0, MLX pin `b7c3dd6d27f4`,
`MLX_CUDA_ARCHITECTURES=121`, `MLX_ENABLE_TF32=0`), release profile, checkpoint
`molmo2-4b`, prompt `Answer with one word only. What color is snow?`, `-n 12`:

| path | returned content | reported tokens |
| --- | --- | --- |
| eager MLX CLI (default backend) | `White` | 1 |
| `MLXCEL_BACKEND=xla` CLI | `White<\|im_end\|>` | 2 |

The same defect shows on the XLA server: a text-only request to a Molmo2 XLA server
returns content `Red<|im_end|>` with `finish_reason=stop`, where the eager server
returns `Red`.

This is not model-specific and not server-specific. It affects every model served or
generated through the XLA backend, on both the CLI and the OpenAI-compatible server.

## Root cause

The eager path treats the terminating id as a control token. `BatchScheduler` checks it
*before* recording or detokenizing, so it never reaches `generated_tokens` or the
incremental detokenizer:

```rust
// src/server/batch/scheduler.rs
if seq.merged_eos.contains(&token_val) {
    // transition to Finished(Stop)
    continue;
}
seq.generated_tokens.push(token_val);
```

Every XLA path records or emits the token *before* testing it:

- `src/lib/mlxcel-xla/src/batch.rs:686` — `XlaBatchEngine::pump` admission: pushes
  `EngineEvent::Token` for the prefill argmax, then computes `finish_reason`.
- `src/lib/mlxcel-xla/src/batch.rs:763` — `XlaBatchEngine::pump` decode: pushes
  `EngineEvent::Token`, then checks `finish_reason`.
- `src/lib/mlxcel-xla/src/lib.rs:420` — `generate_streaming_greedy`: `out.push(next)`
  before the EOS test.
- `src/lib/mlxcel-xla/src/lib.rs:586` — the prepared-input greedy loop, same shape.
- `src/lib/mlxcel-xla/src/batch.rs` — `XlaReferenceEngine::generate`, same shape.

Because `XlaServeWorker` counts one generated token per `EngineEvent::Token`
(`src/server/batch/xla_worker.rs:328`) and detokenizes whatever it receives, the engine
handing over the EOS token is enough to produce both symptoms.

Note the admission path is already inconsistent with itself: when the prefill argmax is
EOS it emits `Token` *and* `Finished`, while the slot is discarded without ever decoding.

## Expected contract

The terminating EOS id is a control token, never output. On every XLA path:

- it must not appear in returned tokens or streamed content,
- it must not count toward `completion_tokens`,
- `finish_reason` must still be `stop`,
- a token that ends generation by hitting the length cap is real output and must still
  be emitted.

## Required implementation

- Make the emit/finish decision one explicit, unit-testable step shared by both
  `pump` sites, so `Length` still emits and `Stop` does not.
- Apply the same ordering to the three single-sequence greedy loops.
- Keep `finish_reason`'s existing "EOS wins over Length" precedence.

## Validation

- Unit tests, no device needed: EOS suppressed and reported as `Stop`; a cap-reached
  non-EOS token emitted and reported as `Length`; EOS as the very first (prefill) token
  emits nothing and finishes.
- Real checkpoint on CUDA: `MLXCEL_BACKEND=xla` CLI output and token count match the
  eager CLI byte for byte on a prompt that terminates on EOS.
- Real checkpoint on CUDA: XLA server non-streaming and streaming content carries no
  EOS text, `finish_reason=stop`, and `completion_tokens` matches the eager server.

## Acceptance criteria

- [ ] No XLA path returns, streams, or counts the terminating EOS id.
- [ ] A length-terminated final token is still emitted and counted.
- [ ] `finish_reason` values are unchanged (`stop` for EOS, `length` for the cap).
- [ ] XLA CLI matches eager CLI content and token count on the pinned fixture.
- [ ] XLA server streaming and non-streaming content contain no EOS text.
- [ ] Unit tests cover the emit/finish decision without a device.

## Out of scope

`SampleParams` carries no per-request stop token ids and `XlaBatchEngine` builds its
EOS set from the checkpoint config alone, so request-level `stop_token_ids` are not
honoured by the XLA engine at all. That is a separate gap and is not addressed here.

Found while qualifying #916 against #871's "text-only behaviour remains token-exact"
criterion. Refs #566, #932.

inureyes:
## Summary

The OpenXLA backend handed the terminating EOS id to its consumers as if it were
generated output. The eager MLX paths drop it. That produced a visible control token
in user-facing content and a completion-token count one too high, on both the CLI
and the OpenAI-compatible server, for every model served through this backend.

Measured on GB10 with `molmo2-4b`, prompt `Answer with one word only. What color is
snow?`, `-n 12`:

| path | content | tokens |
| --- | --- | --- |
| eager MLX CLI | `White` | 1 |
| `MLXCEL_BACKEND=xla` CLI, before | `White<\|im_end\|>` | 2 |
| `MLXCEL_BACKEND=xla` CLI, after | `White` | 1 |

## What changed

The eager `BatchScheduler` tests `merged_eos` *before* recording or detokenizing, so
the id never reaches `generated_tokens` or the incremental detokenizer. Every XLA
path recorded or emitted first and tested after. This makes the ordering match.

- `mlxcel-xla/src/batch.rs`: new `commit_token` returns both whether a sampled token
  is client-visible and why the sequence ended, and both `XlaBatchEngine::pump`
  sites (admission and decode) use it. `finish_reason`'s existing "EOS wins over
  Length" precedence is untouched, and a token that ends generation by reaching the
  cap is still emitted.
- `mlxcel-xla/src/lib.rs`: the four single-sequence greedy generators now share one
  `drive_greedy` driver that applies the same contract once, instead of four copies
  of a push-then-test loop.
- `mlxcel-xla/src/batch.rs`: `XlaReferenceEngine::generate` likewise, so a reference
  token sequence is directly comparable with an eager one.
- `server/batch/xla_worker.rs`: the decode-width metric is now
  `sampled_token_count`, which counts a `Stop` finish as the one sampled token whose
  id was withheld. Without this the throughput metric would silently lose one step
  per sequence.

## Why this matters beyond the visible symptom

Under the reduction-order contract just decided for #932, a family PR is gated on
per-operator parity plus **greedy token-exactness against MLX**. Before this fix
every XLA family would have failed that gate on one trailing token, regardless of
its numerics, so this lands first.

## Validation

Unit tests, no device required:

- `commit_token_withholds_the_terminating_eos_id` — suppressed and reported `Stop`,
  including as the very first (prefill) token and when it is also the cap-th token.
- `commit_token_emits_a_length_terminated_final_token` — a non-EOS token that
  exhausts the budget is still emitted and reported `Length`.
- `stop_finish_without_a_token_event_returns_empty_content_and_no_completion_tokens`
  — a bare `Finished{Stop}` yields empty content, `completion_tokens=0`, and
  `finish_reason=stop`.
- `sampled_token_count_charges_a_withheld_eos_to_the_decode_step` — the metric does
  not shrink.

`cargo test -p mlxcel-xla --lib`: 252 passed.
`cargo test --release --lib --features xla-iree server::batch::xla_worker`: 9 passed.
`cargo fmt --all --check` and `git diff --check` clean.

Real checkpoint on CUDA (GB10, driver 580.159.03, CUDA 13.0, MLX pin
`b7c3dd6d27f4`, `MLX_CUDA_ARCHITECTURES=121`, `MLX_ENABLE_TF32=0`, release):
the CLI table above, and the XLA server's non-streaming and streaming content.

Existing callers that pass an empty EOS set (`tests/xla_prepared_prefill.rs`,
`tests/xla_gemma3n_prefill.rs`) are unaffected: with no EOS id the loops emit
exactly `max_new_tokens` tokens and issue the same number of decode steps as before.

## Out of scope

`SampleParams` carries no per-request stop token ids and `XlaBatchEngine` builds its
EOS set from the checkpoint config alone, so request-level `stop_token_ids` are not
honoured by the XLA engine at all. That is a separate gap, noted in the issue and
not addressed here.

Closes #963
Refs #566, #932, #916

### DIFF
--- a/src/lib/mlxcel-xla/src/batch.rs
+++ b/src/lib/mlxcel-xla/src/batch.rs
@@ -112,7 +112,8 @@ impl From<PreparedInputError> for XlaAdmissionError {
 /// called [`XlaBatchEngine::cancel`] already knows), so it is not a finish reason.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum FinishReason {
-    /// An EOS token was emitted.
+    /// An EOS token was sampled. The id itself is a control token and is not
+    /// emitted as a [`EngineEvent::Token`] (issue #963).
     Stop,
     /// The token budget (`max_new_tokens`) was reached.
     Length,
@@ -230,6 +231,35 @@ fn finish_reason(produced: usize, cap: usize, last: i32, eos: &[i32]) -> Option<
         Some(FinishReason::Length)
     } else {
         None
+    }
+}
+
+/// What committing one sampled token means for its slot.
+#[derive(Debug, PartialEq, Eq)]
+struct TokenCommit {
+    /// Whether the token is client-visible output.
+    emit: bool,
+    /// Why the sequence ended, if it did.
+    finish: Option<FinishReason>,
+}
+
+/// Decide, for a slot that has now produced `produced` tokens against `cap`,
+/// whether `last` is output and whether it ends the sequence.
+///
+/// The terminating EOS id is a control token, not output. The eager
+/// `BatchScheduler` drops it before it reaches `generated_tokens` or the
+/// incremental detokenizer, so the XLA engine must not hand it to consumers
+/// either: `XlaServeWorker` detokenizes every `EngineEvent::Token` it receives
+/// and counts one generated token per event, so emitting the EOS id both leaks
+/// it into content (`White<|im_end|>`) and reports one token too many
+/// (issue #963).
+///
+/// A token that ends the sequence by reaching `cap` is real output and is still
+/// emitted. EOS still wins over Length when a token is both.
+fn commit_token(produced: usize, cap: usize, last: i32, eos: &[i32]) -> TokenCommit {
+    TokenCommit {
+        emit: !eos.contains(&last),
+        finish: finish_reason(produced, cap, last, eos),
     }
 }
 
@@ -683,10 +713,13 @@ impl XlaBatchEngine {
             if needs_history {
                 history.push(first);
             }
-            events.push(EngineEvent::Token {
-                req_id: p.req_id,
-                token: first,
-            });
+            let commit = commit_token(1, p.cap, first, &eos);
+            if commit.emit {
+                events.push(EngineEvent::Token {
+                    req_id: p.req_id,
+                    token: first,
+                });
+            }
             let slot = Slot {
                 req_id: p.req_id,
                 adapter_mode: p.input.adapter_mode(),
@@ -699,7 +732,7 @@ impl XlaBatchEngine {
                 rng,
                 history,
             };
-            if let Some(reason) = finish_reason(slot.produced, slot.cap, first, &eos) {
+            if let Some(reason) = commit.finish {
                 // Finished at its first token: leave the slot free for the next admit.
                 events.push(EngineEvent::Finished {
                     req_id: p.req_id,
@@ -759,9 +792,11 @@ impl XlaBatchEngine {
                 slot.cache_len += 1;
                 slot.produced += 1;
                 let req_id = slot.req_id;
-                let done = finish_reason(slot.produced, slot.cap, nt, &eos);
-                events.push(EngineEvent::Token { req_id, token: nt });
-                if let Some(reason) = done {
+                let commit = commit_token(slot.produced, slot.cap, nt, &eos);
+                if commit.emit {
+                    events.push(EngineEvent::Token { req_id, token: nt });
+                }
+                if let Some(reason) = commit.finish {
                     events.push(EngineEvent::Finished { req_id, reason });
                     *slot_opt = None;
                 }
@@ -1232,6 +1267,10 @@ impl XlaReferenceEngine {
     /// is prefilled in full and its argmax is the first token, then decode advances
     /// from there.
     ///
+    /// The terminating EOS id is a control token and is not returned, so this
+    /// reference sequence is directly comparable with an eager MLX one
+    /// (issue #963).
+    ///
     /// # Errors
     ///
     /// Propagates prefill / decode failures.
@@ -1249,15 +1288,16 @@ impl XlaReferenceEngine {
         if max_new_tokens == 0 {
             return Ok(Vec::new());
         }
-        let first = self.inner.prefill_first(prompt)?;
-        let mut out = vec![first];
+        let mut out = Vec::with_capacity(max_new_tokens);
         let mut cache_len = prompt.len() as i32;
-        let mut cur = first;
-        while out.len() < max_new_tokens && !eos.contains(&cur) {
-            let nt = self.inner.decode(cur, cache_len)?;
-            out.push(nt);
+        let mut cur = self.inner.prefill_first(prompt)?;
+        while !eos.contains(&cur) {
+            out.push(cur);
+            if out.len() >= max_new_tokens {
+                break;
+            }
+            cur = self.inner.decode(cur, cache_len)?;
             cache_len += 1;
-            cur = nt;
         }
         Ok(out)
     }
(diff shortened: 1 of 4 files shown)


inureyes:
## Real-checkpoint before/after on GB10

Both measurements use the **same fixture, the same binary lineage and the same
host**. The only difference is whether the three changed source files carry this
commit or its parent: I reverted exactly
`mlxcel-xla/src/{batch.rs,lib.rs}` and `server/batch/xla_worker.rs` to `HEAD~1`,
rebuilt `--release --features cuda,xla-iree --bin mlxcel`, measured, then restored
them and measured again. No result is carried over from another head or another
model.

Environment: GB10 / Grace-Blackwell, driver 580.159.03, CUDA 13.0, rustc 1.93.1,
MLX pin `b7c3dd6d27f4`, `MLX_CUDA_ARCHITECTURES=121` (the shipped value, confirmed in
the live `CMakeCache.txt`), `MLX_ENABLE_TF32=0`, release profile, IREE
3.12.0rc20260721. Machine idle (`loadavg` 0.4) for both runs.

Prompt `Answer with one word only. What color is snow?`, `max_tokens=12`,
`temperature=0`.

| consumer | before | after |
| --- | --- | --- |
| eager MLX CLI (reference) | `White`, 1 token | `White`, 1 token |
| XLA CLI | `White<\|im_end\|>`, **2 tokens** | `White`, **1 token** |
| XLA server, non-streaming | `White<\|im_end\|>`, `stop`, `completion_tokens=2` | `White`, `stop`, `completion_tokens=1` |
| XLA server, streaming (assembled) | `White<\|im_end\|>`, `stop` | `White`, `stop` |

After the fix the XLA CLI matches the eager CLI byte for byte and token for token,
the server carries no EOS text on either the streaming or the non-streaming path,
`finish_reason` stays `stop`, and `completion_tokens` drops from 2 to the correct 1.

### Checkpoint note

The fixture is `qwen2.5-1.5b-instruct-4bit` plus a `generation_config.json`
containing only the two ids the checkpoint's own `config.json` already declares
(`bos_token_id: 151643`, `eos_token_id: 151645`); every other file is a hardlink to
the original. That file has to be added because both backends read EOS ids **only**
from `generation_config.json` (`loading/mod.rs:314`,
`mlxcel-xla/src/lib.rs:227`) and never fall back to `config.json`, which HF
`transformers` does. On the unmodified mlx-community repack, `eager` and XLA alike
run to the token cap and never reach an EOS at all, so that checkpoint cannot
exercise this path in either direction. That fallback gap is pre-existing, affects
both backends equally, and is not touched here; it is worth its own issue.

The originally reported symptom (`Red<|im_end|>` on a Molmo2 XLA server) cannot be
reproduced from `main` because Molmo2's XLA support lands with #916, not here. The
Molmo2 numbers in the issue were measured on that branch; the table above is the
`main`-based equivalent on a checkpoint `main` can actually serve.

inureyes:
## Amended before merge: a regression I introduced in the first push

Reviewing my own diff, `drive_greedy` had the two break conditions in the wrong
order:

```rust
if out.len() >= max_new_tokens || !on_token(current) { break; }
```

`||` short-circuits, so once the token budget was exhausted the streaming callback
was **never invoked for that final token**. The original loop called `on_token` for
every token including the last, so a streaming consumer would have silently lost the
last piece. The CLI would not have shown it (it renders from the returned vector),
which is exactly the kind of thing that survives a casual test.

Fixed by invoking the callback first, and the ordering requirement is now stated in
the doc comment so it is not "tidied" back:

```rust
if !on_token(current) || out.len() >= max_new_tokens { break; }
```

To make this testable without a device I split the loop out of the session as
`drive_greedy_loop`, taking `decode` as a closure. Four new tests in
`greedy_contract_tests` run under both the plain and the `iree` build:

- `withholds_the_terminating_eos_and_never_streams_it`
- `returns_nothing_when_the_prefill_token_is_eos`
- `streams_the_token_that_exhausts_the_budget` (the regression guard)
- `a_false_callback_stops_generation_keeping_that_token`

`cargo test -p mlxcel-xla --lib`: 254 passed, 11 ignored.
`cargo test --release --features xla-iree --test xla_prepared_prefill --test xla_gemma3n_prefill`:
compiles and runs; all cases are `#[ignore]`d real-checkpoint gates, so this is
compile-level assurance only.
The `unused import: vision::emit_vision_diagnostics` warning is pre-existing on
`main` (verified by stashing this change).

The real-checkpoint before/after table above was measured with the callback in the
pre-amend order. That ordering affects only the streaming callback, not the returned
token vector or the server path (`XlaServeWorker` consumes `EngineEvent`s from the
batch engine, not this loop), so the measured CLI and server results still hold. The
new unit tests cover the callback contract directly.

Issue #898, opened 2026-07-24
Fused paged-attention decode v2: CSR page table, cross-CTA split-KV, and variable-length merge kernels


inureyes:
## Background

mlxcel already ships a fused split-K flash-decoding paged-attention kernel for both backends in `src/lib/mlx-cpp/turbo/paged_attention.cpp`: the Metal body `PAGED_ATTENTION_DECODE_SOURCE` (around line 79) and its CUDA port `PAGED_ATTENTION_DECODE_CUDA_SOURCE` (around line 232), launched through `mlx::core::fast::metal_kernel` / `cuda_kernel` JIT strings. The Rust entry points are `paged_decode_attention_pooled` and the dispatch selectors in `src/lib/mlxcel-core/src/layers.rs` (`select_pooled_paged_dispatch`, `resolve_pooled_paged_dispatch`).

This kernel has a structural scalability limit: its KV split happens **inside a single threadgroup**. The launch shape is grid `(32, NumSplits, B*Hq)` with threadgroup `(32, NumSplits, 1)`, where the 32 lanes partition the head dimension and `NumSplits` SIMD groups stripe KV tokens with online softmax. `NumSplits` is bounded by the 28 KB threadgroup memory budget (see `paged_attention.cpp`, the launch-shape computation around lines 463-505). Consequence: one CTA/threadgroup serves one `(batch, q_head)` pair regardless of context length, so a small batch with a long context leaves most of the GPU idle, and adding context length does not add parallelism.

`docs/adr/0001-paged-attention-gather-vs-fused-kernel.md` quantifies what the production fallback (gather-then-SDPA) costs instead: gather overhead vs the contiguous-SDPA lower bound is under ~15% below ~4096 tokens single-sequence, but ~56% at 16384 and ~67% at 32768; batched decode (batch 4) is already ~48% at 1024 tokens and 2x-3x past 4096. The ADR's stated trigger for building a proper fused path ("single-sequence context past ~16384, or any sustained batched decode") is met by the current server defaults (`--parallel 4` batched decode since v0.4).

The scalable design, standard in the flash-decoding lineage used by modern GPU serving engines, combines four composable pieces:

[...]

Issue #235, opened 2026-06-12
perf(core): chunked slab storage for the paged pool to eliminate growth reallocation entirely


inureyes:
## Problem

A paged pool layer's physical storage is one tensor `[capacity_blocks, block_size, n_kv_heads, head_dim]` per K and V. Growth past capacity reallocates the whole slab: `grow_pool` (src/lib/mlxcel-core/src/cache/paged.rs) allocates a larger tensor with `ffi::zeros` and copies the old contents via `slice_update`. #224 reduced the damage (one presize per prefill span plus eager eval bounding the transient to one layer's old+new pair), but two structural costs remain whenever the pool grows across successive non-shared long prefills:

1. Every growth still copies the entire layer slab (O(pool) memcpy per grow per layer).
2. The freed old slabs form a ladder of strictly increasing sizes in the MLX buffer cache that no later allocation can reuse (the next grow always needs a bigger buffer), so the cache accumulates roughly the sum of the ladder. Measured before the sharing fixes: qwen3-0.6b, 8 sequential pool-extending ~4.2k-token prefills peaked at 5930 MiB vs 3833 MiB on the dense backend (1.55x), with the gap attributable to ladder retention.

Cross-request sharing (#227) removed pool growth for shared-prefix traffic, so the regression no longer shows on the shared scenarios. It still reproduces with DISTINCT long prompts (each prefill extends the pool), e.g. 8 concurrent unique ~4k-token system prompts.

## Proposed solution: chunked slabs

Replace the single per-layer pool tensor with a vector of fixed-size slab tensors:

- `pool_k[layer]`, `pool_v[layer]` become `Vec<UniquePtr<MlxArray>>`, each slab shaped `[POOL_SLAB_BLOCKS, block_size, n_kv_heads, head_dim]`. A physical row maps as `slab = row / POOL_SLAB_BLOCKS`, `row_in_slab = row % POOL_SLAB_BLOCKS`.
- Growth appends a fresh zeros slab. No copy, no free, no transient old+new pair, no ladder: allocator high-water becomes exactly the live capacity. `presize_for_span` appends however many slabs the span needs in one step; the #224 `pool_grow_events` counter becomes a slab-append counter (no slab is ever copied).
[...]

PR #988, opened 2026-07-31
feat(server): route production batched paged decode through the fused v2 kernel

### LINKED ISSUE #899: Route production batched paged decode through the fused v2 kernel, retiring the gather-then-SDPA hot path
Blocked by #898 (fused paged-attention decode v2 kernels and plan).

## Background

The server's paged decode path does not use the fused paged-attention kernel at all. Since #720 the fused kernel is library-only: `mlxcel-server --decode-storage paged` routes pool-backed layers through per-sequence `update_and_fetch_paged` (`src/lib/mlxcel-core/src/cache.rs`, around line 2744) which calls `pool.gather_visible` to materialize a dense per-sequence K/V copy and then runs the ordinary fused SDPA (gather-then-SDPA). `docs/benchmark_results/paged-attention-cuda-port-gb10-2026-07-10.md` records this state, and `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md` measures the cost: ~48% overhead at batch 4 with only 1024 tokens of context, 2x-3x past 4096 tokens, ~56%/~67% single-sequence at 16K/32K.

Since v0.4 the server defaults to batched decode (`--parallel 4`), which is precisely the "sustained batched decode" trigger ADR-0001 set for building the fused path. This issue closes that loop: the v2 kernel from issue #898 becomes the production decode path for paged storage.

Relevant scheduler facts: decode storage backend selection lives in `effective_decode_storage_backend` (`src/server/batch/scheduler.rs`, around line 193; paged only when `max_batch_size > 1 && supports_batching && supports_paged_decode_backend`); the decode step is `execute_decode_step` / `run_decode_tick` (around lines 5863/5911). The env override `MLXCEL_PAGED_ATTENTION_NATIVE` and `parse_native_paged_override` (`layers.rs`, around line 3476) already exist as a dispatch gate and should be preserved as the kill switch.

A design note for the plan/run contract (relevant when MLX's CUDA graph cache is in play): keeping the launch geometry fixed across steps and passing step-varying scalars (like the KV chunk size) through device memory rather than kernel constants lets a captured graph be replayed while batch composition evolves; padded grid slots carry a validity mask and exit early. Adopt this shape for the v2 launch path so graph capture never forces a recapture per step.

## Task

1. **Plan lifecycle in the scheduler**: build the CSR plan (from #898) once per batch-composition change and cache it on the active batch; invalidate on admission, eviction/preemption, sequence finish, and whenever a sequence crosses a page boundary (new block allocated). A cheap incremental update (bump `last_page_len`, occasionally append a page id) is preferred over full rebuilds; measure and pick.
2. **Dispatch**: for models and shapes the v2 kernel supports (f16 KV pool, head_dim/GQA in the validated matrix of #898), replace the `gather_visible` + dense SDPA flow in the batched decode step with a single v2 kernel call over the whole batch. Keep the gather path as the fallback for unsupported shapes and as the `MLXCEL_PAGED_ATTENTION_NATIVE=0` kill-switch target.
3. **Variant coverage on this path**:
   - Sliding-window families (`RotatingKVCache` / `SlidingWindowKVCache` users): restrict each sequence's CSR range to the visible window (adjust indptr/first-page offset and the RoPE position offsets); attention-sink retention (`trim_front_keep_sink`) must keep producing correct visible ranges.
   - Logits soft-cap families: fall back to the gather path initially; note the follow-up.
   - Speculative/MTP verify steps use multi-token queries; they stay on their current path (out of scope, document in code).
4. **Memory accounting**: the v2 workspace (partial V/LSE) must be budgeted under the existing KV-cache budget logic so admission control stays truthful (`block_budget` in `paged.rs`, memory estimate in `src/execution/memory_estimate.rs`).
5. **Docs**: update `docs/CONTINUOUS_BATCHING.md` (backend notes) and add a superseding note to ADR-0001 (its trigger fired; the fused path is now production). Record the decision details in a short ADR if any layout or plan-contract decision deviates from #898.
6. **Rollout**: default ON for paged decode storage once validation passes; `--decode-storage` selection logic itself is unchanged.

## Performance validation (mandatory)

- Server-level benchmark, before/after, same hardware, recorded in `docs/benchmark_results/paged-decode-production-<hw>-<date>.md`:
  - 4 concurrent clients (the documented v0.4 default scenario), aggregate decode throughput and inter-token latency at context lengths 1K, 4K, 16K.
  - Single-sequence long-context decode at 16K and 32K.
- Required outcomes: batched decode aggregate throughput improves at 4K+ context (ADR-0001 predicts the gather overhead removed there is 2x-3x of SDPA time); no scenario in the matrix regresses beyond noise (3%).
- Run on at least one Apple Silicon machine and on GB10 when available.

## Regression guard (mandatory)

- Greedy token parity: on a pinned prompt set (>= 3 model families among Llama3/Qwen3/Gemma, prompts >= 8K tokens, batch 4 with mixed lengths), greedy decode token streams must match the gather-path baseline exactly. Any mismatch is investigated and either fixed or explicitly signed off with a numerical-difference analysis in the PR (matching the repo's pinned-oracle culture).
- `MLXCEL_PAGED_ATTENTION_NATIVE=0` restores the previous gather behavior end to end.
- Full server test suite and `cargo test -p mlxcel-core` pass; dense (`--no-batch` / non-paged) storage paths show zero diff in behavior and performance (spot-check single-client decode within 3%).
- Memory: admission under `--parallel 4` with a constrained KV budget behaves as before (no OOM regressions; workspace accounted).

## References

- `src/server/batch/scheduler.rs`
- `src/lib/mlxcel-core/src/cache.rs` (`update_and_fetch_paged`)
- `src/lib/mlxcel-core/src/cache/paged.rs`
- `docs/adr/0001-paged-attention-gather-vs-fused-kernel.md`
- `docs/CONTINUOUS_BATCHING.md`

Line numbers are indicative as of current `main`; search for the named symbols.

---

Part of epic #909.

inureyes:
## Summary

Makes the fused paged-attention decode v2 kernel from #898 the path every batched paged decode request takes, replacing the per-sequence `gather_visible` + dense SDPA loop that ADR 0001 measured as the paged decode hot path. The switch is gated by a measured token floor, keeps the gather path as the fallback and the parity reference, and preserves `MLXCEL_PAGED_ATTENTION_NATIVE=0` as an end-to-end kill switch.

## The dispatch threshold, and why it is two-regime

#898's M1 Ultra sweep (`docs/benchmark_results/paged-decode-v2-m1ultra-2026-07-31.md`, medians of three repetitions) has exactly one cell where v2 loses to gather: batch 1 at 1024 tokens of context, 0.91x median and below parity in all three repetitions. The plan degenerates to two pages per chunk there and the merge pass plus the workspace round trip costs more than the whole attention does. So the switch cannot be unconditional.

The first cut of this PR summed the launch's visible tokens and required 4096. The production benchmark disproved that formulation: it declined every scenario in the matrix and measured gather against gather. **The loss is a property of batch 1, not of total tokens.** Read the ctx-1024 column: 0.91x at batch 1, 1.41x at batch 4, 1.47x at batch 8, same per-request context and opposite outcome, because a batched launch spreads the same chunk count over more requests and amortizes the merge over more useful work. A total-token floor separates those cells only by accident and with no margin, and the benchmark's nominal 1K scenario delivered 956 tokens per request (`4 * 956 = 3824`), so it was declined despite being the same shape as the measured 1.41x win.

The floor is now stated the way the measurements are:

| launch | floor | evidence |
|---|---|---|
| one request | `MIN_SINGLE_REQUEST_KV_TOKENS` = 4096 visible tokens | 1024 loses (0.91x), 4096 wins (1.08x median, 1.02x-1.21x across repetitions) |
| more than one | `MIN_BATCHED_KV_TOKENS_PER_REQUEST` = 512 per request | 1024 per request wins at batch 4 (1.41x) and batch 8 (1.47x) |

512 rather than 1024 for the batched case is deliberate: 1024 is the *lowest measured* batched context and it wins comfortably, so a floor sitting on top of a measured point declines real workloads that land just under it for no measured reason. Batch 2 and 3 are interpolated, not measured; the trend across batch 1, 4 and 8 is monotone and the mechanism above explains why, so the interpolation is stated in the module docs rather than hidden. `MLXCEL_PAGED_V2_MIN_KV_TOKENS` and `MLXCEL_PAGED_V2_MIN_KV_TOKENS_PER_REQUEST` move the two floors; `MLXCEL_PAGED_ATTENTION_NATIVE=1` bypasses them entirely, which is the supported way to benchmark the declined corner.

## Both decode paths are wired, not just the batched one

`dispatch_sync_decode` routes a batch of one to `decode_single_step`, which calls the model's single-sequence `Attention::forward`, not `forward_split_attention`. Wiring only the batched path left the two single-sequence scenarios in the issue's benchmark matrix unreachable **by construction**, and with staggered prefills it also left most of a batched scenario on gather. Both `forward_split_attention` (whole batch) and `Attention::forward` (batch of one) now take the same entry point.

## Which kernel actually ran is visible at info

The dispatch decision is a returned `PagedDecodeOutcome` carrying the numbers behind it, and the caller announces the first occurrence of each distinct outcome kind at **info**, with no `RUST_LOG` needed. Per kind rather than one global one-shot: a single flag reports only whatever happened first, typically a warmup request, so a later permanent decline for a different reason never surfaces. That is how the first production sweep ran entirely on the gather path without a single line saying so. `scripts/benchmark_paged_decode_production.sh` now exits non-zero if the after arm never logged a fused launch or the before arm did, so a null sweep fails instead of reporting.

Verified on `mlxcel-server` (Llama-3.2-1B-Instruct-4bit, `--parallel 4 --ctx-size 131072`, no `RUST_LOG`): 4 concurrent clients at a nominal 1K prompt log `fused v2 launch (batch 4, 3828 visible KV tokens, 16 chunks, merge on)`; a single client at 16K logs `fused v2 launch (batch 1, 13845 visible KV tokens, ...)`; the same binary under `MLXCEL_PAGED_ATTENTION_NATIVE=0` logs `gather: pinned by MLXCEL_PAGED_ATTENTION_NATIVE` and never a fused launch.

## The thing that was actually blocking this issue

`PagedBlockPool` allocated physical rows in fixed 32-block slabs (#235) and both fused decode kernels read one contiguous pool buffer per side, so any layer past 32 rows was multi-slab and declined. At `block_size` 32 that capped the fused path at 1024 tokens across the **entire batch**, which is below the token floor above: v2 would never have run in production and this issue would have been a no-op.

The slab size is now a per-pool value (`PagedBlockPool::set_slab_blocks`) defaulting to the historical 32, so nothing changes for any caller that does not set it. The server sizes it from the configuration it already reserved KV memory for: `ceil(per_slot_ctx / block_size) * batch`, floored at the old default and clamped by the per-layer share of the paged block budget (`memory_estimate::resolve_paged_slab_blocks`). `MLXCEL_PAGED_SLAB_BLOCKS` overrides it, `0` pins the old default.

The cost is that a layer's first write allocates its whole slab instead of growing into it, front-loading exactly the KV bytes the startup memory estimate already reports for that batch and context. Growth past the slab still appends without copying (#235's property is preserved), so an undersized slab degrades to the pre-#899 gather path rather than failing. This is the main thing to review; the alternative considered and rejected was geometric single-slab growth, which restores the pre-#235 realloc-and-copy behaviour and the ladder of orphaned buffer sizes in the MLX cache that #235 exists to avoid.

**Operational consequence worth flagging**: serving contexts longer than `--ctx-size` implies makes layers outgrow the slab and silently returns them to the gather path. The server logs the resolved value at startup (`Paged KV slab size: N blocks per layer`) and `docs/CONTINUOUS_BATCHING.md` says this plainly.

## What changed

- **`src/lib/mlxcel-core/src/cache/paged_batch_decode.rs`** (new): `paged_batch_decode_attention`, the whole-batch entry point. All-or-nothing contract: it either declines before writing anything (caller's unchanged per-sequence loop runs against an untouched pool) or it owns the step, including running the gather fallback itself when v2 is not the right choice. That is what keeps a fallback from double-writing the pool.
- **`src/lib/mlxcel-core/src/paged_v2/dispatch.rs`** (new): the pure token-floor selector and its env override. `layers::resolve_paged_v2_dispatch` layers `MLXCEL_PAGED_ATTENTION_NATIVE` on top, checked before the selector so the kill switch cannot be out-thought by a shape the selector likes.
- **`src/lib/mlxcel-core/src/paged_v2/plan_cache.rs`** (new): per-layer CSR page table and per-batch chunk plan caching. Building a page table costs one hash lookup per visible page (~131k per decode step for a 32-layer model at batch 4 and 32K context); refreshing the per-request scalars costs four writes per request. The expensive half is kept, the cheap half is **recomputed, never incremented**, because a delta would be one missed invalidation away from feeding the kernel a length that outruns its pages, which is an out-of-bounds read rather than a wrong answer.
- **`src/lib/mlxcel-core/src/cache/paged.rs`**: `PagedBlockPool::paged_decode_batched` (the production entry), `block_epoch` (the cache invalidation counter), `slab_blocks` and `set_slab_blocks`.
- **`src/models/llama3.rs`, `src/models/qwen3.rs`**: `forward_split_attention` calls the new entry point before its existing dense-compat paged block. Between them these two attention modules back Llama, Mistral, Qwen2 / Qwen2.5, Qwen3, Helium, and every VLM whose text backbone is one of those, which is the whole set of families that get pool-backed caches.
- **`src/execution/memory_estimate.rs`**: `resolve_paged_slab_blocks` and the v2 workspace reserve charged to the KV budget.
- **`src/server/model_worker.rs`, `src/server/batch/scheduler.rs`, `src/lib/mlxcel-core/src/cache.rs`**: plumbing the resolved slab size to the lazily created pool.
- **Docs**: ADR 0001 marked superseded in part with a closing addendum, `docs/CONTINUOUS_BATCHING.md` gains an operator-facing section.

## Plan caching and invalidation

Reuse of a cached page table is gated on a per-request page-range fingerprint (`blocks`, `first_page`, `last_page`, `visible`) plus `PagedBlockPool::block_epoch`, a counter bumped by every mutation that can move a block: acquire, release, row assignment, block restore, sequence restore, sequence release, and token trim. It is deliberately **not** bumped by a token append that fits in the current block, which is the decode hot path.

The epoch is what makes the issue's named invalidation events fall out without any scheduler plumbing, because each of them already goes through a pool method that moves blocks: admission (the sequence's first `write_prefill` acquires), eviction and preemption and finish (`release_sequence` releases), page-boundary crossing (`append_tokens` acquires), prompt-cache adopt and detach (`restore_sequence` / row forget). This is a deviation from the issue's wording, which asks for the cache to live on the active batch with scheduler-side invalidation: forwarding events the pool observes first hand would add a coupling that can be forgotten at a new call site, whereas the epoch cannot be. The epoch is global rather than per-layer, so a boundary crossing costs two fully rebuilt steps rather than one; at 32-token pages that is 2 steps in 32, about 6% of the uncached cost.

## Sliding-window CSR ranges

v2 applies no windowing mask of its own, so the window has to be the CSR range. `build_paged_csr_view` (from #898) already emits only the pages from `logical_start / page_size` onward and sets `first_page_offset` to `logical_start % page_size`, so request `r`'s token `i` resolves to absolute position `logical_start + i` and the retired prefix is never addressed. That is the same `[logical_start, len)` window `gather_visible` slices, which is why the two agree. Two properties are documented in the module because both are silent if violated: the window must be **contiguous** (a future paged attention-sink retention, `[0, keep) ++ [start, len)`, is not expressible in one CSR range and must not be routed here; today it cannot be, because `trim_front_keep_sink` returns `0` for pool-backed caches), and **RoPE is applied upstream** from the scheduler's `rope_offsets`, so the view's own `rope_offsets` is not consumed here.

## Workspace budgeting

The workspace is the partial kernel's `(partial_v, lse)` output pair, `PagedDecodePlan::workspace_bytes()`. `resolve_paged_block_budget` now subtracts a reserve before converting the KV byte budget into blocks, so admission reserves blocks it can actually back. The reserve bounds `num_chunks` from the plan's own search invariant (`num_chunks * ctas_per_chunk` lands within a factor of two of the device CTA target, and `ctas_per_chunk >= Hkv`), which makes the dominant term independent of the head counts: `2 * target_ctas * n_rep * (head_dim + 1) * 4`, times two concurrent launches for the decode lookahead pipeline. `n_rep` is bounded at 16 rather than read from config, which over-reserves by a few times in the safe direction. It lands in the low tens of megabytes, well under a tenth of a percent of a serving KV budget; it is charged because the accounting should be truthful, not because it is large.

## Test plan

- [x] `cargo test --release -p mlxcel-core --lib --features metal,accelerate paged_v2::` — 55 passed (dispatch policy, outcome reporting, plan cache reuse and invalidation).
- [x] `cargo test --release -p mlxcel-core --lib --features metal,accelerate cache::paged` — 22 of them for the batched entry point, including the batch-1 fused path, the 4x956 shape the old floor declined, and the multi-slab decline naming its own fix.
- [x] `cargo test --release -p mlxcel-core --lib --features metal,accelerate cache::` — 468 passed.
- [x] `cargo test --release -p mlxcel-core --lib --features metal,accelerate layers::` — 64 passed; `ffi_tests::` — 93 passed.
- [x] `cargo test --release --lib --features metal,accelerate memory_estimate::` — 35 passed.
- [x] `cargo test --release --lib --features metal,accelerate server::batch::scheduler::tests::paged_decode_storage` and `::auto_decode_storage` — 2 passed.
- [x] `cargo test --release --test paged_scheduler_parity --features metal,accelerate -- --ignored` — 4 passed; `paged_real_model_parity` 1, `paged_prefix_share_parity` 7, `paged_kv_serialize_parity` 2, `paged_handoff_parity` (with `test-utils`) 2.
- [x] **Greedy token parity across three families**: `cargo test --release --test paged_decode_v2_greedy_parity --features metal,accelerate -- --ignored --nocapture` at the default batch 4, 8192-token prompts with mixed lengths, 32 decode steps. Qwen3-0.6B, Llama-3.2-1B, and Qwen2.5-0.5B all produce token streams identical to the gather baseline on all four sequences. The test asserts the v2 arm actually built a decode plan, so a silent fallback cannot pass as parity.
- [x] `cargo clippy -p mlxcel-core --lib --tests --features metal,accelerate -- -D warnings` and `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` clean.
- [ ] **Server-level benchmark, not run here.** `scripts/benchmark_paged_decode_production.sh` drives the issue's mandatory matrix (4 concurrent clients at 1K / 4K / 16K, plus single-sequence 16K and 32K) against a freshly started server per arm, with `MLXCEL_PAGED_ATTENTION_NATIVE=0` as the before arm. It must run serialized on a quiet host; a loaded machine produces misleading numbers for a bandwidth-bound path.

## Left out, deliberately

**Logit-softcap families** fall back to gather, as the issue scopes them to a follow-up: neither the v2 kernel nor this function's fallback threads a soft-cap term. **Speculative and MTP verify steps** stay on their current path by construction rather than by a flag, because they arrive with more than one query token and the decline is on `seq_len != 1`. **Multi-slab layers** are declined rather than served; lifting that needs either a slab-aware kernel launch (one partial launch per slab, merged through the existing variable-length merge, which the merge kernel already supports) or a different pool growth strategy, and both are larger than this issue. **CUDA is still unvalidated**: the v2 CUDA JIT bodies from #898 have never been compiled or run, and `DEFAULT_TARGET_CTAS` still needs a GB10 pass.

## Where I was unsure

The **slab sizing policy** is the judgement call with the widest blast radius. `ceil(per_slot_ctx / block_size) * batch` is honest (it is the KV the config already reserved) but it is eager, and I could not measure the allocation-timing effect. A reviewer who prefers laziness over fused-path eligibility should look here first; `MLXCEL_PAGED_SLAB_BLOCKS=0` restores the old behaviour exactly.

The **1.08x cell at batch 1 / 4096** sits on the floor with the least margin of any win in the table. If the server benchmark shows single-sequence 4K regressing, raising `MIN_TOTAL_KV_TOKENS` to 8192 gives up only that cell.

`resolve_paged_slab_blocks` reads the per-slot context from `sched_config.max_kv_size`, which is `None` when `--ctx-size` is unset; it then falls back to `DEFAULT_CTX_LEN` (8192). I did not find a more direct per-slot context figure on the worker, so a server started without `--ctx-size` sizes its slab for 8192 tokens per slot and returns longer sequences to the gather path.

Closes #899


inureyes:
## Root cause of the null benchmark, with evidence

Two independent causes, not one. Both are fixed and both are now verified on a live server.

### 1. The single-sequence decode path was never wired (the one I missed)

`dispatch_sync_decode` sends a batch of one to `decode_single_step`, which calls `forward_with_sequence_id` and therefore the model's single-sequence `Attention::forward`. My change only touched `forward_split_attention`, the batched path. So `batch1_ctx16k` and `batch1_ctx32k` could not reach the fused kernel **by construction**, and neither could any moment in a batched run where only one request was still decoding, which with staggered prefills is most of a scenario. Your `server_after.log` shows exactly that staggering: the batch4_ctx16k prefills complete at 4343ms, 8351ms, 12535ms and 16537ms, roughly 4 seconds apart, against a decode phase far shorter than that gap.

Fixed by calling the same whole-batch entry point with a batch of one from `Attention::forward` in llama3 and qwen3. The decode tensors there are already `[1, H, 1, D]`, which is the shape it serves.

### 2. The floor had the wrong shape, and 4096 does not transfer

You asked me to say plainly whether the floor as specified excludes the scenarios the benchmark measures. **It does, and the formulation was wrong rather than the number.**

`seq_lens` does hold what I assumed, and the single-slab check was not firing: your log shows `Paged KV slab size: 4096 blocks per layer`, and the largest scenario needs about 1732 rows. What went wrong is the shape of the rule. Re-read the #898 ctx-1024 column: 0.91x at batch 1, 1.41x at batch 4, 1.47x at batch 8. Same per-request context, same plan degeneracy, opposite outcome. **The loss is a property of batch 1, not of total tokens.** A total-token floor separates those two cells only by accident, and with no margin: a batched launch at 1024 tokens per request sits exactly on 4096. Your run delivered 956 tokens per request, `4 x 956 = 3824`, and was declined despite being the same shape as the measured 1.41x win.

The floor is now stated the way the measurements are:

| launch | floor | evidence |
|---|---|---|
| one request | 4096 visible tokens | 1024 loses (0.91x), 4096 wins (1.08x) |
| more than one | 512 visible tokens per request | 1024 per request wins at batch 4 (1.41x) and batch 8 (1.47x) |

512 rather than 1024 for the batched case is deliberate: 1024 is the *lowest measured* batched context and it wins comfortably, so putting the floor on top of a measured point declines real workloads landing just under it for no measured reason, which is precisely what happened. Batch 2 and 3 are interpolated, not measured; the trend across batch 1, 4 and 8 is monotone and the mechanism (a batched launch spreads the same chunk count over more requests and amortizes the merge) explains why, so the interpolation is stated in the module docs rather than hidden.

### 3. Why you were debugging blind

You are right that the diagnostics were unusable. They are now a returned `PagedDecodeOutcome` carrying the numbers behind the decision, announced by the caller at **info**, one line per distinct outcome kind. Per kind matters: a single global one-shot reports only whatever happened first, which in a real server is a warmup request, so a later permanent decline for a different reason never surfaces. That is how a full sweep ran on the gather path in silence.

I could not reproduce a broken `RUST_LOG`; `EnvFilter::try_from_default_env()` in `initialize_server_logging` should honour `mlxcel_core=debug` and there is no `max_level` feature compiling the macros out. Rather than keep chasing it, the diagnostics no longer depend on it.

## Verification on a real server

`mlxcel-server`, Llama-3.2-1B-Instruct-4bit, `--parallel 4 --ctx-size 131072`, **no `RUST_LOG` set**:

```text
# 4 concurrent clients, nominal 1K prompt
INFO decode_step{batch_size=4}: mlxcel_core::cache::paged_batch_decode:
  paged decode v2: fused v2 launch (batch 4, 3828 visible KV tokens, 16 chunks, merge on)

# 1 client, 16K prompt (fresh process)
INFO decode_step{batch_size=1}: mlxcel_core::cache::paged_batch_decode:
  paged decode v2: fused v2 launch (batch 1, 13845 visible KV tokens, 16 chunks, merge on)

# same binary, MLXCEL_PAGED_ATTENTION_NATIVE=0
INFO decode_step{batch_size=4}: mlxcel_core::cache::paged_batch_decode:
  paged decode v2: gather: pinned by MLXCEL_PAGED_ATTENTION_NATIVE
```

3828 tokens at batch 4 is the exact shape the old floor declined, and batch 1 at 13845 is the path that did not exist. The kill-switch arm never logs a fused launch, so the two arms are now distinguishable from the log alone.

I did not run a timing sweep. I will note one incidental number only so you know the sweep is worth your time, with the caveat that it is a single unreplicated run on a machine that had just finished a build and is **not a measurement**: at 4 clients on a nominal 1K prompt the fused run reported 75.4 mean decode tok/s against 41.0 for the kill-switch run. Treat that as a smoke signal, nothing more.

## The harness now refuses to produce a null sweep

`scripts/benchmark_paged_decode_production.sh` checks each arm after its scenarios and exits non-zero if the after arm never logged a fused launch, or if the before arm did. It also prints the first few `paged decode v2:` lines per arm, so a decline names itself.

## What to run

```bash
export DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer
MODEL=~/.cache/mlxcel/models/mlx-community/Llama-3.2-1B-Instruct-4bit \
  ./scripts/benchmark_paged_decode_production.sh
```

If it aborts, the reason is on a `paged decode v2: gather: ...` line in `benchmarks/paged_decode_production_logs_<date>/server_after.log` and names what to change.

Greedy parity re-verified after the `Attention::forward` change, across three families at batch 4 with 8192-token prompts and 32 steps:

```bash
cargo test --release --test paged_decode_v2_greedy_parity --features metal,accelerate -- --ignored --nocapture
```

Qwen3-0.6B, Llama-3.2-1B and Qwen2.5-0.5B all identical to the gather baseline on all four sequences.

Issue #973, opened 2026-07-30
fix(core): bound the declared quantization mode string, not just group_size and bits


inureyes:
## Background

PR #972 (issue #958, merged as `db6cce9dc`) bounded the declared `quantization` parameters at the point they are stored. It added `mlxcel_core::layers::validate_quantization_params(group_size, bits)` (`src/lib/mlxcel-core/src/layers.rs:1292`) and called it from the shared loaders plus roughly twenty family-local expert-plane constructors, and it made `QuantizedEmbedding::new` and `QuantizedMultiLinear::new` return `Result`. The defect class it exists to close is this one: a config-derived value passes every Rust-side check, violates an undocumented MLX precondition, and terminates the process at the FIRST forward pass rather than failing at load, because `quantized_matmul`, `gather_qmm`, `dequantize` and `quantized_embedding` cross the cxx bridge as `UniquePtr<MlxArray>` rather than `Result` (`src/lib/mlxcel-core/src/lib.rs:1092`, `:1108`, `:1121`, `:1136`), so the C++ throw is an uncatchable `std::terminate` that `catch_unwind` cannot contain. This issue is the third value in the same `quantization` block, the `mode` string, which #972 deliberately left out of scope. The body of PR #972 records that both review passes on it raised the unbounded `mode` independently, which is why it was filed rather than folded in.

## What is unbounded

`validate_quantization_params` bounds `bits` to `1..=32` and `group_size` to `1..=MAX_QUANT_GROUP_SIZE` (`1 << 20`, `src/lib/mlxcel-core/src/layers.rs:1419`). The `mode` string is bounded in exactly one place in the tree, and that place is family-local: `SUPPORTED_QUANT_SCHEMES` (`src/models/gemma4.rs:3741`) is `["affine", "mxfp4", "nvfp4", "mxfp8"]`, checked by `validate_quantization_scheme` (`src/models/gemma4.rs:3768`) and wired into gemma4's own two loaders only (`src/models/gemma4.rs:3868` and `src/models/gemma4.rs:4134`). That function also inspects `quant_method`, which is the #467 external-format guard and a superset of what a core mode check needs. Nothing in `mlxcel-core` bounds the mode at all.

## What MLX accepts

[...]

PR #989, opened 2026-07-31
fix(models): read the nested quantization block in deepseek::ModelArgs

### LINKED ISSUE #975: fix(models): deepseek::ModelArgs silently drops an inherited quantization block
## Background

PR #972 (issue #958, merged as `db6cce9dc`) bounded the declared `quantization` parameters where they are stored, adding `mlxcel_core::layers::validate_quantization_params(group_size, bits)` (`src/lib/mlxcel-core/src/layers.rs:1292`) and calling it at the shared loaders plus roughly twenty family-local expert-plane constructors, and making `QuantizedEmbedding::new` and `QuantizedMultiLinear::new` return `Result`. The defect class it exists to close: a config-derived value passes every Rust-side check, violates an undocumented MLX precondition, and terminates the process at the FIRST forward pass rather than failing at load, because `quantized_matmul` and `gather_qmm` cross the cxx bridge as `UniquePtr<MlxArray>` rather than `Result` (`src/lib/mlxcel-core/src/lib.rs:1092`, `:1108`), so the C++ throw is an uncatchable `std::terminate` that `catch_unwind` cannot contain.

This issue is the adjacent problem on the same code: #972 bounded the values the loaders store, but on one family the values reaching those loaders are not the ones the checkpoint declared. The declared `quantization` block is silently discarded and hardcoded defaults are used instead.

## Which path inherits from where

This is a nested-sub-config inheritance case in the DeepSeek-OCR VLM loaders, plus the plain text loader for the same config struct. It is not a `text_config` case and not a shared-args conversion.

`src/models/deepseek.rs:35` declares `ModelArgs` with two flat, top-level optional fields and no nested block:

```rust
#[serde(default)]
pub group_size: Option<i32>,

#[serde(default)]
pub bits: Option<i32>,
```

(`:70` and `:73`, read through `group_size()` at `:97` and `bits()` at `:101`, which fall back to `64` and `4`.) There is no `quantization` field on the struct and no `#[serde(deny_unknown_fields)]`.

The three DeepSeek-OCR loaders in `src/loading/vlm_deepseekocr.rs` inherit the root `quantization` OBJECT into `language_config` before deserializing into that struct:

```rust
if !obj.contains_key("quantization")
    && let Some(q) = full_config.get("quantization")
{
    obj.insert("quantization".to_string(), q.clone());
}
```

then

```rust
let args: models::deepseek::ModelArgs = serde_json::from_value(lc)?;
let (gs, bits) = (args.group_size(), args.bits());
```

The three sites are `load_deepseekocr_vlm` (insert at `:78`, deserialize at `:84`, read the pair at `:86`), the Unlimited-OCR loader (`:152` / `:158` / `:160`), and the DeepSeek-OCR 2 loader (`:245` / `:251` / `:253`).

Serde ignores the unknown `quantization` key, so the nested object is discarded on deserialization. `group_size` and `bits` stay `None`, and `group_size()` / `bits()` return the hardcoded `64` / `4` regardless of what the checkpoint declares. The inheritance code runs, succeeds, and accomplishes nothing.

`DeepSeekModel::load` (`src/models/deepseek.rs:602`) has the same shape for a plain `deepseek` `config.json`: it deserializes the whole file into `ModelArgs` at `:609`, so a normal MLX-style nested `"quantization": {"group_size": N, "bits": M}` block is dropped there too. Only a checkpoint that happens to carry flat top-level `group_size` / `bits` keys is read correctly, and no MLX conversion tool emits that spelling.

## The contrast that shows this is a struct bug, not a loader bug

`src/loading/vlm_deepseek_vl2.rs:91` performs the byte-identical inheritance and then deserializes into `models::deepseek_v2::ModelArgs` at `:97`. That one works, because `deepseek_v2::ModelArgs` has a real nested field: `pub quantization: Option<Quantization>` at `src/models/deepseek_v2.rs:114`, backed by `pub struct Quantization { group_size, bits }` at `:118`, with accessors at `:181` and `:188`. The two loaders look the same at the call site and behave differently purely because of the config struct they target.

## Consequence, at load and at first forward

The declared values propagate to four places in this family: the embedding at `src/models/deepseek.rs:624`, the MoE expert planes through `SwitchLinear::from_stacked_parts` (which reads the pair at `:887`), every dense projection through `load_quantized_linear` at `:916`, and the DeepSeek-OCR projector at `src/loading/vlm_deepseekocr.rs:94`. The consequence of substituting `64` / `4` for the declared values differs by path, and both halves matter:

- Dense projections, the embedding, and the OCR projector go through `reconcile_quantization_layout` (`src/lib/mlxcel-core/src/layers.rs:1467`). For affine weights it trusts `group_size` and re-derives `bits` from the tensor shapes. A wrong `bits` is therefore corrected, with only a `tracing::warn!` on the `mlxcel::quant` target (`:1587`). A wrong `group_size` is worse: `infer_quantization_bits` (`:1155`) solves `packed_in * 32 / (num_groups * group_size)` and, for a real group_size-32 4-bit tensor read as group_size 64, lands on `bits = 2`, which is in the accepted `{2,3,4,5,6,8}` set. The resulting `(64, 2)` pair is self-consistent, so MLX's own shape check passes and the kernel runs. The output is silently wrong, with a warning line and no load failure.
- MoE expert planes take no such path. `SwitchLinear::from_stacked_parts` stores the pair verbatim and hands it to `gather_qmm`, so a `group_size` that does not match the tensor fails MLX's `packed_in * 32 == bits * num_groups * group_size` check and aborts uncatchably at the first routed forward pass.

Neither outcome is a load error, which is the point: nothing in the pipeline reports that the declared block was ignored.

## Impact today is latent

Every DeepSeek-family checkpoint currently under `models/` declares exactly the fallback values, so nothing misbehaves right now. `models/deepseek-ocr-4bit` (`model_type: deepseekocr`) and `models/deepseek-ocr-2-4bit` (`model_type: deepseekocr_2`) both declare root `{"group_size": 64, "bits": 4, "mode": "affine"}` and carry no `language_config.quantization`, so both land on `64` / `4` by coincidence. There is no local `model_type: deepseek` checkpoint at all, so `DeepSeekModel::load` is unexercised. An 8-bit export, or a `group_size` of 32 or 128, would be mis-read on every one of these paths.

## Adjacency to #991, flagged rather than asserted

#991 reports that `deepseek_v2` generates degenerate output on `models/deepseek-v2-lite-4bit` while `mlx-lm` is correct on the same weights, and that investigation established `models/deepseek-vl2-small-4bit` runs the same `DeepSeekV2Model::from_weights` correctly. Neither of those checkpoints reaches the code in this issue. `models/deepseek-v2-lite-4bit` is `model_type: deepseek_v2` and loads through `deepseek_v2::ModelArgs`; `models/deepseek-vl2-small-4bit` is `model_type: deepseek_vl_v2` and loads through `src/loading/vlm_deepseek_vl2.rs`, which also targets `deepseek_v2::ModelArgs`. Both therefore read their nested `quantization` block correctly, via the field at `src/models/deepseek_v2.rs:114`. On the evidence available, this issue and #991 are separate defects in sibling modules and do not interact. That is worth re-checking rather than assuming, because the two modules are near-duplicates and a fix here that touches shared config plumbing could easily reach `deepseek_v2`.

## Proposed Solution

Give `deepseek::ModelArgs` the nested field the loaders already assume it has.

1. Add `#[serde(default)] pub quantization: Option<...>` to `src/models/deepseek.rs:35`.
2. Change `group_size()` (`:97`) and `bits()` (`:101`) to prefer the nested block, keeping the flat `group_size` / `bits` fields as a fallback so anything relying on the current spelling keeps loading.
3. Prefer the existing shared type over adding another private copy. `models::config::QuantizationArgs` (`src/models/config.rs:244`) already exists with `Option<i32> group_size`, `Option<i32> bits`, `Option<String> mode`, and `get_group_size()` / `get_bits()` / `get_mode()` accessors that default to `64` / `4` / `"affine"`. It currently has zero production callers. There are 59 near-identical private `pub struct Quantization` declarations under `src/models/`, and the divergence between `deepseek.rs` and `deepseek_v2.rs` is precisely what produced this bug, so adding a sixtieth is the wrong move. Note that `get_mode()` returns the declared string unbounded, which is #973; adopting `QuantizationArgs` here should not silently widen that surface.
4. Add a deserialization test that feeds a `language_config` with an inherited nested `quantization` block through `deepseek::ModelArgs` and asserts the declared values survive, so the inheritance in `src/loading/vlm_deepseekocr.rs` has a regression gate.
5. Consider whether `DeepSeekModel::load` should also reject a `quantization` block it cannot interpret, rather than falling through to defaults. Silence is the failure mode here, so an explicit error is worth more than a correct default.

## Inspection and reproduction

Confirm the drop without building anything. This prints the deserialized struct's view of an inherited block:

```bash
python3 - <<'PY'
import json
# What the loader builds and hands to serde, for the OCR-2 checkpoint.
full = json.load(open("models/deepseek-ocr-2-4bit/config.json"))
lc = dict(full["language_config"])
lc.setdefault("quantization", full["quantization"])
print("inherited into language_config:", lc["quantization"])
print("flat keys serde will actually read:", {k: lc.get(k) for k in ("group_size", "bits")})
PY
```

Expected output: the nested block is present, and both flat keys are `None`, so `group_size()` / `bits()` return `64` / `4`.

Confirm the struct has no nested field and no unknown-field rejection:

```bash
grep -n "pub struct ModelArgs" -A 45 src/models/deepseek.rs | grep -n "quantization\|group_size\|bits\|deny_unknown_fields"
# Contrast with the sibling that works.
grep -n "pub quantization: Option<Quantization>" -A 6 src/models/deepseek_v2.rs
```

Demonstrate the mis-read on a real checkpoint. `models/deepseek-ocr-4bit` declares the fallback values, so edit a copy to declare something else and observe that the output does not change, which it should. Symlink rather than copy to avoid duplicating the weights:

```bash
mkdir -p /tmp/deepseek-ocr-8bitdecl
for f in /Volumes/Monolith/Development/rust/mlxcel/models/deepseek-ocr-4bit/*; do ln -sf "$f" /tmp/deepseek-ocr-8bitdecl/; done
rm /tmp/deepseek-ocr-8bitdecl/config.json
python3 - <<'PY'
import json
c = json.load(open("/Volumes/Monolith/Development/rust/mlxcel/models/deepseek-ocr-4bit/config.json"))
c["quantization"] = {"group_size": 32, "bits": 8, "mode": "affine"}
json.dump(c, open("/tmp/deepseek-ocr-8bitdecl/config.json", "w"), indent=2)
PY
DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer \
./target/release/mlxcel generate -m /tmp/deepseek-ocr-8bitdecl -p "Hello" -n 8
# The `mlxcel::quant` warning only surfaces through a tracing subscriber, and the CLI installs none.
# Only `mlxcel-server` does (`src/server/startup.rs:1009`), so use it when the warning is the signal:
RUST_LOG=mlxcel::quant=warn ./target/release/mlxcel-server -m /tmp/deepseek-ocr-8bitdecl
```

Expected today: identical behavior to the unedited checkpoint, and no reconciliation warning from the server either, because the declared block never reaches the loader. Expected after the fix: the declared `(32, 8)` reaches `reconcile_quantization_layout`, which either warns and reconciles against the real shapes or fails at load, and either way the change is visible. That the CLI has no subscriber at all is part of why this defect is silent: even when the reconciler does fire, the only user-facing surface that reports it is the server.

## Local validation: possible and not possible

Possible here. The quantized MoE checkpoints available under `models/` in the repo root are `gpt-oss-20b-mxfp4`, `deepseek-v2-lite-4bit`, `qwen3-30b-a3b-4bit`, `qwen3-moe-4bit`, `mixtral-8x7b-4bit`, `jamba-v0.1-4bit`, `nemotron-h-30b-4bit` and `llama-4-scout-17b-4bit`. For this issue the relevant ones are `models/deepseek-ocr-4bit` and `models/deepseek-ocr-2-4bit`, the only two local checkpoints that reach `deepseek::ModelArgs`. Both are regression controls: because they declare exactly the fallback pair, their OCR output must be byte-identical before and after the fix, which is the check that the new nested field does not change the values any existing checkpoint loads with. The deserialization behavior itself is fully testable locally as a unit test, with no model on disk.

Not possible here. No local checkpoint declares anything other than `64` / `4` on this path, and there is no local `model_type: deepseek` checkpoint at all, so the positive case (a declared block that differs from the fallback is honoured end to end on real weights) cannot be validated against a genuine artifact. It has to be driven either by an edited config over symlinked weights, as above, or by a unit test. `models/deepseek-v2-lite-4bit` is not usable as a cross-family control right now: it produces degenerate output for an unrelated reason tracked as #991.

## Machine notes

- Pass `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer` on every cargo invocation. Without it cmake fails with `xcrun: error: unable to find utility "metal"`.
- Only the release profile is warm (`target/release` exists, `target/debug` does not). A plain debug `cargo check` triggers a cold MLX C++ build. Use `cargo check --release --features metal,accelerate` and `cargo test --release`.
- A test that actually trips an MLX C++ throw aborts the whole test binary with SIGABRT rather than failing cleanly, because the bridge functions return `UniquePtr<MlxArray>` and cannot propagate the exception. Keep the tests here at the deserialization and load-`Result` level and never run a forward pass in them.

## Acceptance Criteria

- [ ] `deepseek::ModelArgs` deserializes a nested `"quantization": {"group_size": N, "bits": M}` block and `group_size()` / `bits()` return `N` / `M`.
- [ ] All three DeepSeek-OCR loaders in `src/loading/vlm_deepseekocr.rs` see the inherited root block instead of falling back to `64` / `4`, verified at each of the three sites rather than only the first.
- [ ] `DeepSeekModel::load` reads a nested block from a plain `deepseek` `config.json`.
- [ ] A config carrying the flat top-level `group_size` / `bits` keys still loads with those values, so the change is backward compatible with whatever spelling the current fields exist for.
- [ ] A `language_config` that declares its own `quantization` block still wins over the inherited root block, matching the `!obj.contains_key("quantization")` guard the loaders already have.
- [ ] A unit test covers the inheritance path end to end (nested block inserted into `language_config`, deserialized through `deepseek::ModelArgs`, values asserted), not just the struct in isolation.
- [ ] No sixtieth private `Quantization` struct is added. Either `models::config::QuantizationArgs` is adopted or the reason it was not is recorded in the code.
- [ ] `models/deepseek-ocr-4bit` and `models/deepseek-ocr-2-4bit` produce byte-identical OCR output before and after, since both declare the fallback values.
- [ ] Whether this interacts with #991 is checked and stated, rather than assumed either way, before the change lands.
- [ ] `cargo clippy --release --lib --tests --features metal,accelerate -- -D warnings` is clean and `cargo fmt --all -- --check` passes.

## Refs

Follow-up to #958 (PR #972, merged as `db6cce9dc`), which bounded the values the family-local expert loaders store; this is the adjacent problem that the values reaching those loaders are not the ones the checkpoint declared. Adjacent, not established as related: #991. Sibling follow-ups from the same trace: #973 and #974.

rapsealk:
## Summary

`deepseek::ModelArgs` silently dropped the MLX-style nested `"quantization": {"group_size": .., "bits": ..}` block — the struct only declared flat top-level `group_size` / `bits` fields, so the inheritance the three DeepSeek-OCR loaders perform in `vlm_deepseekocr.rs` accomplished nothing and the accessors always fell back to `64` / `4`. This adds the nested field (reusing `deepseek_v2::Quantization`, whose presence is why the byte-identical inheritance in `vlm_deepseek_vl2.rs` already worked) and makes `group_size()` / `bits()` prefer nested-then-flat-then-default.

## Related issues

Closes #975

## Type of change

- [x] `fix` — bug fix

## Test plan

- New regression tests in `src/models/deepseek_tests.rs`:
  - `model_args_reads_inherited_nested_quantization_block` — drives the same `language_config` insert the OCR loaders perform (including the extra `"mode"` key real checkpoints carry) and asserts the declared pair survives deserialization.
  - `model_args_reads_flat_quantization_keys` — the flat spelling keeps loading with its declared values.
  - `model_args_nested_quantization_beats_flat_keys` — pins the documented precedence.
  - `model_args_defaults_without_quantization_declaration` — no declaration keeps 64 / 4, which is what the two in-tree DeepSeek-OCR checkpoints rely on.
- [x] `cargo clippy --all-targets --features metal,accelerate -- -D warnings`
- [x] `cargo test --profile test-fast --features metal,accelerate models::deepseek` / `vision::unlimited_ocr` — 7 + 3 passed (full `--release` suite not run on this machine; see notes)
- [ ] Validated with a real checkpoint: not run — no Metal toolchain on this machine (CPU-only build via `MLXCEL_BUILD_METAL=OFF`). Both in-tree OCR checkpoints declare exactly `{group_size: 64, bits: 4}`, i.e. the fallback values, so the resolved pair — and therefore OCR output — is unchanged for them by construction. Happy to run the confirmation if a maintainer prefers it before merge.

## Notes for reviewers

- `DeepSeekModel::load` needs no change: it deserializes the whole `config.json` into `ModelArgs`, so a plain `deepseek` checkpoint's nested block is now picked up through the same field.
- Issue #975's optional item 3 (factoring the ~dozen private `Quantization` copies under `src/models/` into one shared struct) is deliberately left out — that's a cross-family refactor and a separate PR per the one-logical-change rule. Reusing `deepseek_v2::Quantization` here at least stops these two siblings from diverging again.
- One nuance inherited from `deepseek_v2`: `Quantization`'s fields are required, so a `quantization` block missing `group_size`/`bits` now fails deserialization instead of being ignored. That matches `deepseek_v2`'s existing behavior for the same block and MLX exports always write both keys.

## Checklist

- [x] PR title uses a Conventional Commits prefix
- [x] One logical change per PR
- [ ] Updated `docs/` — not needed (no user-facing behavior change for in-tree checkpoints)
- [ ] Updated `// Used by: ...` comments — no shared function modified

### DIFF
--- a/src/models/deepseek.rs
+++ b/src/models/deepseek.rs
@@ -22,6 +22,7 @@
 //! - routed_scaling_factor for expert outputs (default 1.0)
 //! - Top-k routing with softmax scoring
 
+use crate::models::deepseek_v2::Quantization;
 use crate::models::switch_layers::validate_expert_quantization_params;
 use mlxcel_core::generate::LanguageModel;
 use mlxcel_core::layers::{KVCache, RMSNorm, UnifiedEmbedding, UnifiedLinear};
@@ -66,6 +67,15 @@ pub struct ModelArgs {
     #[serde(default)]
     pub attention_bias: bool,
 
+    // MLX-style nested `"quantization": {"group_size": .., "bits": ..}` block:
+    // present in a plain `deepseek` config.json and inherited into
+    // `language_config` by the DeepSeek-OCR loaders (vlm_deepseekocr.rs).
+    // Takes precedence over the flat keys below.
+    #[serde(default)]
+    pub quantization: Option<Quantization>,
+
+    // Flat top-level spelling, kept as a fallback for checkpoints that
+    // declare `group_size` / `bits` directly on the (sub-)config root.
     #[serde(default)]
     pub group_size: Option<i32>,
 
@@ -95,11 +105,19 @@ impl ModelArgs {
     }
 
     pub fn group_size(&self) -> i32 {
-        self.group_size.unwrap_or(64)
+        self.quantization
+            .as_ref()
+            .map(|q| q.group_size)
+            .or(self.group_size)
+            .unwrap_or(64)
     }
 
     pub fn bits(&self) -> i32 {
-        self.bits.unwrap_or(4)
+        self.quantization
+            .as_ref()
+            .map(|q| q.bits)
+            .or(self.bits)
+            .unwrap_or(4)
     }
 }
 
--- a/src/models/deepseek_tests.rs
+++ b/src/models/deepseek_tests.rs
@@ -40,6 +40,7 @@ fn test_args(n_routed_experts: Option<usize>) -> ModelArgs {
         first_k_dense_replace: 0,
         routed_scaling_factor: 1.0,
         attention_bias: false,
+        quantization: None,
         group_size: None,
         bits: None,
     }
@@ -99,9 +100,10 @@ fn switch_linear_accepts_full_expert_count() {
     assert_eq!(sl.num_experts(), 4);
 }
 
-/// The same minimal config with an explicit quantization pair. DeepSeek v1
-/// declares `group_size` / `bits` as flat top-level keys rather than a nested
-/// `quantization` block, so those are the two fields the guard reads.
+/// The same minimal config with an explicit quantization pair, in the flat
+/// top-level spelling. DeepSeek v1 accepts both a nested `quantization` block
+/// (preferred) and flat `group_size` / `bits` keys; the guard reads the
+/// resolved pair either way.
 fn quant_args(group_size: i32, bits: i32) -> ModelArgs {
     let mut args = test_args(None);
     args.group_size = Some(group_size);
@@ -169,4 +171,85 @@ fn deepseek_switch_linear_rejects_quantization_params_that_would_abort_gather_qm
     );
     SwitchLinear::from_weights(&regular, &quant_args(0, 0), PREFIX)
         .expect("a bf16 expert plane must not be gated on quantization params");
+}
+
+/// Minimal `language_config` JSON in the shape the DeepSeek-OCR checkpoints
+/// ship: required structural fields only, no quantization declaration.
+fn minimal_language_config() -> serde_json::Value {
+    serde_json::json!({
+        "model_type": "deepseek",
+        "vocab_size": 8,
+        "hidden_size": 8,
+        "intermediate_size": 8,
+        "num_hidden_layers": 0,
+        "num_attention_heads": 1,
+        "num_key_value_heads": 1,
+        "max_position_embeddings": 16,
+    })
+}
+
+/// #975: the DeepSeek-OCR loaders (`src/loading/vlm_deepseekocr.rs`) inherit
+/// the root `quantization` object into `language_config` before deserializing
+/// into `ModelArgs`. The struct used to lack the nested field, so serde
+/// silently dropped the block and `group_size()` / `bits()` fell back to
+/// 64 / 4 regardless of what the checkpoint declared. This drives the same
+/// insert the loaders perform and asserts the declared pair survives.
+#[test]
+fn model_args_reads_inherited_nested_quantization_block() {
+    let mut lc = minimal_language_config();
+    let full_config = serde_json::json!({
+        "quantization": {"group_size": 32, "bits": 8, "mode": "affine"}
+    });
+    let obj = lc.as_object_mut().unwrap();
+    if !obj.contains_key("quantization")
+        && let Some(q) = full_config.get("quantization")
+    {
+        obj.insert("quantization".to_string(), q.clone());
+    }
+
+    let args: ModelArgs = serde_json::from_value(lc).expect("language_config must deserialize");
+    assert_eq!(args.group_size(), 32);
+    assert_eq!(args.bits(), 8);
+}
+
+/// A checkpoint that spells the pair as flat top-level keys must keep loading
+/// with those values.
+#[test]
+fn model_args_reads_flat_quantization_keys() {
+    let mut lc = minimal_language_config();
+    let obj = lc.as_object_mut().unwrap();
+    obj.insert("group_size".to_string(), serde_json::json!(128));
+    obj.insert("bits".to_string(), serde_json::json!(8));
+
+    let args: ModelArgs = serde_json::from_value(lc).expect("flat spelling must deserialize");
+    assert_eq!(args.group_size(), 128);
+    assert_eq!(args.bits(), 8);
+}
+
+/// When both spellings are present the nested block wins, matching the
+/// documented precedence on `ModelArgs`.
+#[test]
+fn model_args_nested_quantization_beats_flat_keys() {
+    let mut lc = minimal_language_config();
+    let obj = lc.as_object_mut().unwrap();
+    obj.insert(
+        "quantization".to_string(),
+        serde_json::json!({"group_size": 32, "bits": 8}),
+    );
+    obj.insert("group_size".to_string(), serde_json::json!(128));
+    obj.insert("bits".to_string(), serde_json::json!(4));
+
+    let args: ModelArgs = serde_json::from_value(lc).expect("mixed spelling must deserialize");
+    assert_eq!(args.group_size(), 32);
+    assert_eq!(args.bits(), 8);
+}
+
+/// No declaration at all keeps the family defaults, which is what the two
+/// in-tree DeepSeek-OCR checkpoints rely on.
+#[test]
+fn model_args_defaults_without_quantization_declaration() {
+    let args: ModelArgs = serde_json::from_value(minimal_language_config())
+        .expect("undeclared quantization must deserialize");
+    assert_eq!(args.group_size(), 64);
+    assert_eq!(args.bits(), 4);
 }
--- a/src/vision/unlimited_ocr_tests.rs
+++ b/src/vision/unlimited_ocr_tests.rs
@@ -170,6 +170,7 @@ fn build_wrapper() -> UnlimitedOcrVlModel {
         first_k_dense_replace: 0,
         routed_scaling_factor: 1.0,
         attention_bias: false,
+        quantization: None,
         group_size: None,
         bits: None,
     };


inureyes:
Thanks for catching this, the diagnosis here is exactly right and matches what #975 describes.

I have opened #1016 for the same issue and it supersedes this PR, so I wanted to explain rather than leave it hanging. Three things pushed it toward a separate branch instead of building on this one:

- **The shared type.** #975 asks specifically that no sixtieth private `Quantization` struct be added, and that `models::config::QuantizationArgs` either be adopted or the reason recorded in code. This PR reuses `deepseek_v2::Quantization`, which is reasonable but keeps the family-local copies growing. #1016 adopts the shared type, which also removes its "no production callers yet" status.
- **Optional fields and the declared mode.** `deepseek_v2::Quantization` has required `group_size` / `bits` and no `mode`, so a partially declared block becomes a hard deserialization failure (as your PR body notes) and the declared mode stays invisible to this family. `QuantizationArgs` keeps all three optional and its `get_mode()` runs the allowlist added in #973 / #1006, both of which landed after this branch's base. #1016 uses that to refuse a block-float mode at load, which this affine-only family would otherwise reinterpret.
- **Verifying all three loaders.** The acceptance criteria ask for the inheritance to be checked at each of the three DeepSeek-OCR sites. #1016 collapses the triplicated `language_config` fixup into one helper and tests it once per loader.

It is also a fork branch, so completing it in place would have meant force-pushing over your work.

For what it is worth, the real-checkpoint runs confirmed your reasoning about the in-tree checkpoints: `models/deepseek-ocr-4bit` and `models/deepseek-ocr-2-4bit` both declare exactly `{group_size: 64, bits: 4}`, and their OCR output is byte-identical before and after.

inureyes:
Closing as superseded by #1016, which merged as `a86e6340a`. The diagnosis in this PR was correct and is credited in the comment above; the separate branch was needed because this one predates #1006 (the quantization-mode allowlist) and because completing it in place would have meant force-pushing over your fork. Thanks for finding it.

Issue #1007, opened 2026-08-02
fix(ci): make verify and the nightly test only one of the four workspace members


inureyes:
## Problem / Background

Everything below was measured directly on 2026-08-03 at `main` head `b06cb3872`.

`Makefile` target `verify-test` runs `cargo test --release --features metal,accelerate`, `.github/workflows/nightly-verify.yml` runs `make verify`, and `CONTRIBUTING.md` points contributors at the same target. All three therefore share one blind spot.

`Cargo.toml` declares four workspace members and no `default-members`:

```toml
[workspace]
members = [".", "src/lib/mlxcel-core", "src/lib/mlxcel-surgery", "src/lib/mlxcel-xla"]
resolver = "2"
```

Because the workspace root is itself the `mlxcel` package, a bare `cargo test` there resolves to `cargo test -p mlxcel` and never touches the other three members. Verified:

```
$ cargo test --release --features metal,accelerate --no-run 2>&1 | grep -c 'mlxcel-core'
0
```

Zero `mlxcel-core` test targets are built by the command the gate runs.

### What is not being tested

| member | tests | in the gate? |
|---|---|---|
| `mlxcel` (root) | 5238 | yes |
| `mlxcel-core` | 1354 | **no** |
| `mlxcel-surgery` | 146 | **no** |
| `mlxcel-xla` | 254 | **no** |

`mlxcel-core` is the crate that holds the `cxx` bridge to MLX, `layers.rs`, the KV cache, the quantization loaders and the speculative decoding round loop. It is the least covered and the most consequential.

### It is currently red

`cargo test --release -p mlxcel-core --features metal,accelerate --no-fail-fast` fails on `main`:

```
---- fused_moe_parity_tests::fused_moe_geglu_kernel_matches_references_gemma4_shape stdout ----
fused_moe_geglu parity [metal, seed 886]: vs dense f32 (bf16-rounded) nrms=0e0 nmax=0e0; vs gather_qmm nrms=1.1697728227157964e-2 nmax=4.522727947528232e-2; gather-vs-dense baseline nrms=1.1574492404669066e-2
panicked at src/lib/mlxcel-core/src/fused_moe_parity_tests.rs:369:9
test result: FAILED. 1353 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

[...]

Issue #964, opened 2026-07-30
fix(tests): fused MoE GeGLU parity threshold is tighter than the gather_qmm reference's own jitter


inureyes:
## Summary

`fused_moe_parity_tests::fused_moe_geglu_kernel_matches_references_gemma4_shape` fails deterministically on `main`. The failure is not a kernel defect: the assertion threshold is tighter than the reference implementation's own numerical jitter, which makes it unsatisfiable regardless of how correct the fused kernel is.

## Reproduction

On `main` at commit `14412c13`, Apple M1 Ultra (macOS, Metal backend):

```
cargo test -p mlxcel-core --release --features metal,accelerate
```

Result: `1050 passed; 1 failed`. The failing test is `fused_moe_parity_tests::fused_moe_geglu_kernel_matches_references_gemma4_shape`.

This is not a flake. The test iterates `for seed in [886u64, 887, 888]` and calls `random_seed(seed)`, so the inputs are fully deterministic. It fails on the first seed, 886.

## Measured values at seed 886

These are printed by the test's own diagnostic line:

- fused vs dense f32 (bf16-rounded): `nrms=0e0`, `nmax=0e0`
- fused vs `gather_qmm`: `nrms=1.1697728227157964e-2`, `nmax=4.522727947528232e-2`
- `gather_qmm` vs dense baseline (`nrms_ref_jitter`): `nrms=1.1574492404669066e-2`

## Failing assertion

`src/lib/mlxcel-core/src/fused_moe_parity_tests.rs`, the second assertion in the test (around line 369):

```rust
assert!(
    nrms_gather < 5e-3 && nmax_gather < 5e-2,
    "fused vs gather_qmm reference deviates (seed {seed}): nrms={nrms_gather:e} nmax={nmax_gather:e}"
);
```

The `nrms_gather < 5e-3` bound is the one that fails: actual `1.1698e-2`, roughly 2.3x over the limit.

## Root cause

The threshold sits below the reference's own noise floor, so the assertion cannot be satisfied by any implementation.

[...]

PR #1017, opened 2026-08-04
fix(ci): report unfinished nightly runs and cut the test build cost

### LINKED ISSUE #1000: fix(ci): nightly-verify times out in the build phase and reports nothing when it does
## Problem / Background

`nightly-verify` has two independent defects: its `cargo test` step never finishes inside the job budget, and when the job is killed by that timeout nobody is told. All figures below were measured on 2026-08-02.

### Defect 1: the `cargo test` step never finishes inside the 180-minute budget

`.github/workflows/nightly-verify.yml` sets `timeout-minutes: 180`. A job killed by that timeout is reported by GitHub with conclusion `cancelled`, not `failure`, so in the run list it reads as "someone cancelled it". That is why it has gone unexamined.

The last four runs of the workflow:

| Run | Trigger | Created | Ended | Conclusion | Duration |
|---|---|---|---|---|---|
| 30733081983 | workflow_dispatch | 2026-08-02T04:51:14Z | 2026-08-02T07:51:39Z | cancelled | 3h00m |
| 30712399425 | schedule | 2026-08-01T18:23:19Z | 2026-08-01T21:23:48Z | cancelled | 3h00m |
| 30655703583 | schedule | 2026-07-31T18:34:43Z | (short) | failure | fail-fast at `cli_help_consistency`, issue #962 |
| 30570981898 | schedule | 2026-07-30T18:34:13Z | 2026-07-30T21:34:38Z | cancelled | 3h00m |

Three of the last four hit exactly 3h00m. The only run that produced a verdict did so because `cargo test` is fail-fast and it died early on the harness bug fixed in #996. The workflow has never completed a full pass.

Per-step timings for run 30733081983, from `gh api repos/lablup/mlxcel/actions/runs/30733081983/jobs`:

```
cargo fmt:    04:51:26 -> 04:51:30   (4s,   success)
cargo clippy: 04:51:30 -> 04:52:37   (67s,  success)
cargo test:   04:52:37 -> 07:51:28   (179m, cancelled)
```

The time goes to the build, not to the tests. The run log contains zero `Running tests/...` lines and zero `test result:` lines, and the processes the runner terminated at the timeout were `ld` and `clang`:

```
Terminate orphan process: pid (68682) (ld)
Terminate orphan process: pid (68678) (clang)
```

The 67-second clippy is the load-bearing contrast: the persistent `CARGO_TARGET_DIR` at `$HOME/.cargo-target/mlxcel` is warm. But `cargo clippy` runs `clippy-driver` and leaves metadata rather than linkable artifacts, so `cargo test` still has to codegen the library and link roughly 75 integration-test binaries, each against the large MLX static library, on a shared runner. That is where the three hours go.

For calibration, the same suite on the M1 Ultra development machine completes in well under an hour with a warm `target/`. So this is about the runner and the clippy-then-test artifact split, not about the test count being unreasonable.

### Defect 2: a timed-out run notifies nobody

The workflow has a `Report a red main` step whose purpose, stated in the file's own header comment, is that "a failed scheduled run files (or comments on) a GitHub issue", because "a red Actions run that nobody opens is the same blind spot in a new place".

In run 30733081983 that step's conclusion is `skipped`. Its condition matches failure, and a timeout surfaces as `cancelled`, so the notification path does not fire. Three silent timeouts in four days is precisely the blind spot the workflow was created to close.

### Why this matters now

The workflow exists (see its header comment) because two deterministically failing tests reached `main` and sat there unnoticed, one of them for 26 days. #962 fixed the harness bug that was breaking it, and #953 created it. Neither is sufficient while the job cannot finish and cannot report.

## Proposed Solution

These are options to evaluate and measure, not a decided design.

- Drop `cargo clippy` from the same job, or reorder so the test build is not paying for clippy having warmed a different artifact set. Measure whether running `cargo test --no-run` first, or running clippy after test, changes the total.
- Split fmt/clippy and test into separate jobs, so a slow test build does not consume the whole budget and so the three signals stay independent (the file already states it wants them independent).
- Raise `timeout-minutes`, but only with a measurement showing what a full pass actually costs. Raising it blindly trades a silent timeout for a longer silent timeout.
- Reduce link cost: fewer and larger integration-test binaries, or a `-C link-arg` / linker choice, or `cargo nextest`, which builds the same binaries but may schedule them better. Consolidating test binaries is a real refactor with its own review cost.
- Fix the notification independently of the performance work, and first, since it is small: make the reporting step fire on `cancelled` and on timeout, not only on `failure`. That alone converts a silent failure into a visible one and is worth landing even if the timeout takes longer to solve.

## Acceptance Criteria

- [ ] `nightly-verify` completes a full pass within its budget on the self-hosted runner, demonstrated by at least one green run, with the measured duration recorded in the PR.
- [ ] A run that times out or is cancelled reports somewhere a human will see it, verified by an actual run and not only by reading the condition.
- [ ] The chosen approach is justified with before and after timings from the real runner, not from the development machine.
- [ ] If the fix trades coverage for time (a narrower selector, a split job), what is no longer covered is stated explicitly in the workflow's header comment, which is unusually thorough and should stay accurate.

## Refs

#953 created the workflow. #962 and #996 cover the harness bug that masked this. #997 is a flaky test found in the same suite.

inureyes:
## Summary

`nightly-verify` had two independent defects: the `cargo test` step never finished inside the 180-minute job budget, and when the job was killed by that budget nobody was told. This fixes both. Coverage of the other three workspace members (#1007) is deliberately out of scope, and #964 is being fixed separately.

## Phase 1: make an unfinished run visible

The old `Report a red main` step was conditioned on `failure()` and reported nothing on the three-hour runs.

**A correction to the premise this was scoped on.** The step was not skipped because a job timeout tears the job down before later steps can run. It was skipped because the *condition did not match*: a `timeout-minutes` kill surfaces as cancellation, and `failure()` is false while a job is cancelled. Later steps do still run. A throwaway probe on a one-minute job settled it directly:

| probe step | condition | conclusion after the timeout |
|---|---|---|
| `sleep 300` | (none) | `cancelled` |
| `always step` | `if: always()` | **success, it ran** |
| `cancelled step` | `if: cancelled()` | **success, it ran** |
| `failure step` | `if: failure()` | `skipped` |

The probe also showed job **outputs survive** a timeout cancellation: the downstream job read both `steps.sleeper.outcome` (`cancelled`) and a custom output set before the sleep.

So a step-level condition fix would have covered the timeout case. Reporting still moves to a separate `report` job, for three failure modes no in-job step can reach:

1. **Fate-sharing, and this one actually happened.** The 2026-08-03 scheduled run (30842184319) failed in "Ensure build tools" because `brew` was not on the self-hosted runner's PATH, and the in-job reporting step then died the same way, `brew install gh` exiting 127. A genuinely red run reported nothing, for a reason unrelated to timeouts.
2. **Runner loss.** If the self-hosted runner disappears mid-job, no step in that job runs at all.
3. **The post-cancellation window is bounded.** The probe's step was one `echo`. A reporter that has to install `gh` and make several API calls is far likelier to be cut off.

The new job runs on `ubuntu-latest` (ships `gh`, no Homebrew needed, and keeps reporting off the scarce macOS runner), uses `needs: verify` + `if: always()`, and reads `needs.verify.result`. `failure` and `cancelled` produce **different titles**, so the two kinds dedup onto separate tracking issues:

- `failure` produces `[nightly-verify] main is red`
- `cancelled` produces `[nightly-verify] the run did not finish`

The existing body composition is preserved (step-outcome table, commit, run URL) with per-step outcomes passed across the boundary via job `outputs`, plus the `verify` job result, the trigger, and the job duration. Duration is what separates "ran out of budget" from "a person cancelled it", so the body states it and points at `timeout-minutes` rather than hardcoding 180, which would drift.

### `workflow_dispatch` reporting: now included

Reversed from the previous rule. That rule assumed a manual dispatch has someone watching, so filing at them is noise. The 2026-08-02 dispatch disproved it: it ran three hours, timed out, told nobody, and was only found because someone went looking. Nobody watches a job for three hours. The reintroduced noise is bounded because repeat dispatches comment on one deduplicated issue rather than filing new ones, and the body names the trigger so a reader can dismiss a run they cancelled themselves.

## Phase 2: cut the build cost

`verify-test` now builds under `[profile.test-fast]` instead of `[profile.release]`. `--profile test-fast` and `--release` are mutually exclusive, so the flag is replaced, not added.

The budget was going to codegen and linking, not to tests: run 30733081983 logged zero `Running tests/` lines and zero `test result:` lines, and the processes the runner terminated were `ld` and `clang`. `[profile.release]` is fat LTO plus `codegen-units = 1`, measured at 4 to 6 minutes per incremental rebuild of the ~390k-line main crate, paid across roughly 77 test binaries.

### The tradeoff, stated plainly

`opt-level = 3` is unchanged, so the optimized MLX numerics the suite depends on are the same. What is dropped is fat LTO and `codegen-units = 1`, which exist to tune a *shipped* binary.

**Still covered:** everything the suite asserts, under optimized codegen, with `metal,accelerate`, on the real hardware class. `panic = "unwind"` is inherited from release, so the `catch_unwind` isolation tests behave identically.

**No longer covered:** a defect that reproduces only under fat LTO or single-unit codegen. The f16 reduction-order jitter class this repo has hit before is the realistic example. `cargo test --release --features metal,accelerate` by hand remains the tool for chasing one, and is documented as such.

**Unaffected:** what ships. `release.yml` still builds and links the shipping artifacts under `[profile.release]`.

`verify-clippy` is left alone. It has no `--release` today, takes 67s warm on the runner, and changing it would risk a cold debug MLX build for no measured gain.

### Two operational notes

- **The first nightly after this lands will still be slow, and that is expected.** A different profile means a different `OUT_DIR` for the `mlxcel-core` build script, so the first run pays a full cold MLX C++ build against the persistent `$HOME/.cargo-target/mlxcel`. Do not read that run as a failed fix. Locally this cost was confirmed: a fresh 596M `target/test-fast/build/mlxcel-core-*/out` was produced.
- **Disk.** The persistent target dir now holds two profile trees rather than one, order of 10 GiB each (locally `target/release` is 9.9G and `target/test-fast` is 10G). `release.yml`'s 7-day prune is `rm -rf` of the whole directory, not a per-profile clean, so both are already bounded by it and no third rule is needed. A second cost worth naming: the nightly no longer refreshes `[profile.release]` artifacts, so `release.yml`'s first build after a prune is colder than it used to be.

## Local measurement, and its caveat

**Development machine, M1 Ultra, not the runner.** The self-hosted runner is a different and slower shared machine, so treat this as directional only.

```
cargo test --profile test-fast --features metal,accelerate --no-run
```

Cold for this profile, including the full MLX C++ build and the link of every root test binary: **14m56s wall clock**, exit 0, 76 test executables. For contrast, on the runner the release-profile equivalent did not finish in 179 minutes on 2026-08-02.

**The runner numbers are not mine to produce.** The workflow has not been dispatched from this branch. The PR is ready to have the nightly dispatched against it for before and after.

### One data point that arrived after the issue was written

Run 30761020020 (2026-08-02 schedule) **succeeded**, the first full pass the workflow has ever completed. Its `cargo test` step ran 2h54m34s inside the 180-minute budget: about 5 minutes of headroom, roughly 3%. That does not weaken the case, it sharpens it. The gate was passing on a knife edge, and #1007 proposes adding 1754 tests and three more link steps to it.

## What changed

- `.github/workflows/nightly-verify.yml`: new `report` job on `ubuntu-latest` with `needs: verify` and `if: always()`; `outputs` on `verify` to carry step outcomes across the boundary; the in-job `Report a red main` step removed; `issues: write` moved to the job that needs it; test step renamed; header comment updated for the profile tradeoff, the reporting rationale, the disk and cache-warming costs, and a correction to its stale "~30 min" figure that the 2h56m completed pass contradicts.
- `Makefile`: `verify-test` builds `--profile test-fast`; the "how these differ" comment block and the `test-fast` section comment updated.
- `Cargo.toml`: comment-only. `[profile.test-fast]` now documents that it backs `verify-test`, and the "do not use for anything that ships" note no longer points at `verify-test`.
- `CONTRIBUTING.md`, `docs/installation.md`, `.github/PULL_REQUEST_TEMPLATE.md`: the documented gate matches what the gate now runs.
- `.github/workflows/ci.yml`, `.github/workflows/release.yml`: comment-only, the quoted pre-push checklists said `test(release)`.

No Rust source is touched.

## Test plan

- [x] `cargo clippy --all-targets --profile test-fast --features metal,accelerate -- -D warnings` clean. Run under `test-fast` rather than a cold debug build because the diff contains no Rust changes and the `Cargo.toml` change is comment-only.
- [x] `cargo test --profile test-fast --features metal,accelerate --no-run`: exit 0, 76 executables, 14m56s cold.
- [x] Workflow YAML parses; both jobs and the `if:` expression resolve as intended.
- [x] The reporting script extracted and exercised as a standalone script against mocked `gh` and `date` in three states: timeout (`cancelled`, partial outputs, duration resolved to 180m21s from the real 30733081983 timings), genuine failure (`failure`, all outcomes present), and fully degraded (API unreachable and every output empty). All three exit 0 and emit a complete report; the degraded one reports `unknown` per step rather than failing.
- [x] Job-timeout semantics and output survival established by a throwaway probe run, branch deleted afterward.
- [ ] Runner before and after timings, owned by the maintainer dispatching the nightly against this branch.

Closes #1000

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -35,7 +35,7 @@ For inference changes, real-checkpoint validation is required — synthetic-only
 
 - [ ] `cargo fmt --all -- --check` (enforced by CI — violations block merge)
 - [ ] `cargo clippy --all-targets -- -D warnings`
-- [ ] `cargo test --release`
+- [ ] `cargo test --profile test-fast`
 - [ ] `cargo deny check`
 - [ ] Validated with a real checkpoint (specify which, e.g. `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`): ...
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@
 #     for why it is nightly rather than per-PR.
 #
 # Quality gate still lives locally at PR time. Pre-push checklist:
-#   make verify         # fmt + clippy(metal,accelerate, -D warnings) + test(release)
+#   make verify         # fmt + clippy(metal,accelerate, -D warnings) + test(test-fast)
 #   make verify-clean   # same, after `cargo clean` — use when clippy's
 #                       # per-crate cache may be hiding a regression
 #
--- a/.github/workflows/nightly-verify.yml
+++ b/.github/workflows/nightly-verify.yml
@@ -163,12 +163,36 @@ jobs:
       - name: Ensure build tools
         run: |
           set -euo pipefail
+
+          # Put Homebrew on PATH before looking for anything. A non-login shell
+          # on this runner does not source the profile that `brew shellenv`
+          # writes, so neither `cmake` nor `brew` itself is resolvable by
+          # default. Runs 30842184319 (2026-08-03 schedule) and 30870790330
+          # (2026-08-04 dispatch) both died here in about one second: `cmake`
+          # was not found, the fallback ran `brew install cmake`, and `brew`
+          # was not found either, so the step exited 127. Both prefixes are
+          # listed because Apple Silicon installs under /opt/homebrew and
+          # Intel under /usr/local; adding a directory that does not exist is
+          # harmless.
+          export PATH="/opt/homebrew/bin:/usr/local/bin:$PATH"
+
           # cmake is required by the mlxcel-core and sentencepiece-sys build scripts.
           if ! command -v cmake >/dev/null 2>&1; then
+            if ! command -v brew >/dev/null 2>&1; then
+              echo "::error::Neither cmake nor brew is on PATH after adding the Homebrew prefixes." >&2
+              echo "This is a runner provisioning problem, not a code failure." >&2
+              echo "PATH=$PATH" >&2
+              exit 1
+            fi
             echo "cmake not found, installing via Homebrew..."
             brew install cmake
           fi
           echo "cmake: $(cmake --version | head -1)"
+
+          # Every later step gets the same PATH; `export` above is scoped to
+          # this step's shell only.
+          echo "/opt/homebrew/bin" >> "$GITHUB_PATH"
+          echo "/usr/local/bin" >> "$GITHUB_PATH"
 
       # rust-toolchain.toml pins the channel (and the rustfmt/clippy
       # components) that every cargo invocation below actually resolves to, so
@@ -256,6 +280,10 @@ jobs:
           COMMIT_SHA: ${{ github.sha }}
           EVENT_NAME: ${{ github.event_name }}
           VERIFY_RESULT: ${{ needs.verify.result }}
+          # A `workflow_dispatch` can target any branch, so the report must not
+          # claim `main` is red when it was verifying something else.
+          REF_NAME: ${{ github.ref_name }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
           # These do survive a `timeout-minutes` cancellation, so the usual
           # timeout report is fully populated. They are empty only if the runner
           # is lost outright, which the `unknown` default below covers.
@@ -265,12 +293,33 @@ jobs:
         run: |
           set -euo pipefail
 
+          # Name the ref when it is not the default branch. A dispatch against a
+          # PR branch says nothing about `main`, and a title claiming otherwise
+          # sends the reader to the wrong place. Keeping the bare title for the
+          # default branch also keeps the dedup key stable for the scheduled
+          # runs, which are the ones that matter; a branch run gets its own
+          # issue and does not reuse `main`'s.
+          if [ "$REF_NAME" = "$DEFAULT_BRANCH" ]; then
+            SUBJECT='main'
+            SCOPE=''
+          else
+            SUBJECT="\`$REF_NAME\`"
+            SCOPE=" ($REF_NAME)"
+          fi
+
           if [ "$VERIFY_RESULT" = "cancelled" ]; then
-            TITLE='[nightly-verify] the run did not finish'
+            TITLE="[nightly-verify] the run did not finish${SCOPE}"
             HEADLINE='The nightly `make verify` run did not finish. GitHub reports a `timeout-minutes` kill and a human cancel with the same `cancelled` conclusion, so compare the duration below against `timeout-minutes` in `.github/workflows/nightly-verify.yml`: at or just under the budget means the job ran out of time, anything well short of it means someone cancelled the run.'
           else
-            TITLE='[nightly-verify] main is red'
-            HEADLINE='The nightly `make verify` run on `self-hosted-macos-26-arm64` failed. At least one of the three gates below is genuinely red on the commit named at the bottom.'
+            TITLE="[nightly-verify] ${SUBJECT} is red${SCOPE}"
+            # Deliberately not "at least one of the three gates is red": a setup
+            # step (toolchain, build tools) can fail the job before any gate
+            # runs, and then the gates report `failure` only because cargo was
+            # never installed. Run 30870790330 was exactly that, and a headline
+            # asserting a genuine code failure would have pointed the reader at
+            # the wrong thing. Say the job failed and let the table and the log
+            # say where.
+            HEADLINE='The nightly `make verify` run on `self-hosted-macos-26-arm64` failed. Check the first failing step in the run log before reading the gate table below: a setup step that fails leaves the gates reporting `failure` for want of a toolchain rather than because the code is red.'
           fi
           export TITLE
 
--- a/.github/workflows/nightly-verify.yml
+++ b/.github/workflows/nightly-verify.yml
@@ -238,8 +238,9 @@ jobs:
       ${{ always()
           && github.repository == 'lablup/mlxcel'
           && (needs.verify.result == 'failure' || needs.verify.result == 'cancelled') }}
-    # GitHub-hosted, which is what makes reasons 1 and 3 above actually go away:
-    # `ubuntu-latest` ships `gh` preinstalled, so nothing here needs Homebrew.
+    # Being a separate job is what answers reasons 2 and 3; being GitHub-hosted
+    # is what answers reason 1, since `ubuntu-latest` ships `gh` preinstalled
+    # and nothing here needs Homebrew.
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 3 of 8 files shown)


inureyes:
## Runner measurement

Dispatched on this branch after the two follow-up commits (`a6f872806` Homebrew PATH, `dbd5ee463` DEVELOPER_DIR). Run [30871982545](https://github.com/lablup/mlxcel/actions/runs/30871982545), `success`.

| step | before (`--release`, run 30761020020) | after (`--profile test-fast`, run 30871982545) |
|---|---|---|
| `cargo fmt` | 4s | 4s |
| `cargo clippy` | 71s | 71s |
| **`cargo test`** | **2h54m34s** | **48m54s** |
| job total | 2h56m | 50m18s |
| share of the 180-minute budget | 97% | **27%** |

A 3.6x reduction on the step that was consuming the entire budget.

**The "after" number is an upper bound, not the steady state.** The preceding run on this branch (30870981405) died in the Metal error before it could finish populating the `test-fast` tree, so this run paid most of a cold MLX C++ build on top of the test link. A run against an already-warm `test-fast` tree should come in lower.

## The reporting path, verified end to end on the runner

Three observations across three dispatches, covering both directions:

| run | verify result | report job | outcome |
|---|---|---|---|
| 30870790330 | `failure` (died in `Ensure build tools`, 16s) | **success** | filed #1018 |
| 30870981405 | `failure` (clippy and test, Metal) | **success** | commented on the same issue |
| 30871982545 | `success` | **skipped** | silent, as intended |

The first of those is the fate-sharing case this PR's design argument rests on, and it was not hypothetical: the job died at `brew`, which is exactly where the old in-job reporter would have died too (it ran `brew install gh`). The reporter on `ubuntu-latest` was unaffected.

## Two runner defects this surfaced, both fixed here

Neither was caused by the profile change; both were latent and would have surfaced on their own the next time `release.yml` pruned the persistent target directory.

1. **Homebrew is not on PATH** for a non-login shell on the runner, so `Ensure build tools` exited 127 looking for `brew` after failing to find `cmake`. This had already taken down the 2026-08-03 scheduled run.
2. **`xcrun` cannot resolve `metal`**, because `xcode-select` points at the Command Line Tools. Every warm run skipped the MLX C++ build and never noticed; the first cold build failed immediately. This is the same condition `docs/mla-absorbed-decode.md` tells developers to fix with `DEVELOPER_DIR`.

Also corrected: the report titled itself `main is red` for a run verifying a PR branch, and its headline asserted that one of the three gates was genuinely red when the job had actually died in a setup step and the gates only reported `failure` for want of a toolchain.

Issue #1008, opened 2026-08-03
fix(test): mlxcel-core release test binary intermittently hangs forever, blocking later runs via the cargo lock


inureyes:
## Summary

The `mlxcel-core` release test binary intermittently hangs and never terminates, spinning at 100% CPU indefinitely. It has been observed three times over three days on the same machine. Because the hung process keeps the cargo file lock for its target directory, every subsequent `cargo test` in that working tree blocks behind it, so the failure also silently prevents later test runs from executing at all.

This matters more than an ordinary flake because this repository's CI does not run the test suite. A local full-suite run is the project's only real correctness gate, and a gate that intermittently hangs forever is not a gate.

## Observations

Three occurrences, all the same test binary (`target/release/deps/mlxcel_core-693b1aae6e646b8d`), all pinned at approximately 100% CPU, all requiring `kill -9`:

| elapsed when found | working tree |
|---|---|
| 2 days 02:44:59 | `mlxcel-2` |
| 2 days 11:34:45 | `wt-epic-909-issue-899` (worktree, since removed; the process held the deleted inode) |
| 7 hours 43:44 | `wt-epic-909-issue-904` |

Spinning rather than blocked: the processes consumed a full core each for their whole lifetime, which distinguishes this from a deadlock on a mutex.

Independently, during the same period an implementer working on #902 reported a full-suite run stalling on `layers::tests::causal_attention_*` while `pgrep` showed test binaries from two different working trees competing for the GPU. That is the only named lead so far and has not been confirmed.

## Conditions

Every observed occurrence happened while the machine was under heavy parallel load: several git worktrees of this repository building and running tests concurrently, load average between 18 and 25 on a 20-core M1 Ultra. The suite completes normally in about 20 to 30 seconds of test time when the machine is quiet, and repeated quiet-machine runs during the same period all completed.

That points at a contention-dependent failure rather than a deterministic one, and it means reproduction likely requires inducing GPU or memory pressure rather than simply re-running the suite.

## Impact

Two effects, the second easy to misdiagnose:

1. A full core is consumed indefinitely. Two of these ran for over two days before being noticed, which also silently degraded every performance measurement taken on the machine during that window.
2. The hung process holds the cargo lock for its target directory. A later `cargo test` in the same tree blocks waiting for it and produces no output, which reads exactly like the new run itself hanging or dying silently. During epic #909 this was initially misdiagnosed as shell pipe buffering; the actual cause was a new run queued behind a hung old one.

[...]

Issue #997, opened 2026-08-02
fix: Resolve flaky text_only_forward_produces_finite_logits in granite4_vision and hunyuan_vl parity tests


inureyes:
## Problem / Background

Two real-model VLM parity tests, both named `text_only_forward_produces_finite_logits`, fail intermittently when the whole root-package test suite runs, and pass every time when their target is run in isolation.

- `tests/granite4_vision_parity.rs:100` (model `granite-4.0-3b-vision-4bit`), assertion `text-only logits must be finite`.
- `tests/hunyuan_vl_parity.rs:105` (model `hunyuanocr-mlx-4bit`), same assertion.

Both tests follow the same shape: `mlxcel::load_model(&dir)`, tokenize `"Hello, world."`, `LanguageModel::make_caches`, `LanguageModel::forward`, `eval`, slice the last position, `max_all`, then assert `item_f32(&max).is_finite()`. The observed failure is therefore a non-finite (NaN or inf) maximum logit on a text-only forward pass.

These failures are unrelated to the fix in #996 (issue #962). They surfaced while running `cargo test --release --features metal,accelerate --no-fail-fast` as that issue's fourth acceptance criterion, and are deliberately not folded into that PR.

## Evidence

All observations on the same machine (Apple Silicon, macOS, `--features metal,accelerate`, release profile), against branch `fix/issue-962-nightly-verify-red` at commit `d32f684ca`.

| Run | Command | `granite4_vision_parity` | `hunyuan_vl_parity` | Rest of suite |
|-----|---------|--------------------------|---------------------|---------------|
| Whole-suite 1 | `cargo test --release --features metal,accelerate --no-fail-fast` (73 result lines) | 3 passed, 0.97s | 4 passed, 1.25s | green |
| Whole-suite 2 | identical command | 2 passed / 1 failed, 0.82s | 3 passed / 1 failed, 0.67s | green (5216 passed, 274 ignored) |
| Isolated x3 | `--test granite4_vision_parity` | 3/3 passed, three consecutive attempts | | |
| Isolated x3 | `--test hunyuan_vl_parity` | | 4/4 passed, three consecutive attempts | |

Two things that are worth ruling out up front:

- The only source difference between whole-suite run 1 and run 2 was an edit to `tests/cli_help_consistency.rs` and `tests/common/mod.rs`. Neither failing test spawns a binary; both reach `common::repo_model_dir`, which that change did not touch. The change is not a plausible cause.
[...]

PR #1021, opened 2026-08-04
fix(ci): widen the verify gate to all four workspace members

### LINKED ISSUE #1007: fix(ci): make verify and the nightly test only one of the four workspace members

inureyes:
## Summary

`make verify-test` and `make verify-clippy` passed no package selector. The workspace root is itself the `mlxcel` package, so both resolved to `-p mlxcel` and neither built `mlxcel-core`, `mlxcel-surgery` or `mlxcel-xla`. Both targets now pass `--workspace`, `verify-test` also passes `--no-fail-fast`, and the lint debt that surfaces behind the wider scope is cleared.

## Scope of the hole, measured

| | test binaries | doctest targets | tests passed |
|---|---|---|---|
| root only (old gate) | 76 | 1 | 5284 |
| `--workspace` (new gate) | 84 | 4 | 7100 |

The 8 skipped binaries are `mlxcel-core`'s lib test, `mlxcel-xla`'s lib test, and `mlxcel-surgery`'s lib test plus its five `tests/*_integration.rs`. `mlxcel-core` alone carried 1354 of the missing tests, and it is the crate holding the MLX `cxx` bridge, `layers.rs`, the KV cache and the quantization loaders.

## `--workspace` rather than `default-members`

`default-members` would have worked, and is rejected for two reasons.

It re-scopes every bare cargo invocation in the repository at once. `release.yml` runs `cargo build --release --target aarch64-apple-darwin --locked` with no selector, so a `default-members` covering all four would start compiling the default-off `mlxcel-xla` into every release build. `scripts/run_quality_gate.sh` has three more bare invocations.

It also moves the gate's scope into a manifest field that the gate's own command line does not mention. A scope you cannot read off the command is the defect this issue is about, so hiding it in `Cargo.toml` would be repeating the mistake somewhere new. The `[workspace]` stanza carries a comment recording the decision so the next person does not add one.

## Feature resolution

`cargo tree --workspace --features metal,accelerate -e features -i mlxcel-core` resolves as follows, so this is not a complication:

| member | resolved features |
|---|---|
| `mlxcel` (root) | `default` (`surgery`) + `metal` + `accelerate` |
| `mlxcel-core` | `metal` + `accelerate`, through the root's forwarding, one shared build |
| `mlxcel-surgery` | `default` (empty) |
| `mlxcel-xla` | `default` (empty); `iree` stays off |

`mlxcel-xla`'s `iree` staying off is what keeps the gate free of an IREE distribution: `src/lib/mlxcel-xla/build.rs` returns early on a missing `CARGO_FEATURE_IREE` rather than compiling the native shim. The code behind `iree`, `diagnostics` and `micro-oracle` therefore remains outside the gate, which the docs now say.

## Concurrency, the #1008 question

`--workspace` puts the `mlxcel-core` suite and the root suite in one invocation, which is the arrangement that corrupts results in #1008. It does not reproduce, because cargo builds every test binary and then runs them one at a time.

Sampled every 0.5s across a full workspace run: **217 of 217 samples showed exactly one test binary running**, across 75 distinct binaries, with `mlxcel_core-cdfd4c46d62ff555` observed alone. Doctests are not a second path to it either: `mlxcel-core` executes 2 doctests and both are pure-CPU `hardware::kv_cache_bytes*`, and rustdoc merges them into one binary.

The `no_other_mlxcel_core_test_binary_is_sharing_the_gpu` guard would not have caught the root-suite case by itself, since it matches only a second `mlxcel-core` binary. The Makefile, `docs/installation.md` and `CONTRIBUTING.md` therefore record the serialization that a parallel test runner such as `cargo nextest` would have to re-establish.

## `--no-fail-fast`

Only matters now that the run covers four members: without it the first failing test binary ends the run and hides the other three behind whatever failed first, which for a nightly that costs the better part of an hour is the wrong trade. Cargo still exits non-zero, so the gate is no weaker, and compile errors stop the run either way.

## Lint debt cleared

Six clippy errors, all pre-existing, all in `mlxcel-xla`'s lib-test target. `mlxcel-core` and `mlxcel-surgery` were already clean. The "7" in the issue thread reconciles as these six plus cargo's own `error: could not compile mlxcel-xla (lib test) due to 6 previous errors` summary line.

| lint | site |
|---|---|
| `unused_imports` | `emitter/mod.rs:157`, the `emit_vision_diagnostics` re-export |
| `clippy::manual_is_multiple_of` | `emitter/phi4_audio.rs:1003`, `emitter/qwen2_vl.rs:227`, `emitter/qwen2_vl.rs:458` |
| `clippy::identity_op` | `emitter/qwen2_vl.rs:827`, a `48 * 64 + 0` index in a test |
| `clippy::items_after_test_module` | `emitter/vision.rs:35` |

The unused import is allowed rather than deleted, and the `#[cfg(any(test, feature = "diagnostics"))]` gate kept, because that `test` arm is the only thing type-checking `emit_vision_impl(config, true)` in a build without `diagnostics`, and `diagnostics` implies `iree`, which needs an IREE dist and is therefore never on in CI. The four sibling re-export blocks above it already carry the same attribute for the same reason. The `identity_op` is fixed with a `(row, column)` accessor so the packed-bias assertions keep saying which cell they mean instead of degrading to a bare `48 * 64`. The `items_after_test_module` fix moves the test module to the end of `emitter/vision.rs` with no change to its contents.

## Timings, development machine

M1 Ultra, warm build, `--profile test-fast --features metal,accelerate --no-fail-fast`. **These are development-machine numbers; the runner is slower and shared, and its measurement is a separate nightly dispatch that this PR does not trigger.**

| run | wall clock | result |
|---|---|---|
| `cargo test --profile test-fast --features metal,accelerate --no-fail-fast` | **1m54s** | 5284 passed, 0 failed, 274 ignored |
| `cargo test --profile test-fast --workspace --features metal,accelerate --no-fail-fast` | **2m21s** | 7100 passed, 0 failed, 290 ignored |
| delta | **+27s (+23.7%)** | +8 binaries, +1816 tests |

Added build cost, isolated: deleting exactly the 8 member test executables and re-running `--no-run` rebuilt only the three members, root untouched, in **17.3s**. A no-op `--no-run` is 0 to 1s at either scope.

So on a warm tree the whole change costs about 44s, split roughly 17s of build and 27s of execution. The runner's 48m54s is dominated by the root build, which is unchanged; the new members add 8 link steps out of 84. `nightly-verify.yml`'s header records what to do if the budget ever stops fitting: split `verify-test` into per-member steps before dropping members from it.

## Test plan

- [x] `make verify-fmt`: pass (4s)
- [x] `make verify-clippy`: pass (22s), workspace clean at `-D warnings`
- [x] `make verify-test`: pass (217s), 84 binaries + 4 doctest targets, **7100 passed, 0 failed, 290 ignored**
- [x] Coverage demonstrated: the run lists `mlxcel_core-…`, `mlxcel_surgery-…`, `mlxcel_xla-…` and the five `mlxcel-surgery` integration binaries, plus `Doc-tests` for all four members
- [x] Concurrency: 217/217 samples at one running test binary, `mlxcel-core`'s observed alone
- [x] Negative control, the criterion the issue asks for. A `Debug`-bound error planted in an `mlxcel-core` test, the #973 defect class, then reverted:
  - old `cargo clippy --all-targets --features metal,accelerate -- -D warnings`: **exit 0**, passes with the error present
  - new `make verify-clippy`: **exit 2**, `error[E0277]: NotDebug doesn't implement std::fmt::Debug` at `src/lib/mlxcel-core/src/streams.rs:289`, `could not compile mlxcel-core (lib test)`
- [x] `python3 scripts/ci/check_cross_repo_refs.py` clean; every `#N` in the diff verified against `lablup/mlxcel`
- [x] No `models/` path written, moved or read for write

Not applicable: no inference code changes, so no real-checkpoint validation. The 84-binary suite includes the model parity tests, which self-skip without weights.

## Notes for reviewers

`CLAUDE.md` is a symlink to `AGENTS.md` and both are gitignored, so the equivalent update to the local agent contract is not in this diff.

The known-flaky `granite4_vision_parity` and `ernie4_5_moe_vl_parity` tests (#997) passed in all three full runs here.

Closes #1007

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -34,8 +34,8 @@ For inference changes, real-checkpoint validation is required — synthetic-only
 -->
 
 - [ ] `cargo fmt --all -- --check` (enforced by CI — violations block merge)
-- [ ] `cargo clippy --all-targets -- -D warnings`
-- [ ] `cargo test --profile test-fast`
+- [ ] `cargo clippy --workspace --all-targets -- -D warnings`
+- [ ] `cargo test --workspace --profile test-fast`
 - [ ] `cargo deny check`
 - [ ] Validated with a real checkpoint (specify which, e.g. `mlx-community/Qwen3.5-0.8B-OptiQ-4bit`): ...
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,8 @@
 #     for why it is nightly rather than per-PR.
 #
 # Quality gate still lives locally at PR time. Pre-push checklist:
-#   make verify         # fmt + clippy(metal,accelerate, -D warnings) + test(test-fast)
+#   make verify         # fmt + clippy(workspace, metal,accelerate, -D warnings)
+#                       # + test(workspace, test-fast)
 #   make verify-clean   # same, after `cargo clean` — use when clippy's
 #                       # per-crate cache may be hiding a regression
 #
--- a/.github/workflows/nightly-verify.yml
+++ b/.github/workflows/nightly-verify.yml
@@ -46,6 +46,40 @@
 # are `#[ignore]`d. What this run actually gates is the weight-free unit and
 # contract suite, which is exactly where both known failures lived.
 #
+# WHICH MEMBERS IT COVERS
+#
+# All four, as of #1007, because `verify-clippy` and `verify-test` now pass
+# `--workspace`. They did not before, and the omission was not a small one: the
+# workspace root is itself the `mlxcel` package, so a bare `cargo test` there
+# resolves to `-p mlxcel` and never builds mlxcel-core, mlxcel-surgery or
+# mlxcel-xla at all. This job was running 5238 of the 6992 tests in the
+# repository and calling that the suite. mlxcel-core held 1354 of the missing
+# ones and is the crate with the MLX `cxx` bridge, layers.rs, the KV cache and
+# the quantization loaders.
+#
+# The clippy half was the subtler one. `--all-targets` without `--workspace`
+# does not compile a member's *test* target either, so test-only lint debt and
+# test-only compile errors were equally invisible: six clippy errors were
+# sitting in mlxcel-xla's lib-test target while this job passed clean, and five
+# `Debug`-bound compile errors in new mlxcel-core tests once passed both this
+# job and the local gate. `verify-fmt` never had the hole, because
+# `cargo fmt --all` was already workspace-wide.
+#
+# Feature resolution across the members is not a complication here.
+# mlxcel-core resolves to metal + accelerate through the root package's
+# forwarding, so one build of it serves every member. mlxcel-surgery and
+# mlxcel-xla resolve to their empty defaults, which matters most for
+# mlxcel-xla: its `iree` feature stays off, its build script skips the native
+# shim, and this job therefore needs no IREE distribution. The code behind
+# `iree`, `diagnostics` and `micro-oracle` stays outside the gate for the same
+# reason.
+#
+# On the budget: the switch to `[profile.test-fast]` in #1000 took the
+# `cargo test` step from 2h54m34s to 48m54s (run 30871982545), and the scope
+# widening spends part of that headroom back. If it ever stops fitting, split
+# `verify-test` into per-member steps before dropping members from it; a member
+# that is not in the gate is a member nobody is testing.
+#
 # WHICH PROFILE THE TESTS BUILD UNDER, AND WHAT THAT GIVES UP
 #
 # `make verify-test` builds under `[profile.test-fast]`, not `[profile.release]`
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,7 +149,8 @@ jobs:
       # machine in a fraction of the time.
       #
       # Quality gate lives locally now. Pre-push checklist:
-      #   make verify         # fmt + clippy(metal,accelerate, -D warnings) + test(test-fast)
+      #   make verify         # fmt + clippy(workspace, metal,accelerate, -D warnings)
+      #                       # + test(workspace, test-fast)
       #   make verify-clean   # same, after `cargo clean` — use when clippy's
       #                       # per-crate cache may be hiding a regression
       #
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,22 +33,26 @@ Thank you for your interest in contributing to mlxcel! This document covers the 
    ```bash
    # macOS (Apple Silicon)
    cargo build --release --features metal,accelerate
-   cargo test --profile test-fast --features metal,accelerate
+   cargo test --workspace --profile test-fast --features metal,accelerate
 
    # Linux / CUDA
    cargo build --release --features cuda
-   cargo test --profile test-fast --features cuda
+   cargo test --workspace --profile test-fast --features cuda
    ```
 4. Run the local quality gates:
    ```bash
    cargo fmt --all -- --check   # gated at PR time; fmt violations block merge
-   cargo clippy --all-targets --features metal,accelerate -- -D warnings   # NOT gated at PR time; yours to run
-   cargo test --profile test-fast --features metal,accelerate              # NOT gated at PR time; yours to run
+   cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings   # NOT gated at PR time; yours to run
+   cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast   # NOT gated at PR time; yours to run
    cargo deny check             # gated at PR time (advisories + licenses + sources)
    ```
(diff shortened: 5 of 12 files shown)


inureyes:
## Runner measurement

Run [30881430828](https://github.com/lablup/mlxcel/actions/runs/30881430828), `success`, **88 binaries / 7100 passed / 0 failed**.

| step | before this PR (root only) | with `--workspace` |
|---|---|---|
| `cargo fmt` | 4s | 14s |
| `cargo clippy` | 71s | 2m28s |
| **`cargo test`** | **48m54s** | **22m52s** |
| job total | 50m18s | **25m49s** |
| share of the 180-minute budget | 27% | **13%** |

Against the original `--release` baseline (run 30761020020, `cargo test` 2h54m34s at 97% of budget) the gate is now **7.6x faster while covering four workspace members instead of one**, and 7100 tests instead of 5284.

**These two columns are not a controlled A/B and should not be read as one.** The before column is run 30871982545 on `lablup-macmini01`; this run landed on `lablup-macmini13`, a different machine with a different cache state. What the numbers do establish is the thing that matters for the acceptance criterion: the expanded gate finishes comfortably inside the budget, measured on the runner, on the machine that was least prepared for it.

## Runner pool

This run is also the first successful one on `lablup-macmini13`. Three machines answer to `self-hosted-macos-26-arm64` (`macmini01`, `macmini12`, `macmini13`) and until today only `macmini01` was provisioned, which is why runs that looked like flaky infrastructure were really a coin flip over assignment. `macmini12` and `macmini13` have since had Homebrew, cmake, Xcode and the Metal toolchain installed. The `Ensure build tools` guard added in #1017 is what identified this: it printed `xcode-select -p: /Library/Developer/CommandLineTools` and an empty `Xcode installs found:` instead of letting cmake fail hundreds of lines later with a header error.

## One caveat, tracked separately

A local full-workspace run hit `drafter::tests::auto_detect_emits_info_log_for_default_fallback` failing once under load (3 of 3 passing alone, 3 of 3 passing for the standalone core lib binary). Filed as #1023. It did not reproduce on this runner run. Together with #997 it means the widened gate has two known intermittent sources, and both should be resolved rather than tolerated: an intermittently red nightly gets ignored, and an ignored gate is the blind spot the nightly exists to close.

PR #1025, opened 2026-08-05
fix(models): mamba and mamba2 treat a scales-without-biases embedding as non-quantized

### LINKED ISSUE #976: fix(models): mamba and mamba2 treat a scales-without-biases embedding as non-quantized
## Problem / Background

`MambaModel::from_weights` and `Mamba2Model::from_weights` decide whether the embedding table is quantized by requiring **both** `.scales` and `.biases`:

```rust
let embed_scales = weights
    .remove("backbone.embeddings.scales")
    .or_else(|| weights.remove("model.embed_tokens.scales"));
let embed_biases = weights
    .remove("backbone.embeddings.biases")
    .or_else(|| weights.remove("model.embed_tokens.biases"));

let embeddings = if let (Some(scales), Some(biases)) = (embed_scales, embed_biases) {
    UnifiedEmbedding::Quantized(mlxcel_core::layers::QuantizedEmbedding::new(
        embed_weight, scales, biases, group_size, bits,
    ))
} else {
    UnifiedEmbedding::Regular(mlxcel_core::layers::Embedding::new(embed_weight))
};
```

`src/models/mamba.rs:534` through `src/models/mamba.rs:551`, and the identical block at `src/models/mamba2.rs:763` through `src/models/mamba2.rs:780`.

Affine quantization stores zero-point `biases`; the block-float schemes (`mxfp4`, `nvfp4`, `mxfp8`) do not, by design. That is exactly what `infer_quantization_mode` (`src/lib/mlxcel-core/src/layers.rs:1030`) encodes, and why `QuantizedEmbedding::from_weights` (`src/lib/mlxcel-core/src/layers.rs:253`) keys the quantized branch on `.scales` alone and infers the mode from bias presence:

```rust
let has_biases = weights.get(&format!("{}.biases", prefix)).is_some();
let mode = infer_quantization_mode(has_biases, group_size, bits);
Self::from_weights_with_mode(weights, prefix, group_size, bits, mode)
```

So for a block-float export, the two mamba files take the `else` branch and build `Embedding::new(embed_weight)` over the raw packed tensor. `Embedding::forward` (`src/lib/mlxcel-core/src/layers.rs:193`) calls `ffi::embedding`, which is a bare `mlx::core::take(weight, indices, 0)` (`src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:2976`). No dequantization happens: the lookup returns a `uint32` array whose last dimension is the packed width (`hidden_size * bits / 32`), not `hidden_size`. Nothing rejects it at load; the model reports as loaded and the failure surfaces downstream, either as a shape or dtype rejection inside a later MLX op (which crosses the cxx bridge as an uncatchable abort at the first forward pass) or as nonsense hidden states. Either way the load succeeds when it should not.

These two files also bypass the shared reconciler entirely, which is the point #958 raised for them: `QuantizedEmbedding::new` (`src/lib/mlxcel-core/src/layers.rs:235`) is a plain field-filling constructor that hardcodes `mode: "affine"` and calls neither `reconcile_quantization_layout_logged` nor `validate_quantization_params`. The `group_size` and `bits` come from `MambaConfig::group_size()` / `bits()` (`src/models/mamba.rs:126` / `:134`, `src/models/mamba2.rs:76` / `:84`), which read a private `Quantization { group_size, bits }` struct (`src/models/mamba.rs:44`, `src/models/mamba2.rs:31`) with no `mode` field at all, so even an affine mamba export cannot express a block-float mode.

### Currently latent

No mamba or mamba2 checkpoint in the recommended set ships a block-float embedding, so this does not reproduce today. It becomes live the moment an `mxfp4` / `nvfp4` / `mxfp8` mamba export appears, and it will fail as garbage output rather than as a load error, which is the expensive way to find it.

## Proposed Solution

Route both files through the shared loader instead of hand-building the layer:

1. Replace the `if let (Some(scales), Some(biases))` gate at `src/models/mamba.rs:541` and `src/models/mamba2.rs:770` with a `.scales`-only check, and construct via `QuantizedEmbedding::from_weights` (or `from_weights_with_mode` with a mode from `mlxcel_core::layers::infer_quantization_mode(has_biases, group_size, bits)`), so `biases` stays optional the way the struct already allows (`src/lib/mlxcel-core/src/layers.rs:227`).
2. Going through `from_weights` also picks up `reconcile_quantization_layout_logged` (`src/lib/mlxcel-core/src/layers.rs:1175`) and therefore `validate_quantization_params`, which closes the mamba half of #958 as a side effect rather than as a separate bounds check bolted on.
3. The two loaders currently `remove()` the tensors from the `WeightMap` while `QuantizedEmbedding::from_weights` takes `(&WeightMap, prefix)`. The prefix differs by checkpoint (`backbone.embeddings` vs `model.embed_tokens`), so resolve the prefix first and pass it, rather than keeping the by-hand extraction.
4. Consider making `QuantizedEmbedding::new` private or `#[doc(hidden)]`, or removing it in favour of the fallible constructors, so the next family that builds one directly cannot reintroduce either problem. #958 raised the same question.

## Acceptance Criteria

- [ ] A mamba or mamba2 checkpoint whose embedding carries `.scales` and no `.biases` loads as `UnifiedEmbedding::Quantized` with the inferred block-float mode, not as `UnifiedEmbedding::Regular` over the packed tensor.
- [ ] An affine mamba / mamba2 embedding (`.scales` plus `.biases`) continues to load and generate identically.
- [ ] A non-quantized mamba / mamba2 embedding (no `.scales`) still takes the `Embedding::new` path.
- [ ] Both files reach `QuantizedEmbedding` through a path that runs `validate_quantization_params`, so a hostile `"bits": 0` or `"group_size": 0` fails at load rather than aborting at the first forward pass.
- [ ] `models/falcon-mamba-7b-4bit` (or another mamba checkpoint present under `models/`) produces unchanged output, confirming the affine path is untouched.
- [ ] `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` pass.

## Refs

Follow-up to #958 (PR #972). #958 listed `mamba.rs` and `mamba2.rs` under Scope for the direct `QuantizedEmbedding::new` bypass; this issue covers the second defect on the same two lines, that the branch is gated on bias presence rather than scales presence. Same abort class as #929.

inureyes:
## Summary

`MambaModel::from_weights` and `Mamba2Model::from_weights` decided whether the embedding table was quantized by requiring both `.scales` and `.biases`. Affine quantization stores zero-point biases; the block-float modes (mxfp4, nvfp4, mxfp8) carry none by design, which is exactly what `infer_quantization_mode` keys on. A block-float embedding export therefore fell to the else branch and built a plain `Embedding` over the raw packed uint32 table. `Embedding::forward` is a bare `mlx::core::take` with no dequantization, so the lookup returned a uint32 array whose last axis was the packed width (`hidden_size * bits / 32`) rather than `hidden_size`. Nothing rejected it at load: the model reported as loaded and the failure surfaced later as garbage hidden states, or as an MLX C++ throw crossing the cxx bridge as an uncatchable `std::terminate` at the first forward pass.

The defect is latent today because no mamba or mamba2 checkpoint in the recommended set ships a block-float embedding. It goes live the moment one appears, and it would present as wrong output rather than as a load error, which is the expensive way to find it.

Both files now resolve the embedding prefix (`backbone.embeddings` or `model.embed_tokens`, one `contains_key` probe) and hand the whole map to `UnifiedEmbedding::from_weights`, which gates on `.scales` alone and infers the mode from bias presence.

## This supersedes PR #972

#972, merged for issue #958, declined to route mamba through the shared loader and instead made the hand-built `QuantizedEmbedding::new` fallible. Two of its four stated reasons do not hold:

- "takes the table out of the map by `remove`, so it cannot address a single-prefix map loader": the prefix is resolvable at the call site with one `contains_key`, and the `remove` maintains no invariant. The same function already loads in_proj / x_proj / dt_proj / out_proj / lm_head through `UnifiedLinear::from_weights(&weights, ...)`, which copies and leaves the keys in the map, and there is no leftover-weight check anywhere in the load path.
- "treats `.biases` as optional where mamba requires both": that requirement is the bug this PR fixes, so the reason is circular.

The other two reasons are real, and this PR accepts them as costs:

- Routing costs one lazy `ffi::copy` (`mlx::core::copy`) of the embedding table instead of moving it out of the map. That is a lazy-array node share rather than an element copy, and it is what every other model already pays.
- The loader stores the reconciled pair rather than the declared one. For a uniform-quant checkpoint reconciliation is a no-op and emits no warning; for a mixed-precision one it is a strict improvement, since the declared top-level pair is the wrong one for an embedding quantized at a different bit width.

Routing through the shared loader also picks up `reconcile_quantization_layout` and therefore `validate_quantization_params`, which is the bound `QuantizedEmbedding::new` was carrying on its own since #958.

## New load-time rejection for these two families

Both files now call `validate_embedding_table` (`src/models/gpt2.rs`) right after the embedding loads. This is a NEW rejection for mamba and mamba2: it requires a 2-D table with non-zero rows, `vocab_size <= rows`, and, for a non-quantized table, a width equal to `hidden_size` (for a quantized table it reconstructs the dequantized width as `scales.shape(-1) * group_size` through `validate_quantized_packing`). It is the same guard BailingMoe, Gpt2, GptBigCode and GptNeoX already run, so this is not a new policy, only a new caller.

It is also what gives the regression tests an observable. `MambaModel::embeddings` is private and the test module is a sibling rather than a descendant, so a test cannot inspect the variant, and a forward pass on a regressed load aborts the test binary instead of failing it. With the guard in place the assertion is simply that the block-float load succeeds, because the regressed path (a `Regular` embedding over the packed table) now fails at load.

## `QuantizedEmbedding::new` removed

After this change it has zero production callers across all four workspace members (verified by grep). It is removed, so the next family cannot rebuild the same defect by hand-building an affine-only layer that requires a `biases` argument. `QuantizedMultiLinear::new` stays: it is pre-emptive and also callerless, but it takes `Option<biases>` and so cannot express this bug.

## What changed

- `src/models/mamba.rs`, `src/models/mamba2.rs`: prefix resolution plus `UnifiedEmbedding::from_weights`, then `validate_embedding_table`. The missing-embedding error still names both spellings rather than regressing to the loader's single-prefix "Weight not found". The stale comment describing the `remove` chain is replaced with one that says why the gate is `.scales`-only.
- `src/lib/mlxcel-core/src/layers.rs`: `QuantizedEmbedding::new` removed, with the reason recorded on `from_weights`; the three doc references to it retargeted; the `QuantizedEmbedding` half of `hand_built_quantized_layers_bound_their_declared_params` retargeted at `QuantizedEmbedding::from_weights` over a synthetic `WeightMap`, keeping the `QuantizedMultiLinear::new` half intact.
- `src/models/mamba_tests.rs`, `src/models/mamba2_tests.rs`: the existing fixture closures hoisted into shared helpers, the existing test's doc comment corrected, and three new tests per family (see below).
- `src/models/gpt2.rs`, `src/lib/mlxcel-core/src/layers.rs`: "Used by" lists updated to include Mamba and Mamba2.
- `docs/adding-models.md`: `QuantizedEmbedding::new` removed from the table of loaders that already carry the quant bound, with a note explaining why hand-building an embedding is no longer the answer for a multi-spelling prefix.

## Tests

Per family:

- The block-float regression (`*_loads_a_block_float_embedding_as_quantized`): mxfp4 at `packed_in` 1 and mxfp8 at `packed_in` 2, both dequantizing to `hidden_size` 8, `.scales` present and `.biases` absent, load must succeed. nvfp4 is not covered because group_size 16 cannot describe a width of 8 with a non-zero number of groups, and inventing a second `hidden_size` for it would have been dishonest geometry.
- Both prefix spellings (`*_resolves_both_embedding_prefix_spellings`): affine and block-float both load under `model.embed_tokens.*`, and a checkpoint carrying neither spelling fails with a message naming both.
- The guard itself (`*_rejects_an_embedding_table_narrower_than_hidden_size`): a float table whose width is not `hidden_size` errors at load with a message naming `hidden_size`.
- The existing float-table case in `*_rejects_quantization_params_that_would_abort_the_embedding_lookup` now inserts a genuine [8, 8] float table instead of reusing the packed [8, 1] one with its scales stripped, because the new guard correctly rejects the latter. The guard was not weakened to accommodate the old fixture.
- The hostile-pair loop over `switch_layers::HOSTILE_QUANT_PARAMS` is unchanged and still passes: with `.biases` present the inferred mode is affine, and `reconcile_quantization_layout` calls `validate_quantization_params` before it looks at any shape.

The block-float tests were confirmed to fail on regression by temporarily restoring the bias-requiring gate in `mamba.rs`: both new cases failed with `backbone.embeddings.weight is [8, 1] but hidden_size is 8`, which is a load error rather than an abort. The temporary change was reverted before committing.

## Test plan

Commands actually run, all green:

- [x] `cargo fmt --check`
- [x] `cargo clippy -p mlxcel-core --lib --tests -- -D warnings`
- [x] `cargo check -p mlxcel-core --lib --tests`
- [x] `cargo test --lib models::mamba` (14 passed, 2 ignored; the ignored two are the pre-existing `requires serial MLX execution` conv-state tests)
- [x] `cargo test -p mlxcel-core --lib layers::tests` (72 passed), including `hand_built_quantized_layers_bound_their_declared_params`
- [x] `cargo test --lib models::gpt2::tests` (27 passed), since the shared guard gained callers

Orchestrator-side, after the above:

- [x] `cargo clippy --workspace --all-targets -- -D warnings` (all four workspace members, every target), exit 0
- [x] `cargo fmt --all -- --check`, the exact CI gate
- [x] Real-checkpoint shape verification of the new guard against `falcon-mamba-7b-4bit`, `mamba2-1.3b-4bit` and `mamba2-130m`, see the comment below

Pre-existing and unrelated: a bare `cargo test --workspace --lib` on a Linux/CUDA box built without `--features cuda` aborts the root lib test binary with `terminate called ... [cuda_kernel] No CUDA back-end` (SIGABRT) after roughly 1926 tests. This was confirmed to reproduce identically on `main` at e1ea5c2f with the same warm target directory (1927 tests, same abort), so it is a property of this build configuration rather than of this branch. Every candidate test in the in-flight window passes when run individually. CI does not use this configuration.

Not run: the acceptance criterion comparing `models/falcon-mamba-7b-4bit` output before and after. No mamba or mamba2 checkpoint is present under `models/` on this machine, so the affine path was not confirmed against real weights. The affine path is covered by the synthetic positive control in the existing guard test, which is weaker evidence. `cargo clippy --all-targets -- -D warnings` and the full workspace test suite were also not run here; only `--lib --tests` was, plus the narrow test filters above.

Closes #976

### DIFF
--- a/docs/adding-models.md
+++ b/docs/adding-models.md
@@ -238,9 +238,18 @@ What already carries the bound, so you inherit it for free:
 | `reconcile_quantization_layout` | every `UnifiedLinear` / `UnifiedEmbedding` |
 | `SwitchLinear::from_stacked_parts` | MoE experts via the shared `switch_layers` loader |
 | `QuantizedMultiLinear::{new, from_weights}` | MLA `embed_q` / `unembed_out` |
-| `QuantizedEmbedding::new` | hand-built embeddings that skip the map loader |
 | `FusedQKVLinear::from_weights_separate_with_mode` | fused QKV projections |
 | `infer_mla_quantization_params` | the MLA `kv_b_proj` decomposition in `sanitize_weights` |
+
+There is deliberately no hand-built `QuantizedEmbedding` constructor on that
+list any more. One existed for Mamba / Mamba2, which resolve their table under
+two possible prefixes and so looked unable to address the single-prefix map
+loader. It hardcoded `mode: "affine"` and required a `biases` argument, and that
+is exactly how those two families came to treat a block-float embedding
+(`.scales`, no `.biases`) as non-quantized (issue #976). If your checkpoint
+spells the embedding prefix more than one way, resolve the prefix with a
+`contains_key` probe and pass it to `UnifiedEmbedding::from_weights`; do not
+hand-build the layer.
 
 What you must do yourself:
 
--- a/src/lib/mlxcel-core/src/layers.rs
+++ b/src/lib/mlxcel-core/src/layers.rs
@@ -76,7 +76,7 @@ impl QuantizedWeight {
     /// Create a new quantized weight with explicit mode.
     ///
     /// Fallible on purpose (issue #973), on the same reasoning that made
-    /// [`QuantizedEmbedding::new`] fallible for the declared `group_size` /
+    /// [`QuantizedMultiLinear::new`] fallible for the declared `group_size` /
     /// `bits` pair. This is the one way to build a `QuantizedWeight` with a
     /// caller-chosen mode without going through
     /// [`reconcile_quantization_layout`], and therefore without the allowlist it
@@ -250,53 +250,17 @@ pub struct QuantizedEmbedding {
 }
 
 impl QuantizedEmbedding {
-    /// Create a new quantized embedding layer (affine mode) from tensors the
-    /// caller already owns.
-    ///
-    /// Fallible on purpose (issue #958). This is the one way to build a
-    /// `QuantizedEmbedding` without going through
-    /// [`Self::from_weights_with_mode`], and therefore without
-    /// [`reconcile_quantization_layout`] and the bound it now carries. A caller
-    /// that hand-builds the layer is exactly the caller whose declared
-    /// `group_size` / `bits` have never been looked at, and the stored pair goes
-    /// straight to `quantized_embedding` / `quantized_matmul`, which cross the
-    /// cxx bridge as `UniquePtr<MlxArray>` rather than `Result`: a C++ throw
-    /// there is an uncatchable abort at the first forward pass rather than a
-    /// load error. Returning `Result` puts the bound on the path the next family
-    /// that builds one directly will take, which is the whole reason the
-    /// signature changed rather than the two existing call sites being patched.
-    /// It is a strong default rather than an enforced invariant: the fields
-    /// above are `pub`, so a struct literal still bypasses this entirely.
-    /// Nothing in the tree does that today.
-    ///
-    /// This bounds the declared pair only; it does not reconcile against the
-    /// tensor shapes the way `from_weights_with_mode` does, because the shapes
-    /// arrive here already detached from the prefix the reconciler names in its
-    /// divergence warning.
-    ///
-    /// Used by: Mamba, Mamba2 (both take the table out of the `WeightMap` by
-    ///          `remove`, under either the `backbone.embeddings` or the
-    ///          `model.embed_tokens` spelling, so neither can address the
-    ///          single-prefix map loader)
-    pub fn new(
-        weight: UniquePtr<MlxArray>,
-        scales: UniquePtr<MlxArray>,
-        biases: UniquePtr<MlxArray>,
-        group_size: i32,
-        bits: i32,
-    ) -> Result<Self, String> {
-        validate_quantization_params(group_size, bits)?;
-        Ok(Self {
-            weight,
-            scales,
-            biases: Some(biases),
-            group_size,
-            bits,
-            mode: "affine".to_string(),
-        })
-    }
-
     /// Load from weight map
+    ///
+    /// There is deliberately no by-hand `new` beside this. One existed for
+    /// Mamba / Mamba2, which took their table out of the `WeightMap` by
+    /// `remove` under two possible prefixes and so could not address a
+    /// single-prefix map loader; it hardcoded `mode: "affine"` and required a
+    /// `biases` argument, which is precisely how those two families came to
+    /// treat a block-float embedding (`.scales`, no `.biases`) as
+    /// non-quantized (issue #976). Both now resolve the prefix and come through
+    /// here, and the constructor is gone so the next family cannot rebuild the
+    /// same defect. Prefer resolving a prefix over hand-building a layer.
     pub fn from_weights(
         weights: &crate::weights::WeightMap,
         prefix: &str,
@@ -485,7 +449,7 @@ impl UnifiedEmbedding {
     /// a packed table describes, which `is_quantized()` alone cannot give them.
     ///
     /// Used by: the shared `validate_embedding_table` guard (BailingMoe, Gpt2,
-    ///          GptBigCode, GptNeoX)
+    ///          GptBigCode, GptNeoX, Mamba, Mamba2)
     pub fn quantized(&self) -> Option<&QuantizedEmbedding> {
         match self {
             Self::Quantized(e) => Some(e),
@@ -1312,8 +1276,7 @@ pub struct ReconciledQuant {
 /// Used by: every quantizing family through [`reconcile_quantization_layout`]
 ///          (dense projections and quantized embeddings) and through
 ///          `SwitchLinear::from_stacked_parts` (MoE experts); the by-hand
-///          constructors [`QuantizedEmbedding::new`],
-///          [`QuantizedMultiLinear::new`] and
+///          constructors [`QuantizedMultiLinear::new`] and
 ///          [`QuantizedMultiLinear::from_weights`], which never reach the
 ///          reconciler; every family-local MoE expert loader through
 ///          `crate::models::switch_layers::validate_expert_quantization_params`
@@ -1808,7 +1771,7 @@ pub struct QuantizedTensorShapes<'a> {
 ///
 /// Used by: BailingMoe weight validation (token table, dense projections,
 ///          stacked experts) and the shared `validate_embedding_table` guard
-///          (BailingMoe, Gpt2, GptBigCode, GptNeoX)
+///          (BailingMoe, Gpt2, GptBigCode, GptNeoX, Mamba, Mamba2)
 pub fn validate_quantized_packing(
     label: &str,
     shapes: &QuantizedTensorShapes<'_>,
@@ -2821,9 +2784,15 @@ impl QuantizedMultiLinear {
     /// Create a new quantized multi-linear layer from tensors the caller
     /// already owns.
     ///
-    /// Fallible for the same reason as [`QuantizedEmbedding::new`] (issue
-    /// #958): the stored pair reaches `quantized_matmul` unmediated, and nothing
-    /// else on this path bounds it.
+    /// Fallible on purpose (issue #958): the stored pair reaches
+    /// `quantized_matmul` unmediated, without
+    /// [`reconcile_quantization_layout`] and the bound it carries, and nothing
+    /// else on this path bounds it. Unlike the `QuantizedEmbedding` constructor
+    /// this once mirrored (removed in issue #976), it takes `biases` as an
+    /// `Option` and so cannot express a mode that contradicts the planes the
+    /// checkpoint ships. Nothing in the tree calls it today; it is a strong
+    /// default for the next caller that hand-builds one, not an enforced
+    /// invariant, since the fields are `pub`.
     pub fn new(
         weight: UniquePtr<MlxArray>,
         scales: UniquePtr<MlxArray>,
@@ -5328,24 +5297,40 @@ mod tests {
         }
     }
 
-    /// The two by-hand constructors are the only way to build a quantized layer
-    /// without going through `reconcile_quantization_layout`, which is exactly
-    /// how `mamba` / `mamba2` reached `quantized_embedding` with an unbounded
-    /// pair (issue #958). They are fallible so the bound cannot be skipped, and
-    /// this drives the real constructors rather than the pure helper, so the
-    /// guard is exercised where a checkpoint actually reaches it. Losing it turns
-    /// a rejected load into an uncatchable abort at the first forward pass in
-    /// production; here the assertions are on the constructor result, so a
-    /// regression fails cleanly rather than aborting the test binary.
+    /// A quantized layer built without going through
+    /// `reconcile_quantization_layout` stores the declared `group_size` / `bits`
+    /// verbatim, which is exactly how `mamba` / `mamba2` reached
+    /// `quantized_embedding` with an unbounded pair (issue #958). Both halves
+    /// here drive a real entry point rather than the pure helper, so the guard
+    /// is exercised where a checkpoint actually reaches it, and both assert on
+    /// the constructor result with no forward pass, so a regression fails
+    /// cleanly rather than aborting the test binary.
+    ///
+    /// The embedding half targets `QuantizedEmbedding::from_weights`, the path
+    /// production now uses: the by-hand `QuantizedEmbedding::new` was removed in
+    /// issue #976, because requiring a `biases` argument is what made a
+    /// block-float table look non-quantized to the two families that called it.
+    /// `QuantizedMultiLinear::new` stays and is still hand-built here: it takes
+    /// `Option<biases>` and so cannot express that defect, and it remains the
+    /// bound for the next caller that builds one directly.
     #[test]
     fn hand_built_quantized_layers_bound_their_declared_params() {
+        use crate::weights::WeightMap;
+
         // Honest affine 4-bit geometry: packed_in * 32 == bits * groups * gs,
         // i.e. 8 * 32 == 4 * 1 * 64.
         let plane = |last: i32| ffi::from_slice_f32(&vec![0.0; (4 * last) as usize], &[4, last]);
+        let embed_weights = || {
+            let mut weights = WeightMap::new();
+            weights.insert("embed.weight".to_string(), plane(8));
+            weights.insert("embed.scales".to_string(), plane(1));
+            weights.insert("embed.biases".to_string(), plane(1));
+            weights
+        };
 
         // Positive control first, so a guard that rejects everything quantized
         // cannot pass this test.
-        QuantizedEmbedding::new(plane(8), plane(1), plane(1), 64, 4)
+        QuantizedEmbedding::from_weights(&embed_weights(), "embed", 64, 4)
             .expect("an honest affine pair must still build a quantized embedding");
         QuantizedMultiLinear::new(plane(8), plane(1), Some(plane(1)), 64, 4)
             .expect("an honest affine pair must still build a quantized MultiLinear");
@@ -5357,10 +5342,10 @@ mod tests {
             (0, 4, "group_size"),
             (-64, 4, "group_size"),
         ] {
-            let err = QuantizedEmbedding::new(plane(8), plane(1), plane(1), group_size, bits)
+            let err = QuantizedEmbedding::from_weights(&embed_weights(), "embed", group_size, bits)
                 .err()
                 .unwrap_or_else(|| {
-                    panic!("QuantizedEmbedding::new must reject {group_size} / {bits}")
+                    panic!("QuantizedEmbedding::from_weights must reject {group_size} / {bits}")
                 });
             assert!(err.contains(field), "unhelpful error: {err}");
 
--- a/src/models/gpt2.rs
+++ b/src/models/gpt2.rs
@@ -851,7 +851,7 @@ pub(crate) fn dim_eq(dim: i32, expected: usize) -> bool {
 /// `std::terminate` at the first forward pass. The reconstruction is shared with
 /// the dense projections via `mlxcel_core::layers::validate_quantized_packing`.
 ///
-/// Used by: BailingMoe, Gpt2, GptBigCode, GptNeoX
+/// Used by: BailingMoe, Gpt2, GptBigCode, GptNeoX, Mamba, Mamba2
 pub(crate) fn validate_embedding_table(
     table: &UnifiedEmbedding,
     key: &str,
--- a/src/models/mamba.rs
+++ b/src/models/mamba.rs
@@ -25,6 +25,7 @@ use std::path::Path;
 
 use super::model_owned::ModelOwnedSequenceState;
 use super::recurrent_snapshot::{push_optional, restore_optional};
+use crate::models::gpt2::validate_embedding_table;
 
 /// Parse eos_token_id from config.json. Can be a single int, an array of ints, or absent.
 /// Used by: Mamba, Mamba2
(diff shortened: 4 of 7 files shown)


inureyes:
## Real-checkpoint verification of the new `validate_embedding_table` guard

The PR body notes that no mamba or mamba2 checkpoint is present locally, so the acceptance criterion naming `models/falcon-mamba-7b-4bit` could not be executed. That gap is now closed for the part that actually matters, which is whether the new load-time rejection would refuse a checkpoint that loads today.

Tensor shapes come from each repo's safetensors header (two HTTP Range requests: bytes 0-7 give the little-endian u64 header length, then the header JSON maps tensor name to dtype and shape), and the config values from `raw/main/config.json`. No weights were downloaded.

| checkpoint | config | `embeddings.weight` | `embeddings.scales` | biases |
|---|---|---|---|---|
| `mlx-community/falcon-mamba-7b-4bit` | `falcon_mamba`, vocab 65024, hidden 4096, gs 64, bits 4 | U32 `[65024, 512]` | F16 `[65024, 64]` | present |
| `mlx-community/mamba2-1.3b-4bit` | `mamba2`, vocab 50288, hidden 2048, gs 64, bits 4 | U32 `[50288, 256]` | F16 `[50288, 32]` | present |
| `mlx-community/mamba2-130m` | `mamba2`, vocab 50288, hidden 768, no quantization | F16 `[50288, 768]` | absent | absent |

Against the four checks the guard applies:

| checkpoint | 2-D | `vocab_size <= rows` | width | packing invariant | verdict |
|---|---|---|---|---|---|
| falcon-mamba-7b-4bit | yes | 65024 <= 65024 | `scales.shape(-1) * gs` = 64 * 64 = 4096 == hidden_size | 512 * 32 == 4 * 64 * 64 (16384) | loads |
| mamba2-1.3b-4bit | yes | 50288 <= 50288 | 32 * 64 = 2048 == hidden_size | 256 * 32 == 4 * 32 * 64 (8192) | loads |
| mamba2-130m | yes | 50288 <= 50288 | cols 768 == hidden_size (non-quantized branch) | n/a, float table | loads |

All three pass, so the guard adds no rejection for the checkpoints this family is validated against today. It exercises both branches of the guard as well: two affine quantized tables and one float table.

The same arithmetic covers the "loads and generates identically" criterion for the affine path. Both quantized checkpoints satisfy `packed_in * 32 == bits * groups * group_size` exactly at their declared `group_size` 64 / `bits` 4, so `reconcile_quantization_layout` returns the declared pair unchanged with `reconciled = false` and emits no warning. The reconciled pair this PR now stores is byte-identical to the declared pair the old hand-built path stored, which is why routing through the shared loader cannot change what these two checkpoints produce.

Still not executed: an actual generation run on real weights. The above is shape-level ground truth, not an output comparison.

inureyes:

### Intent

Stop `mamba` / `mamba2` from gating the quantized embedding branch on `.biases` presence, so a block-float export (`.scales`, no `.biases`) loads as `UnifiedEmbedding::Quantized` instead of as a plain `Embedding` over the raw packed uint32 table.

Verdict: the PR implements issue #976 correctly and completely. No CRITICAL or HIGH finding, nothing auto-fixed, no commit added.

### 1. Prefix resolution: dropping the per-tensor `.or_else()` chain is not a regression

The old chain did not support mixed spellings, it silently paired tensors from different prefixes. Both pathological orderings get strictly better:

- `weight` under `backbone.embeddings`, `.scales` under `model.embed_tokens`: the old code built a `QuantizedEmbedding` from mismatched planes, which MLX rejects in `validate_quantized_input` as an uncatchable abort at the first forward pass. The new code resolves `backbone.embeddings`, finds no `.scales` there, builds a `Regular` embedding over the packed table, and `validate_embedding_table` refuses it at load naming `hidden_size`.
- The mirrored ordering behaves the same way.

No real checkpoint can carry mixed spellings here, because both files hardcode `backbone.layers.{i}.*` and `backbone.norm_f.weight` for everything else (`src/models/mamba.rs:571`, `src/models/mamba2.rs:807`). A genuinely transformers-spelled checkpoint would fail at `backbone.layers.0.mixer.conv1d.weight` regardless of how the embedding is spelled, so the `model.embed_tokens` branch is only reachable at `num_hidden_layers: 0`, which is what the new `*_resolves_both_embedding_prefix_spellings` fixture uses. Theoretical only, and the change is the safer direction.

### 2. `validate_embedding_table` cannot refuse a checkpoint that works on `main`

Checked all four conditions against the failure modes they replace:

- 2-D and non-zero rows: mamba embedding tables are always `[rows, packed_or_width]`.
- `vocab_size <= rows`: padded-vocab checkpoints go the safe direction (config `vocab_size` at or below the padded row count). The `state-spaces` lineage (`vocab_size` 50277 against a 50280-row table) passes; all three validated checkpoints have `vocab_size == rows`.
- Width: for a float table `cols == hidden_size`; for a quantized one `scales.shape(-1) * reconciled_group_size == hidden_size` plus MLX's own `validate_quantized_input` predicate, mirrored in `validate_quantized_packing`. Anything failing either throws inside MLX at the first forward pass on `main`, so the guard converts an abort into a load error rather than refusing a working load.
- `falcon_mamba` specifically: `model_type == "falcon_mamba"` only flips `use_bcdt_rms` in `compute_time_step_rank` (`src/models/mamba.rs:117`). Nothing about the variant changes the embedding geometry, and 65024 rows / `64 * 64 == 4096 == hidden_size` passes.
- Tied vs untied `lm_head` is not reachable by the guard: it inspects only the embedding table, and `lm_head` still loads through `UnifiedLinear::from_weights`.

The one class that "works on `main` and now fails" is a config whose `vocab_size` exceeds the table rows and whose prompts happen never to emit a high token id. That is an out-of-bounds gather on `main`, so refusing it is the point of the guard.

### 3. Affine behavior is preserved, and no new warning fires

- `infer_quantization_mode(true, _, _)` returns `"affine"`, identical to the removed constructor's hardcoded mode.
- The affine branch of `reconcile_quantization_layout` trusts `group_size` and re-derives `bits` through `infer_quantization_bits`, which returns `caller_bits` unchanged when `packed_in * 32 == num_groups * group_size * bits`. That holds exactly for `falcon-mamba-7b-4bit` (512 * 32 == 64 * 64 * 4) and `mamba2-1.3b-4bit` (256 * 32 == 32 * 64 * 4), so `reconciled == false` and `reconcile_quantization_layout_logged` emits no `tracing::warn!`.
- The `group_size == 16 && bits == 4` NVFP4-repack special case is unreachable for both, since both declare `group_size` 64.

So the reconciled pair the loader now stores is the declared pair for every validated checkpoint. The only genuinely new work on the affine path is `validate_quantization_params`, which is what issue #958 asked for.

### 4. `QuantizedEmbedding::new` removal is complete

`git ls-files | xargs grep -l 'QuantizedEmbedding::new'` returns one file, `src/lib/mlxcel-core/src/layers.rs`, and the single hit is the doc comment at line 5310 recording the removal. That covers all four workspace members, `tests/`, `examples/`, `benchmarks/`, `python/`, `spike/` and the docs. `src/lib/mlxcel-surgery/src/ops/scale.rs:264` mentions `QuantizedEmbedding::from_weights` in prose only. `mlxcel-core` is not published, and the struct fields stay `pub`, so this is not an external break.

### 5. The new tests do fail on a reverted fix, and they fail rather than abort

Every new assertion is on the `Result` of `MambaModel::from_weights` / `Mamba2Model::from_weights`. There is no forward pass anywhere in the new tests, so nothing can abort the test binary in place of failing it.

The revert claim is provable from evidence already on the board, without re-running it: `*_rejects_an_embedding_table_narrower_than_hidden_size` is green, which establishes that a `Regular` embedding over an `[8, 1]` table produces a load error containing `hidden_size`. Restoring the bias-requiring gate routes `block_float_embedding_weights(prefix, 1)` to exactly that `Regular` path, so `*_loads_a_block_float_embedding_as_quantized` panics on its `.expect`. The mxfp8 case lands the same way at `[8, 2]`. The error text the author reports matches `validate_embedding_table`'s non-quantized branch verbatim.

One coupling worth recording: the block-float tests get their observable from the guard, so if `validate_embedding_table` were ever dropped from these two loaders they would go green again even with the bias gate restored. The test doc comments say so explicitly, so it is a stated tradeoff rather than an oversight.

### 6. The same defect class elsewhere

A full sweep of the tracked tree found no second instance of the exact defect, a both-planes-required predicate choosing Quantized over Regular for an embedding. Nearby findings, all out of scope for #976:

- `src/models/deepseek.rs:971` (`SwitchLinear::from_stacked_parts`) is the only surviving `(Some(scales), Some(biases)) => Quantized, _ => Regular` for MoE experts. It is already masked: `DeepSeekModel::from_weights` rejects `mode != "affine"` up front at `src/models/deepseek.rs:690` with a message that names this exact failure. The residual hole is that `ModelArgs::quantization_mode()` returns `"affine"` when the `quantization` block is absent entirely. Fixing it properly means giving the local `SwitchLinear::Quantized` an `Option` biases and a `mode` field, not flipping the match. (MEDIUM, separate issue)
- `src/models/kimi_linear.rs:169-190` (`MultiLinear::forward`) gates correctly on `.scales` alone and keeps `Option` biases, then passes a null biases pointer with a hardcoded `"affine"` mode, with no `validate_quantization_biases` at `from_weights`. That is an uncatchable abort rather than a load error, the same shape `gpt_oss.rs:463` already fixed. Cheapest high-value follow-up. (MEDIUM, separate issue)
- Roughly 18 family-local MoE expert loaders require `.biases` and return a clean `Err` without it. They refuse a block-float export rather than mis-loading it, which is the safe direction, but each carries a non-`Option` biases and a hardcoded `"affine"` at `gather_qmm`. (LOW, one sweep)
- Four MLA `kv_b_proj` sanitizers (`deepseek_v3.rs:1138`, `deepseek_v32.rs:920`, `kimi_linear.rs:1280`, `longcat_flash_ngram.rs:1096`) gate on `.scales` then `.unwrap()` the biases, so a block-float plane panics instead of erroring. `youtu_vl_lm_sanitize.rs:71` is the already-corrected sibling. (LOW)

Confirmed clean: `switch_layers.rs:461`, `gpt_oss.rs:445`, `bailing_moe.rs:879`, `nemotron_h.rs:2187`, the fused-QKV bias triple in `layers.rs:2358`, `mlxcel-surgery`, and the shape-only bias checks in `gemma3n.rs` / `helium.rs`.

### Verification

- [x] Integrated into the project code flow (both `MambaModel::from_weights` and `Mamba2Model::from_weights`, which are the live entry points for `ModelType::Mamba` / `ModelType::Mamba2` in `src/loading/nonstandard.rs` and `src/loading/special.rs`)
- [x] Project conventions followed (`use crate::models::gpt2::validate_embedding_table` matches `bailing_moe.rs:105` and `gpt_neox.rs:69` exactly)
- [x] Existing modules reused where applicable (`UnifiedEmbedding::from_weights`, `validate_embedding_table`, `validate_quantized_packing`, `HOSTILE_QUANT_PARAMS`)
- [x] No unintended structural changes (one public constructor removed, which issue #976 asked for; no renames, moves or API-surface churn beyond it)
- [x] Tests pass

### Remaining items

- [ ] The `models/falcon-mamba-7b-4bit` output-comparison acceptance criterion is still not executed, since no mamba checkpoint is present locally. Shape-level ground truth from the safetensors headers stands in for it, and the arithmetic above shows the affine path stores the same pair it stored before, but that is not an output comparison.

inureyes:

Scope: the 7-file diff at `c68ffdae` against `main` @ `e1ea5c2f`, read under a hostile-or-malformed-checkpoint threat model (config-declared integers and tensor shapes crossing the cxx bridge, where an MLX C++ throw is an uncatchable `std::terminate` rather than a `Result`).

**No CRITICAL and no HIGH findings. Nothing auto-fixed; the branch is unchanged.**

### Config-declared integers reaching arithmetic: clean, and strictly stronger than the path it replaces

`group_size` / `bits` are bounded by `validate_quantization_params` (`bits` in `[1, 32]`, `group_size` in `[1, 2^20]`) at the top of `reconcile_quantization_layout`, before any shape is read. The removed `QuantizedEmbedding::new` called that same bound but nothing else; the shared path adds `validate_quantization_mode` and `validate_quantization_biases` on top of it, and reconciles against the tensor shapes. Every multiply on the new path is `checked_mul` (`infer_quantization_bits`, `infer_quantization_group_size`, `packed_in.checked_mul(32 / bits)`) or widened to `i64` (the MLX-mirror predicate in `validate_quantized_packing`). `vocab_size` and `hidden_size` are `usize` and reach only comparisons and one `checked_mul`. No overflow is reachable in debug or release.

One property worth recording since two more families now exercise it: the block-float branch of the reconciler derives `group_size = in_features / num_groups` without re-bounding it against `MAX_QUANT_GROUP_SIZE`. It is safe, because `num_groups * derived_gs == in_features == packed_in * (32 / bits)` is already `checked_mul`-bounded to `i32`, so MLX's `scales.shape(-1) * group_size` in C++ `int` cannot overflow. Pre-existing shared-loader behavior, not introduced here.

### Panic reachability: clean, at parity with the old path

The new path is `contains_key` then `get().map(ffi::copy).ok_or_else()` then a `let [rows, cols] = shape.as_slice() else` refutation then length-guarded slicing. The `*shape.last().unwrap()` inside the reconciler is guarded by its own `is_empty()` early return, and `validate_quantized_packing` rejects a rank mismatch before it slices. The only `expect` on the path is `UnifiedEmbedding::weight()`, which asserts a `std::make_unique` result, is pre-existing, and is shared with the four families already running this guard. A malformed weight map produces an `Err`, not an abort.

### Out-of-bounds gather: closed for these two families, and this is the main security gain

Confirmed against the vendored MLX source rather than assumed. Both `ffi::embedding` and `ffi::quantized_embedding` are `mlx::core::take(..., 0)` (`mlx_cxx_bridge.cpp:2980,2999`), and the gather resolves an index with `offset_neg_idx(idx, size) = (idx < 0) ? idx + size : idx` (`mlx/backend/cpu/indexing.cpp:19`), which has no upper bound. `validate_embedding_table` rejects `vocab_size > weight.shape[0]`, and on the quantized branch `validate_quantized_packing` additionally requires `scales.shape[..rank-1] == weight.shape[..rank-1]`, so the second `take(scales, flat_indices, 0)` inside `quantized_embedding` is bounded by the same row count.

No bypass exists. `MambaModel::from_weights` and `Mamba2Model::from_weights` are the only construction sites (`src/loading/special.rs:149,156`), the guard runs unconditionally on both immediately after the embedding loads, and neither model type can be built by struct literal from outside its module.

Pre-existing and global, stated for completeness rather than as a finding against this PR: `config.vocab_size` bounds sampled tokens only indirectly (sampling clamps to the logits width, which for a tied embedding is the table row count). Prompt token ids come from the tokenizer, so a checkpoint whose tokenizer vocabulary exceeds its table rows is still an out-of-bounds read no matter what `config.json` declares. That is equally true of Gpt2, GptNeoX, GptBigCode and BailingMoe.

### `QuantizedEmbedding` construction after the removal: clean

Two struct literals exist and both are inside `layers.rs`: `from_weights_with_mode` (post-validation) and `clone_shared` (copies already-validated fields). There is no `QuantizedEmbedding { .. }` anywhere else in the workspace and no remaining reference to `QuantizedEmbedding::new` outside one doc comment that explains its removal. The `pub` fields remain a theoretical bypass, which the doc comment on `QuantizedMultiLinear::new` now states explicitly.

### Performance: the PR body's claim is correct

`ffi::copy` is `mlx::core::copy` (`mlx_cxx_bridge.cpp:825`), which builds a lazy `Copy` node (`mlx/ops.cpp:303`). `Copy::eval` is `out.copy_shared_buffer(inputs[0])` (`mlx/backend/common/common.cpp:57`), reached by both `eval_cpu` and `eval_gpu`. It is a buffer alias, not an element copy. For `falcon-mamba-7b-4bit` that is roughly 133 MB of packed table plus two 16.6 MB planes that are aliased rather than duplicated; the delta is three `array` handles.

Peak load memory is unchanged for a second reason as well: `weights` was already retained for the whole of `from_weights` before this change, since every `UnifiedLinear::from_weights(&weights, ...)` borrows it. Swapping the embedding's three `remove` calls for `get` does not extend the map's lifetime by a single statement.

`validate_embedding_table` is O(rank) shape arithmetic. `array_shape` reads `arr.inner.shape()` metadata (`mlx_cxx_bridge.cpp:300`), which forces no evaluation and touches no element data. There is no data-dependent scan anywhere in the guard.

### MEDIUM

**The `vocab_size` clause of the new guard has no test for either family.** The three new tests per family cover the block-float regression, both prefix spellings, and the `hidden_size` width clause, but nothing exercises `claimed_rows > rows`. That is the clause that closes the out-of-bounds gather, and it is the one place a caller can silently mis-wire the guard by passing the wrong config field into the `claimed_rows` slot. Every peer family that adopted this guard shipped exactly such a test: `gpt_neox_tests.rs:1191 from_weights_rejects_a_config_that_overstates_the_token_table`, `gpt2_tests.rs:554`, `gpt_bigcode_tests.rs:558`, `helium_tests.rs:544`. The shared helper is therefore covered; what is not covered is that Mamba and Mamba2 wire `config.vocab_size` into it. With the fixtures this PR already hoisted, mirroring the GptNeoX test is about ten lines per family: insert a `[3, 8]` float table under `backbone.embeddings.weight`, assert the load fails naming `vocab_size`, then assert a padded `[64, 8]` table still loads.

### LOW

**The prefix-mixing hole is closed too, and the PR body does not claim it.** The old code resolved `weight`, `scales` and `biases` through three independent `or_else` chains, so a checkpoint carrying `backbone.embeddings.weight` alongside `model.embed_tokens.scales` would have built one `QuantizedEmbedding` out of two unrelated tables. Resolving a single prefix and passing it to the shared loader makes all three planes come from the same table by construction. Worth a line in the body, since it is a second latent defect this fixes.

**Mamba and Mamba2 still have no config-scalar guard.** Unlike Gpt2, GptBigCode, GptNeoX and BailingMoe, which all call `args.validate()` before touching the checkpoint, these two validate nothing. `Vec::with_capacity(config.num_hidden_layers)` (`mamba.rs:576`, `mamba2.rs:805`) is an allocation sized straight from `config.json`, and `intermediate_size = num_heads * head_dim` then `conv_dim = intermediate_size + 2 * n_groups * ssm_state_size` (`mamba2.rs`) are unchecked `usize` products that wrap in release and panic in debug. Pre-existing, untouched by this PR, and narrower than before now that the embedding surface is guarded. Natural follow-up issue rather than a change request here.

### Verdict

Approve on security and performance grounds. The change removes a real out-of-bounds-read and garbage-output path for two families, replaces a hand-built constructor that could only express affine with the shared loader that reconciles and bounds, and pays for it with a lazy buffer alias that costs no measurable memory at load. The only thing I would ask for before merge is the `vocab_size` rows test, and that is a coverage gap rather than a defect.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: added `mamba_rejects_a_config_that_overstates_vocab_size` (`src/models/mamba_tests.rs`) and `mamba2_rejects_a_config_that_overstates_vocab_size` (`src/models/mamba2_tests.rs`), closing the MEDIUM coverage gap flagged in the security review. Both drive `MambaModel::from_weights` / `Mamba2Model::from_weights` with a `[3, 8]` table under `backbone.embeddings.weight` against the existing `vocab_size: 8` fixture config, assert the load fails naming `vocab_size`, then assert a `[64, 8]` padded table still loads. Both reuse the fixtures this PR already hoisted (`embedding_config`, `embedding_test_base_weights`) and assert on the load `Result` only, no forward pass.
- **Documentation**: none needed. `docs/adding-models.md` already covers the `QuantizedEmbedding::new` removal and does not enumerate per-clause test coverage, so there is nothing for this addition to update.
- **Lint/Format**: no issues found; nothing to fix.

### Validation
- `cargo fmt --all -- --check`: clean
- `cargo check --lib --tests`: clean
- `cargo test --lib models::mamba`: 16 passed, 2 pre-existing ignored (serial-MLX conv-state tests)
- Confirmed both new tests fail cleanly (not abort) when the `claimed_rows > rows` clause in `validate_embedding_table` is temporarily disabled, then reverted before committing.

Commit `118ddb9f` on top of `c68ffdae`.

PR #1027, opened 2026-08-05
fix(models): MLA kv_b_proj sanitizers reject scales-without-biases

### LINKED ISSUE #1026: fix(models): the MLA kv_b_proj sanitizers panic on a scales-without-biases plane
## Problem / Background

Four MLA `kv_b_proj` sanitizers decide the plane is quantized by `weights.contains_key("<prefix>.kv_b_proj.scales")`, then unconditionally `.unwrap()` the `.biases` removal:

- `src/models/kimi_linear.rs:1281` (`let biases = weights.remove(...kv_b_proj.biases).unwrap();`)
- `src/models/deepseek_v3.rs:1149`
- `src/models/deepseek_v32.rs:929`
- `src/models/longcat_flash_ngram.rs:1105`

Affine quantization stores zero-point `.biases`; the block-float modes (`mxfp4`, `nvfp4`, `mxfp8`) carry none by design, which is exactly what `mlxcel_core::layers::infer_quantization_mode` keys on. A `kv_b_proj` exported block-float therefore satisfies the `.scales` gate and then panics on the `.unwrap()`. A panic is worse than the load error it should be: in the server it takes down the process rather than rejecting one model load. Note this is a Rust panic, not the uncatchable C++ `std::terminate` class, because it fires before `dequantize` is reached.

Each of the four then calls `dequantize` with a hardcoded `"affine"` mode, so even a checkpoint that carried both planes could not express a block-float `kv_b_proj`.

### Already solved once, in-tree

`src/models/youtu_vl_lm_sanitize.rs:66-73` is the corrected fifth sibling of this exact code and already does the right thing:

```rust
let b_key = format!("{}.kv_b_proj.biases", prefix);
let b = weights.remove(&b_key).ok_or_else(|| {
    format!(
        "layer {layer_idx}: kv_b_proj has scales but no biases at key `{b_key}`; \
         the checkpoint may be corrupted or only partially converted"
    )
})?;
```

All five sites already share `mlxcel_core::layers::infer_mla_quantization_params` for the packed-pair bound (issue #958), so this is the one remaining divergence between them.

### Currently latent

Verified against the real checkpoints by reading their safetensors headers, no weights downloaded: `mlx-community/Kimi-Linear-48B-A3B-Instruct-4bit` and `-8bit` both ship `kv_b_proj` with `weight`, `scales` and `biases` (declared `"mode": "affine"`), and `moonshotai/Kimi-Linear-48B-A3B-Instruct` ships `weight` only. So no shipped checkpoint takes the panic path today. It goes live the moment a block-float MLA export appears, and it presents as a process kill rather than a load error.

### Second, lower-severity item in the same file

`kimi_linear.rs`'s private `MultiLinear::forward` (`src/models/kimi_linear.rs:171-186`) passes a null `biases` pointer to `quantized_matmul` under a hardcoded `"affine"` mode, and `MultiLinear::from_weights` (`src/models/kimi_linear.rs:231`) never calls `validate_quantization_biases`. That combination is what MLX throws "Biases must be provided for affine quantization" on, and it crosses the cxx bridge as an uncatchable abort.

This branch is **currently unreachable, not a live defect**. The sanitizer at `src/models/kimi_linear.rs:1324-1325` always dequantizes `kv_b_proj` and inserts `embed_q.weight` / `unembed_out.weight` as dense tensors with no `.scales` ("Store as dense MultiLinear weights (no re-quantization)"), and those two tensors exist in no shipped checkpoint, so `is_quantized` is always false there. It is worth correcting anyway so a future change to the sanitizer cannot silently reintroduce a live abort. `src/models/gpt_oss.rs:463` is the in-tree precedent for the fix (store an inferred mode rather than hardcoding `"affine"`, and call `mlxcel_core::layers::validate_quantization_biases`).

## Proposed Solution

1. Replace the four `.unwrap()`s with the `ok_or_else` form from `youtu_vl_lm_sanitize.rs:66-73`, keeping the message wording consistent across all five sites so the five converge rather than diverge further.
2. In `kimi_linear.rs`'s private `MultiLinear`, derive the mode with `mlxcel_core::layers::infer_quantization_mode(biases.is_some(), group_size, bits)`, store it, pass it to `quantized_matmul` instead of the hardcoded `"affine"`, and call `mlxcel_core::layers::validate_quantization_biases` at load, following `gpt_oss.rs:463`.
3. Consider whether the five near-identical MLA decomposition blocks should share a helper. This is a question for the implementer to evaluate and report on rather than a required change, since the blocks differ in their config accessors.

## Acceptance Criteria

- [ ] A `kv_b_proj` carrying `.scales` and no `.biases` produces a load error naming the missing key, not a panic, in all four families.
- [ ] The error message matches the wording already used in `youtu_vl_lm_sanitize.rs`.
- [ ] An affine `kv_b_proj` (both planes present) continues to load and dequantize identically in all four families.
- [ ] `kimi_linear`'s `MultiLinear` no longer hardcodes `"affine"` and validates the bias plane against the mode at load.
- [ ] Unit tests cover the scales-without-biases rejection for at least `kimi_linear` and one deepseek family, asserting on the load `Result` and not running a forward pass.

## Refs

Follow-up to #976 (PR #1025), which addressed the same defect class (bias presence used as a proxy for quantized-ness) in the mamba and mamba2 embedding loaders. Found by the review sweep on PR #1025; see that PR's review comments for the full inventory. Related: #958, #929, #973.

inureyes:
## Summary

Four MLA `kv_b_proj` sanitizers decided the plane was quantized by probing for `.scales`, then unconditionally `.unwrap()`ed the `.biases` removal. Affine quantization stores zero-point `.biases`; the block-float modes (mxfp4, nvfp4, mxfp8) carry none by design, which is exactly what `infer_quantization_mode` keys on. A block-float `kv_b_proj` therefore satisfies the `.scales` gate and then panics. All four now use the `ok_or_else` form already in tree at `src/models/youtu_vl_lm_sanitize.rs:66-73`, the corrected fifth sibling of this exact code, with the same message wording so the five converge instead of diverging further.

This is latent today. No shipped checkpoint takes the path: `mlx-community/Kimi-Linear-48B-A3B-Instruct-4bit` and `-8bit` both ship `kv_b_proj` with `weight`, `scales` and `biases`, and `moonshotai/Kimi-Linear-48B-A3B-Instruct` ships `weight` only. It goes live the moment a block-float MLA export appears, and it presents as a process kill rather than a load error.

## What changed

- `src/models/kimi_linear.rs`, `src/models/deepseek_v3.rs`, `src/models/deepseek_v32.rs`, `src/models/longcat_flash_ngram.rs`: the `.biases` removal is now `ok_or_else(...)?` with the wording from `youtu_vl_lm_sanitize.rs`. **No signature changed**: every enclosing `sanitize_weights` already returns `Result<WeightMap, String>` and already used `?` for `infer_mla_quantization_params`, so the `?` had a home at all four sites. This was flagged as the interesting part of the task, and it turned out to be a non-event.
- The `dequantize` call stays hardcoded `"affine"` at all five sites. A genuine block-float `kv_b_proj` is still unsupported; what changes is that it is refused by name instead of unwinding. See "What this deliberately does not do" below.
- `src/models/youtu_vl_lm_sanitize.rs`: comment only. It carried a terse `M1:` review-artifact label; it now states the same rationale as the other four, so the five are identical in behavior, message and comment. The code there was already correct.
- `src/models/kimi_linear.rs` private `MultiLinear`: gained a `mode: &'static str` field, derived at load with `mlxcel_core::layers::infer_quantization_mode(biases.is_some(), group_size, bits)` and passed to `quantized_matmul` in place of the hardcoded `"affine"`, plus a `mlxcel_core::layers::validate_quantization_biases` call at load. `src/models/gpt_oss.rs:463` is the precedent.

## The `MultiLinear` change is reachable today

Correcting an earlier claim in the other direction, per the security review: the branch is **reachable today**, not merely a bound against a future sanitizer change. `sanitize_weights` (`src/models/kimi_linear.rs:1323-1325`) only inserts a dense `embed_q.weight` / `unembed_out.weight` pair inside its `if weights.contains_key(&kv_b_key)` guard, where `kv_b_key` is `{attn_prefix}.kv_b_proj.weight`; when a checkpoint carries no `kv_b_proj.weight` at all, that whole decomposition block is skipped, so any `embed_q` / `unembed_out` planes the checkpoint shipped itself pass through sanitization untouched. `MlaAttention::from_weights` (`src/models/kimi_linear.rs:576-577`) then calls this loader unconditionally for every non-linear-attention layer, so a checkpoint that ships pre-decomposed, quantized `embed_q.weight` plus `embed_q.scales` (and the matching pair for `unembed_out`) with no `kv_b_proj.weight` to decompose from lands here with `is_quantized == true` today. This branch must not be deleted as dead code.

One honest caveat on the `validate_quantization_biases` call: with the mode inferred from `biases.is_some()` on the line above, the check is consistent by construction and cannot fire as written. It is kept because it is the assertion that couples the two helpers, so the day the mode comes from anywhere else (a declared `quantization.mode`, a per-prefix override) the contradiction is caught where the mode is stored rather than inside `quantized_matmul`, where it is an uncatchable abort. The comment at the call site says exactly this rather than implying it is load-bearing.

## What this deliberately does not do

Supporting a genuine block-float `kv_b_proj` would mean threading the inferred mode into the `dequantize` call at all five sites and verifying the decomposition against a real block-float MLA export. That is a feature, not this bug fix, and the acceptance criteria here ask for a load error naming the missing key. Worth recording: for a genuine block-float export the shared message ("the checkpoint may be corrupted or only partially converted") understates the situation slightly, since such a checkpoint is well-formed and merely unsupported here. Wording consistency across the five sites was judged the more valuable property, per the issue.

## On issue item 3, the shared helper

Evaluated, and the recommendation is **yes, but not in this PR**. Detail so the next person does not have to re-derive it.

It is more feasible than the issue's caveat suggests. The differing config accessors are a non-issue: all five resolve to `i32` at the call site already, so a helper taking `(weights, prefix, num_heads, qk_nope, v_head, kv_lora_rank, layer_label)` and returning `Result<(wk, wv), String>` absorbs the whole block, and the per-family loop (kimi skipping linear-attention layers, longcat iterating two sub-attentions per layer) stays at the call site where it belongs. That is roughly 225 lines across five files collapsing to about 110. These five sites have now been touched twice for the same defect class, once for #958 and once here, which is the strongest argument for consolidating them.

What makes it wrong to do here is that the five are not actually identical, and unifying them forces three behavioral decisions that this PR is not equipped to make:

- `youtu_vl_lm_sanitize.rs` cross-checks `w_full` against `[num_heads * head_dim, kv_lora_rank]` before reshaping; the other four do not. A shared helper either adds a new load-time rejection to four families or drops an existing guard from one.
- `kimi_linear.rs` uses `mlxcel_core::slice` + `swap_axes` where the other four use `slice_axis` + `transpose_axes`. Equivalent, so this one is free.
- `kimi_linear.rs` does **not** `copy()` the `wk` / `wv` slices contiguous, where the other four do, with a comment in youtu saying it exists so `MultiLinear`'s matmul has well-formed strides regardless of the upstream backend. Unifying silently changes kimi's contiguity. That deserves to be the change under review, not a rider.

A refactor that adds a new rejection to four families and changes the contiguity of a fifth belongs in a PR where that is the headline. Happy to file it as a follow-up issue on request.

## Separately, a sibling that this PR does not touch

`mlxcel_core::layers::QuantizedMultiLinear` (`src/lib/mlxcel-core/src/layers.rs:2775`) has the identical defect in `forward` and `forward_no_transpose`: a null biases pointer under a hardcoded `"affine"`, with `from_weights` gating on `.scales` alone and taking `biases` as an `Option` with no validation. It is reached by DeepSeek V3, DeepSeek V3.2, GLM4 MoE Lite and LongCat Flash NGram through `MultiLinear::from_weights`, which also gates on `.scales` alone.

Its reachability story is arguably better than the `kimi_linear` one fixed here, because all four sanitizers skip decomposition when `embed_q.weight` already exists, so a checkpoint shipping a pre-decomposed quantized `embed_q` reaches it directly. Left out of scope on purpose: it is a different crate, the reachability claim needs verifying against what `mlx_lm.convert` actually emits for a pre-decomposed MLA model, and adding it would blur this PR's story. Flagging it rather than silently fixing or silently skipping it.

## Test plan

Commands actually run on this branch, all green:

- [x] `cargo clippy --lib --tests -- -D warnings`, exit 0
- [x] `cargo test --lib models::kimi_linear` (4 passed)
- [x] `cargo test --lib models::deepseek_v3::` (14 passed)
- [x] `cargo test --lib models::deepseek_v32` (14 passed)
- [x] `cargo test --lib models::longcat_flash_ngram` (2 passed)
- [x] `cargo test --lib models::youtu_vl` (4 passed), since its comment was touched and it shares the message wording

New tests, one rejection case per family rather than the two families the issue required:

- `kimi_linear_sanitize_rejects_a_kv_b_proj_with_scales_and_no_biases`
- `quantized_kv_b_proj_rejects_scales_with_no_biases` (deepseek_v3)
- `quantized_kv_b_proj_rejects_scales_with_no_biases` (deepseek_v32)
- `sanitize_rejects_scales_with_no_biases` (longcat_flash_ngram), which pins the full key because this is the one family whose prefix carries a sub-attention index (`self_attn.0`), the site where a copy of the fix could most easily blame the wrong tensor
- `kimi_linear_multi_linear_infers_its_quantization_mode_from_the_biases_plane`, pinning the stored mode across affine, mxfp4, nvfp4 and mxfp8 planes and confirming a dense projection stays unquantized

Each assertion is on the load `Result` and its message, and no test runs a forward pass, because a forward pass on a bad load aborts the test binary instead of failing the test. Each rejection test keeps a both-planes-present positive control first, so a check that rejected every quantized `kv_b_proj` could not pass it. The per-family `build` fixture closures were hoisted into shared `affine_kv_b_proj_weights` / `block_float_kv_b_proj_weights` helpers so both tests in each family drive the same geometry.

Failure was confirmed rather than assumed: temporarily restoring the `.unwrap()` in `kimi_linear.rs` made the new test fail with `called Option::unwrap() on a None value` at that exact line, which confirms both that the failure mode is a Rust panic (not the `std::terminate` class, as the issue states) and that the test catches it. The temporary change was reverted before committing, and a grep confirms no `.unwrap()` remains on any `kv_b_proj` biases removal.

Not run here: `cargo clippy --workspace --all-targets -- -D warnings` and the full test suite, both left to the orchestrator. `mlxcel-core` was not modified, so no `-p mlxcel-core` run was made. No real MLA checkpoint is present on this machine, so the affine positive path was confirmed only against the synthetic UINT32 fixtures, not against real weights.

Closes #1026

### DIFF
--- a/src/models/deepseek_v3.rs
+++ b/src/models/deepseek_v3.rs
@@ -1136,9 +1136,23 @@ impl DeepSeekV3Model {
                 let s = weights
                     .remove(&format!("{}.kv_b_proj.scales", prefix))
                     .unwrap();
-                let b = weights
-                    .remove(&format!("{}.kv_b_proj.biases", prefix))
-                    .unwrap();
+                // `is_quantized` gates on `.scales` alone, and the block-float
+                // modes (mxfp4 / nvfp4 / mxfp8) ship scales with no zero
+                // points, so a block-float export satisfies that gate and
+                // arrives here carrying no `.biases` plane. The `.unwrap()`
+                // this replaces turned that into a panic during sanitization,
+                // which in the server takes the process down rather than
+                // rejecting one model load (issue #1026). `dequantize` below is
+                // hardcoded `"affine"` and so could not decompose such a plane
+                // in any case; what this buys is a load error naming the key
+                // that is missing.
+                let b_key = format!("{}.kv_b_proj.biases", prefix);
+                let b = weights.remove(&b_key).ok_or_else(|| {
+                    format!(
+                        "layer {l}: kv_b_proj has scales but no biases at key `{b_key}`; \
+                         the checkpoint may be corrupted or only partially converted"
+                    )
+                })?;
 
                 // Solve the packed pair from the shapes and bound it before it
                 // reaches `dequantize`. The shared helper checks each divisor
@@ -1978,6 +1992,44 @@ mod tests {
         }
     }
 
+    /// An affine 4-bit `kv_b_proj` for every layer in `cfg`.
+    ///
+    /// Honest geometry at `kv_lora_rank` 16: packed_in * 32 == bits *
+    /// num_groups * group_size (2 * 32 == 4 * 1 * 16). `test_config_dense_direct_q`
+    /// gives num_attention_heads 2, qk_nope_head_dim 4, v_head_dim 4, so
+    /// rows = 2 * 8 = 16.
+    ///
+    /// The packed plane is UINT32, not a float ramp: `dequantize` rejects any
+    /// other packed dtype by throwing, and that throw is the uncatchable abort
+    /// these guards exist to keep out of the forward path.
+    fn affine_kv_b_proj_weights(cfg: &DeepSeekV3Config) -> WeightMap {
+        let heads = cfg.num_attention_heads as i32;
+        let head_dim = (cfg.qk_nope_head_dim + cfg.v_head_dim) as i32;
+        let rows = heads * head_dim;
+        let mut weights = WeightMap::new();
+        for l in 0..cfg.num_hidden_layers {
+            let prefix = format!("model.layers.{l}.self_attn.kv_b_proj");
+            weights.insert(
+                format!("{prefix}.weight"),
+                mlxcel_core::zeros(&[rows, 2], mlxcel_core::dtype::UINT32),
+            );
+            insert_ramp(&mut weights, &format!("{prefix}.scales"), &[rows, 1]);
+            insert_ramp(&mut weights, &format!("{prefix}.biases"), &[rows, 1]);
+        }
+        weights
+    }
+
+    /// The same plane an mxfp4 / nvfp4 / mxfp8 export ships: `.scales`
+    /// present, `.biases` absent, because the block-float modes carry no zero
+    /// points.
+    fn block_float_kv_b_proj_weights(cfg: &DeepSeekV3Config) -> WeightMap {
+        let mut weights = affine_kv_b_proj_weights(cfg);
+        for l in 0..cfg.num_hidden_layers {
+            weights.remove(&format!("model.layers.{l}.self_attn.kv_b_proj.biases"));
+        }
+        weights
+    }
+
     /// The MLA `kv_b_proj` pair `DeepSeekV3Model::sanitize_weights` decomposes
     /// is solved from `kv_lora_rank` and two tensor axes rather than declared,
     /// and every input is untrusted: `kv_lora_rank` comes from `config.json`
@@ -1994,36 +2046,13 @@ mod tests {
     /// a regression fails cleanly here instead of aborting the test binary.
     #[test]
     fn quantized_kv_b_proj_rejects_a_kv_lora_rank_no_packing_can_describe() {
-        // Honest affine 4-bit geometry: packed_in * 32 == bits * num_groups *
-        // group_size (2 * 32 == 4 * 1 * 16). `test_config_dense_direct_q`
-        // gives num_attention_heads 2, qk_nope_head_dim 4, v_head_dim 4, so
-        // rows = 2 * 8 = 16.
-        let build = |cfg: &DeepSeekV3Config| {
-            let heads = cfg.num_attention_heads as i32;
-            let head_dim = (cfg.qk_nope_head_dim + cfg.v_head_dim) as i32;
-            let rows = heads * head_dim;
-            let mut weights = WeightMap::new();
-            for l in 0..cfg.num_hidden_layers {
-                let prefix = format!("model.layers.{l}.self_attn.kv_b_proj");
-                // UINT32, not a float ramp: `dequantize` rejects any other
-                // packed dtype by throwing, and that throw is the uncatchable
-                // abort this guard exists to keep out of the forward path.
-                weights.insert(
-                    format!("{prefix}.weight"),
-                    mlxcel_core::zeros(&[rows, 2], mlxcel_core::dtype::UINT32),
-                );
-                insert_ramp(&mut weights, &format!("{prefix}.scales"), &[rows, 1]);
-                insert_ramp(&mut weights, &format!("{prefix}.biases"), &[rows, 1]);
-            }
-            weights
-        };
-
         // Positive control first, so a guard that rejected every quantized
         // kv_b_proj could not pass this test.
         let mut honest = test_config_dense_direct_q();
         honest.kv_lora_rank = 16;
-        let sanitized = DeepSeekV3Model::sanitize_weights(build(&honest), &honest)
-            .expect("an honest quantized kv_b_proj must still sanitize");
+        let sanitized =
+            DeepSeekV3Model::sanitize_weights(affine_kv_b_proj_weights(&honest), &honest)
+                .expect("an honest quantized kv_b_proj must still sanitize");
         assert!(sanitized.contains_key("model.layers.0.self_attn.embed_q.weight"));
 
         // A zero `kv_lora_rank` is Rust integer division by zero on the very
@@ -2031,18 +2060,20 @@ mod tests {
         // than after.
         let mut zero_rank = honest.clone();
         zero_rank.kv_lora_rank = 0;
-        let err = DeepSeekV3Model::sanitize_weights(build(&zero_rank), &zero_rank)
-            .err()
-            .unwrap_or_else(|| panic!("kv_lora_rank 0 must be refused, not divided by"));
+        let err =
+            DeepSeekV3Model::sanitize_weights(affine_kv_b_proj_weights(&zero_rank), &zero_rank)
+                .err()
+                .unwrap_or_else(|| panic!("kv_lora_rank 0 must be refused, not divided by"));
         assert!(err.contains("kv_lora_rank"), "unhelpful error: {err}");
 
         // A large `kv_lora_rank` truncates the solved bit width to 0, which
         // is the divisor MLX would then divide by.
         let mut wide_rank = honest.clone();
         wide_rank.kv_lora_rank = 4096;
-        let err = DeepSeekV3Model::sanitize_weights(build(&wide_rank), &wide_rank)
-            .err()
-            .unwrap_or_else(|| panic!("a solved bit width of 0 must be refused"));
+        let err =
+            DeepSeekV3Model::sanitize_weights(affine_kv_b_proj_weights(&wide_rank), &wide_rank)
+                .err()
+                .unwrap_or_else(|| panic!("a solved bit width of 0 must be refused"));
         assert!(err.contains("bits"), "unhelpful error: {err}");
 
         // A non-quantized kv_b_proj carries no packing, so the pair is never
@@ -2060,5 +2091,45 @@ mod tests {
         );
         DeepSeekV3Model::sanitize_weights(float_only, &wide_rank)
             .expect("a float kv_b_proj must not be gated on quantization params");
+    }
+
+    /// Issue #1026: `sanitize_weights` decides `kv_b_proj` is quantized on
+    /// `.scales` alone, then used to take `.biases` with `.unwrap()`.
+    ///
+    /// Affine stores zero points; the block-float modes (mxfp4 / nvfp4 /
+    /// mxfp8) do not, which is exactly what `infer_quantization_mode` keys on.
+    /// A block-float `kv_b_proj` therefore satisfies the `.scales` gate and
+    /// arrives at the `.biases` removal with nothing to take, and the
+    /// `.unwrap()` made that a panic during weight sanitization. In the server
+    /// that takes the process down rather than rejecting one model load, which
+    /// is strictly worse than the load error it should be.
+    ///
+    /// The decomposition still dequantizes as `"affine"`, so a genuine
+    /// block-float `kv_b_proj` is not supported either way: what this pins is
+    /// that it is refused by name instead of unwinding. The assertion is on
+    /// the returned `Result` and no forward pass runs, so a regression fails
+    /// cleanly here.
+    #[test]
+    fn quantized_kv_b_proj_rejects_scales_with_no_biases() {
+        let mut cfg = test_config_dense_direct_q();
+        cfg.kv_lora_rank = 16;
+
+        // Positive control first, so a check that rejected every quantized
+        // kv_b_proj could not pass this test.
+        let sanitized = DeepSeekV3Model::sanitize_weights(affine_kv_b_proj_weights(&cfg), &cfg)
+            .expect("a kv_b_proj carrying both planes must still sanitize");
+        assert!(sanitized.contains_key("model.layers.0.self_attn.embed_q.weight"));
+
+        let err = DeepSeekV3Model::sanitize_weights(block_float_kv_b_proj_weights(&cfg), &cfg)
+            .err()
+            .expect("scales with no biases must be refused at load, not unwrapped");
+        assert!(
+            err.contains("model.layers.0.self_attn.kv_b_proj.biases"),
+            "the error must name the key that is missing, got: {err}"
+        );
+        assert!(
+            err.contains("scales but no biases"),
+            "the wording must match the other four MLA sanitizers, got: {err}"
+        );
     }
 }
--- a/src/models/deepseek_v32.rs
+++ b/src/models/deepseek_v32.rs
@@ -918,9 +918,23 @@ impl DeepSeekV32Model {
                 let s = weights
                     .remove(&format!("{}.kv_b_proj.scales", prefix))
                     .unwrap();
-                let b = weights
-                    .remove(&format!("{}.kv_b_proj.biases", prefix))
-                    .unwrap();
+                // `is_quantized` gates on `.scales` alone, and the block-float
+                // modes (mxfp4 / nvfp4 / mxfp8) ship scales with no zero
+                // points, so a block-float export satisfies that gate and
+                // arrives here carrying no `.biases` plane. The `.unwrap()`
+                // this replaces turned that into a panic during sanitization,
+                // which in the server takes the process down rather than
+                // rejecting one model load (issue #1026). `dequantize` below is
+                // hardcoded `"affine"` and so could not decompose such a plane
+                // in any case; what this buys is a load error naming the key
+                // that is missing.
+                let b_key = format!("{}.kv_b_proj.biases", prefix);
+                let b = weights.remove(&b_key).ok_or_else(|| {
+                    format!(
+                        "layer {l}: kv_b_proj has scales but no biases at key `{b_key}`; \
+                         the checkpoint may be corrupted or only partially converted"
+                    )
+                })?;
                 // Solve the packed pair from the shapes and bound it before it
                 // reaches `dequantize`. The shared helper checks each divisor
                 // before dividing: `kv_lora_rank` is a config field and the
--- a/src/models/deepseek_v32_tests.rs
+++ b/src/models/deepseek_v32_tests.rs
@@ -539,6 +539,41 @@ fn deepseek_v32_switch_linear_rejects_quantization_params_that_would_abort_gathe
     }
 }
 
+/// An affine 4-bit `kv_b_proj` for every layer in `args`.
+///
+/// Honest geometry at `kv_lora_rank` 16: packed_in * 32 == bits * num_groups *
+/// group_size (2 * 32 == 4 * 1 * 16).
+///
+/// The packed plane is UINT32, not a float weight: `dequantize` rejects any
+/// other packed dtype by throwing, and that throw is the uncatchable abort
+/// these guards exist to keep out of the forward path.
+fn affine_kv_b_proj_weights(args: &ModelArgs) -> WeightMap {
+    let num_heads = args.num_attention_heads as i32;
+    let head_dim = (args.qk_nope_head_dim + args.v_head_dim) as i32;
+    let rows = num_heads * head_dim;
+    let mut weights = WeightMap::new();
+    for l in 0..args.num_hidden_layers {
+        let prefix = format!("model.layers.{l}.self_attn.kv_b_proj");
+        weights.insert(
+            format!("{prefix}.weight"),
+            mlxcel_core::zeros(&[rows, 2], mlxcel_core::dtype::UINT32),
+        );
+        weights.insert(format!("{prefix}.scales"), f32_weight(&[rows, 1]));
+        weights.insert(format!("{prefix}.biases"), f32_weight(&[rows, 1]));
+    }
+    weights
+}
+
+/// The same plane an mxfp4 / nvfp4 / mxfp8 export ships: `.scales` present,
+/// `.biases` absent, because the block-float modes carry no zero points.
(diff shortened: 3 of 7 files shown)


inureyes:

### Intent

Stop four MLA `kv_b_proj` sanitizers from panicking on a `.scales`-without-`.biases` plane, converge their message wording on the already-correct `youtu_vl_lm_sanitize.rs` site, and stop `kimi_linear`'s private `MultiLinear` from hardcoding `"affine"`.

### Findings Addressed

None. No CRITICAL or HIGH finding was raised, so nothing was auto-fixed and the branch is unchanged at `cc28a0d0`.

### Verified

**Partial mutation on the new error path is not observable.** Each site `?`-returns after `.weight` and `.scales` are already removed, but all five `sanitize_weights` / `sanitize_text_weights` take `WeightMap` (a plain `HashMap<String, UniquePtr<MlxArray>>`) **by value**, so the half-stripped map is dropped inside the function on the error path and the borrow checker makes it unreachable to the caller. Every production caller propagates rather than continuing: `loading/special.rs:245` and `:255` (kimi, longcat), `loading/vlm_kimi_vl.rs:98` and `distributed/pipeline/stage_executor/deepseek_v3.rs:84` (dsv3), `models/glm_moe_dsa.rs:259` and `distributed/pipeline/stage_executor/glm_moe_dsa.rs:42` (dsv32), `loading/vlm_youtu_vl.rs:121` (youtu). No fallback loader retries: `loading/mod.rs:669` propagates the `Special` route error with `?`. The only non-propagating call sites are in tests. `special.rs` additionally passes a `copy_weight_map`, so the caller's own map is never touched.

**Interpolated index is the layer index at every site.** dsv3 `l` from `for l in 0..num_layers`, dsv32 `l` from `for l in 0..args.num_hidden_layers`, longcat `l` from `for l in 0..args.num_layers` (the inner `for i in 0..2` sub-attention index is carried by `b_key`, not by the `layer {l}` prefix), youtu `layer_idx`, kimi `l` from `for l in 0..config.num_hidden_layers`. `b_key` / `biases_key` is the same string that was probed in all five. The two line-continuation styles (`\` after `key ` in kimi, after `; ` in the other four) both strip the newline plus leading whitespace, so the rendered strings are identical.

**Affine behavior is preserved.** The only non-comment change in the four sanitizers is `remove(&format!(...)).unwrap()` becoming `let b_key = format!(...); remove(&b_key).ok_or_else(...)?`, same key, same value on the `Some` path. `dequantize` arguments are untouched. `youtu_vl_lm_sanitize.rs` is comment-only (`+10/-2`, entirely inside one `//` block). Each of the four new tests keeps a both-planes-present positive control that asserts `embed_q.weight` is produced, so the affine path is exercised, not just asserted.

**The `MultiLinear` mode is threaded on every path.** kimi's private `MultiLinear` has a single `forward(&self, x, transpose: bool)` with a single `quantized_matmul`, now carrying `self.mode`. There is no `forward_no_transpose` sibling to miss; both `transpose` polarities are used (`kimi_linear.rs:510`, `:522`, `:525`, `:526`) and both go through that one call. The affine case is byte-identical to before (`biases.is_some()` implies mode `"affine"` and a non-null pointer); the block-float case now loads instead of aborting in MLX.

**Tests.** All assert on the load `Result` and its message; none runs a forward pass. Rejection coverage is all four families rather than the two the issue required. Positive controls are present in every one, so a guard that rejected every quantized `kv_b_proj` could not pass. The fixture hoisting into `affine_kv_b_proj_weights` / `block_float_kv_b_proj_weights` is faithful: no assertion, dtype, or geometry was dropped from the pre-existing `#958` tests.

Ran locally on this branch, all green: `models::kimi_linear` (4), `models::deepseek_v3::` (14), `models::deepseek_v32` (14), `models::longcat_flash_ngram` (2), `models::youtu_vl` (4), plus downstream consumers `models::glm_moe_dsa` (5) and `loading::` (230). `cargo fmt --all -- --check` and `cargo clippy --lib --tests -- -D warnings` clean.

### Scope discipline

Issue item 3 asked the implementer to evaluate a shared helper and report, not build one. No helper was built: the diff is 7 files, all under `src/models/`, no new function in `mlxcel-core`. All three stated blockers check out:

1. **youtu's extra `w_full` shape cross-check.** Correct, and understated. `youtu_vl_lm_sanitize.rs:113-125` cross-checks `w_full` against `[num_heads * head_dim, kv_lora_rank]`; dsv3, dsv32, longcat and kimi go straight from `w_full` to `reshape`. Unifying would not merely add a rejection to four families, it would convert an MLX `reshape` throw (an uncatchable abort) into a load error for them. That is an improvement, and exactly the kind that should be the headline of its own PR.
2. **`slice` + `swap_axes` vs `slice_axis` + `transpose_axes`.** Correct, and correctly called free. `mlxcel-core/src/utils.rs:43` shows `slice_axis` maps `end == -1` to the full axis length, which is what kimi's explicit `v_last_dim` computes.
3. **kimi does not `copy()` the slices contiguous.** Correct. dsv3, dsv32, longcat and youtu all `copy()` both `wk` and `wv`; kimi inserts the `swap_axes` result and the raw `slice` result directly.

### Remaining items (reported, not fixed)

- **MEDIUM, pre-existing, out of scope.** `mlxcel_core::layers::QuantizedMultiLinear::forward` (`src/lib/mlxcel-core/src/layers.rs:2856`) and `forward_no_transpose` (`:2878`) hardcode `"affine"` with a possibly-null biases pointer, and `from_weights` (`:2825`) takes `.biases` as an unvalidated `Option` while `MultiLinear::from_weights` (`:3279`) gates the quantized branch on `.scales` alone. Reached by DeepSeek V3, DeepSeek V3.2, GLM4 MoE Lite and LongCat Flash NGram. The reachability claim in the PR body holds structurally: all four sanitizers `continue` when `embed_q.weight` already exists, so a checkpoint shipping a pre-decomposed quantized `embed_q` bypasses the fix here and lands there. Same defect class as #1026 but a different crate and a different key, and the issue's acceptance criteria do not name it. Correctly flagged rather than silently fixed or silently skipped; worth a follow-up issue.
- **LOW, introduced here, no change requested.** `validate_quantization_biases` in `MultiLinear::from_weights` is a tautology: `mode` is derived from `biases.is_some()` on the line above, and `infer_quantization_mode` only ever returns one of the four modes `validate_quantization_mode` accepts, so neither half can fire. Keeping it is the right call. The 10-line comment says outright that it cannot fire and why it is there, so it misleads nobody, and the `gpt_oss.rs` `ExpertLinear` precedent it cites is the load-bearing version of the same pairing (there the mode is declared, so the check has teeth). Noted only because the comment is longer than the code it guards.
- **LOW, pre-existing.** `youtu_vl_lm_sanitize.rs:113` still carries an `M2:` review-artifact label. This PR converged the `M1:` one on the biases site; `M2:` sits on the shape cross-check, which has no sibling in the other four, so leaving it is defensible.
- **LOW, pre-existing.** The comment "the separate shape cross-check below the solve" in the non-quantized leg of `quantized_kv_b_proj_rejects_a_kv_lora_rank_no_packing_can_describe` describes a check that exists only in youtu. It is accurate in `youtu_vl_lm_tests.rs:262` and was copied to the other four by the #958 tests, before this PR (present on `main` at `deepseek_v3.rs:2051`). This PR only re-plumbed those tests onto the hoisted fixture.
- **LOW.** The dense leg of `kimi_linear_multi_linear_infers_its_quantization_mode_from_the_biases_plane` asserts `!loaded.is_quantized` but not `loaded.mode == "affine"`. The field is inert on that path, so this is cosmetic.

### Verification

- [x] All stated requirements implemented (6 of 6 acceptance criteria, plus issue item 3 evaluated and reported)
- [x] Integrated into project code flow (every production caller propagates the new error to the load failure)
- [x] Project conventions followed (inline format args, `ok_or_else` over `unwrap`, `thiserror`-free `Result<_, String>` matching the surrounding loaders)
- [x] Existing modules reused where applicable (`infer_quantization_mode`, `validate_quantization_biases`, `infer_mla_quantization_params`, no reimplementation)
- [x] No unintended structural changes (no signature changed, no file moved, no new abstraction)
- [x] Tests pass

inureyes:

Threat model applied: a hostile or malformed checkpoint reaching model-loading code that feeds checkpoint tensors and config-declared integers across the cxx bridge, where a C++ throw is an uncatchable `std::terminate` and a Rust panic in the server is a process-level denial of service.

**Verdict: approve. No CRITICAL and no HIGH findings, and none introduced by this PR.** Four MEDIUM/LOW items below, all reported rather than fixed; three are pre-existing and one is documentation-only.

### 1. The panic is removed, not moved

Confirmed at all five sites. Between the `.scales` gate and `dequantize` the only remaining `.unwrap()`s are the two that are provably `Some`: `weights.remove(&kv_b_key)` and `weights.remove(<scales key>)`, each preceded by a `contains_key` on the identical key with no intervening mutation of `weights`. The diff adds no `.unwrap()`, no `.expect()`, no `panic!`, no direct slice indexing and no new arithmetic to any non-test line; every such token in the added lines is inside a comment or a test. `infer_quantization_mode` is pure and total, and `validate_quantization_biases` returns `Result`.

### 2. Resource handling on the new early return: clean

All four `sanitize_weights` take `mut weights: WeightMap` (`HashMap<String, UniquePtr<MlxArray>>`) by value. On the `?` path, `w` and `s` (already removed) and `weights` itself drop normally, each `UniquePtr` running the cxx deleter exactly once. The raw pointer `&*b as *const _` is constructed only on the success path, is non-owning, and is consumed synchronously by `dequantize`; nothing is `forget`ten or `into_raw`ed. No leak, no double free, no null dereference, no `UniquePtr` misuse. The partial mutation is unobservable because the map was moved in.

### 3. Error message: clean, and safer than the brief assumed

The interpolated `b_key` is not checkpoint-derived at any of the five sites. It is synthesized from the loop index and literals (`format!("model.layers.{l}.self_attn")` and friends) and is only ever used as a lookup key, so its length is bounded by `log10(num_hidden_layers)` and it echoes no attacker-controlled bytes. Rust's `format!` takes its format string as a compile-time literal, so `{b_key}` is an argument and not a directive: no format-string injection is expressible even if the key were attacker-controlled. No path, no secret, no internal address.

### 4. The guard's completeness, and the prior reviewer's bypass claim

**Confirmed, and it is broader than stated. Pre-existing, not introduced here.** `mlxcel_core::layers::QuantizedMultiLinear` reaches `quantized_matmul` / `dequantize` with `"affine"` hardcoded and a null `biases` pointer (`src/lib/mlxcel-core/src/layers.rs:2857`, `:2874`, `:2890`). `MultiLinear::from_weights` (`:3288`) gates on `.scales` alone and `QuantizedMultiLinear::from_weights` (`:2833`) takes `biases` as an `Option` with no `validate_quantization_biases` call. There is no global bias-plane sweep anywhere in the load path.

Two routes, not one:

- The `embed_q.weight` bypass the prior reviewer identified, at `deepseek_v3.rs:1123`, `deepseek_v32.rs:907` and `longcat_flash_ngram.rs:1083`. A pre-decomposed quantized `embed_q` skips decomposition and is loaded directly.
- **GLM4 MoE Lite has no `kv_b_proj` sanitizer at all.** It loads `embed_q` / `unembed_out` straight from the checkpoint through the shared `MultiLinear::from_weights` (`glm4_moe_lite.rs:326-330`). No bypass is needed; that is the front door.

Failure mode is MLX's `validate_mode_with_type` throwing `Biases must be provided for affine quantization` at the first MLA forward, which crosses cxx as `UniquePtr<MlxArray>` rather than `Result` and is therefore an uncatchable abort, strictly worse than the Rust panic this PR removes. Severity MEDIUM: latent, since it needs a checkpoint shipping a pre-decomposed block-float MLA and nothing shipped does. Correctly and honestly declared out of scope in the PR body; a follow-up issue is the right home for it, not a rider here.

### 5. Performance: clean, and it is load-time

All five sites are inside `sanitize_weights`, which runs once per model load, and `MultiLinear::from_weights` runs once per layer at construction. Nothing here is on the decode path. Per layer the change is O(1). The `format!` for `b_key` is not a new allocation: the pre-PR code already built the same `String` inline as the `remove` argument, so hoisting it to a binding is the same single allocation, now also reachable by the error closure, which itself allocates only on failure. `infer_quantization_mode` returns `&'static str` and allocates nothing. The new `mode` field adds 16 bytes to a per-layer struct. In `MultiLinear::forward`, `self.mode` replaces a `"affine"` literal with a `&'static str` load at identical cost.

### Remaining findings

**MEDIUM, pre-existing.** The `QuantizedMultiLinear` sibling described in item 4.

**MEDIUM, pre-existing.** Four of the five sites hand `w_full` to `reshape(&[num_heads, head_dim, -1])` and `slice_axis(.., qk_nope_head_dim, ..)` without cross-checking the tensor against the config first. Only `youtu_vl_lm_sanitize.rs:112-118` validates `[num_heads * head_dim, kv_lora_rank]` before reshaping. A mismatch is an MLX throw, so an uncatchable abort rather than a load error. `dequantize` likewise does not validate the packed dtype: a `kv_b_proj.weight` shipped as FLOAT32 throws, which this PR's own test fixtures document in their comments. Related: `head_dim = (qk_nope_head_dim + v_head_dim) as i32` is a `usize` add over two untrusted config fields, so it panics in an overflow-checked build and wraps in release before truncating. The PR body identifies exactly this asymmetry as its reason for declining the shared-helper refactor, which is the right call: that refactor would add a rejection to four families and change a fifth's contiguity, and deserves to be its own headline.

**LOW, introduced here, documentation only.** The doc comment at `kimi_linear.rs:216-224` and the matching PR body section state the private `MultiLinear` quantized branch is "not reachable today". Under this threat model it is reachable. The kimi decomposition block is entered only when `self_attn.kv_b_proj.weight` is present (`kimi_linear.rs:1313`); a checkpoint shipping `self_attn.embed_q.weight` plus `self_attn.embed_q.scales` and no `kv_b_proj.weight` skips it entirely, and `MlaAttention::from_weights` (`:565`) then hands those checkpoint-supplied planes to the private loader with `is_quantized == true`. The `mode` change is therefore a live hardening of the same class as the `.biases` fix, not merely a bound on future edits. The code is correct either way; the risk is only that a future maintainer reads "not reachable" and removes it as dead code.

**LOW, not a regression.** The inferred mode is not cross-checked against MLX's mode-fixed geometry. MLX requires `group_size == 32` for `mxfp4` and `mxfp8` and `16` for `nvfp4`, while `infer_quantization_mode(false, 64, 8)` returns `"mxfp8"`, which the new `kimi_linear_multi_linear_infers_its_quantization_mode_from_the_biases_plane` pins explicitly. Kimi's private `MultiLinear` calls only `validate_expert_quantization_params`, a range check, and never `reconcile_quantization_layout`, so such a plane still reaches `quantized_matmul` and aborts. This is not a regression: the same input aborted unconditionally before, so the change strictly reduces the abort surface, and the gap is shared by every `infer_quantization_mode` caller in tree rather than being specific to this PR.

**LOW, acknowledged in the PR body.** For a genuine block-float export, "the checkpoint may be corrupted or only partially converted" mis-describes a well-formed but unsupported checkpoint. Wording consistency across the five sites is the better property here, so the trade the PR made is the right one.

### Notes

No code changes were made: nothing reached the CRITICAL or HIGH bar that this pass auto-fixes, and every MEDIUM is pre-existing and outside the changed files. Working tree left clean.

PR #1055, opened 2026-08-06
fix(cuda): call libdevice cosf/sinf in the fused RoPE kernel

### LINKED ISSUE #1049: fix(cuda): fused_rope_append diverges from the graph RoPE reference at offsets near 2^17

inureyes:
## Summary

The CUDA branch of the fused q/k RoPE + KV-append kernel called the SFU intrinsics `__cosf` / `__sinf`, while MLX's own `mlx/backend/cuda/rope.cu` calls `cos(float)` / `sin(float)`, which resolve to libdevice `cosf` / `sinf`. Both sides compute a bit-identical fp32 `theta`, so the trig call was the entire divergence. The SFU range reduction only carries enough bits for the documented accuracy over `[-pi, pi]`, and RoPE angles are unreduced, so the error grew with the absolute position and surfaced only at the two large-offset cases in `fused_rope_parity_tests`.

The fix is to call what MLX calls. The tolerance is not touched. Metal is not touched: `rope.metal` uses `metal::fast::cos` / `metal::fast::sin`, the Metal body matches it, and both degrade together.

## Root cause, against a float64 host reference

The decisive measurement is the `p = 0` dimension pair, where `d = 0` so `inv_freq` is exactly 1.0 on host and device alike and `theta` equals the absolute position with no fp32 representation error to confound the comparison. The applied phase is recovered from each path's own output as `arg(r1 + i*r2) - arg(x1 + i*x2)`, averaged over the 32 query heads that share one `theta`.

Applied-phase error in radians versus the exact float64 angle:

| offset | `__cosf`/`__sinf` (before) | `cosf`/`sinf` (after) | `theta * 2^-23` |
|---|---|---|---|
| 4096 | -1.80e-4 | -5.00e-5 | 4.88e-4 |
| 16384 | -6.65e-4 | -1.09e-6 | 1.95e-3 |
| 32768 | -1.33e-3 | -1.68e-5 | 3.91e-3 |
| 65536 | -2.66e-3 | 4.32e-6 | 7.81e-3 |
| 98304 | -7.01e-3 | -2.86e-7 | 1.17e-2 |
| 131071 | -1.16e-2 | 7.81e-6 | 1.56e-2 |

The libdevice column stays at order 1e-5 rad with no trend. Over the full 12-offset sweep in the benchmark doc, which spans a 128x change in angle, it scatters between 2.9e-7 and 5.0e-5 rad and puts its largest entry at offset 4096 rather than at 131071. That is the floor the method predicts: fp16 output rounding of about 2^-11 relative gives one head roughly 5e-4 rad, and averaging the 32 heads that share one `theta` brings it under 1e-4 rad. The SFU column instead grows about 300x over that same range. So MLX was the accurate side and the fused kernel was not, which is what licenses fixing the kernel rather than raising the tolerance.

Whole-tensor deviation from float64 (q, B=1 L=1, the same normalized rms / max the parity test asserts on) at offset 131071: `__cosf` 3.24e-3 / 3.61e-2, `cosf` 1.81e-3 / 2.27e-2, fp16 floor 2.09e-4 / 9.66e-4. Both columns sit above the fp16 floor because `theta` itself is an fp32 value at that magnitude; that part is computed identically by both GPU paths and cancels in the fused-versus-graph comparison, which is why the `p = 0` table is the part that matters.

## Error versus offset

Growth that tracks the offset, every point inside the `theta * 2^-23` envelope (ratio to the envelope 0.28 to 0.90), sign stable, tabulated at 12 offsets from 1024 to 131071 in the benchmark doc. No threshold and no wraparound. It is not monotonic point to point: 49152, 81920 and 114688 each read higher than the next offset up, by up to 24%, which is what an error that depends on where the angle falls inside the SFU's reduction period looks like. The trend across the range is unambiguous and that is what the issue asked to distinguish. The high max-to-rms ratio the issue noted is explained by `theta` falling off as `base^-d` across the head dimension: only the low-`p` pairs carry a large enough argument to lose accuracy.

## Agreement with the graph path after the fix

Raw output bytes against `slice` / `reshape` / `transpose` / `fast_rope`:

| shape | offset | q bytes differing |
|---|---|---|
| B=1 L=1 | 0, 4096, 131071 | 0 / 8192 each |
| B=1 L=3 | 131069 | 1 / 24576 |
| B=1 L=17 | 65536 | 2 / 139264 |
| B=1 L=64 | 131000 | 26 / 524288 |

Single-token decode is byte-identical at every offset. Multi-token windows differ on about 0.005% of elements at the last fp16 bit, which is nvcc's FMA-contraction freedom over `x1*costheta - x2*sintheta` in two separately compiled kernels, and is exactly the divergence class the tolerance was written for.

## Cost

Four arms interleaved inside one process (three trig choices plus the unfused graph as an unchanged control), arm order rotated each round, 9 repetitions per shape, across three quiet-host sessions. Interleaving was necessary: an earlier split-across-builds A/B moved the control arm by 4x between runs and was unusable.

Three of the eight sequence lengths swept are reported: the decode shape and the two prefill lengths where the arms were stable enough across sessions for a ratio to mean anything. The intermediate lengths were too noisy on the control arm to quote and are left out rather than averaged in.

Per-call microseconds, median of 9, three sessions:

| seq | `cosf`/`sinf` | `__cosf`/`__sinf` | graph control | ratio |
|---|---|---|---|---|
| 1 | 17.9 / 16.8 / 18.9 | 17.2 / 16.1 / 18.7 | 41.1 / 38.1 / 43.6 | 1.04x / 1.05x / 1.01x |
| 4096 | 1092 / 1028 / 961 | 1054 / 804 / 850 | 1390 / 1481 / 1453 | 1.04x / 1.28x / 1.13x |
| 8192 | 1701 / 1593 / 1544 | 1504 / 1354 / 1394 | 2251 / 2237 / 2186 | 1.13x / 1.18x / 1.11x |

At the decode shape the kernel is launch-bound and the change costs 0.2 to 0.7us of 17-19us, smaller than the control arm's own drift across the same sessions (38.1 to 43.6us). At prefill scale it costs 11-18% at seq 8192, where all four arms are tight, on a kernel that is bandwidth-bound (about 200 MB in 1.6 ms, roughly half this host's bandwidth). The seq 4096 row is looser at 4%, 28% and 13%: its `__cosf`/`__sinf` arm alone moved 804 to 1054us between sessions, a spread wider than the effect, so that row bounds the cost rather than measuring it. Across every prefill cell reported the cost spans 4% to 28% and the figure to carry forward is the seq 8192 one. The fusion still beats the graph it replaces in every row, so the simple fix stands.

`sincosf` was measured as a third arm and lands on the SFU arm's time, recovering essentially the whole cost. It was not adopted, and the terms of that trade are recorded in the benchmark doc rather than asserted, because they are worth revisiting. Against it: `sincosf` is not bit-identical to the two separate calls (one differing q element at L=17, offset 65536), and byte-identical single-token decode against the `fast_rope` graph is the property this fix rests on. For it: the cost is not confined to microbenchmarks. `Attention::forward` in `src/models/llama3.rs` calls `forward_fused_rope_append` for every window, prefill as well as decode, with no `l == 1` guard, so the seq 8192 percentage is paid on the wired path whenever the flag is on and the checkpoint is non-quantized; only at the decode shape is the saving genuinely invisible. Bit-identity wins while the path ships opt-in and off by default, so nobody pays the prefill percentage today. If the default is ever flipped, `sincosf` is the first thing to re-measure.

## Token-level impact near 131072

`qwen2.5-0.5b-bf16` (non-quantized, so it takes the fused path), 130967-token prompt, greedy, 128 tokens generated at absolute positions 130967 to 131094, with the two kernel variants selectable in one binary:

| Measure | Value |
|---|---|
| Tokens identical, SFU vs libdevice | 128 / 128 |
| Chosen-token logprob shift | max 0.1875, median 0.0000, mean 0.0142 nats |
| Top1 minus top2 margin | min 0.125, p10 0.500, median 4.75 nats |
| Steps with margin below the largest observed shift | 2 / 128 |

Measurable in the logits, no token flips here, but 2 of 128 steps sat inside its range so a flip is not structurally excluded. The reason the effect is small in token terms: a phase error `phi` at frequency `inv_freq` is indistinguishable from a position error of `phi / inv_freq`, and the SFU error scales as `position * inv_freq * 2^-23`, so the equivalent position error is about `position * 2^-23` for every dimension, roughly 0.016 tokens at position 131072.

## Correction to the issue text

The issue and its technical-considerations section state that the fused kernel defaults ON. It does not: `FUSED_ROPE_APPEND_DEFAULT` in `src/lib/mlxcel-core/src/layers.rs` is `false`, and the kernel is opt-in via `MLXCEL_FUSED_ROPE_APPEND=1`. Nothing shipped ran the buggy kernel by default, so there is no production long-context exposure to account for and this fix repairs no user-visible defect. It also means the kill-switch A/B the issue suggested was not available as described: with the default off, `=0` is the shipping behavior, so every measurement above runs the fused path explicitly under `MLXCEL_FUSED_ROPE_APPEND=1` and compares the two trig variants inside it. This PR does not change the default; flipping it is a separate decision.

Separately, `examples/fused_norm_rope_microbench` on GB10 measures the fusion at 1.02x to 2.06x across the nine-cell sweep, every cell at or above parity, unlike the Apple Silicon sweep that set the default to off. Flipping the default is a separate decision with its own evidence bar and is not part of this fix; the numbers are recorded in the benchmark doc so the question is not lost.

## What changed

- `src/lib/mlx-cpp/turbo/fused_rope_append.cpp`: CUDA body calls `cosf` / `sinf` instead of `__cosf` / `__sinf`. The Metal body is unchanged. The block comment above the CUDA body is rewritten: it dropped the stale UNVALIDATED note (which also pointed at the wrong parity suite, `fused_norm_parity_tests.rs`, the norm one, and at a test name, `fused_rope_append_matches_graph_rope`, that no file defines), states the rule the file follows (call whatever MLX's RoPE kernel for the same backend calls), and records why this substitution must not be reverted.
- `docs/benchmark_results/fused-rope-cuda-trig-gb10-2026-08-06.md`: the full measurement record.

No tolerance change, no test change, no Metal change.

## Test plan

- [x] `cargo test --release --features cuda -p mlxcel-core --lib fused_rope_parity -- --test-threads=1` on GB10 (sm_121): 9 passed, 0 failed. Same command reproduced the 7-passed / 2-failed baseline on `018b5c16` before the change, with the exact numbers the issue quotes.
- [x] Float64 host-reference probe, both GPU paths against the same reference: 12 offsets from 1024 to 131071 for the `p = 0` applied phase, 5 offsets from 511 to 131071 for the whole-tensor deviation.
- [x] Byte-level agreement with the `fast_rope` graph across six shape and offset combinations.
- [x] Four-arm interleaved timing A/B, 9 repetitions per shape, 8 shapes, 3 sessions; 3 shapes reported, the rest too noisy on the control arm to quote.
- [x] End-to-end greedy generation at positions 130967 to 131094 on `qwen2.5-0.5b-bf16`.

## Tolerance premise, corrected in `11104b32`

`assert_close` in `src/lib/mlxcel-core/src/fused_rope_parity_tests.rs` justified its 2e-3 / 1.2e-2 budget with "a pair of fp32 multiply-adds around `fast::cos` / `fast::sin` in both paths, so the only divergence available is the final rounding to the activation dtype". After this change the CUDA path calls no `fast::` intrinsic, and the byte-level table above shows the real residual is nvcc's FMA contraction across two separately compiled kernels rather than dtype rounding alone.

The tolerance itself is unchanged and correct; only the comment behind it was wrong. It now states what actually bounds the budget, names the per-backend rule (the same function MLX's own RoPE kernel uses for that backend), and records that the premise was wrong once, so a future large-offset failure reads as the trig calls having drifted apart again rather than as a tolerance that needs widening. Issue #1049 named that premise directly and rules out widening as the resolution, so leaving the stale rationale in place would have preserved the exact misreading the issue warns about.

The parity suite still passes 9/9 after the edit (6.43s), and `cargo fmt --all -- --check` is clean.

Closes #1049


inureyes:

### Intent

Replace the CUDA SFU intrinsics in the JIT'd fused RoPE + KV-append kernel with libdevice `cosf`/`sinf` so the fused path matches MLX's `rope.cu`, and justify that against a float64 reference rather than by loosening the tolerance.

The two-line code change is correct and the float64 methodology holds up. The findings are all in the evidence write-up.

### Verified, no finding

- **The `p = 0` isolation is genuine.** At `p = 0` the kernel computes `d = 0.0`, so `exp2f(-0.0 * log2(base))` is exactly 1.0, `rope_params[1]` is scale 1.0, and every position up to 131071 is below 2^24 and therefore exact in fp32. `theta` carries zero representation error at that pair, so the residual is the trig call alone. This is the one claim the whole PR rests on and it is sound.
- **Averaging over 32 heads is legitimate and load-bearing.** All 32 query heads share one `theta` at fixed `(t, p)`. Each head's phase is recovered from its own fp16 output pair, so per-head noise is about 2^-11 relative, roughly 5e-4 rad; averaging 32 of those predicts a floor under 1e-4 rad, which is where the libdevice column actually sits (2.9e-7 to 5.0e-5). Method and data agree.
- **It really is float64, not path-versus-path.** Each arm's applied phase is recovered from its own input/output pair and compared to the exact angle, which is what #1049 required.
- **The residual claim is maintained everywhere.** The doc, the PR body and the C++ comment all keep "fused now matches MLX, and MLX is as accurate as an fp32 angle allows" and none of them slips into "fused now matches float64".
- **No leftover A/B scaffolding.** The only `MLXCEL_` token in the diff is a deleted line. The temporary switch is gone.
- **No tolerance change.** `assert_close` is still `nrms < 2e-3 && nmax < 1.2e-2` and no test file is touched.
- **Metal untouched.** `metal::fast::cos` / `metal::fast::sin` still stand at lines 168-169.
- **The comment's backend rule is factually right.** `mlx/backend/cuda/rope.cu` calls `cos(float)`/`sin(float)`; `mlx/backend/metal/kernels/rope.metal` calls `metal::fast::cos`/`sin`. It now points at `fused_rope_parity_tests.rs`, which exists and holds the suite.
- **The rule is not at risk from fast math.** MLX's NVRTC path passes only `--gpu-architecture` plus include paths, and there is no `use_fast_math` anywhere in MLX's CUDA backend or CMake, so `cosf` will not be remapped to `__cosf` behind the fix. The same convention already exists at `mlx_cxx_kernels.cpp:1631` ("Precise expf (not `__expf`)").
- **Default not flipped**, no AI attribution, no em dashes, arithmetic in the tables checks out (envelope ratios 0.28-0.90, 200 MB / 1.6 ms is 126 GB/s, the fusion beats the graph in all nine cells).

### Findings Addressed

- [x] **(HIGH) Prefill cost understated and inconsistent with its own table.** "11-25%" against cells of 3.6 / 27.9 / 13.1 / 13.1 / 17.7 / 10.8%: the top was 3 points low and the bottom silently excluded the most favourable cell. Now reports 11-18% at seq 8192 where the arms are tight, flags seq 4096 as noise-bounded (its `__cosf` arm alone moved 804 to 1054us between sessions, wider than the effect), and gives 4-28% as the honest span.
- [x] **(HIGH) Three of eight swept shapes reported, with no disclosure.** The test plan said 8 shapes, the table showed 3, and nothing connected them. Both the doc and the body now state the selection and its reason before the table.
- [x] **(HIGH) The `sincosf` rejection rested on a false scoping claim.** "invisible at the decode shape the path is wired for" is not what the code does: `Attention::forward` in `src/models/llama3.rs` calls `forward_fused_rope_append` for every window with no `l == 1` guard, so prefill runs it and the seq 8192 percentage is real. The trade is now written out on both sides, with the trigger for revisiting it (a default flip or a prefill-heavy profile) named.
- [x] **(MEDIUM) "Growth is smooth" contradicted by the table.** 49152, 81920 and 114688 each read higher than the next offset up, by up to 24%. Rewritten as a trend inside the envelope with no threshold, no wraparound and no sign change, which is what #1049 asked to distinguish, and the non-monotonicity is explained by the SFU reduction period rather than smoothed over.
- [x] **(MEDIUM) The libdevice floor quoted 5x tighter than measured.** "flat at the 1e-5 rad floor" against a column spanning 2.9e-7 to 5.0e-5. Now states 5e-5 as the bound and derives it instead of asserting it.
- [x] **(MEDIUM) `docs/environment-variables.md` still attributed the CUDA rotation to `rope.metal`.** True of the layout and formula, not of the trig call after this change. The row now states the per-backend rule and links the new benchmark doc, which nothing referenced before.
- [x] **(LOW) "pointed at a test file name that does not exist" is itself wrong.** `fused_norm_parity_tests.rs` exists; it is the norm suite, the wrong file. The test name `fused_rope_append_matches_graph_rope` is the part nothing defines. Corrected in the body. The same sentence is in the commit message and cannot be fixed without a rewrite, but this repo squash-merges with the PR body, so what lands on `main` is correct.
- [x] **(LOW) Offset-count inconsistency.** Body said "15 offsets from 1024", test plan said "15 offsets from 0", the doc tabulates 12 from 1024. Body now matches the doc: 12 offsets for the `p = 0` phase, 5 for the whole-tensor deviation.

### Remaining Items

- [ ] **(MEDIUM) The tolerance rationale in `fused_rope_parity_tests.rs:81` is now stale.** `assert_close` still justifies its budget with "a pair of fp32 multiply-adds around `fast::cos` / `fast::sin` in both paths, so the only divergence available is the final rounding to the activation dtype". After this change the CUDA path calls no `fast::` intrinsic, and this PR's own byte-level table shows the real residual is nvcc FMA contraction across two separately compiled kernels. #1049 named this premise as not covering the case. Not fixed here because editing a `.rs` file would dirty the fingerprint of the crate currently under `cargo test`; recorded in the PR body under "Not addressed here".
- [ ] **(LOW) The block comment does not mention `sincosf`.** A future reader who reaches only the "must not be reverted" note will not learn that `sincosf` was measured and rejected. One line pointing at `docs/benchmark_results/fused-rope-cuda-trig-gb10-2026-08-06.md` would close it. Deferred for the same rebuild reason.
- [ ] **(LOW) The comment says `cosf`/`sinf` "stayed at the 1e-5 rad measurement floor".** 5e-5 is the measured bound. Same deferral.
- [ ] **(LOW) Table rounding.** seq 1 session 2 is 16.8 / 16.1 = 1.043, printed as 1.05x. Harmless if the underlying medians carry more digits than the table.

### Verification

- [x] Integrated into project code flow (kernel string is the one the JIT path compiles; benchmark doc now reachable from `environment-variables.md`)
- [x] Project conventions followed (matches the existing precise-math rule at `mlx_cxx_kernels.cpp:1631`; benchmark filename matches `{topic}-{host}-{date}.md`)
- [ ] Tests pass: `cargo test --release --features cuda -p mlxcel-core --lib fused_rope_parity` was not run in this review. The build lock was held and the review was restricted to non-building tools, so the 9/9 result is taken from the PR's test plan and not independently reproduced. No commit in this review touches Rust, C++, or any tolerance, so the suite's result is unchanged by it.

PR #1057, opened 2026-08-07
fix(tests): make tp_e2e scaling efficiency assertions deterministic

### LINKED ISSUE #1050: [nightly-verify] main is red
The nightly `make verify` run on `self-hosted-macos-26-arm64` failed. Check the first failing step in the run log before reading the gate table below: a setup step that fails leaves the gates reporting `failure` for want of a toolchain rather than because the code is red.

| Field | Value |
|---|---|
| `verify` job result | `failure` |
| `make verify-fmt` | `success` |
| `make verify-clippy` | `success` |
| `make verify-test` | `failure` |
| Trigger | `schedule` |

Duration of the `verify` job: **54m12s**.

Commit: `892806f383c539f182712e0217de9751dbf2ba78`
Run: https://github.com/lablup/mlxcel/actions/runs/31035477592

An `unknown` step outcome means the job published no outcome for that step. A `timeout-minutes` cancellation still publishes them, so `unknown` points at the runner being lost outright rather than at the budget.

Reproduce locally with `make verify` (or the single failing target above).
This issue is reused by subsequent nightly reports of the same kind while it stays open.

inureyes:
## Summary

This fixes the nightly-verify tracker issue: the Aug 5 nightly run failed `make verify-test` because of a flaky assertion in `tests/tp_e2e.rs`, not a code regression. The assertion is replaced with a deterministic check of the same computation.

## Root cause

The nightly `make verify` run on `self-hosted-macos-26-arm64` (run https://github.com/lablup/mlxcel/actions/runs/31035477592, commit `892806f383c539f182712e0217de9751dbf2ba78`) failed only in `make verify-test`, with 1 of 36 tests in `tests/tp_e2e.rs` failing: `e2e_scaling_analysis_basic` panicked at `tests/tp_e2e.rs:674` with "scaling efficiency should be near or below 1.0" (the assertion `r.scaling_efficiency <= 1.1`).

`run_scaling_analysis` in `src/distributed/tensor_parallel/benchmark.rs` measures wall-clock time via `spin_wait` busy-loops at microsecond scale and `Instant::now()`. `scaling_efficiency` is `throughput(TP=n) / (throughput(TP=1) * n)`. With the test config (`layer_compute_time` 200us, `allreduce_time` 20us) the nominal efficiency is around 0.83 at TP=2 and 0.71 at TP=4, but if the TP=1 baseline run gets preempted on a loaded CI runner (parallel test threads, scheduler jitter), its measured throughput deflates and the apparent TP=2/4 efficiency inflates past the 1.1 bound. This is a busy-wait microbenchmark flake, not a regression: the assertion and the benchmark code were unchanged since the test was introduced in `5dc894ca`, and the prior nightly run was green.

## Fix

`e2e_scaling_analysis_basic` no longer asserts an absolute wall-clock-dependent bound on `scaling_efficiency`. Instead it recomputes the expected efficiency directly from the throughputs the analysis itself recorded, using the same `baseline * tp_size / first_tp` formula that `run_scaling_analysis` uses, and asserts the stored `scaling_efficiency` matches within `1e-9`, plus that it is finite and positive. This validates the efficiency computation rather than the CI scheduler, so it cannot flake under load while still catching any regression in the formula itself. A comment in the test explains why an absolute bound is not asserted. No changes were made to `src/distributed/tensor_parallel/benchmark.rs`, and the ratio assertion in `e2e_crossover_larger_models_benefit_more` (line 742) was left untouched.

## Verification

Ran on the target machine (Linux aarch64, GB10) with the pinned toolchain and the release profile with the cuda feature, per project conventions:

`cargo test --release --features cuda --test tp_e2e` -> `test result: ok. 36 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.29s`

`cargo clippy --release --features cuda --test tp_e2e -- -D warnings` -> completes with `Finished \`release\` profile [optimized] target(s)` and no clippy diagnostics on the Rust code (the only warnings printed are pre-existing MLX C++ build warnings from the `mlxcel-core` native build, unrelated to this change).

Closes #1050

### DIFF
--- a/tests/tp_e2e.rs
+++ b/tests/tp_e2e.rs
@@ -692,6 +692,39 @@ fn e2e_scaling_analysis_basic() {
             "tp_size={tp_size}: scaling efficiency {actual} should be finite and positive"
         );
     }
+
+    // Physical invariant: scaling efficiency cannot exceed 1.0, because
+    // all-reduce overhead (present only for tp_size > 1) can only add time.
+    // `spin_wait` busy-loops until `elapsed >= duration`, so it only ever
+    // overshoots its requested duration, and `Instant`-based measurement
+    // only ever adds overhead. So measured throughput can only be <= the
+    // throughput implied by a zero-overshoot, zero-overhead nominal timing
+    // model -- deterministically, regardless of runner load. Compute the
+    // tp_size=1 nominal throughput analytically from the config, mirroring
+    // `run_tp_benchmark`'s accounting exactly, and check that no result's
+    // measured throughput exceeds `nominal_throughput(1) * tp_size`.
+    let batch = config.batch_sizes.first().copied().unwrap_or(1);
+    let seq_len = config.seq_lengths.first().copied().unwrap_or(128);
+    let prefill_steps = (seq_len + batch - 1) / batch.max(1);
+    let step_time_1 = config.layer_compute_time.as_secs_f64();
+    let nominal_time_1 = (prefill_steps + config.decode_steps as usize) as f64 * step_time_1;
+    let tokens_1 = (batch * config.decode_steps as usize) as f64;
+    let nominal_throughput_1 = tokens_1 / nominal_time_1;
+
+    for r in &analysis.results {
+        if r.tp_size <= 1 {
+            continue;
+        }
+        let nominal_bound = nominal_throughput_1 * r.tp_size as f64;
+        let ratio = r.throughput_tok_per_sec / nominal_bound;
+        assert!(
+            ratio <= 1.0 + 1e-9,
+            "tp_size={}: measured throughput {} exceeds the deterministic nominal \
+             upper bound {nominal_bound} (ratio={ratio})",
+            r.tp_size,
+            r.throughput_tok_per_sec,
+        );
+    }
 }
 
 #[test]
--- a/tests/tp_e2e.rs
+++ b/tests/tp_e2e.rs
@@ -678,15 +678,18 @@ fn e2e_scaling_analysis_basic() {
     let baseline = analysis.results[0].throughput_tok_per_sec;
     let first_tp = analysis.results[0].tp_size as f64;
     for r in &analysis.results {
-        let ideal = baseline * r.tp_size as f64 / first_tp;
+        let tp_size = r.tp_size;
+        let ideal = baseline * tp_size as f64 / first_tp;
         let expected = r.throughput_tok_per_sec / ideal;
+        let actual = r.scaling_efficiency;
         assert!(
-            (r.scaling_efficiency - expected).abs() < 1e-9,
-            "scaling efficiency should match baseline*tp_size/ideal computation"
+            (actual - expected).abs() < 1e-9,
+            "tp_size={tp_size}: stored scaling efficiency {actual} should equal \
+             throughput/ideal {expected}"
         );
         assert!(
-            r.scaling_efficiency.is_finite() && r.scaling_efficiency > 0.0,
-            "scaling efficiency should be a finite positive number"
+            actual.is_finite() && actual > 0.0,
+            "tp_size={tp_size}: scaling efficiency {actual} should be finite and positive"
         );
     }
 }
--- a/tests/tp_e2e.rs
+++ b/tests/tp_e2e.rs
@@ -669,11 +669,24 @@ fn e2e_scaling_analysis_basic() {
     let factor_1_to_2 = analysis.scaling_factor(0, 1).unwrap();
     assert!(factor_1_to_2 > 0.0);
 
-    // Scaling efficiency should be <= 1.0 (due to AR overhead).
-    for r in &analysis.results[1..] {
+    // Recompute the expected scaling efficiency from the throughputs the
+    // analysis itself recorded and compare against the stored value. This
+    // validates the efficiency computation rather than the wall-clock
+    // measurements: throughputs come from a spin_wait-based simulation, so
+    // an absolute bound (e.g. "efficiency <= 1.1") is nondeterministic on
+    // loaded CI runners where the tp_size=1 baseline run can be preempted.
+    let baseline = analysis.results[0].throughput_tok_per_sec;
+    let first_tp = analysis.results[0].tp_size as f64;
+    for r in &analysis.results {
+        let ideal = baseline * r.tp_size as f64 / first_tp;
+        let expected = r.throughput_tok_per_sec / ideal;
         assert!(
-            r.scaling_efficiency <= 1.1,
-            "scaling efficiency should be near or below 1.0"
+            (r.scaling_efficiency - expected).abs() < 1e-9,
+            "scaling efficiency should match baseline*tp_size/ideal computation"
+        );
+        assert!(
+            r.scaling_efficiency.is_finite() && r.scaling_efficiency > 0.0,
+            "scaling efficiency should be a finite positive number"
         );
     }
 }


inureyes:

### Intent
> Replace the wall-clock-dependent `scaling_efficiency <= 1.1` assertion in `e2e_scaling_analysis_basic` with a deterministic check of the same computation, so nightly `make verify-test` stops flaking under runner load.

### Verification of the fix

The test recomputation is an exact mirror of `run_scaling_analysis` (`src/distributed/tensor_parallel/benchmark.rs:548-557`): `ideal = baseline * tp_size / first_tp`, `efficiency = throughput / ideal`, with the same operand order, so the comparison is bit-exact and the `1e-9` tolerance is generous rather than load-bearing. `analysis.results[0].tp_size` is the same value as `config.tp_sizes[0]` because `run_scaling_analysis` pushes results in config order, so sourcing `baseline` and `first_tp` from `results[0]` is faithful.

Regression sensitivity was checked empirically with a mutation: making `run_scaling_analysis` refresh `baseline_throughput` on every iteration (instead of only the first) makes `e2e_scaling_analysis_basic` fail at `tests/tp_e2e.rs:683`. The removed `<= 1.1` bound would not have caught that mutation, since the mutated efficiency is 0.5. The source file was restored and the suite re-verified green afterwards.

The assertion no longer reads any wall-clock quantity except through values that appear on both sides of the equality, so it is timing-independent. Widening the loop to include `results[0]` also adds a real check that the baseline entry is normalized to exactly 1.0.

### Findings Addressed
- [x] Assertion messages carried no values, so a CI failure would show which invariant broke but none of the numbers behind it, the same triage gap that made the original nightly failure expensive to diagnose. The messages now name `tp_size` and print the stored and recomputed efficiencies, matching the style of `e2e_benchmark_tp2_faster_than_tp1_with_heavy_compute`. The old text "baseline*tp_size/ideal computation" also mixed the ideal formula with the division applied to it, and now reads `throughput/ideal`. (LOW, commit `f1fbdea0`)

### Remaining Items (no action required)
- [ ] The `ideal > 0.0` guard in `run_scaling_analysis` is not mirrored by the test. It is unreachable here because line 665 already asserts every throughput is positive, so a zero throughput fails earlier with a clearer message. (LOW)
- [ ] `first_tp` is 1.0 in every scaling-analysis test config, so a regression that dropped the `/ first_tp` divisor would still pass. Pre-existing coverage gap, not introduced by this PR. (LOW)
- [ ] Three sibling assertions remain wall-clock comparative: `tests/tp_e2e.rs:755`, `tests/tp_e2e.rs:571`, and the 30s bound at `tests/tp_e2e.rs:850`. Their nominal margins are roughly 60 percent, 96 percent, and 2500x, against the roughly 32 percent margin that actually flaked, so they are far less exposed. Worth a follow-up only if the nightly re-reds on one of them. (LOW)

### Verification
- [x] Integrated into project code flow (test runs under `make verify-test`)
- [x] Project conventions followed (inline format args, epsilon comparison over `==`)
- [x] No unintended structural changes (test-only; `src/distributed/tensor_parallel/benchmark.rs` untouched)
- [x] Tests pass: `cargo test --release --features cuda --test tp_e2e` 36 passed, `cargo clippy --release --features cuda --test tp_e2e -- -D warnings` clean, `cargo fmt --check` clean

No CRITICAL or HIGH findings.

inureyes:

Scope: `git diff main...HEAD` (`tests/tp_e2e.rs`, +20/-4). Verified on the PR branch with the pinned toolchain, release profile, cuda feature, narrow target.

**No CRITICAL or HIGH findings.** Nothing security relevant in this diff: test only, no `unsafe`, no I/O, no network, no process spawning, no dependency change, no `src/` change. No performance impact either: the diff adds arithmetic over a 3 element vector and no additional `spin_wait` or benchmark run, so test runtime is unchanged.

Independent verification on this branch:

- `cargo test --release --features cuda --test tp_e2e` -> `36 passed; 0 failed`, 0.29s
- `cargo clippy --release --features cuda --test tp_e2e -- -D warnings` -> clean

The root cause analysis holds. With `layer_compute_time` 200us and `allreduce_time` 20us the nominal step times are 200us / 120us / 70us, giving efficiency 0.833 at tp=2 and 0.714 at tp=4, and nominal tp=1 total time is about 27ms (128 prefill steps plus 8 decode steps). The flake is one directional: `spin_wait` can only overshoot, so preemption can only make a measured run slower than nominal. Because the measured tp=1 throughput is the denominator of `scaling_efficiency`, only baseline deflation can inflate the ratio, and breaching the old 1.1 bound needs roughly 9ms of extra stall inside the tp=1 run. That is plausible on a loaded self hosted runner where the 36 tests in this file busy spin in parallel and oversubscribe cores.

### MEDIUM: the new assertion drops the invariant the old one encoded

The replacement recomputes the implementation's own expression and compares, so it verifies arithmetic consistency but no longer verifies the modeled physics. The regression class the old bound caught, efficiency climbing past 1.0 because the all reduce overhead model or the tp_size scaling broke, is now uncovered in this test. It is also a change detector: a future edit to the formula in `benchmark.rs` accompanied by a matching edit here would pass.

This is not blocking, and the flake fix is correct as far as it goes. A deterministic way to keep the invariant, if you want it back in a follow up: nominal efficiency depends only on the configured durations, `layer_compute / (tp * (layer_compute / tp + allreduce))`, since the prefill and decode step counts cancel out of the ratio. Asserting `measured_throughput(tp) / (nominal_throughput(1) * tp) <= 1.0 + eps` leaves only one measured quantity, in the numerator, where `spin_wait` overshoot can push the value down but never up. That bound cannot flake under load and still fails if the model regresses. It needs `nominal_throughput(1) = total_tokens / ((prefill_steps + decode_steps) * layer_compute_time)` with `prefill_steps = seq_len / batch_size`, so the test would set `seq_lengths` and `batch_sizes` explicitly rather than inheriting them from `Default`.

### LOW: the `ideal > 0.0` guard is not mirrored

`run_scaling_analysis` computes `if ideal > 0.0 { throughput / ideal } else { 0.0 }`, but the test computes `let expected = r.throughput_tok_per_sec / ideal;` unguarded. If a baseline throughput of 0.0 were ever reachable, `expected` would be NaN while `actual` would be 0.0, and the failure message would read "should equal NaN" instead of naming the real problem.

Not reachable in practice: the preceding loop already asserts `r.throughput_tok_per_sec > 0.0` for every result, which pins `baseline > 0.0` and therefore `ideal > 0.0` before this block runs. Reporting it for the record rather than requesting a change, since a guard here would be dead code. Worth noting separately that the `else { 0.0 }` branch in `benchmark.rs` stays untested, which is a pre-existing gap and not something this PR introduced.

### LOW: tp=1 iteration is self referential, and assertion order

Widening the loop from `results[1..]` to `results` adds a tp=1 iteration where `ideal == baseline` and `expected == 1.0`, matching the value the implementation assigns by construction. It can only fail if the baseline is zero or NaN, so it mostly pins "baseline efficiency is exactly 1.0". Harmless.

Minor: the finite and positive check runs after the equality check. If `actual` were NaN the equality assert fires first with the less useful message. Swapping the two would give the better diagnostic.

### Checked and clear

- `analysis.results[0]` indexing is guarded by the preceding `assert_eq!(analysis.results.len(), 3)`, so no out of bounds panic.
- The `1e-9` tolerance is generous but harmless. Both sides evaluate the identical f64 expression in the identical order, and Rust does not contract to FMA by default, so the difference is exactly 0.0.
- `usize as f64` casts are lossless at these magnitudes.
- `e2e_crossover_larger_models_benefit_more`, deliberately left untouched, is not the same flake class. `run_crossover_analysis` overrides the compute and all reduce times per hidden size (`200 * f^2` us and `50 * f` us), so nominal efficiency is about 0.5 at hidden 2048 against 0.8 at hidden 8192. That 0.3 gap is far outside scheduler noise.
- The other wall clock assertions in this file have similar margins: about 2x for `e2e_benchmark_tp2_faster_than_tp1_with_heavy_compute`, 16x for `e2e_benchmark_ttft_scales_with_seq_length`, and a 30s budget against a few ms of work for `e2e_ci_benchmark_completes_quickly`. No further flake candidates found in this target.

Verdict: no blocking findings. The two LOW items and the MEDIUM coverage note are for the author's judgment.

Issue #854, opened 2026-07-22
feat(models): Florence-2 vision-language fusion + full weight loading (sub of #850)


inureyes:
## Summary

The fusion stage that joins the DaViT vision features (#853) with the BART text stack (#852), plus complete weight loading for the whole Florence-2 model. Florence-2 concatenates image features with the task-prompt embeddings to form the encoder input rather than scattering into placeholder slots.

## Part of epic

#850

## Dependencies

#852 (BART seq2seq engine + text core), #853 (DaViT vision backbone)

## Upstream reference

- florence2.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/florence2.py
- vision.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/vision.py
- language.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/language.py
- config.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/config.py

## Architecture notes

- LearnedPositionEmbedding2D: adds 2D positions to the image features.
- `image_projection`: matmul projecting the image feature dim to text `d_model`.
- `_encode_image`: DaViT features to 2D positions to projection.
- `_merge_input_ids_with_image_features`: CONCATENATE image features with the task-prompt embeddings to form the encoder input, and build the joint attention mask. Florence-2 concatenates rather than scattering into placeholder slots.
- `from_weights` + `sanitize`: BART encoder-decoder weight-key naming, DaViT conv channels-last remap, mixed precision (vision often bf16 while text is quantized).

## Implementation plan

- Consume the vision tower from #853 and the encoder / decoder from #852.
- Implement fusion in the Florence-2 top-level model module under `src/models/florence2/`.
- Complete `from_weights` / `sanitize`: handle the BART encoder-decoder key naming and the DaViT conv channels-last remap, and keep the vision weights bf16 while the text weights are quantized (mixed precision), mirroring the mixed-precision handling in existing VLM loaders.

## Acceptance criteria

- [x] `_encode_image` (LearnedPositionEmbedding2D + image_projection) implemented.
- [x] `_merge_input_ids_with_image_features` concatenates image + prompt embeddings and builds the joint attention mask.
- [x] `from_weights` + `sanitize` complete: BART naming, DaViT channels-last remap, mixed precision.
- [x] Weights from a real checkpoint load without shape errors.
[...]

PR #1063, opened 2026-08-07
feat(models): Florence-2 DaViT vision backbone

### LINKED ISSUE #853: feat(models): Florence-2 DaViT vision backbone (sub of #850)
## Summary

The DaViT hierarchical vision backbone for Florence-2: dual attention (spatial windowed plus channel) over 4 stages, producing the image features that the fusion stage concatenates with the task prompt.

## Part of epic

#850

## Dependencies

none

## Upstream reference

- vision.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/vision.py

## Architecture notes

- ConvEmbed patch embedding per stage: `patch_size = [7,3,3,3]`, `patch_stride = [4,2,2,2]`, `patch_padding = [3,1,1,1]`, with `patch_prenorm`.
- DepthWiseConv2d positional convs: `conv_at_attn`, `conv_at_ffn`.
- SpatialBlock: windowed attention (window_partition / window_reverse, `window_size = 12`, WindowAttention).
- ChannelBlock: ChannelAttention.
- 4 stages: `depths = [1,1,9,1]`, `num_heads = [4,8,16,32]`, `qkv_bias`, `mlp_ratio = 4`.
- Output image features with `hidden_size = 768`.
- Parse `VisionConfig` (`model_type: davit`).

## Implementation plan

- Add the DaViT encoder following the tree's vision-encoder layout under `src/vision/encoders/` (and/or the Florence-2 family directory `src/models/florence2/`).
- Reuse windowed-attention helpers under `src/vision/encoders/`: the window_partition / window_reverse patterns from qwen2_5_vl and kimi_vl.
- Reuse the conv layer helpers used by the FastViT and gemma3n conv paths for ConvEmbed and DepthWiseConv2d.
- DaViT conv weights need a channels-last remap in the sanitize step. Mirror the conv channels-last remap already used by the FastViT / gemma3n loaders.

## Acceptance criteria

- [x] `VisionConfig` (`model_type: davit`) parses.
- [x] ConvEmbed, DepthWiseConv2d, SpatialBlock (WindowAttention), and ChannelBlock (ChannelAttention) implemented across the 4 stages.
- [x] DaViT conv weights channels-last remap handled in sanitize.
- [x] DaViT forward on a real image produces features matching the mlx-vlm reference (per-stage shapes plus numerical parity).
- [x] Integration: the backbone is exposed through the API that the fusion sub-issue (#S3) calls, so it plugs into the model forward path rather than living as a standalone script. Full model-path wiring is completed by #S3, which depends on this.

## Effort

MEDIUM-HIGH.

inureyes:
## Summary

Implements the Florence-2 DaViT ("Dual Attention Vision Transformer") vision backbone, the pixel-side half of the model that will feed the BART seq2seq engine landed in #1060. Four stages, each a strided `ConvEmbed` patch embedding followed by `depths[i]` blocks that alternate windowed spatial attention with channel attention, every half sandwiched between two depthwise positional convolutions. For the `Florence-2-base-ft` geometry and a 768x768 input the progression is 192x192 @ 128 -> 96x96 @ 256 -> 48x48 @ 512 -> 24x24 @ 1024, so the backbone emits `[B, 576, 1024]`.

Scope is the backbone only. Fusion / projection, the processor, and end-to-end validation are separate sub-issues that depend on this one.

## What changed

- `src/vision/encoders/florence2_davit.rs` (new): `Florence2VisionConfig` (parse + validate), the `sanitize` channels-last remap, `Florence2DaViT` with `from_weights` / `load` / `forward` / `forward_stages`, and `impl VisionEncoder`.
- `src/vision/encoders/florence2_davit_blocks.rs` (new): `ConvEmbed`, `DepthWiseConv2d`, `Mlp`, `SpatialBlock`, `ChannelBlock`, `Block`.
- `src/vision/encoders/florence2_davit_attention.rs` (new): `WindowAttention` and `ChannelAttention`. Split out of the blocks file to stay under the 500-line limit.
- `src/vision/encoders/florence2_davit_tests.rs` (new): checkpoint-free unit tests for config parsing / validation, the sanitize pass and its idempotency, a synthetic-weight forward through the window pad / crop path, and the missing-weight error path.
- `tests/florence2_vision_parity.rs` (new): numerical parity against the mlx-vlm reference on the real checkpoint, gated on checkpoint presence.
- `src/vision/encoders/deepseekocr_sam.rs`: `window_partition` / `window_unpartition` reused rather than duplicated; their `// Used by:` comments now name Florence-2 DaViT, per `docs/code-guidelines.md`.
- `src/vision/encoders/mod.rs`, `src/models/florence2/mod.rs`, `src/models/mod.rs`: module registration and re-export, so the fusion sub-issue can reach `Florence2DaViT` from the Florence-2 family directory alongside `Florence2TextModel`.

## API for the fusion sub-issue

`Florence2VisionConfig::from_model_config(&Value)` (descends into `vision_config`, also accepts a bare vision-config object, mirroring `Florence2TextConfig::from_model_config`), `Florence2DaViT::from_weights(&WeightMap, &Florence2VisionConfig, prefix)` with `FLORENCE2_VISION_PREFIX = "vision_tower."`, `Florence2DaViT::load(&Path)` (config + `vision_tower.*` tensors + sanitize + bf16 -> f16 per the Apple Silicon precision policy), `forward` -> `[B, N, dim_embed[-1]]`, `forward_stages` -> per-stage `(tokens, (H, W))`, and `impl VisionEncoder`.

Every one of these was checked against `models/Florence-2-base-ft-bf16` (config.json and the actual safetensors keys and shapes) and against the upstream source, not inferred from the issue prose.

1. **`vision_config.model_type` is the empty string `""`, not `"davit"`.** The issue says to parse a `VisionConfig` with `model_type: davit`. Gating on that literal rejects every real Florence-2 checkpoint. Parsing here accepts `""`, `"davit"`, and an absent key (matching the upstream `VisionModel.__init__` guard, which allows `["davit", ""]`) and rejects anything else; the family is identified by the parent `model_type: florence2` plus the structural fields. Covered by a unit test that asserts all three spellings parse and a bogus one errors.
2. **There is no `hidden_size` in `vision_config`, and the backbone output is not 768.** The issue says "output image features with `hidden_size = 768`". The checkpoint has `projection_dim: 768` at both the top level and inside `vision_config`, while the last stage's `dim_embed` is `1024`, and `image_projection` is a `[1024, 768]` tensor. So the backbone emits `[B, N, 1024]`; the 1024 -> 768 projection is `image_projection` + `image_proj_norm` and belongs to the fusion sub-issue. `projection_dim` is parsed and retained on the config but never applied here. (Upstream's `VisionConfig` dataclass does carry a `hidden_size: int = 768` field, but `vision.py` never reads it.)
3. **The `window_partition` / `window_reverse` reuse pointer in the issue is wrong.** The issue says to reuse the patterns "from qwen2_5_vl and kimi_vl". Those encoders window differently. The helpers with matching semantics live only in `src/vision/encoders/deepseekocr_sam.rs`, and that is what this PR reuses (including the pad-inside-partition / crop-inside-unpartition contract, which lets the reference's explicit `mx.pad` and `x[:, :H, :W, :]` steps drop out).
4. **The issue omits four present, load-bearing config fields.** `num_groups: [4,8,16,32]`, `image_pos_embed: {type: "learned_abs_2d", max_pos_embeddings: 50}`, `visual_temporal_embedding: {type: "COSINE", max_temporal_embeddings: 100}`, and `image_feature_source: ["spatial_avg_pool","temporal_avg_pool"]`. Only `num_groups` affects this forward: it is `ChannelAttention.groups` per stage. The other three are consumed exclusively by `Florence2._encode_image` in the fusion stage, and upstream carries them on `ModelConfig` rather than `VisionConfig` while the real checkpoint puts them inside `vision_config`. All four are parsed and retained on `Florence2VisionConfig` so the fusion sub-issue reads them off one struct.
5. **The upstream `ChannelAttention` inline comment mis-states the attention shape.** It annotates `mx.matmul(q.transpose(0, 1, 3, 2), k)` as `(B, groups, N, N)`. The operands are `(B, g, C/g, N)` and `(B, g, N, C/g)`, so the product is `(B, groups, C/groups, C/groups)`: channel-by-channel, which is the whole point of channel attention. This port follows the code, and the doc comment on `ChannelAttention` records the discrepancy so nobody "corrects" it back.
6. **The `mlx-community` bf16 export already ships conv weights channels-last, so `sanitize` is a no-op on it.** `vision_tower.convs.0.proj.weight` is `[128, 7, 7, 3]` and `...dw.weight` is `[128, 3, 3, 1]` in the actual file, i.e. already `(O, kH, kW, I)`. The pass is still implemented in full (with the reference's `check_array_shape` guard on `convs.*.proj.weight` and the `shape[1] < shape[-1]` guard on `blocks.*.dw.weight`) so a raw PyTorch export converts correctly, and a unit test asserts both the transposing and the idempotent branches.
7. **`ChannelAttention` scales queries by the token count, not the head dimension.** `q = q * (float(N) ** -0.5)` where `N` is up to ~37k at stage 0. Unusual enough to look like a bug; it is not, and it is commented as deliberate. The softmax there runs in f32 and casts back, since the reduction axis is long.
8. **`DropPath` is training-only.** `drop_path_rate: 0.1` is present in the checkpoint and is parsed onto the config, but `self.training` is false at inference so the module is the identity. Implemented as a no-op with a comment saying why the config value does not reach the forward.

`docs/supported-models.md` and `src/models/detection.rs` are deliberately untouched: this sub-issue does not make Florence-2 a loadable family (fusion does), so advertising it in the support table or wiring it into architecture detection now would claim a capability the binary does not have. Both belong with the fusion sub-issue.

## Validation

**Input choice.** A deterministic synthetic pixel tensor, not a decoded image. Florence-2 image preprocessing is the processor sub-issue's scope, so reproducing it byte-identically in both runtimes is not possible yet, and what parity actually requires is that both sides see identical input bytes. Both sides build `x[0, c, h, w] = ((h * 768 + w) * 3 + c) % 251 / 251.0 - 0.5` in f32 (every step a single IEEE-754 operation over exactly represented integers, so the two implementations agree bit for bit) and then cast to f16.

**Reference.** The mlx-vlm florence2 `VisionModel` (https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/vision.py) run in a scratch venv against `models/Florence-2-base-ft-bf16`, with `vision_tower.*` weights put through `VisionModel.sanitize` and cast bf16 -> f16 to match mlxcel's Apple Silicon precision policy. Captured per-stage shape, first-16 activations, and mean / std; the regeneration recipe is recorded in the test's module doc.

**Tolerance.** Per-stage absolute, scaled to the activation magnitude at that stage, because both sides run f16 with different op ordering: `2e-3 / 1.2e-2 / 6e-2 / 1.2e-2` against stage peaks of `1.54 / 8.03 / 65.5 / 3.97`. Each is a small multiple of one f16 ulp at its stage's peak and at least 25x below that stage's own standard deviation. Nothing was loosened to make a failure pass; the tolerances were set after the first run already agreed.

**Observed max absolute deviation** (Apple Silicon, `test-fast` profile):

| Stage | Shape | Max abs deviation | Tolerance | Mean (ours / ref) | Std (ours / ref) |
|-------|-------|-------------------|-----------|-------------------|------------------|
| 0 | `[1, 36864, 128]` | 3.37e-4 | 2e-3 | -0.001657 / -0.001656 | 0.119353 / 0.119339 |
| 1 | `[1, 9216, 256]` | 4.29e-3 | 1.2e-2 | 0.010804 / 0.010804 | 0.290855 / 0.290888 |
| 2 | `[1, 2304, 512]` | 7.38e-4 | 6e-2 | 0.074498 / 0.074503 | 1.458913 / 1.458821 |
| 3 | `[1, 576, 1024]` | 3.14e-3 | 1.2e-2 | -0.487351 / -0.487382 | 0.524581 / 0.524704 |

## Test plan

- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib vision::encoders::florence2_davit` (7 passed)
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_vision_parity -- --nocapture` (2 passed)
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_text_parity` (1 passed, no regression from the re-export)
- [x] `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings` (clean)
- [x] `python3 scripts/ci/check_cross_repo_refs.py` (clean)

Closes #853


inureyes:

### Intent

> Port the Florence-2 DaViT vision backbone (four stages of ConvEmbed plus alternating windowed-spatial and channel attention) from the mlx-vlm reference, backbone only, exposed through an API that the fusion sub-issue #854 can call.

### Line-by-line faithfulness check against the reference

Compared `src/vision/encoders/florence2_davit{,_blocks,_attention}.rs` against the mlx-vlm `VisionModel` (https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/vision.py). Every item below was traced independently, not taken from the PR description.

- **qkv split**: `split_qkv` reproduces `reshape(B,N,3,H,D).transpose(2,0,3,1,4)` exactly via slice + reshape + `transpose(0,2,1,3)`.
- **WindowAttention**: head-dim scale `(dim/num_heads)^-0.5`, `q @ k^T`, softmax over the last axis, `transpose(0,2,1,3)` before the reshape, then window merge and crop. Matches.
- **ChannelAttention query scale**: `(N as f32).powf(-0.5)` on the token count, matching `q * (float(N) ** -0.5)`. Present, and the doc comment on the struct plus the inline comment both mark it deliberate.
- **ChannelAttention output shape**: `matmul(q^T, k)` over `(B,g,C/g,N)` and `(B,g,N,C/g)` gives `(B,g,C/g,C/g)`. The transpose chain `matmul(attn, v^T) -> (0,1,3,2) -> (0,2,1,3) -> reshape(B,N,C)` is character-for-character the reference chain. The doc comment records that the upstream inline `(B, groups, N, N)` annotation is wrong, so nobody reverts it.
- **ConvEmbed pre-norm vs post-norm**: the 3-D branch norms before the reshape when `pre_norm`, the 4-D branch transposes NCHW to NHWC, and the post-conv norm runs only when `!pre_norm`. Both norm widths come from the checkpoint tensor itself (`convs.{i}.norm.weight`), so stage 0 gets `embed_dim` and stages 1-3 get `dim_embed[i-1]` by construction rather than by an index the code could get wrong. The synthetic-weight unit test builds each stage's norm at the width the branch requires, so a swap would fail the test.
- **GELU**: `mlxcel_core::gelu` resolves to `x * 0.5 * (1 + erf(x / sqrt(2)))` in `mlx_cxx_bridge.cpp:2648`, the exact erf form that `nn.GELU()` defaults to. Not `gelu_approx`, not `gelu_sigmoid`.
- **PreNorm residual with `norm = None`**: `conv1` / `conv2` are `add(x, conv.forward(x, size))`, so the residual is added on the depthwise path too.
- **Block ordering**: SpatialBlock is conv1, window_attn, conv2, ffn; ChannelBlock is conv1, channel_attn, conv2, ffn; Block is spatial then channel. Matches.
- **Weight key paths**: `conv1.fn.dw`, `window_attn.norm` / `window_attn.fn`, `ffn.norm` / `ffn.fn.net.fc1`, `blocks.{stage}.{depth}.{spatial,channel}_block` all match the reference module attribute names.
- **DepthWiseConv2d**: kernel 3, stride 1, padding 1, `groups = channels`, matching `DepthWiseConv2d(dim, 3, 1, 1)` with `groups=dim_in`.
- **sanitize**: `is_channels_last_conv` is `check_array_shape` verbatim. I checked the transposing branch by hand for the path with no real-checkpoint coverage: raw PyTorch `convs.0.proj.weight` `[128,3,7,7]` fails `kH == kW` and becomes `[128,7,7,3]`; `convs.1.proj.weight` `[256,128,3,3]` fails `128 == 3` and becomes `[256,3,3,128]`; `dw.weight` `[C,1,3,3]` satisfies `shape[1] < shape[3]` and becomes `[C,3,3,1]`. Every result is `(C_out, kH, kW, C_in/groups)`, which is what MLX `conv2d` wants. Re-running the pass on each result is a no-op, so `load` cannot double-transpose. The `-large` geometry was checked the same way and has no ambiguous case.
- **bf16 to f16**: `convert_bf16_weights` runs at the weight-loading boundary in `load`, after `sanitize`, exactly as `Florence2TextModel::load` does.
- **Config defaults**: cross-checked against the reference `VisionConfig` dataclass and against `models/Florence-2-base-ft-bf16/config.json`. `in_chans`, `window_size`, `mlp_ratio`, `qkv_bias`, `conv_at_attn`, `conv_at_ffn`, and `patch_prenorm` defaults all match upstream; the checkpoint genuinely omits `hidden_size`, `image_size`, and `num_classes`, and genuinely carries `model_type: ""`.

### Integration and conventions

- `Florence2DaViT`, `Florence2VisionConfig`, and `FLORENCE2_VISION_PREFIX` are re-exported through `src/models/florence2/mod.rs` and `src/models/mod.rs`, so #854 reaches both halves of the model from one place.
- The API shape matches `Florence2TextModel` on the same file: `from_model_config(&Value)`, `from_weights(&WeightMap, config, prefix) -> Result<Self, String>`, `load(&Path) -> Result<Self>`, `config()`. `load` even mirrors the text loader's body structure (`sanitize_config_json`, `load_weights_from_dir_filtered`, `convert_bf16_weights`).
- `impl VisionEncoder` is present and returns `VisionEncoderOutput { hidden_states }`; the unit test asserts the trait entry point agrees with the inherent `forward`.
- `window_partition` / `window_unpartition` are reused from `deepseekocr_sam.rs` rather than duplicated, and their `// Used by:` comments were updated. The reuse is semantically correct: the in-tree helpers pad with zeros inside `window_partition` and crop inside `window_unpartition`, which collapses the reference's explicit `mx.pad` and `x[:, :H, :W, :]` into the same behavior, including the reference's choice to let padded positions participate in the softmax.
- Plain `conv2d` rather than `try_conv2d` matches all 34 existing call sites in `src/`; `try_conv2d` is documented as being for the audio path whose shapes are runtime-derived.
- No `references/...` path, no bare `mlx_lm/...` path in the new files. The `mlx-vlm \`mlx_vlm/models/florence2/vision.py\`` doc line is the tree's established form (18 other files use it) and is paired with the full GitHub URL on the next line, so it is not a dead link.
- No em dashes and no AI attribution in the new files, the commit message, or the PR body.
- File sizes: 458 / 373 / 165 / 408 / 324 lines, all under the 500-line limit. Splitting the attention pair into its own file to stay under it was the right call.

### Scope

No scope creep found. `projection_dim`, `image_pos_embed`, `visual_temporal_embedding`, and `image_feature_source` are parsed and retained on the config but never read by the forward, which is correct for a backbone-only sub-issue. Nothing the backbone itself needs was deferred.

Leaving `docs/supported-models.md` and `src/models/detection.rs` untouched is the right judgment. `mlxcel list` is documented as the authoritative answer to "is this checkpoint supported?", and adding Florence-2 to the detection table now would make that answer wrong: without fusion and the processor the binary cannot load or run the family. Both belong with #854.

### Findings

No CRITICAL or HIGH findings. Nothing was changed; the branch is unmodified by this review.

- **LOW**: `ChannelAttention` uses `softmax_precise` (`mx.softmax(..., precise=true)`) where the reference uses the default non-precise softmax. This is a deliberate, commented numerical improvement for the long reduction axis, and parity holds at the pinned tolerances, so it is fine as-is. Worth knowing it is a conscious deviation and not a transcription of the reference.
- **LOW**: the real-checkpoint parity test never exercises the window pad and crop path. At 768x768 the stage grids are 192, 96, 48, and 24, all multiples of `window_size = 12`, so `pad_h == pad_w == 0` at every stage. The synthetic unit test does drive the pad path (`window_size = 6` against 8x8 and 4x4 grids) but only asserts shapes, not values. Residual risk is low because the helpers already carry a round-trip test in `deepseekocr_sam.rs` and are exercised numerically by the DeepSeek-OCR SAM encoder.
- **LOW**: `from_weights` requires pre-sanitized weights and says so in its doc, but cannot enforce it (it borrows the map, while `sanitize` consumes one). On an `mlx-community` export a missed `sanitize` is a silent no-op; on a raw PyTorch export it is a hard conv shape failure. Worth a line in #854's checklist since that loader will read the checkpoint once and call both `from_weights` entry points.
- **LOW**: `num_groups` falls back to `num_heads` when absent, where the reference dataclass defaults to `[4, 8, 16, 32]`, and `drop_path_rate` defaults to 0.0 against the reference's 0.1. Both are inert for real checkpoints (the field is present, and drop path is training-only).
- **LOW**: `patch_size`, `in_chans`, `mlp_ratio`, and `qkv_bias` are parsed but never cross-checked against the loaded tensors, so a config that disagrees with the checkpoint surfaces as an MLX shape error rather than a named Rust error. Consistent with every other encoder in the tree, so not worth changing here.

### Verification

- [x] Integrated into project code flow (backbone re-exported and reachable; full model-path wiring is #854's deliverable, as the issue states)
- [x] Tests pass (7 unit, 2 real-checkpoint parity, 1 text parity; clippy and fmt clean)

inureyes:

Reviewed as model-inference code: the realistic threat is a checkpoint (`config.json` + safetensors) that a user downloaded, not a network or auth surface. No CRITICAL findings. Fixed in bae77b1f.

### HIGH: hostile config reaches MLX `conv2d` and aborts the process

`patch_stride`, `patch_padding`, and `patch_size` went from `config.json` straight into `mlxcel_core::conv2d`. MLX validates conv geometry eagerly at graph build and throws; the plain `conv2d` binding is not declared `-> Result`, so the throw crossed a boundary that cannot carry it and terminated the process. The tree already has `try_conv2d` for exactly this case (`src/audio/nemotron_h_nano_omni/encoder.rs:198`, `src/audio/encoder.rs:179`), but the vision forward path has to stay infallible for the `VisionEncoder` trait, so the check belongs at parse time instead. `validate()` now requires `patch_size >= 1`, `patch_stride >= 1`, `patch_padding >= 0`, and `in_chans >= 1`.

### HIGH: `depths` drove an unbounded allocation

`from_weights` sized a `Vec` from `config.depths[stage]` before looking up a single weight, so `"depths": [2147483647, ...]` became an allocation-failure abort rather than an error return. Bounded by `MAX_STAGE_DEPTH = 1024` against a real maximum of 9.

### HIGH: `from_weights` trusted an unvalidated config

`Florence2VisionConfig` has public fields, so a caller could build one that never went through `from_vision_config`. `num_heads` of zero is an integer divide-by-zero at `florence2_davit_attention.rs:80`. `from_weights` now re-runs `validate()` first.

### HIGH: the `sanitize` precondition was advisory only

`from_weights` documented that weights must already be sanitized but could not enforce it (it borrows the map while `sanitize` consumes one). On a raw PyTorch export the un-remapped conv weight reached `conv2d` and aborted. `ConvEmbed` and `DepthWiseConv2d` now check the weight against the channels-last layout `sanitize` produces, using the per-stage channel widths from the config, so the load fails with the offending key named and a pointer to `sanitize`. On `mlx-community` exports skipping `sanitize` is genuinely a no-op for conv layout (`is_channels_last_conv` already returns true), so the silent-wrong-output case was not reachable there; the hard-failure case is the one that mattered.

### LOW: `expect()` in library code

`split_qkv` drained a 3-element `Vec` through three `expect()` calls. Replaced with `std::array::from_fn`, which is infallible by construction.

### LOW: redundant materialization of the input tensor

`forward` and `forward_stages` opened with `mlxcel_core::copy(pixel_values)` purely to get an owned handle. `mlx::core::copy` is a real primitive (allocate plus copy kernel), and no stage read the copy. Removed. This is a correctness-neutral cleanup, not a measured speedup: it is one copy of a 3.4 MB pixel tensor against a backbone that runs 12 blocks, so it is below the noise floor of any honest measurement on this box.

### Assessed, no change needed

**`softmax_precise` in `ChannelAttention`.** Not a performance concern. In MLX, `precise` is an accumulation-type flag on the same fused softmax kernel, not an f32 round trip, and the output dtype is `at_least_float(input)` either way. It also does not run on a long axis: the attention matrix is `(B, groups, C/groups, C/groups)`, so the softmax reduction is 32 elements wide for every Florence-2-base stage. What runs over N (up to 36864 at stage 0) is the matmul that produces the logits, which is the reason the logits can be large and the reason f32 accumulation is cheap insurance. Keep it. One dead line: the following `astype(&attn, array_dtype(x))` is always a no-op, since `softmax_precise` already returns the input dtype. Harmless, left in place as defensive.

**Consecutive transposes at `florence2_davit_attention.rs:160-161`.** They compose to a single `transpose_axes(&out, &[0, 3, 1, 2])`, but MLX `Transpose` is a zero-copy stride rewrite, so collapsing them saves nothing. The subsequent `reshape` of a non-contiguous array is the one copy, and it happens either way. Left alone to keep the line-by-line correspondence with the upstream reference.

**No eval or `item_*` in the forward path.** Confirmed by grep across all three implementation files. The lazy graph is never forced mid-forward.

**No `unsafe` blocks.** Neither `einsum` nor `scaled_dot_product_attention` is used here.

### Remaining, not fixed

- **MEDIUM: rank assumption on `pixel_values`.** `forward` and `forward_stages` index `shape[2]` and `shape[3]` without checking rank, so a 3-D input panics with a bare index-out-of-bounds. `ConvEmbed::forward` makes the same assumption on its non-token-major branch. This is the convention across the encoder tree (`siglip.rs:374` and `:414` do the same), the `VisionEncoder` trait is infallible so it cannot be turned into a `Result` locally, and the input comes from the preprocessor rather than the checkpoint. Changing it is a tree-wide convention decision, not a PR-local one.
- **MEDIUM: wrong-shaped non-conv tensors still abort.** A LayerNorm or Linear weight with the wrong width fails inside MLX at eval rather than at load. Fixing it properly means threading an expected width through `layer_norm_from_weights` and `Linear::from_weights`, which is shared infrastructure and belongs in its own change. The conv weights, which are the ones `sanitize` actually rewrites, are covered.
- **Informational: `is_channels_last_conv` is ambiguous when `in_channels == kernel_size`.** Ported verbatim from the reference. It cannot trigger for this family (stage 0 is `in=3, k=7`; later stages are `in>=128, k=3`), and the new load-time shape check would not distinguish the ambiguous case either, since both readings give the same shape.

### Verification

Parity against `models/Florence-2-base-ft-bf16` reports deviations identical to the pre-fix run, with no tolerance touched: stage 0 3.4e-4 (tol 2e-3), stage 1 4.3e-3 (tol 1.2e-2), stage 2 7.4e-4 (tol 6e-2), stage 3 3.1e-3 (tol 1.2e-2). 10 unit tests pass (3 new: hostile patch geometry and depths, hand-built config revalidation, unsanitized conv weights), 2 parity tests pass, `clippy --lib --tests -D warnings` clean, `cargo fmt --check` clean.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: added `synthetic_forward_exercises_window_padding_at_real_geometry` to `src/vision/encoders/florence2_davit_tests.rs`. It builds the backbone at the real `Florence-2-base-ft` per-stage widths/heads/groups/patch settings (`window_size = 12`) with `depths` cut to one block per stage for speed, drives a synthetic 704x704 input whose stage grids (176/88/44/22) do not divide 12, and asserts the stage shape progression plus a NaN/Inf finiteness check on the output. This closes the coverage gap left by the checkpoint-gated parity test, whose 768x768 input lands every stage grid on an exact multiple of 12 and so never exercises `window_partition`'s pad branch or `window_unpartition`'s crop branch. Existing coverage (10 unit tests + 2 real-checkpoint parity tests) still passes unchanged.
- **Dead code**: removed the `astype(&attn, array_dtype(x))` after `softmax_precise` in `ChannelAttention::forward` (`src/vision/encoders/florence2_davit_attention.rs`). MLX's `Softmax` primitive on the reduce-last-axis fast path (always taken here since the axis is always `-1`) returns `at_least_float(input dtype)`, which already equals `attn`'s dtype, so the cast was a provable no-op. Replaced the "cast back" comment, which implied a dtype change that never happened, with an accurate one. Real-checkpoint parity is confirmed byte-behavior-identical before and after.
- **Documentation**: reviewed the module-level and field-level doc comments in `florence2_davit.rs`, `florence2_davit_blocks.rs`, and `florence2_davit_attention.rs`. They already state the API contract, name `projection_dim` / `image_pos_embed` / `visual_temporal_embedding` / `image_feature_source` as parsed-but-not-applied fusion-stage fields, and document that `sanitize` must run before `from_weights`, so no changes were needed there. Confirmed `docs/supported-models.md` and `src/models/detection.rs` remain untouched, per the deliberate scoping to issue #854/#855.
- **Cross-repo references**: `python3 scripts/ci/check_cross_repo_refs.py` is clean, and no new file cites `references/...`, bare `mlx_vlm/...`, or `mlx_lm/...` as a standalone path; all upstream citations use the full `https://github.com/Blaizzy/mlx-vlm/blob/main/...` URL form.

### Final verification (all pass)
- `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib vision::encoders::florence2_davit` — 11 passed (10 existing + 1 new)
- `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_vision_parity -- --nocapture` — 2 passed, deviations unchanged from before this change
- `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings` — clean
- `cargo fmt --check` — clean

Issue #856, opened 2026-07-22
feat(models): Florence-2 end-to-end integration and real-checkpoint validation (sub of #850)


inureyes:
## Summary

The capstone that wires Florence-2 into mlxcel's model detection, registration, and generation paths, then validates the whole model on a real checkpoint across caption + OCR + detection modes. No new model math lives here; it connects the pieces from #852, #853, #854, and #855 into the runtime and proves the integrated deliverable.

## Part of epic

#850

## Dependencies

#852 (seq2seq engine + text core), #853 (DaViT backbone), #854 (fusion + weight loading), #855 (processor + task prompts + location tokens)

## Upstream reference

- florence2.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/florence2.py
- config.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/config.py

## Architecture notes

- This sub-issue contains no new model math. It connects the pieces from #852 to #855 into the runtime and proves the integrated model.

## Implementation plan

- Add the `"florence2"` arm to `src/models/detection.rs`.
- Register in `src/model_metadata.rs` via `for_each_model_registration!`.
- Wire the seq2seq VLM decode path into `generate` / `generate_vlm`.
- Add the TP / distributed arch-string if applicable.
- Update `docs/supported-models.md`.
- Add `_tests.rs` unit tests (DaViT shapes, cross-attention, task-prompt parse).
- Validate on a real checkpoint (`mlx-community/Florence-2-base-ft-bf16` or `-4bit`) across caption + OCR + detection modes via `./target/release/mlxcel generate -m <model> --image <img> -p "<task prompt>"`, confirming `mlxcel list` reports florence2.
- Follow `docs/adding-models.md`.

## Acceptance criteria

- [x] `"florence2"` arm added to `src/models/detection.rs`; registered in `src/model_metadata.rs` via `for_each_model_registration!`.
- [x] seq2seq VLM decode path wired into `generate` / `generate_vlm`; `mlxcel list` reports florence2.
- [x] TP / distributed arch-string added if applicable.
- [x] `docs/supported-models.md` updated; `_tests.rs` unit tests added.
- [x] Real checkpoint loads and produces correct output for at least caption + OCR + detection modes.
- [x] Integration: the epic deliverable is a functionally integrated model reachable from the CLI and server, not standalone modules.

## Effort

MEDIUM.

[...]

Issue #852, opened 2026-07-22
feat(models): Florence-2 BART-style seq2seq encoder-decoder engine + text core (sub of #850)


inureyes:
## Summary

The reusable encoder-decoder inference path plus the Florence-2 text core: the architecturally new foundation for the epic. Florence-2 is BART-style seq2seq, which mlxcel does not have outside Whisper. This sub-issue builds the bidirectional encoder, the causal decoder with encoder cross-attention, and the dual KV cache that the seq2seq decode loop needs.

## Part of epic

#850

## Dependencies

none

## Upstream reference

- language.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/language.py
- config.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/config.py

## Architecture notes

- Bidirectional encoder: `Florence2EncoderLayer` = self-attention + LayerNorm, BART post-norm ordering.
- Causal decoder: `Florence2DecoderLayer` = causal self-attention + encoder cross-attention + LayerNorm.
- DUAL KV cache: a self-attention cache that grows per decode step, plus a cross-attention cache whose K/V are computed once from the encoder output and reused on every decode step.
- `shift_tokens_right`, `decoder_start_token_id = 2`.
- Learned absolute position embeddings, `max_position_embeddings = 1024`.
- GELU FFN, `d_model = 768`, 6 encoder layers + 6 decoder layers, 8 attention heads.
- Parse `TextConfig`.

## Implementation plan

- Use `src/models/whisper/` as the structural template. It is mlxcel's only encoder-decoder precedent (cross-attention decoder + encoder-output caching). Mirror its seq2seq decode loop and encoder-output cache, adapting from audio-to-text to token-to-token.
- Reuse the in-tree LayerNorm from `src/lib/mlxcel-core/src/layers.rs`.
- Add the Florence-2 text module under a new family directory `src/models/florence2/`.
- Implement the dual-cache decode: cross-attention K/V computed once from the encoder output and cached, self-attention K/V appended per step.

## Acceptance criteria

- [x] `TextConfig` (BART encoder-decoder) parses from a real Florence-2 config.
- [x] `Florence2EncoderLayer` and `Florence2DecoderLayer` implemented with BART post-norm ordering.
- [x] Dual KV cache (self-attention + cross-attention) implemented and exercised over multiple decode steps.
[...]

PR #1064, opened 2026-08-07
feat(models): Florence-2 vision-language fusion and full weight loading

### LINKED ISSUE #854: feat(models): Florence-2 vision-language fusion + full weight loading (sub of #850)

inureyes:
## Summary

Joins the DaViT vision tower (#1063) to the BART seq2seq engine (#1060) into `Florence2Model`, the assembled Florence-2 vision-language model, and completes weight loading for the whole checkpoint. Florence-2 concatenates rather than scattering: the tower output is pooled into a short feature sequence, projected into the text embedding space, and placed in front of the task-prompt embeddings to form the encoder input, with no image placeholder token in the prompt at all.

## What changed

- `src/models/florence2/fusion.rs` (new): `LearnedPositionEmbedding2D` (learned column half first, then row half, over a square feature grid) and `PositionalEmbeddingCosine1D` (the `visual_temporal_embed.pos_idx_to_embed` table, taken from the checkpoint when present and synthesized from the closed form otherwise), plus `additive_attention_mask`, which turns a 0/1 mask into the `[batch, 1, 1, seq]` additive form attention consumes.
- `src/models/florence2/checkpoint.rs` (new): `Florence2Config` (both halves plus `image_token_id`, defaulting to the text vocabulary size) and `sanitize`, which runs the DaViT conv channels-last remap over the whole weight map, drops the BART `final_logits_bias` buffer, and fills `model.shared.weight` from an `embed_tokens` copy for exports that materialize only the tied table.
- `src/models/florence2/model.rs` (new): `Florence2Model` with `encode_image` / `encode_image_features` (`_encode_image`: 2-D grid position embedding, temporal embedding, `image_feature_source` pooling, `image_projection` 1024 to 768, `image_proj_norm`), `embed_prompt`, `merge_input_ids_with_image_features`, `encode` / `encode_with_image_features` / `encode_fused`, `make_cache`, `decode`, and `generate_greedy`.
- `src/models/florence2/{encoder,layers}.rs`: the encoder and its blocks now take an optional additive attention mask, so the joint mask the fusion stage builds reaches the encoder's self-attention instead of being dropped. `encode_tokens` / `encode_embeds` pass `None`, leaving the text-only numerics unchanged; the new `Florence2TextModel::encode_embeds_with_mask` is what the fused path drives.
- `src/models/florence2/florence2_fusion_tests.rs` (new): 37 unit tests covering both positional embeddings, the additive mask, all three `sanitize` passes, config parsing, load-time validation, and the whole `pixels -> features -> concat -> encoder -> decoder logits` chain on a tiny synthetic model (one block-free DaViT stage plus a one-layer BART stack), so the integrated path is exercised without a checkpoint.
- `tests/florence2_fusion_parity.rs` (new): gated end-to-end parity against the mlx-vlm florence2 `Model` on `models/Florence-2-base-ft-bf16`, skipping cleanly when the checkpoint is absent.

## Precision and dtype handling

The fused path pins its activation dtype from the text stack and casts the projected image features to it, and casts pixel input to the vision tower's own dtype, so a checkpoint whose two halves land on different dense dtypes does not silently promote the fused sequence to f32. Weight loading applies the usual Apple Silicon policy of promoting bf16 to f16.

Quantized Florence-2 checkpoints are rejected at load with a named error rather than half-loaded. Neither half of this family has a quantized code path: the BART text stack and the DaViT tower are both built from dense `Linear` / `Embedding`, so a packed uint32 weight would reach MLX and abort the process instead of surfacing as an error. The check runs right after config parsing, before any weight file is read.

## Deviations and findings

- **Upstream `image_feature_source` order is wrong for real checkpoints.** mlx-vlm's `ModelConfig` reads `image_feature_source`, `image_pos_embed`, and `visual_temporal_embedding` from the *top level* of `config.json`, but every real Florence-2 checkpoint stores them inside `vision_config`. Upstream therefore falls back to its dataclass defaults, and its `image_feature_source` default is `["temporal_avg_pool", "spatial_avg_pool"]`, the reverse of the checkpoint's `["spatial_avg_pool", "temporal_avg_pool"]`. That would put the pooled token at the end of the 577-token feature sequence instead of the front. mlxcel reads the config (as the HuggingFace Florence-2 implementation does) and refuses to guess when the field is absent; the parity run passes the checkpoint's own values to the reference so both sides execute the same recipe. A unit test pins the ordering behavior.
- **The temporal embedding is not a no-op for still images.** With one frame it contributes row 0 of the cosine table, which is `sin(0), cos(0), ...` = `0, 1, 0, 1, ...`, not the zero vector. Skipping it would shift every odd channel by 1.0 before projection. Pinned by a unit test.
- **`image_projection` is a bare tensor, not a `Linear`.** Upstream registers it as a raw `[in, out]` parameter applied as a right-hand matmul, so it is loaded and applied without a transpose or bias.
- The issue text says `projection_dim: 768` is the fusion-stage width and the backbone emits `dim_embed[-1]`; the checkpoint config confirms both (1024 to 768). No correction needed there.
- The joint attention mask reaches the encoder's self-attention only. Decoder cross-attention still passes no mask, matching the fact that this path drives a single image and a single prompt with no padding; a genuinely padded batch would need the mask threaded through cross-attention as well, which is not implemented here. The doc comment on `merge_input_ids_with_image_features` states this.
- **The all-ones mask is skipped rather than applied.** `merge_input_ids_with_image_features` builds the mask the acceptance criteria ask for, but it is all ones by construction, and converting it would add a provably all-zero `[batch, 1, 1, seq]` tensor into every encoder layer's `[batch, heads, seq, seq]` logits (roughly 24M elementwise f16 adds per 585-token encode across 6 layers, all computing an identity). The general entry point is `encode_fused(inputs_embeds, attention_mask)`, which converts a caller-supplied mask; `encode_with_image_features` passes `None`. That removes provably redundant arithmetic rather than trading accuracy for speed, so no wall-clock figure is claimed and none was measured. `encoder_mask_is_identity_when_all_ones_and_load_bearing_otherwise` pins both directions: all-ones must equal `None`, and zeroing a position must change the output.
- **Untrusted `config.json` integers are range-checked before they reach MLX.** MLX throws C++ exceptions that cannot cross the `cxx` bridge, so an out-of-range shape aborts the process instead of returning `Err`. `max_temporal_embeddings` is bounded (it drives an eager host allocation on the synthesis fallback), the loaded `embed_positions` table is checked to actually cover `max_position_embeddings` on both encoder and decoder, and `Florence2TextConfig` now rejects non-positive `d_model` / head counts / layer counts and out-of-vocabulary `pad_token_id` / `decoder_start_token_id`. `additive_attention_mask` returns `Result` instead of indexing an unchecked rank.
- Runtime registration (`src/models/detection.rs`, `src/model_metadata.rs`, `docs/supported-models.md`, the `generate` / `generate_vlm` wiring, `mlxcel list`) is deliberately untouched: that is #856's scope.

## Test plan

- [x] `cargo test --profile test-fast --features metal,accelerate --lib models::florence2::`: 51 passed. 37 in the new `florence2_fusion_tests.rs`, and 14 in `florence2_tests.rs` (7 pre-existing text-engine tests still green after the encoder mask change, plus 7 new config-validation tests).
- [x] `cargo test --profile test-fast --features metal,accelerate --test florence2_fusion_parity`: 2 passed against the real `models/Florence-2-base-ft-bf16` checkpoint. Worst deviations vs the mlx-vlm reference: 2.2e-3 on image features (absmax 8.6), 1.5e-2 on encoder hidden states (absmax 45.8), 6.6e-3 on first-step logits (absmax 21.2). Mean and standard deviation agree to 1e-4 or better on all three, and the greedy token ids match exactly (`[0, 879, 27740, 868]`, decoding to `<s>unanswerable`, the sensible caption for a procedurally generated noise field).
- [x] `cargo test --profile test-fast --features metal,accelerate --test florence2_text_parity --test florence2_vision_parity`: 3 passed, confirming the #852 and #853 pins are unmoved.
- [x] `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings`: clean.
- [x] `python3 scripts/ci/check_cross_repo_refs.py`: OK.

Closes #854

### DIFF
--- a/src/models/florence2/checkpoint.rs
+++ b/src/models/florence2/checkpoint.rs
@@ -1,0 +1,109 @@
+// Copyright 2025-2026 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Florence-2 checkpoint plumbing: whole-model config parsing and the
+//! weight-key normalization that has to happen before either half of the
+//! model is built.
+//!
+//! Kept apart from `model.rs` because these two run once at load time and
+//! have nothing to do with the forward path.
+//!
+//! Reference:
+//! https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/florence2.py
+
+use anyhow::Result;
+use serde_json::Value;
+
+use mlxcel_core::weights::WeightMap;
+
+use crate::vision::encoders::florence2_davit;
+
+use super::{Florence2TextConfig, Florence2VisionConfig};
+
+/// Weight-key prefix of the language model inside a Florence-2 checkpoint.
+pub(crate) const FLORENCE2_TEXT_PREFIX: &str = "language_model.";
+
+/// Both halves of a Florence-2 `config.json` plus the fusion-stage fields.
+#[derive(Debug, Clone)]
+pub struct Florence2Config {
+    pub text: Florence2TextConfig,
+    pub vision: Florence2VisionConfig,
+    /// Placeholder id the processor may emit for the image. Florence-2 does
+    /// not actually splice image features into placeholder slots, so any
+    /// occurrence is dropped before the prompt is embedded. Upstream defaults
+    /// this to the text vocabulary size (51289), one past the last real id.
+    pub image_token_id: i32,
+}
+
+impl Florence2Config {
+    /// Parse a full Florence-2 `config.json` (`model_type: florence2`).
+    pub fn from_model_config(config: &Value) -> Result<Self> {
+        let text = Florence2TextConfig::from_model_config(config)?;
+        let vision = Florence2VisionConfig::from_model_config(config)?;
+        let image_token_id = config
+            .get("image_token_id")
+            .or_else(|| config.get("image_token_index"))
+            .and_then(Value::as_i64)
+            .map(|v| v as i32)
+            .unwrap_or(text.vocab_size);
+        Ok(Self {
+            text,
+            vision,
+            image_token_id,
+        })
+    }
+}
+
+/// Normalize a full Florence-2 checkpoint's weight keys.
+///
+/// Three passes, in order:
+///
+/// 1. The DaViT conv channels-last remap and `position_ids` drop from
+///    [`florence2_davit::sanitize`]. Both of its key guards are specific
+///    enough (`convs`/`proj.weight`, `blocks`/`dw.weight`) that running them
+///    over the whole checkpoint leaves `language_model.*` untouched, and both
+///    are idempotent, so this is a no-op on the `mlx-community` exports that
+///    already ship channels-last conv weights.
+/// 2. Drop `final_logits_bias`. HuggingFace BART carries it as a registered
+///    buffer of zeros; it is not a parameter and there is no LM-head bias in
+///    this implementation.
+/// 3. BART shared-embedding naming. The encoder and decoder token tables are
+///    tied to `model.shared` and exports vary in which of the three they
+///    materialize, so fill `model.shared.weight` from an `embed_tokens` copy
+///    when only the latter is present.
+pub fn sanitize(weights: WeightMap) -> WeightMap {
+    let weights = florence2_davit::sanitize(weights);
+
+    let mut out = WeightMap::with_capacity(weights.len());
+    for (key, value) in weights {
+        if key.contains("final_logits_bias") {
+            continue;
+        }
+        out.insert(key, value);
+    }
+
+    let shared = format!("{FLORENCE2_TEXT_PREFIX}model.shared.weight");
+    if !out.contains_key(&shared) {
+        let source = [
+            format!("{FLORENCE2_TEXT_PREFIX}model.encoder.embed_tokens.weight"),
+            format!("{FLORENCE2_TEXT_PREFIX}model.decoder.embed_tokens.weight"),
+        ]
+        .into_iter()
+        .find_map(|key| out.get(&key).map(|w| mlxcel_core::copy(w)));
+        if let Some(tensor) = source {
+            out.insert(shared, tensor);
+        }
+    }
+    out
+}
--- a/src/models/florence2/decoder.rs
+++ b/src/models/florence2/decoder.rs
@@ -51,6 +51,25 @@ impl Florence2Decoder {
             .ok_or_else(|| {
                 format!("Florence-2 weight not found: {prefix}.embed_positions.weight")
             })?;
+        // `Florence2Model` bounds both the fused encoder input length and the
+        // decode offset against `config.max_position_embeddings`, and those
+        // guards are only sound if the loaded table actually covers that
+        // bound. Without this check a `config.json` that inflates
+        // `max_position_embeddings` (or `d_model`) passes them and reaches MLX
+        // as an out-of-range slice, which throws across the cxx bridge and
+        // aborts the process instead of returning `Err`. Sum in i64 so the
+        // comparison cannot overflow on a hostile value.
+        let position_shape = mlxcel_core::array_shape(&embed_positions);
+        let required_rows = POSITION_OFFSET as i64 + config.max_position_embeddings as i64;
+        if position_shape.len() != 2
+            || position_shape[1] != config.d_model
+            || (position_shape[0] as i64) < required_rows
+        {
+            return Err(format!(
+                "Florence-2 {prefix}.embed_positions.weight: expected at least [{required_rows}, {}] (max_position_embeddings {} + POSITION_OFFSET {POSITION_OFFSET}), got {position_shape:?}",
+                config.d_model, config.max_position_embeddings
+            ));
+        }
         let layernorm_embedding = layer_norm(weights, &format!("{prefix}.layernorm_embedding"))?;
 
         let mut layers = Vec::with_capacity(config.decoder_layers as usize);
--- a/src/models/florence2/encoder.rs
+++ b/src/models/florence2/encoder.rs
@@ -77,7 +77,15 @@ impl Florence2Encoder {
 
     /// Encode `[batch, seq, d_model]` input embeddings into encoder hidden
     /// states of the same shape.
-    pub(crate) fn forward(&self, inputs_embeds: &MlxArray) -> UniquePtr<MlxArray> {
+    ///
+    /// `mask` is the additive attention mask (`[batch, 1, 1, seq]`, `0` for a
+    /// real key and `-inf` for a padded one) applied by every block's
+    /// self-attention. Unpadded input passes `None`.
+    pub(crate) fn forward(
+        &self,
+        inputs_embeds: &MlxArray,
+        mask: Option<&MlxArray>,
+    ) -> UniquePtr<MlxArray> {
         let seq = mlxcel_core::array_shape(inputs_embeds)[1];
         let pos = mlxcel_core::slice(
             &self.embed_positions,
@@ -87,7 +95,7 @@ impl Florence2Encoder {
         let x = mlxcel_core::add(inputs_embeds, &pos);
         let mut x = self.layernorm_embedding.forward(&x);
         for layer in &self.layers {
-            x = layer.forward(&x);
+            x = layer.forward(&x, mask);
         }
         x
     }
--- a/src/models/florence2/encoder.rs
+++ b/src/models/florence2/encoder.rs
@@ -56,6 +56,25 @@ impl Florence2Encoder {
             .ok_or_else(|| {
                 format!("Florence-2 weight not found: {prefix}.embed_positions.weight")
             })?;
+        // `Florence2Model` bounds both the fused encoder input length and the
+        // decode offset against `config.max_position_embeddings`, and those
+        // guards are only sound if the loaded table actually covers that
+        // bound. Without this check a `config.json` that inflates
+        // `max_position_embeddings` (or `d_model`) passes them and reaches MLX
+        // as an out-of-range slice, which throws across the cxx bridge and
+        // aborts the process instead of returning `Err`. Sum in i64 so the
+        // comparison cannot overflow on a hostile value.
+        let position_shape = mlxcel_core::array_shape(&embed_positions);
+        let required_rows = POSITION_OFFSET as i64 + config.max_position_embeddings as i64;
+        if position_shape.len() != 2
+            || position_shape[1] != config.d_model
+            || (position_shape[0] as i64) < required_rows
+        {
+            return Err(format!(
+                "Florence-2 {prefix}.embed_positions.weight: expected at least [{required_rows}, {}] (max_position_embeddings {} + POSITION_OFFSET {POSITION_OFFSET}), got {position_shape:?}",
+                config.d_model, config.max_position_embeddings
+            ));
+        }
         let layernorm_embedding = layer_norm(weights, &format!("{prefix}.layernorm_embedding"))?;
 
         let mut layers = Vec::with_capacity(config.encoder_layers as usize);
--- a/src/models/florence2/florence2_fusion_tests.rs
+++ b/src/models/florence2/florence2_fusion_tests.rs
@@ -602,6 +602,44 @@ fn from_weights_rejects_mismatched_image_projection() {
     assert!(err.contains("image_projection"), "unexpected error: {err}");
 }
 
+/// `image_projection` has no fallback (it is a bare tensor with no `Linear`
+/// wrapper to synthesize a default from), so a checkpoint that omits it
+/// entirely must fail with the key named, distinct from the
+/// shape-mismatch case above.
+#[test]
+fn from_weights_rejects_missing_image_projection() {
+    let mut map = tiny_weights();
+    map.remove("image_projection");
+    let err = expect_err(Florence2Model::from_weights(
+        &map,
+        tiny_config(&["spatial_avg_pool"]),
+    ));
+    assert!(err.contains("image_projection"), "unexpected error: {err}");
+}
+
+/// The fusion stage cannot run without a position-embedding spec: unlike the
+/// optional embedding *type* checks below, a missing `image_pos_embed` /
+/// `visual_temporal_embedding` key in `vision_config` has to fail load with
+/// the missing field named rather than defaulting to some embedding kind.
+#[test]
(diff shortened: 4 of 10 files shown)


inureyes:

### Intent

> Join the DaViT tower (#1063) to the BART seq2seq engine (#1060) into `Florence2Model`, implement the fusion stage (2-D position embedding, temporal embedding, `image_feature_source` pooling, `image_projection`, concatenation with the prompt), and complete whole-checkpoint weight loading. Runtime registration is #856's scope.

### Fusion math verified against upstream

Traced line by line against `https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/florence2.py`:

- `LearnedPositionEmbedding2D`: column half first then row half, column broadcast over height and row over width, matches `_encode_image`'s `mx.concatenate([x_emb[None], y_emb[:, None]], axis=-1)`.
- `PositionalEmbeddingCosine1D`: sin/cos interleaved with `exp(-ln(10000) * j / embed_dim)`, matches; the checkpoint table wins over recomputation, which is correct because upstream registers the buffer as a module parameter and `models/Florence-2-base-ft-bf16` ships `visual_temporal_embed.pos_idx_to_embed`.
- `_encode_image` reshape/pool ordering and `x @ image_projection` then `image_proj_norm` match; the `image_projection` bare-tensor handling (no transpose, no bias) is right.
- `_merge_input_ids_with_image_features`: concatenation order and the prompt image-token filter match.
- The `image_feature_source` ordering finding in the PR description is confirmed against the checkpoint: `vision_config.image_feature_source` is `["spatial_avg_pool", "temporal_avg_pool"]` while upstream's dataclass default is the reverse, and upstream reads the field from the top level where it does not exist. Reading the config is correct.

Sanitization is complete for this checkpoint: the DaViT conv guards (`convs` + `proj.weight`, `blocks` + `dw.weight`) do not collide with any `language_model.*` key, and the `model.shared` fill covers exports that materialize only `embed_tokens`. Verified against all 666 keys in the real checkpoint.

### Findings Addressed

- [x] Quantized-checkpoint handling was described but had no backing code path (MEDIUM). Neither half of the family has a quantized layer, so the `config_has_quantization_metadata` branch was silently skipping the bf16 to f16 promotion for a checkpoint whose packed uint32 weights would reach MLX and abort the process. `Florence2Model::load` now rejects it with a named error before any weight file is read, with a unit test.
- [x] `references/mlx-vlm` local-checkout path committed in the parity test's module doc (LOW). Reworded to the public GitHub URL, per CLAUDE.md "Referencing upstream source".
- [x] `merge_input_ids_with_image_features` doc claimed the joint mask lets a padded batch flow through "the same path" (LOW). It reaches encoder self-attention only; `Florence2Attention::cross_attention` passes `None` where upstream passes `encoder_attention_mask`. Numerically irrelevant today (the mask is always all ones, and upstream's own all-ones float mask is a constant additive shift softmax cancels), but the doc now states the limitation.
- [x] Em dashes in the PR body (12 occurrences, MEDIUM convention). The squash merge uses the PR body as the commit body, so these would have landed in `main`. Body rewritten; also refreshed the now-stale mixed-precision section and the test counts.

Fix commit: `3b3352e`.

### Remaining Items

- [ ] Em dashes in the `088b1c4` commit message body (3 occurrences, LOW). Not amended: the squash merge discards this message in favour of the PR body, which is now clean, and rewriting the branch head would force-push an open PR for no effect on `main`.
- [ ] `Florence2Model::vision_tower()` has no caller (LOW, informational). Left in place as the natural sibling of `text_model()`; #856 will use it.
- [ ] `generate_greedy` implements neither `forced_bos_token_id: 0` nor the checkpoint's `num_beams: 3` (LOW, informational). Correct for a parity and development entry point; #856 owns generation wiring.

### Verification

- [x] All stated requirements implemented (all six acceptance criteria in #854 are met, with fusion math, concatenation, sanitize, real-checkpoint load, and mlx-vlm parity all pinned by tests)
- [x] No placeholder/mock code remaining (the synthetic tensors are test fixtures, not shipped defaults; no TODO, stub, or hardcoded activation anywhere in the new modules)
- [x] Integrated into project code flow (`Florence2Model::load` reads a real checkpoint and `generate_greedy` drives tower to fusion to encoder to the same dual-KV-cache decode loop the text engine uses; runtime registration is out of scope per #856)
- [x] Project conventions followed (upstream cited by GitHub URL, English only, no AI attribution, no em dashes, tests beside the code in `_tests.rs`, integration tests under `tests/`)
- [x] Existing modules reused where applicable (`mlxcel_core::layers::{Embedding, Linear, LayerNorm}`, `florence2_davit::sanitize`, `crate::models::convert_bf16_weights`, `sanitize::config_has_quantization_metadata`, the existing `Florence2SeqCache` decode machinery; no reimplementation found)
- [x] No unintended structural changes (the only edits to existing files are the additive-mask parameter threaded through `Florence2Encoder::forward` and `Florence2EncoderLayer::forward`, both defaulting to `None` so text-only numerics are unchanged, plus the `argmax_last_position` visibility bump and the `pub use` additions)
- [x] Tests pass: 37 unit tests in `models::florence2::`, 2 in `florence2_fusion_parity`, 3 in `florence2_text_parity` + `florence2_vision_parity`, `clippy -D warnings` clean, `cargo fmt --check` clean, `check_cross_repo_refs.py` OK

inureyes:

Reviewed against the threat model that a Florence-2 checkpoint (`config.json` + safetensors) is untrusted input, and that an MLX C++ throw crossing the `cxx` bridge aborts the process instead of surfacing as `Err`.

Fixed in 0e8f1d0d. Three HIGH and one MEDIUM addressed; two MEDIUM and three LOW reported below without a code change.

### Fixed

**HIGH: unbounded host allocation from `max_temporal_embeddings`** (`fusion.rs`)
The field comes from `vision_config.visual_temporal_embedding` and was only checked for `< 1`. On the documented synthesis fallback (any checkpoint that omits `visual_temporal_embed.pos_idx_to_embed`, including the unit-test fixture), `PositionalEmbeddingCosine1D::compute` eagerly allocates `max_seq_len * embed_dim` host floats and fills them with a nested sin/cos loop. `"max_temporal_embeddings": 2000000000` against the 1024-wide base-ft tower asks for roughly 8 TB and ~10^12 transcendental evaluations: an out-of-memory abort or an indefinite hang driven by a data file. Now bounded by `MAX_TEMPORAL_EMBEDDINGS = 65536` against a real-export value of 100.

**HIGH: `max_position_embeddings` trusted as the position-slice bound, never checked against the loaded table** (`encoder.rs`, `decoder.rs`)
`encode_with_image_features` rejects `seq > config.text.max_position_embeddings` and `generate_greedy` breaks at `cache.offset() >= max_position_embeddings`, then both reach `slice(embed_positions, [POSITION_OFFSET, 0], [POSITION_OFFSET + seq, d_model])`. Nothing validated that `embed_positions.weight` actually covers that bound, and the field is `unwrap_or(1024)`. A config that inflates `max_position_embeddings` or `d_model` (or a short table) passes both guards and reaches MLX as an out-of-range slice. Both `from_weights` paths now validate rank, width against `d_model`, and rows against `POSITION_OFFSET + max_position_embeddings`, summed in `i64`. That is what makes the runtime guards sound.

`decoder.rs` is outside this PR's diff, but its position slice is reached by the new `generate_greedy`, so the guard belongs there too. The two blocks are textually parallel.

**HIGH: `Florence2TextConfig` accepted non-positive and absurd shape fields** (`mod.rs`)
Reached by the new `Florence2Model::load`. Concretely: `"encoder_attention_heads": 0` made the existing divisibility check *itself* panic with a divide by zero before it could reject anything; `"encoder_layers": -1` reached `Vec::with_capacity` as `usize::MAX` and aborted with a capacity overflow; `"d_model": 0` produced `head_dim == 0`, an infinite attention scale, and a zero-width head reshape into MLX; `pad_token_id` and `decoder_start_token_id` are fed to the shared embedding as literal gather indices, so an out-of-vocabulary or negative value was an out-of-range gather. All now range-checked, ahead of the divisibility check. `bos_token_id` and `eos_token_id` are deliberately left unchecked and the comment says so: they are compared, never indexed, and some exports carry sentinels there.

**MEDIUM: panic-by-indexing in `additive_attention_mask`** (`fusion.rs`)
Read `shape[0]`/`shape[1]` with no rank check. Now returns `Result`.

Six unit tests added, one per rejection.

### Reported, not changed

**MEDIUM (perf): the joint attention mask is always dense, and the resulting no-op is paid in every encoder layer.**
`merge_input_ids_with_image_features` returns `ones(...)` unconditionally, so `additive_attention_mask` always produces an all-zero `[batch, 1, 1, seq]` tensor, and `qkv_attention` then broadcasts it into the `[b, h, lq, lk]` logits in each of the 6 encoder layers. For base-ft that is a `[1, 12, 583, 583]` f16 add per layer, roughly 24M redundant adds and 6 full-size temporaries per encode, for an arithmetic identity. Back-of-envelope this is on the order of 15% of encoder time and a low single-digit percentage of prefill, but that is an estimate, not a measurement.

Not fixed here deliberately. Every available fix has a design tradeoff the author should pick (skip by construction and the encoder-side mask plumbing goes dead; detect density and you buy a small `eval` sync per encode; return density from the merge step and the new public signature changes), and CLAUDE.md is explicit that a perf change lands on a measurement through the production call path, not on a reviewer's arithmetic. Worth its own issue with a benchmark.

**MEDIUM: prompt token ids are not bounded against `vocab_size` before the embedding gather.**
`embed_prompt` filters the image placeholder and hands the rest to `Embedding::forward`. An id at or past `vocab_size` (mismatched tokenizer, hand-built prompt) becomes an out-of-range MLX gather. This is codebase-wide rather than specific to this PR, so it is flagged rather than fixed in a Florence-2 change.

**LOW:**
- `side * side` in `encode_image_features` can overflow `i32` for a token count near `i32::MAX` (debug panic, release wrap). Unreachable with any real tensor; the wrap would be caught by the `!= num_tokens` comparison anyway.
- `image_token_id` and `max_temporal_embeddings` use `as i64 -> as i32`, so `4294967396` silently becomes `100`. Cosmetic given the new bounds.
- `argmax_last_position` and `Florence2Encoder::forward` index `array_shape(...)[1]` without a rank check. Pre-existing, same class as the `additive_attention_mask` fix above.

### Validation

```
cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings   # clean
cargo test   --profile test-fast --features metal,accelerate --lib models::florence2::      # 43 passed, 0 failed
cargo test   --profile test-fast --features metal,accelerate --test florence2_fusion_parity # 2 passed, 0 failed
cargo fmt --check                                                                            # clean
```

The parity tests run against a real checkpoint, so the load-time validation is confirmed not to reject a genuine `Florence-2-base-ft` export. No fixture needed loosening: both `tiny_weights` helpers already sized `embed_positions` at `max_position_embeddings + 2` rows.

inureyes:
## PR Finalization Complete

### Summary

- **Tests**: Reviewed all fallible paths in `fusion.rs` / `checkpoint.rs` / `model.rs` against the existing 44-test suite in `florence2_fusion_tests.rs` and `florence2_tests.rs`, and added 7 unit tests for the branches that had no coverage: a checkpoint missing `image_projection` entirely (distinct from the existing wrong-shape case), a config missing `image_pos_embed` or `visual_temporal_embedding`, `encode_image_features` rejecting a feature width that disagrees with `dim_embed[-1]`, an unsupported `image_feature_source` entry surfacing at pooling time, `merge_input_ids_with_image_features` rejecting a non-3-D image-feature tensor and an incompatible prompt-embedding shape, and `encode_fused` rejecting a wrong-rank `inputs_embeds`. Unit suite is now 51 passed (was 44). Also re-ran the gated integration suite against the real `models/Florence-2-base-ft-bf16` checkpoint (`florence2_fusion_parity`, `florence2_text_parity`, `florence2_vision_parity`): all 6 pass unchanged.
- **Documentation**: Reviewed the module-level `//!` docs in `fusion.rs`, `checkpoint.rs`, `model.rs`, `encoder.rs`, `decoder.rs`, and `mod.rs` against the implementation; found them accurate and complete, no changes needed. Confirmed no reference paths were left as local `references/...` paths (all are full GitHub URLs) and the existing README/NOTICE mlx-vlm attribution already covers this port.
- **Lint/Format**: `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings` and `cargo fmt --check` were both already clean before this pass; no fixes were needed.

### Validation

- `cargo test --profile test-fast --features metal,accelerate --lib models::florence2::`: 51 passed, 0 failed.
- `cargo test --profile test-fast --features metal,accelerate --test florence2_fusion_parity`: 2 passed against `models/Florence-2-base-ft-bf16`.
- `cargo test --profile test-fast --features metal,accelerate --test florence2_text_parity --test florence2_vision_parity`: 3 passed.
- `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `python3 scripts/ci/check_cross_repo_refs.py`: OK.

PR #1069, opened 2026-08-07
feat(models): Florence-2 processor, task prompts, and location tokens

### LINKED ISSUE #855: feat(models): Florence-2 processor, task prompts, and location tokens (sub of #850)
## Summary

The Florence-2 processor layer: task-prompt tokens and their expansions, image preprocessing, tokenizer wiring, and location-token handling with post-processing into structured coordinates for the spatial tasks.

## Part of epic

#850

## Dependencies

#854 (vision-language fusion + full weight loading)

## Upstream reference

- processing_florence2.py: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/processing_florence2.py

## Architecture notes

- Task-prompt tokens and their prompt expansions: `<CAPTION>`, `<DETAILED_CAPTION>`, `<MORE_DETAILED_CAPTION>`, `<OCR>`, `<OCR_WITH_REGION>`, `<OD>`, `<DENSE_REGION_CAPTION>`, `<CAPTION_TO_PHRASE_GROUNDING>`, `<REGION_TO_SEGMENTATION>`, and the rest of the task set.
- Image preprocessing (resize / normalize) per the processor.
- Tokenizer wiring.
- Location tokens `<loc_0>` .. `<loc_999>`: emitted by the spatial tasks and post-processed into boxes / quad-boxes / regions.

## Implementation plan

- Implement the processor / task-prompt stage in the Florence-2 module under `src/models/florence2/`, feeding the fused model from #854.
- Reference the coordinate / OCR post-processing in the existing OCR VLM modules (dots_ocr and paddleocr under `src/models/`) as a partial reference for the location-token to box / region conversion.
- Wire the tokenizer and image preprocessing into the VLM input path so a task prompt drives the correct mode end to end.

## Acceptance criteria

- [x] Task-prompt tokens and their expansions handled for the full task set.
- [x] Image preprocessing (resize / normalize) matches the processor.
- [x] Location tokens parse / emit correctly and post-process into boxes / quad-boxes / regions.
- [x] Task-prompt round trip works; OCR and detection outputs parse into correct structured coordinates on a real image.
- [x] Integration: the processor is wired into the model's actual input/output path (task prompt in, structured result out through the fused model from #854), not a standalone parser.

## Effort

MEDIUM.

inureyes:
## Summary

Completes the Florence-2 request path on top of the fused model from #854. A task marker and an image go in; a structured result with coordinates in original-image pixels comes out.

Florence-2 selects its behavior with a marker such as `<OD>` or `<CAPTION>`, and **none of those markers is in the vocabulary**. The processor expands each one into the English sentence the model was actually trained on (`<OD>` becomes "Locate the objects with category name in the image."), tokenizes that, preprocesses the image to 768x768, generates, and parses the answer's `<loc_N>` tokens back into boxes, quadrilaterals, and polygons against the image's pre-resize extent.

The integration entry point is `Florence2Processor::run(&model, task, input, image, max_new_tokens) -> Florence2Output`, which is what #856 will reach from the CLI.

## What changed

**New, under `src/models/florence2/`**

- `tasks.rs`: the fifteen task markers, their prompt expansions copied byte for byte from the checkpoint, and routing to a post-processing kind. Seven tasks interpolate a caller-supplied input.
- `coords.rs`: `Florence2ImageSize`, `Florence2BoundingBox`, `Florence2QuadBox`, `Florence2Polygon`, the bin-to-pixel dequantizers, and the inverse so a caller can build the `<loc_a><loc_b><loc_c><loc_d>` region string the region-input tasks expect.
- `scan.rs`: literal string scanners (sequence-marker stripping, leading-phrase extraction, `<loc_N>` reading) plus the phrase-grounding stopword blacklist.
- `parse.rs`: the four reachable parsers over compiled-once patterns.
- `postprocess.rs`: `Florence2TaskResult` and the task dispatch. `Florence2TaskResult::parse` is model-free, so an answer obtained elsewhere parses without loading a checkpoint.
- `processor.rs`: `Florence2Processor`, tokenizer plus image processor plus the end-to-end `run`.

**New elsewhere**

- `src/vision/processors/florence2.rs`: `Florence2ImageProcessor`, reading `preprocessor_config.json` rather than hard-coding it, and refusing configurations it would silently mishandle (center cropping, a non-bicubic resample, a zero standard deviation, a zero size).
- `tests/florence2_postprocess_parity.rs`, `tests/florence2_processor_parity.rs`.
- Unit tests beside each new module.

**Moved**

- `Florence2TextConfig` from `src/models/florence2/mod.rs` to `src/models/florence2/checkpoint.rs`, next to the `Florence2Config` that contains it. Pure relocation, no behavior change; it keeps `mod.rs` under the 500-line limit now that the module list has grown.

## Design notes

`Florence2BoundingBox` is deliberately not a reuse of `rt_detr_v2::Detection`. That type carries a confidence score and a `label: usize` index into a fixed class list; Florence-2 emits no per-box score and its labels are open-vocabulary strings decoded from the answer, so reusing it would mean two fabricated fields on every box. There is no polygon or quadrilateral type anywhere in the tree to reuse.

`Florence2ImageSize` has named fields rather than being a `(u32, u32)` tuple because upstream's own docstring disagrees with its code about the order (see corrections below), and passing the 768x768 resized extent instead of the original silently scales every coordinate.

Two upstream regexes were replaced with hand-rolled scanners, each with its equivalence argument recorded in the code. The phrase-string pattern `^\s*(.*?)(?=<od>|...|<loc_)` becomes a forward scan for the first stop marker; the two behaviors that make it non-trivial (`.` not matching a newline, and `<sep>` not being in the stop list) are reproduced and tested. `<loc_(\d+)>` becomes a digit scan, since it is a fixed literal on the innermost path of every parser. Everything else uses `fancy-regex`, which matches Python's backtracking semantics.

- The issue suggests using the dots_ocr and paddleocr modules as a partial reference for location-token to box conversion. **They contain no coordinate parsing at all.** Both are plain text-emitting VLMs. The only box type in the tree is `rt_detr_v2::Detection`, and there is no polygon type. Nothing was reusable.
- `<REGION_PROPOSAL>` routes to the `bboxes` post-processing kind, not `description_with_bboxes`. Same parser, but with `allow_empty_phrase` set, which matters: its answers carry no category names.

## Corrections to upstream

Verified against the checkpoint's own `processing_florence2.py` and the mlx-vlm sources.

- **`post_process_generation`'s docstring is wrong about coordinate order.** It documents `image_size` as "height x width", but every call site unpacks it as `image_width, image_height = image_size`. It is `(width, height)`. This is why the Rust type uses named fields.
- **The `od` parse task's configured PATTERN can never match.** It is `r'([a-zA-Z0-9 ]+)<loc_(\\d+)>...'`, and the doubled backslash inside a raw string matches a literal `\` followed by `d`. It is unreachable in practice: no task routes to `'od'`, and every parser except the OCR one overwrites its `pattern` argument on its first lines, so only the OCR pattern is ever read from config.
- **The OCR area filter never runs.** It is gated on `area_threshold > 0` and the shipped `AREA_THRESHOLD` is `0.00`.
- **`with_box_at_start` is false at every call site reachable from `post_process_generation`,** so that polygon branch is dead.
- **The `<ground>` / `<obj>` prefix strip is a no-op twice over.** The second assignment reads `pharse_text` rather than the result of the first, discarding the `<ground>` strip, and neither marker is in the checkpoint's vocabulary.
- **A newline in a phrase silently deletes the whole chunk.** `^\s*(.*?)(?=...)` is not multi-line and `.` does not match a newline, so if a newline sits between the leading whitespace and the first location token, `re.search` fails and the chunk is dropped. Reproduced deliberately and pinned by `od_newline`.
- **`[^<]+` can start one byte inside a `<loc_N>` token.** A bare location run fed to the labelled parser therefore yields the literal label `loc_1>` and a box shifted by one bin, rather than no result. Reproduced, pinned by `od_bare_locs`, and the reason `<REGION_PROPOSAL>` needs the empty-phrase parser.
- **mlx-vlm's florence2 model applies stochastic depth at inference unless you call `model.eval()`.** mlx's `nn.Module` starts in training mode and upstream's DaViT `DropPath` only short-circuits on `not self.training`. Without it the vision tower output is not reproducible. Recorded in the parity test's module docs so the next person does not lose an hour to it.
- **`AutoTokenizer.from_pretrained` fails on this checkpoint**, because it routes through `AutoConfig`, which does not know the custom `florence2_language` model type. `BartTokenizerFast` directly works.
- Re-confirmed from #854: upstream's `ModelConfig` reads `image_feature_source`, `image_pos_embed` and `visual_temporal_embedding` from the top level of `config.json`, but real checkpoints store them under `vision_config`, and the dataclass default for `image_feature_source` is the reverse order.

## Deliberately not implemented

Each is unreachable from `post_process_generation`, and each is documented in `parse.rs` with the reason rather than silently omitted: `parse_od_from_text_and_spans` and the PATTERN config indirection, the OCR area filter, the `with_box_at_start` polygon branch, and the `<ground>` / `<obj>` prefix strip. `decode_with_spans` is also not ported; its character spans exist only for score attribution, which no reachable task uses.

One intentional deviation: `Florence2Task::expand` rejects an input supplied to a task that takes none, and a missing input on a task that requires one. Upstream produces `"What is the region ?"` in the second case and lets the model answer nonsense, which is indistinguishable from a real prompt after the fact.

One knowingly accepted difference: `$` in the polygon-run pattern matches only at end of string, while Python's also matches just before a trailing newline. Location runs are not emitted with trailing newlines.

## Validation

Against the real checkpoint (`models/Florence-2-base-ft-bf16`) and the real upstream Python, in a scratch virtualenv with `torch`, `transformers`, `mlx`, `numpy` and `Pillow`. The post-processing reference imports the checkpoint's own `processing_florence2.py` **unmodified**, so it is upstream behavior rather than a reimplementation agreeing with itself.

- **Post-processing**: 32 cases across all five result shapes, all matching upstream exactly, including every edge case listed above.
- **Preprocessing**: `tests/fixtures/test_image.png` through the checkpoint's `CLIPImageProcessor` versus the Rust processor. Maximum absolute deviation **exactly 0.0** across all 1,769,472 values; the `image` crate's `CatmullRom` and PIL's `BICUBIC` agree bit for bit on this 224 to 768 upscale. The test tolerance is kept loose anyway, since that agreement is a property of two independent resampling implementations.
- **Tokenization**: exact id match on all six tested task prompts.
- **End to end**: upstream mlx-vlm florence2 with weights cast bf16 to f16. Byte-identical answers on all six tasks, including `<OD>` producing `poster<loc_0><loc_0><loc_998><loc_998>` and `<OCR_WITH_REGION>` producing a full eight-token quadrilateral, and identical parsed coordinates for both.

No unexplained Rust-versus-Python divergence was found anywhere in this work.

## Test plan

- [x] `cargo check --profile test-fast --features metal,accelerate --lib --tests`
- [x] `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings` (zero warnings)
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib florence2`
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_postprocess_parity`
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_processor_parity`
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_fusion_parity` (unchanged by the `Florence2TextConfig` move)
- [x] `cargo fmt --check`

Closes #855

### DIFF
--- a/src/models/florence2/checkpoint.rs
+++ b/src/models/florence2/checkpoint.rs
@@ -22,14 +22,14 @@
 //! Reference:
 //! https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/florence2.py
 
-use anyhow::Result;
+use anyhow::{Result, anyhow};
 use serde_json::Value;
 
 use mlxcel_core::weights::WeightMap;
 
 use crate::vision::encoders::florence2_davit;
 
-use super::{Florence2TextConfig, Florence2VisionConfig};
+use super::Florence2VisionConfig;
 
 /// Weight-key prefix of the language model inside a Florence-2 checkpoint.
 pub(crate) const FLORENCE2_TEXT_PREFIX: &str = "language_model.";
@@ -106,4 +106,166 @@ pub fn sanitize(weights: WeightMap) -> WeightMap {
         }
     }
     out
+}
+
+/// Upper bound accepted for `encoder_layers` / `decoder_layers`. Real
+/// Florence-2 exports ship 6 (base) or 12 (large). The cap exists because
+/// [`encoder::Florence2Encoder::from_weights`] and
+/// [`decoder::Florence2Decoder::from_weights`] size a `Vec` from this field
+/// before they look up a single weight, so a negative value out of a hostile
+/// `config.json` becomes `usize::MAX` and aborts with a capacity overflow.
+const MAX_LAYERS: i32 = 256;
+
+/// Upper bound accepted for `max_position_embeddings`. Florence-2 ships 1024.
+///
+/// The field bounds two separate allocations: the position-table slice the
+/// encoder and decoder take (validated against the loaded table at load time)
+/// and the O(n^2) host buffer [`layers::additive_causal_mask`] fills for a
+/// multi-token decoder call. An unbounded value out of a hostile `config.json`
+/// is an out-of-memory abort rather than an error return.
+const MAX_POSITION_EMBEDDINGS: i32 = 65_536;
+
+/// BART encoder-decoder shape parameters, parsed from the `text_config`
+/// object of a Florence-2 `config.json` (`model_type: florence2_language`).
+#[derive(Debug, Clone)]
+pub struct Florence2TextConfig {
+    pub d_model: i32,
+    pub encoder_layers: i32,
+    pub decoder_layers: i32,
+    pub encoder_attention_heads: i32,
+    pub decoder_attention_heads: i32,
+    pub encoder_ffn_dim: i32,
+    pub decoder_ffn_dim: i32,
+    pub vocab_size: i32,
+    pub max_position_embeddings: i32,
+    pub scale_embedding: bool,
+    pub pad_token_id: i32,
+    pub bos_token_id: i32,
+    pub eos_token_id: i32,
+    pub decoder_start_token_id: i32,
+}
+
+fn config_i32(config: &Value, key: &str) -> Option<i32> {
+    config.get(key).and_then(Value::as_i64).map(|v| v as i32)
+}
+
+impl Florence2TextConfig {
+    /// Parse from the `text_config` sub-object itself.
+    pub fn from_text_config(config: &Value) -> Result<Self> {
+        let require = |key: &str| -> Result<i32> {
+            config_i32(config, key)
+                .ok_or_else(|| anyhow!("Florence-2 text_config missing field: {key}"))
+        };
+        let d_model = require("d_model")?;
+        let parsed = Self {
+            d_model,
+            encoder_layers: require("encoder_layers")?,
+            decoder_layers: require("decoder_layers")?,
+            encoder_attention_heads: require("encoder_attention_heads")?,
+            decoder_attention_heads: require("decoder_attention_heads")?,
+            encoder_ffn_dim: config_i32(config, "encoder_ffn_dim").unwrap_or(4 * d_model),
+            decoder_ffn_dim: config_i32(config, "decoder_ffn_dim").unwrap_or(4 * d_model),
+            vocab_size: require("vocab_size")?,
+            max_position_embeddings: config_i32(config, "max_position_embeddings").unwrap_or(1024),
+            scale_embedding: config
+                .get("scale_embedding")
+                .and_then(Value::as_bool)
+                .unwrap_or(false),
+            pad_token_id: config_i32(config, "pad_token_id").unwrap_or(1),
+            bos_token_id: config_i32(config, "bos_token_id").unwrap_or(0),
+            eos_token_id: config_i32(config, "eos_token_id").unwrap_or(2),
+            decoder_start_token_id: config_i32(config, "decoder_start_token_id").unwrap_or(2),
+        };
+        // Every field below becomes a shape, an index, a divisor, or a `Vec`
+        // capacity somewhere downstream. MLX throws on an out-of-range shape
+        // and the throw crosses an FFI boundary that cannot carry it, so a bad
+        // value aborts the process rather than surfacing as an error; these
+        // checks are what keep an untrusted `config.json` from getting that
+        // far. They also have to run *before* the divisibility check below,
+        // which would itself divide by zero on `encoder_attention_heads: 0`.
+        if parsed.d_model < 1 {
+            return Err(anyhow!(
+                "Florence-2 text_config d_model must be positive, got {}",
+                parsed.d_model
+            ));
+        }
+        if parsed.encoder_attention_heads < 1 || parsed.decoder_attention_heads < 1 {
+            return Err(anyhow!(
+                "Florence-2 text_config attention heads must be positive, got {} enc / {} dec",
+                parsed.encoder_attention_heads,
+                parsed.decoder_attention_heads
+            ));
+        }
+        if !(1..=MAX_LAYERS).contains(&parsed.encoder_layers)
+            || !(1..=MAX_LAYERS).contains(&parsed.decoder_layers)
+        {
+            return Err(anyhow!(
+                "Florence-2 text_config layer counts {} enc / {} dec outside 1..={MAX_LAYERS}",
+                parsed.encoder_layers,
+                parsed.decoder_layers
+            ));
+        }
+        if parsed.encoder_ffn_dim < 1 || parsed.decoder_ffn_dim < 1 {
+            return Err(anyhow!(
+                "Florence-2 text_config ffn dims must be positive, got {} enc / {} dec",
+                parsed.encoder_ffn_dim,
+                parsed.decoder_ffn_dim
+            ));
+        }
+        if parsed.vocab_size < 1 {
+            return Err(anyhow!(
+                "Florence-2 text_config vocab_size must be positive, got {}",
+                parsed.vocab_size
+            ));
+        }
+        if !(1..=MAX_POSITION_EMBEDDINGS).contains(&parsed.max_position_embeddings) {
+            return Err(anyhow!(
+                "Florence-2 text_config max_position_embeddings {} outside 1..={MAX_POSITION_EMBEDDINGS}",
+                parsed.max_position_embeddings
+            ));
+        }
+        // Both of these reach the shared embedding gather as literal token ids
+        // (`generate_greedy` seeds the decoder with `decoder_start_token_id`,
+        // `shift_tokens_right` substitutes `pad_token_id`), so an
+        // out-of-vocabulary or negative value is an out-of-range gather.
+        // `bos_token_id` and `eos_token_id` are deliberately left unchecked:
+        // they are only ever compared, never used as an index, and some
+        // exports legitimately carry sentinel values there.
+        if !(0..parsed.vocab_size).contains(&parsed.pad_token_id) {
+            return Err(anyhow!(
+                "Florence-2 text_config pad_token_id {} outside 0..vocab_size {}",
+                parsed.pad_token_id,
+                parsed.vocab_size
+            ));
+        }
+        if !(0..parsed.vocab_size).contains(&parsed.decoder_start_token_id) {
+            return Err(anyhow!(
+                "Florence-2 text_config decoder_start_token_id {} outside 0..vocab_size {}",
+                parsed.decoder_start_token_id,
+                parsed.vocab_size
+            ));
+        }
+        if parsed.d_model % parsed.encoder_attention_heads != 0
+            || parsed.d_model % parsed.decoder_attention_heads != 0
+        {
+            return Err(anyhow!(
+                "Florence-2 d_model {} not divisible by attention heads ({} enc / {} dec)",
+                parsed.d_model,
+                parsed.encoder_attention_heads,
+                parsed.decoder_attention_heads
+            ));
+        }
+        Ok(parsed)
+    }
+
+    /// Parse from a full Florence-2 `config.json` (`model_type: florence2`),
+    /// descending into its `text_config` sub-object. A bare text config (no
+    /// `text_config` key) is also accepted so the sub-object can be passed
+    /// directly.
+    pub fn from_model_config(config: &Value) -> Result<Self> {
+        match config.get("text_config") {
+            Some(text) => Self::from_text_config(text),
+            None => Self::from_text_config(config),
+        }
+    }
 }
--- a/src/models/florence2/checkpoint.rs
+++ b/src/models/florence2/checkpoint.rs
@@ -110,19 +110,20 @@ pub fn sanitize(weights: WeightMap) -> WeightMap {
 
 /// Upper bound accepted for `encoder_layers` / `decoder_layers`. Real
 /// Florence-2 exports ship 6 (base) or 12 (large). The cap exists because
-/// [`encoder::Florence2Encoder::from_weights`] and
-/// [`decoder::Florence2Decoder::from_weights`] size a `Vec` from this field
-/// before they look up a single weight, so a negative value out of a hostile
-/// `config.json` becomes `usize::MAX` and aborts with a capacity overflow.
+/// [`super::encoder::Florence2Encoder::from_weights`] and
+/// [`super::decoder::Florence2Decoder::from_weights`] size a `Vec` from this
+/// field before they look up a single weight, so a negative value out of a
+/// hostile `config.json` becomes `usize::MAX` and aborts with a capacity
+/// overflow.
 const MAX_LAYERS: i32 = 256;
 
 /// Upper bound accepted for `max_position_embeddings`. Florence-2 ships 1024.
 ///
 /// The field bounds two separate allocations: the position-table slice the
 /// encoder and decoder take (validated against the loaded table at load time)
-/// and the O(n^2) host buffer [`layers::additive_causal_mask`] fills for a
-/// multi-token decoder call. An unbounded value out of a hostile `config.json`
-/// is an out-of-memory abort rather than an error return.
+/// and the O(n^2) host buffer [`super::layers::additive_causal_mask`] fills
+/// for a multi-token decoder call. An unbounded value out of a hostile
+/// `config.json` is an out-of-memory abort rather than an error return.
 const MAX_POSITION_EMBEDDINGS: i32 = 65_536;
 
 /// BART encoder-decoder shape parameters, parsed from the `text_config`
--- a/src/models/florence2/coords.rs
+++ b/src/models/florence2/coords.rs
@@ -1,0 +1,209 @@
+// Copyright 2025-2026 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Florence-2 location tokens and the bin-to-pixel dequantizers.
+//!
+//! Florence-2 encodes every spatial answer as `<loc_N>` tokens. The vocabulary
+//! holds exactly 1000 of them, `<loc_0>` .. `<loc_999>`, occupying a
+//! contiguous id range, and each one names a bin index on a 1000-bin axis
+//! normalized to the *original* image extent. Turning a bin back into a pixel
+//! coordinate therefore needs the pre-resize image size, which is why the
+//! processor keeps it alongside the pixel tensor.
+//!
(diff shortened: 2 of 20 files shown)


inureyes:

### Intent

Port the Florence-2 processor layer for #855: the fifteen task markers and their prompt expansions, 768x768 image preprocessing, tokenizer wiring, and `<loc_0>`..`<loc_999>` post-processing into boxes, quad-boxes and polygons, with `Florence2Processor::run` as the end-to-end entry point the capstone #856 will reach from the CLI.

### Verification performed

I checked the port line by line against the checkpoint's own `processing_florence2.py` (the authoritative upstream for this checkpoint, published at https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/processing_florence2.py) rather than against the PR description.

- **Prompt expansions**: all fifteen strings byte-identical to `task_prompts_without_inputs` / `task_prompts_with_input`, including the comma in "What is the text in the image, with regions?" and the missing trailing period on "Locate {input} in the image with mask" and "What is the polygon mask of region {input}".
- **Post-processing routing**: matches `tasks_answer_post_processing_type` for all fifteen entries, including `<REGION_PROPOSAL>` to `bboxes` (not `description_with_bboxes`) and the `allow_empty_phrase=True` that the `polygons` parse task carries.
- **Dequantization**: the `+ 0.5` bin-center term is present in both `dequantize_box` and `dequantize_coordinates`; 1000 bins per axis; x scaled by width and y by height; the f64-quotient-narrowed-to-f32 `size_per_bin` reproduces torch's promotion of an int64 tensor plus a Python float. `quantize` reproduces `floor` mode with the same clamp.
- **Stopword blacklist**: diffed programmatically against `_create_black_list_of_phrase_grounding` (lines 552-578). Upstream's 100 literals with 28 repeats reduce to 72 unique entries; the Rust list is exactly those 72, sorted for binary search, with the sortedness and the count both pinned by `blacklist_is_sorted_for_binary_search`. Nothing truncated.
- **Parsers**: `pure_text` strips only `<s>`/`</s>` while the box parsers also strip `<pad>` (upstream's asymmetry, reproduced); the OCR pattern is `(.+?)` plus exactly eight `<loc_>`; the chunkers, the phrase scan's stop list, the ASCII drop-not-replace, the one-instance-per-4-loc-group fan-out, the phrase-grounding grouping and blacklist ordering, the `<poly>`/`</poly>` split gated on both delimiters being present, the `((?:<loc_\d+>)+)(?:<sep>|$)` run pattern, and the odd-length tail drop all match. The two hand-rolled scanners that replace upstream regexes carry a correct equivalence argument: I re-derived the `^\s*(.*?)(?=...)` backtracking behaviour (greedy `\s*` can consume a leading newline, `.` cannot cross one afterwards, `<sep>` is absent from the stop list) and `leading_phrase` reproduces it in every case I could construct.
- **The `<ground>`/`<obj>` strip** is not merely inert for the stated reason: the chunk pattern is `[^<]+` followed only by `<loc_\d+>` runs, so neither marker can appear anywhere inside a chunk. The omission is safe by construction, which is stronger than the justification in `parse.rs` claims.
- **Preprocessing**: `resize_exact` to 768x768 with no aspect preservation, rescale 1/255, ImageNet mean/std (not CLIP, not SigLIP), NCHW `[B,3,768,768]`, and the config read from `preprocessor_config.json` with hard errors on center-crop, non-bicubic resample, zero std and zero size rather than silent mishandling.
- **Tokenizer**: `<loc_N>` id equals `50269 + N` contiguously through `<loc_999>` = 51268, verified directly against `added_tokens.json` (zero mismatches over all 1000). `encode(prompt, add_special_tokens = true)` and `decode(ids, skip_special_tokens = false)` are the correct polarity, and the existing `MlxcelTokenizer` / `load_tokenizer` are reused rather than reimplemented.
- **Original-size threading**: `Florence2Processor::run` takes `(width, height)` from `preprocess_with_sizes`' pre-resize `original_sizes` and passes it to post-processing, never the 768x768 extent. `preprocess_with_sizes` pushes `(image.width(), image.height())` in that order, pinned by `reports_the_original_size_before_resizing` on a 640x480 and a 100x200 image. `detection_and_ocr_parse_into_reference_coordinates` additionally re-parses the same answer at a different declared size and asserts the coordinates move.
- **The `Florence2TextConfig` relocation** is byte-for-byte identical apart from the `anyhow!` import and dropping the now-unneeded `super::Florence2TextConfig` use. `mod.rs` re-exports it and `src/models/mod.rs` is unchanged for that name, so existing consumers are unaffected. `mod.rs` lands at 350 lines and every new file is under 500.
- **Error handling on the public surface**: a with-input task called with `None` and a no-input task called with `Some` are both rejected at `Florence2Task::expand` (a deliberate, documented divergence from upstream's "What is the region ?"); an unknown task string errors from `FromStr`; a truncated loc run is discarded by the 4-at-a-time grouping exactly as the regex would; an empty generation yields empty results with no panic; a non-RGB image goes through `to_rgb8`; a zero-size image is guarded in `quantize` rather than dividing by zero. No `unwrap()`, `expect()`, `TODO`, placeholder or mock anywhere in the new library code.
- **Scope**: `src/models/detection.rs`, `src/model_metadata.rs`, `docs/supported-models.md`, `mlxcel list` and the `generate` wiring are untouched, as #856 requires. No behavioral change reaches the already-merged #852/#853/#854 code.

### Findings addressed

- [x] `src/models/florence2/scan.rs:171` and `src/models/florence2/parse.rs:161` (LOW) `location_bins` documented that a `<loc_N>` digit run too wide for `i32` is "accepted here rather than rejected, matching upstream's plain `int()`", but `parse::<i32>()` failed and the bin was silently dropped while the scan advanced past the token. Dropping one bin re-pairs every bin after it in the same chunk, so the four-at-a-time grouping and the eight-token OCR record would misassign coordinates rather than produce one wrong number; `parse_ocr` separately discarded the whole record where Python keeps it. Both now route through a shared `scan::parse_bin` that saturates, so the token stays counted and the grouping stays aligned. Unreachable on real output (the vocabulary carries only `<loc_0>`..`<loc_999>`), and the parity suite is unchanged. Fixed in c761645d with a regression test.

### Remaining items (reported, not fixed)

- [ ] `src/models/florence2/checkpoint.rs:110-124` (LOW) The relocated doc comments still carry `[encoder::Florence2Encoder::from_weights]`, `[decoder::Florence2Decoder::from_weights]` and `[layers::additive_causal_mask]`, which resolved from `mod.rs` but not from `checkpoint.rs`. They sit on private items so default `cargo doc` never tries to resolve them and no CI job checks intra-doc links; left alone to keep the relocation byte-identical. `super::`-qualifying them is the fix whenever that file is next touched.
- [ ] `src/vision/processors/florence2.rs:54-63` (LOW) All `Florence2ImageProcessor` fields are `pub`, so the validation `from_raw` performs (non-zero size, non-zero std, bicubic-only) can be bypassed after construction. The unit tests rely on this to override `height`/`width`. Not a defect today; worth private fields plus accessors if the type ever escapes the crate.
- [ ] `tests/florence2_processor_parity.rs:57` (LOW) The end-to-end fixture is square (224x224), so a width/height transposition inside `run` would not show up there. The ordering is covered indirectly by `reports_the_original_size_before_resizing` and by the non-square cases in `florence2_postprocess_parity.rs`, so this is a coverage gap rather than a risk; a non-square fixture would close it outright.
- [ ] `src/vision/processors/florence2.rs:198-203` (LOW) An RGBA input is resized in RGBA and then flattened by `to_rgb8`, while PIL would convert to RGB first. Only observable on an image with a non-opaque alpha channel, which Florence-2 checkpoints are not exercised with.

### Verification checklist

- [x] All stated requirements implemented (all five acceptance criteria in #855)
- [x] No placeholder, mock or TODO code remaining
- [x] Integrated into the project code flow (`Florence2Processor::run` threads task expansion, tokenize, preprocess, `Florence2Model::generate_greedy`, decode-with-specials and post-process against the original image size, and is exercised end to end against the real checkpoint)
- [x] Project conventions followed (`_tests.rs` beside the implementation, no `unwrap`/`expect` in library code, inline format args, every file under 500 lines, upstream cited only by GitHub URL)
- [x] Existing modules reused where applicable (`MlxcelTokenizer`, `load_tokenizer`, the `ImageProcessor` trait; the two cases where reuse was rejected, `rt_detr_v2::Detection` and a shared image processor, are argued correctly in the code)
- [x] No unintended structural changes (the `Florence2TextConfig` move is a pure relocation with working re-exports)
- [x] Tests pass: clippy zero-warning, `cargo fmt --check` clean, 132 `--lib florence2` unit tests, `florence2_postprocess_parity` (6), `florence2_processor_parity` (4, against the real checkpoint)

inureyes:

Threat model applied: the parsed text is decoder output, which is attacker influenced in a served deployment because the image and the task input string are user supplied, so all parser input was treated as untrusted. Scope was the diff under `src/models/florence2/`, `src/vision/processors/florence2*.rs`, and the two gated integration tests. No CRITICAL or HIGH findings.

### Fixed (a862e2e)

**MEDIUM - unbounded allocation from `preprocessor_config.json` (`src/vision/processors/florence2.rs:154`).** `size` is the only field in this file that becomes an allocation: `preprocess_with_sizes` sizes its host buffer as `batch * 3 * height * width` f32 before it reads a pixel. It was validated as non-zero but had no upper bound. A hostile checkpoint could make that a multi-gigabyte allocation and an out-of-memory abort rather than an error return, and a value past `usize::MAX / 12` (for example `4294967296`) wraps the product in a release build, leaving a short buffer for a write loop that still runs to the configured extent, so `data[base + c * plane + y * width + x]` panics. Added `MAX_SIZE_EDGE = 8192`, mirroring the `MAX_LAYERS` and `MAX_POSITION_EMBEDDINGS` bounds this PR already applies to `text_config` in `src/models/florence2/checkpoint.rs`. Real exports ship 768.

**Regression guard - `adversarial_answers_do_not_panic` (`src/models/florence2/florence2_parse_tests.rs`).** The scanners in `scan.rs` walk byte offsets rather than characters over text that can carry arbitrary Unicode, so the byte-offset slicing is the highest-risk area in the diff. The review found it correct, and this test is what keeps it correct: multi-byte characters against every boundary a scanner slices at (the leading-whitespace cut, the stop-marker lookahead, the `<loc_` digit run, the bare `loc_N>` prefix strip), malformed location tokens with no digits or no terminator, nested `<loc_` opens that exercise the scanner's advance rule, digit runs far wider than i32, unbalanced polygon and separator markers, and long repetition runs, each through all four parsers in both `allow_empty_phrase` modes and through `post_process` for all fifteen tasks.

### Verified clean

**Catastrophic backtracking: not reachable.** `fancy_regex::Regex::new` (0.17.0, `lib.rs::new_options`) compiles to `RegexImpl::Wrap`, a plain `regex`-crate regex, whenever `analyze()` reports the tree is not `hard`. `hard` means lookaround, backreference, atomic group, possessive quantifier, or `\K`. None of the seven patterns here use any of those, so all seven run on the linear-time engine and the backtrack limit never applies. This covers the three patterns worth worrying about: `(?:<loc_\d+>){4,}`, `((?:<loc_\d+>)+)(?:<sep>|$)`, and the non-greedy `(.+?)` before eight location tokens in `OCR_RECORD`. The new adversarial test includes 4096-token runs and 2048 `<loc_1><sep>` repetitions; the whole 133-test florence2 suite finishes in 0.30s.

**UTF-8 boundary slicing: safe at every site.** `leading_phrase` takes its offsets from `char_indices()` and `whitespace_end` is the start of the `trim_start()` suffix, so both are character boundaries. `location_bins` only ever sets its cursor just past the ASCII `<loc_` prefix or just past the ASCII `>`, and it advances by at least 5 bytes on the no-match branch, so it is boundary safe and cannot loop. `strip_bare_loc_prefix` slices past an ASCII digit run and an ASCII `>`. `ascii_only` filters over `chars()`.

**No panics in the parse path.** No indexing, `unwrap`, or `expect` on attacker-controlled data. `parse_bin` saturates to `i32::MAX` instead of unwrapping. All coordinate arithmetic is f32, so no integer overflow; `quantize` returns NaN through `clamp` (which only panics on a NaN bound, not a NaN value) and the `as i32` cast saturates. `florence2_loc_token_id` guards `n < 1000` before `50269 + n`. `PHRASE_GROUNDING_BLACKLIST` is genuinely sorted, so the binary search is sound, and `blacklist_is_sorted_for_binary_search` already pins it.

**Complexity of the hand-rolled scanners is linear.** `location_bins` advances monotonically, so the repeated `text[index..].find` calls do not overlap and total O(n). `leading_phrase` does at most 8 `starts_with` probes per character, all of which fail on the first byte inside a `[^<]+` run, and it is called once per chunk over non-overlapping chunks. `parse_polygons` nests three iterations but each level partitions its parent, so the total stays linear in the answer length.

**Resource use on the hot path.** All seven regexes are `LazyLock`, compiled once. `Florence2Processor` holds one tokenizer and one image processor for its lifetime; nothing is reloaded per call. No large buffer is cloned in the request path.

### Reported, not fixed (LOW)

- `src/vision/processors/florence2.rs:213` - the pixel loop calls `rgb.get_pixel(x, y)` per pixel (about 590k bounds-checked calls) and does a divide per channel per pixel. Indexing `rgb.as_raw()` by row would remove the bounds checks, and hoisting `1.0 / std[c]` would remove 1.8M divides. Left alone deliberately: hoisting the reciprocal changes f32 rounding, and this PR's numerical parity against the Python reference is established. Not worth reopening for what is a low single-digit percentage of a request dominated by the DaViT forward pass.
- `src/models/florence2/scan.rs:121` - `strip_sequence_markers` chains three `String::replace` calls, so it copies the whole answer three times. One pass would do. Negligible at Florence-2 answer sizes.
- `src/models/florence2/parse.rs:217` - `parse_boxes` clones `cat_name` once per box, so a chunk with a long phrase and many boxes is quadratic in the answer length (worst case about `N^2 / 112` bytes). Bounded in practice: `generate_greedy` stops at `max_position_embeddings`, which is 1024 in the shipped config and capped at 65536 by this PR's own validation, putting the realistic worst case in the hundreds of kilobytes. Would need `Arc<str>` to fix, which is not worth the API churn at that bound.
- `src/vision/processors/florence2.rs:201` - `image.clone()` on the `do_resize: false` path copies the whole decoded image. Unreachable with any shipped config.

### For the follow-up wiring (#856)

`preprocess_with_sizes` accepts already-decoded `DynamicImage` values, which is the same boundary every other processor under `src/vision/processors/` uses, so decompression-bomb defense is the caller's responsibility. When the generate and server paths are wired up, user-supplied image bytes must go through `decode_request_image_with_limits` / `ImageInputLimits` in `src/server/model_worker.rs` (16384x16384 and a 512 MB decode budget by default) exactly as the other VLM families do. Nothing in this PR reaches a request path yet, so there is no exposure today.

Also worth noting for that issue: `Florence2Task::expand` interpolates the caller's `input` string into the encoder prompt for the seven tasks that take one. That is the feature, not a bug, but it means the task input is prompt-controlling, and the server surface should treat it as untrusted request data rather than as a trusted internal string.

Issue #778, opened 2026-07-13
fix(diffusion_gemma): demote tool-parser markers from special tokens at load


inureyes:
## Summary
DiffusionGemma checkpoints can register the tool-parsing markers `<|tool_call>`, `<tool_call|>`, `<|"|>`, `<|channel>`, `<channel|>` as special added tokens in the tokenizer. When they are special, they encode atomically instead of as the multi-token text the model was trained on, and decoding with skip-special-tokens strips them, so the server tool-call parser (`src/server/tool_calls/parser.rs`) never sees the markers. Demote these five markers to non-special at DiffusionGemma tokenizer load.

## Implementation plan
1. Verify the premise first on the actual checkpoint: inspect `tokenizer.json` / `tokenizer_config.json` (`added_tokens_decoder`) of the published DiffusionGemma checkpoint and confirm the five markers are flagged special. If they are not, close this issue with a comment stating the finding.
2. If confirmed: in the tokenizer load path (`src/tokenizer/mod.rs`, special-token registration around lines 37-44 and 358-363), add a model-scoped demotion hook applied for the diffusion_gemma family: keep the tokens registered as added tokens (so they do not split during encode) but clear their special flag (so decode retains them).
3. Watch the interaction with the reasoning-channel detection in `src/tokenizer/thinking.rs` (the `<|channel>` / `<channel|>` pair): thinking extraction must keep working after demotion; add a test covering it.
4. Tests: encode/decode round-trip keeps the markers; skip-special decode retains them; a tool-call parse over a synthetic diffusion output containing `<|tool_call>` markup succeeds.

## Acceptance criteria
- The premise check is documented in the issue or PR.
- Tool-augmented generation on DiffusionGemma parses tool calls; thinking-extraction regression tests pass.

Effort: low.

PR #1070, opened 2026-08-07
feat(models): add LocateAnything (locateanything) VLM support

### LINKED ISSUE #847: feat(models): add LocateAnything (locateanything) VLM support
## Summary

LocateAnything is a ~3B generative grounding VLM: a MoonViT vision tower feeds an MLP connector into a Qwen2-style decoder-only text backbone. Standard decode is ordinary autoregressive text generation whose output interleaves coordinate control tokens (box / coord / ref markers) encoding located regions; the checkpoint also ships a specialized parallel-box-decoding path (`pbd`) with multi-token-prediction heads (`n_future_tokens: 6`) for fast box emission. The text side (`model_type: qwen2`) is already fully supported in mlxcel, so the new work is the MoonViT tower, the connector, image-token splicing, and the grounding output handling. This is the strongest first VLM pick of the batch because the language model is done.

Concretizes the LocateAnything entry recorded as tracking-only in #524, filing it as a port work item.

## Upstream reference

- Directory: https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/locateanything
- Vision (MoonViT tower): https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/vision.py
- Language (Qwen2 decoder): https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/language.py
- Top-level model + connector: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/locateanything.py
- Parallel box decoding (grounding heads): https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/pbd.py
- Config: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/config.py
- Processor + image processor: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/processing_locateanything.py , https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/locateanything/image_processing_locateanything.py

## Public checkpoint

Confirmed available on HuggingFace:

- `nvidia/LocateAnything-3B` (original)
- `mlx-community/LocateAnything-3B-4bit` (MLX, primary validation target)
- `andai-labs/LocateAnything-3B-MLX` (alternate MLX conversion)

## Architecture notes

- `model_type: locateanything` with a `qwen2` text sub-config and a `moonvit` vision sub-config; `image_token_index: 151665`.
- Vision: MoonViT tower with a SwiGLU MLP; a connector projects pooled ViT features into the text embedding space.
- Text: Qwen2 decoder-only stack, already supported in mlxcel.
- Grounding contract: output text carries coordinate / box control tokens. The value of the model is the grounding output, so coordinate-token handling matters. The `pbd` path adds multi-token-prediction heads (`n_future_tokens: 6`) for parallel box emission; a minimal port can start with plain autoregressive decode (raw coordinate tokens in the text) and add the parallel path later.

## Implementation plan

- Reuse the existing Qwen2 decoder in `src/models/qwen2.rs` for the language model verbatim; the text arm already exists in `src/models/detection.rs` (`"qwen2" => Ok(ModelType::Qwen2)`).
- MoonViT already exists in-tree for Kimi-VL: reuse `src/vision/encoders/kimi_vl.rs` (encoder) and the VLM wiring in `src/vision/kimi_vl.rs` + `src/loading/vlm_kimi_vl.rs` as the closest template. Confirm the LocateAnything MoonViT config keys against Kimi-VL and diverge only where they differ.
- Reuse qwen2_vl-style patch / RoPE ViT helpers (`src/vision/encoders/qwen2_5_vl.rs`, `src/vision/qwen2_vl.rs`) where the MoonViT patch / RoPE logic overlaps.
- Add a new `src/vision/locateanything.rs` VLM module + `src/loading/vlm_locateanything.rs` loader following the shape of the Kimi-VL pair.
- Image-token splice: expand `image_token_index: 151665` placeholders and scatter projected features via the generic merge path in `src/vision/merge.rs` / `src/models/multimodal_placeholders.rs`.
- Grounding: for the first landing, emit coordinate tokens as plain text (no post-processing) and validate against a grounded prompt; scope the `pbd` parallel-box path and coordinate-to-box post-processing as a follow-up.

- [ ] Text config (`qwen2`) + vision config (`moonvit`) parse from `config.json`.
- [ ] MoonViT vision encoder + MLP connector implemented (reusing the Kimi-VL MoonViT).
- [ ] Image-token splice / merge: `image_token_index: 151665` placeholder expansion + feature scatter.
- [ ] `from_weights` + `sanitize` handle mixed precision (vision often plain bf16 while text is quantized) and any conv / patch weight channels-last remap.
- [ ] Arch-string arm `"locateanything"` added in `src/models/detection.rs`.
- [ ] TP / distributed arch-string + `generate_vlm` summary updated if applicable.
- [ ] `_tests.rs` unit tests (config parse, connector shapes, placeholder expansion).
- [ ] `docs/supported-models.md` updated.
- [ ] Real-checkpoint validation: `./target/release/mlxcel generate -m models/LocateAnything-3B-4bit --image  -p "<grounding prompt>"` produces correct grounded output (coordinate tokens referencing the right region). Follow `docs/adding-models.md`.
- [ ] Follow-up scoped (not blocking this issue): `pbd` parallel-box-decoding path + coordinate-token-to-box post-processing.

## Effort

MEDIUM-HIGH. The language side is already done; the delta is the MoonViT tower (largely present via Kimi-VL), the connector, and the grounding output contract. Best first VLM pick of this batch.

inureyes:
## Summary

Adds native support for LocateAnything (`model_type: locateanything`), NVIDIA's ~3B generative grounding VLM: a MoonViT vision tower and an MLP connector feeding a Qwen2 text decoder whose output interleaves `<ref>`/`<box>` markers and the 1001 coordinate tokens `<0>`..`<1000>` (ids 151677..152677). Plain autoregressive decode already produces those, so no special detokenization is involved. Validated end to end on `mlx-community/LocateAnything-3B-4bit`, where the first emitted box is byte-identical to the output the model card documents for its own example image.

## What changed

**Reuse over duplication.** The tower is the same MoonShot MoonViT-SO-400M that Kimi-VL embeds, so `src/vision/encoders/kimi_vl.rs` is reused rather than copied. Its upstream LocateAnything implementation diverges in exactly two places, both numeric: the LayerNorm epsilon is `1e-5` rather than `1e-6` (already a config field) and the block MLP uses `nn.GELU(approx="precise")`, MLX's tanh approximation, rather than the exact erf form. The activation becomes `MoonViTMlpActivation` on the vision config, defaulting to the erf form, so Kimi-VL and Kimi-VL 2.5 are byte-unchanged. `mlxcel_core` exposes no elementwise tanh GELU (both `gelu` and, despite its name, `gelu_approx` are erf-based), so it is synthesised from primitives the way `models::kokoro::ops::gelu_new` does. The text side is `models::Qwen2Model` verbatim.

**New files.** `src/vision/locateanything.rs` (VLM runtime + `multi_modal_projector` connector), `src/vision/processors/locateanything.rs` (native-resolution processor), `src/multimodal/locateanything_prompt.rs` (image-marker expansion), `src/loading/vlm_locateanything.rs` and `src/loading/vlm_locateanything_quant.rs` (loader), each with a `_tests.rs` beside it.

**Where LocateAnything is genuinely not Kimi-VL.** The connector normalizes over the flattened merged patch (`vision_hidden * merge_h * merge_w`, 4608 in the released checkpoint, which its `layer_norm.weight` shape confirms) rather than over the 1152-wide vision hidden dim. The processor resizes each side UP to the next multiple of `merge * patch` where Kimi-VL centre-crops DOWN, and normalizes with `mean = std = 0.5` rather than the CLIP statistics. Neither is shared.

**Image-token splice.** `<image-N>` is what the chat template renders, and it is plain text rather than a vocabulary token, so the rewrite into ` + <IMG_CONTEXT> * (grid_h * grid_w / merge_length) + </img>` happens on the rendered prompt and is re-encoded, matching upstream's `re.sub`. A token-level splice covers `--no-chat-template` runs. The runtime then verifies the encoded stream carries exactly as many `<IMG_CONTEXT>` ids as the tower will emit feature rows, so a drift cannot silently scatter image features into text positions.

**Control-plane wiring.** `ModelType::LocateAnythingVLM` with the `"locateanything"` detection arm (which has to win at the top level, because the text sub-config also says `"qwen2"`), `for_each_model_registration!` entry, `LoadedModel` variant, `VlmRuntimeRef::LocateAnything`, the `try_load_vlm_model_from_dir` route, a `generate_vlm` summary line, and the TP arch-string table (LocateAnything's backbone is `llama`-family; TP is refused for VLM-kind models earlier, this just keeps the dispatch total).

## Two real-checkpoint gaps the Python reference did not show

Both were found by running the actual checkpoint, not by reading `mlx_vlm`.

**The conversion is `mixed_4_8`, not uniform 4-bit.** The model card explains why: pure 4-bit on the tied `embed_tokens` destroys coordinate-token precision. 18 of the 36 layers store `v_proj` at 8 bits while `q_proj` and `k_proj` stay at 4. Every per-tensor loader reconciles its own width from tensor shapes and is fine, but `FusedQKVLinear::from_weights_separate` concatenates the three packed planes along axis 0 and infers one width from `q_proj`, which is a hard shape error inside MLX for a mixed layer (`[2048, 256]` against `[256, 512]`). The loader dequantizes the three planes of an affected layer, which is exact. It deliberately does not requantize the narrow planes up to the wider width: MLX's affine quantizer snaps the group scale onto the larger-magnitude edge rather than using a plain `(max - min) / (2^bits - 1)`, so a 4-bit group does not land on the 8-bit grid and a round trip perturbs the weights by up to half an 8-bit step (measured 3.7e-3 on a synthetic plane before switching approaches). The cost is roughly 190 MB on the released 3B checkpoint. Note this is not LocateAnything-specific: any family on the shared fused-QKV loader would hit the same wall on an `mlx_lm --quant-predicate mixed_4_8` conversion. Fixing it properly needs a `FusedQKVLinear` that can hold per-projection widths, which is a shared-layer change and is left as a follow-up.

**The checkpoint ships no `tokenizer.json`.** Neither `nvidia/LocateAnything-3B` nor its MLX conversions exported a fast tokenizer, only `vocab.json` + `merges.txt` + `added_tokens.json` + `tokenizer_config.json`, so `load_tokenizer` had nothing to read and the model could not load at all. A new last-resort fallback reconstructs the tokenizer the way `transformers`' `Qwen2Converter` does, component for component: byte-level BPE with no unk token and no subword prefix or suffix, NFC normalizer, `Sequence[Split(PRETOKENIZE_REGEX, isolated), ByteLevel(add_prefix_space, use_regex=false)]`, ByteLevel decoder, ByteLevel post-processor with `trim_offsets = false`. It is gated on `tokenizer_class` naming the Qwen2 tokenizer as well as on both files being present, because `vocab.json` + `merges.txt` is the generic GPT-2 slow-tokenizer pair and another family shipping it must not be tokenized with Qwen2's rules. Added tokens are registered in ascending id order and then verified against the ids the checkpoint declares, since `tokenizers` assigns them sequentially from the base vocab size and a gap would shift every later token silently; LocateAnything's 1038 added tokens include the 1001 coordinate tokens that carry its box output, where an off-by-one would corrupt every box rather than fail loudly.

## Real-checkpoint validation

GB10, CUDA (sm_121), release build, `mlx-community/LocateAnything-3B-4bit`, on the COCO image the model card's own usage example uses (`http://images.cocodataset.org/val2017/000000039769.jpg`).

```
$ ./target/release/mlxcel generate -m models/mlx/locateanything-3b-4bit \
    --image 000000039769.jpg -p "Detect all objects in the image." -n 128 --temp 0.0
LocateAnything: inserted 1 image block(s) (414 total image tokens)
<ref>object</ref><box><64><152><273><244></box><box><520><160><580><392></box>
[Generated 15 tokens in 1.68s = 8.91 tok/s]
```

The first box `<64><152><273><244>` is byte-identical to the `<ref>remote</ref><box><64><152><273><244></box>` the model card documents for this image, and the `object` label instead of `remote` is exactly the semantic generalization the card documents for this 4-bit conversion. The second box is 333,77,371,188 in pixels against a COCO ground-truth remote of 333,72,368,187.

Referring and multi-object queries on the same image:

```
-p "...the region this sentence describes: the cat on the right."
-> <ref>the cat on the right</ref><box><541><49><1000><778></box>     (346,24,640,373 px; GT 345,23,640,368)

-p "Detect all the cats in the image."
-> <ref>the cats in the image</ref><box><9><112><494><989></box><box><541><52><1000><775></box>
                                    (6,54,316,475 px; GT 13,52,314,470)  (346,25,640,372 px; GT 345,23,640,368)
```

`./target/release/mlxcel arch` lists `LocateAnything (MoonViT + MLP connector + Qwen2, grounding)`.

The tokenizer was separately validated against a `transformers` `Qwen2Tokenizer` oracle built from the same checkpoint directory: 7 cases covering the rendered ChatML prompt, a `<ref>`/`<box>` grounding string, repeated whitespace and newline runs, tabs, CJK and accented Latin text, and the apostrophe contractions the pre-tokenize regex special-cases. All 7 encode token-exact and decode round-trip exactly. That harness was a throwaway and is not part of the diff.

## Finalization pass

One review-fix seam had no regression test: the bf16-ordering fix (densify the mixed-precision QKV planes before the Apple Silicon bf16 -> f16 pass, not after) was previously validated only by reading the diff, because the host that landed it could not run its own test suite in that session. `load_locateanything_vlm`'s two inline passes are now `reconcile_mixed_precision_weights`, a private function that takes the Apple Silicon gate as a `convert_bf16: bool` parameter instead of reading `hardware::get_hardware()` internally, so a test can force the conversion branch independent of the host it runs on (this GB10 box is CUDA, not Apple Silicon, and would otherwise never execute that branch). Two new tests in `src/loading/vlm_locateanything_tests.rs` quantize a bf16-sourced dense plane the way the released checkpoint's mixed-precision layers do (confirming the scales stay bf16, which is the premise of the whole bug), densify it, and assert the densified planes land on f16 only when the conversion flag is set; the control test proves the f16 dtype comes from the conversion pass and not from densification itself. No behavior changed, only the call composition became a named, independently testable unit.

## Test plan

- [x] `cargo test --release --features cuda --lib locateanything -- --test-threads=1` (53 passed, 0 failed)
- [x] `cargo test --release --features cuda --lib tokenizer -- --test-threads=1` (75 passed, 0 failed, 5 ignored)
- [x] `cargo test --release --features cuda --lib kimi_vl -- --test-threads=1` (25 passed, 0 failed)
- [x] `cargo test --release --features cuda --lib detection -- --test-threads=1` (80 passed, 0 failed)
- [x] `cargo test --release --features cuda --lib model_metadata -- --test-threads=1` (8 passed, 0 failed)
- [x] `cargo test --release --features cuda --lib loading:: -- --test-threads=1` (255 passed, 0 failed)
- [x] `cargo test --release --features cuda --lib vision:: -- --test-threads=1` (314 passed, 0 failed, 17 ignored)
- [x] `cargo test --release --features cuda --lib multimodal:: -- --test-threads=1` (187 passed, 0 failed, 17 ignored)
- [x] `cargo test --release --features cuda --test kimi_vl_parity -- --test-threads=1` (3 passed, 0 failed)
- [x] `cargo clippy --features cuda --lib --tests -- -D warnings` (clean)
- [x] `cargo fmt --all`
- [x] Real-checkpoint generation above

Not run here, left to the broader gate: the full `cargo test --workspace` and any non-CUDA / Apple Silicon path. The Apple Silicon bf16-to-f16 branch in the loader mirrors the InternVL one line for line but was not exercised on this Linux box.

## Follow-up (explicitly out of scope for #847)

- The `pbd` parallel box-decoding path with its multi-token-prediction heads (`n_future_tokens = 6`), and coordinate-token-to-box post-processing. Plain autoregressive decode emitting coordinate tokens as text was the target here.
- A `FusedQKVLinear` that holds per-projection quantization widths, which would remove the dequantize-on-mixed-precision workaround for every family, not just this one.

Closes #847


inureyes:

### Intent

Add native `locateanything` VLM support (MoonViT tower reused from Kimi-VL + MLP connector + Qwen2 text decoder), plain autoregressive grounding decode only, with `pbd` and coordinate-to-box post-processing explicitly deferred.

### Findings Addressed

- [x] **(HIGH)** `src/loading/vlm_locateanything.rs`: the Apple Silicon bf16 to f16 pass ran *before* `densify_mixed_precision_qkv`. MLX's `dequantize` returns an array carrying the scales' dtype, and the keep-predicate deliberately holds `.scales` / `.biases` at bf16, so every dense q/k/v plane the densification inserted was a fresh bf16 tensor created after the only pass that would have converted it: 54 tensors (18 mixed layers x 3 projections) on the released 3B checkpoint, defeating the exact M5 JIT-crash mitigation the loader's own module doc says it exists to apply. Densification now runs first. Independently confirmed by the Codex reviewer as its sole P1.
- [x] **(MEDIUM)** `src/tokenizer/mod.rs` `read_added_tokens_sorted`: `special` defaulted to `true` in both branches, inverting HuggingFace's `AddedToken` default. The `added_tokens.json` branch was worse, forcing *every* entry special although that file carries no flags at all. Since `AddedVocabulary`'s `special_tokens_set` is insert-only (already documented in this file for #778), a content-bearing token registered special can never be demoted and is stripped from every `skip_special_tokens` decode. Both defaults are now `false`, with two new unit tests covering both branches. LocateAnything is unaffected either way: its `added_tokens_decoder` carries an explicit `special` on all 1038 entries.
- [x] **(LOW)** `src/vision/processors/locateanything.rs`: `merged_token_count` reimplemented `multimodal::locateanything_prompt::merged_token_count` and was reachable only from its own unit test. It now delegates, per the reuse direction in #847.

Fixed in `5ad34371`.

### Remaining Items

- [ ] **(MEDIUM, pre-existing, deliberately not fixed here)** The no-placeholder fallback in `locateanything_prompt.rs:216-220` splices the image block after `prompt_tokens[0]`, which under `--no-chat-template` is an ordinary BPE token rather than a ChatML framing token, so the block can land inside the first word. Raised by the Codex reviewer. This is not introduced by this PR: `internvl_prompt.rs:118`, `smolvlm_prompt.rs:133`, and `kimi_vl_prompt.rs:134` all use the identical `let head = prompt_tokens[0]` splice. Changing it for LocateAnything alone would make it inconsistent with three sibling families, so it belongs in a follow-up against the shared pattern.
- [ ] **(LOW)** The new tokenizer path does not call `ensure_bos_post_processor`, unlike the `tokenizer.json` branch. Only reachable for a Qwen2 slow-tokenizer checkpoint declaring `add_bos_token: true`; no regression, since the path cannot run when `tokenizer.json` exists.

### Notes on the three regression-risk areas

**1. Tokenizer gating (`src/tokenizer/mod.rs`): verified safe.** The fallback cannot preempt any checkpoint that previously loaded a real tokenizer. `load_tokenizer` returns unconditionally from the `tokenizer.json` branch at line 1236, and the new `is_qwen2_slow_tokenizer_dir` check sits at line 1284, after `tokenizer.model`, `*.tiktoken`, and `tokenizer.jsonl`. It additionally requires `tokenizer_class` to be exactly `Qwen2Tokenizer` or `Qwen2TokenizerFast`, so an absent or Qwen2-adjacent class falls through unchanged rather than being tokenized with Qwen2's rules. Checked against a real local checkpoint: `moondream2` ships `vocab.json` + `merges.txt` (the generic GPT-2 pair) but declares `CodeGenTokenizer`, so it is excluded by the class gate *and* by the `tokenizer.json` early return. The one theoretical ordering interaction, preempting the `moondream3` remote-tokenizer fallback, requires a moondream3 checkpoint with no `tokenizer.json` that declares a Qwen2 class; moondream3 is CodeGen-family, so this is unreachable in practice. The reconstruction itself matches `transformers` `Qwen2Converter` component for component (BPE with no unk / no prefix / no suffix, NFC, `Sequence[Split(regex, isolated, invert=false), ByteLevel(add_prefix_space, use_regex=false)]`, ByteLevel decoder, ByteLevel post-processor with `trim_offsets=false`).

**2. Mixed 4/8-bit handling (`vlm_locateanything_quant.rs`): logic correct, one ordering bug (fixed).** Dequantization is exact by construction and the unit tests assert `max_abs_diff == 0.0` on all three planes. Scoping is right: only prefixes ending `.self_attn` that carry `q_proj.scales` are considered, per-projection layouts are reconciled from actual tensor shapes, and a layer is skipped when all three agree on both `bits` and `group_size`, so uniform-4-bit and uniform-8-bit layers stay packed (both covered by positive-control tests). The `unsafe` block is correctly scoped with a SAFETY comment; the raw biases pointer is null or borrowed from a live entry in the same map for the call's duration. The roughly 190 MB cost is documented in the module doc and logged via `tracing::info!`. The only defect was the loader-side call ordering above.

**3. Kimi-VL MoonViT refactor: byte-unchanged.** `MoonViTMlpActivation` derives `Default` with `#[default]` on `Gelu` (the erf form) and the config field carries `#[serde(default)]`, so a Kimi-VL `config.json`, which has no `mlp_activation` key, deserializes to the exact prior behavior at `vlm_kimi_vl.rs:60`. `GeluTanh` is set only in `LocateAnythingVisionConfig::to_moonvit_config`, from a struct literal, never from Kimi-VL's deserialization path. `tests/kimi_vl_parity.rs` was not weakened: the only change is adding `mlp_activation: MoonViTMlpActivation::Gelu` to the `tiny_config` struct literal (mandatory once the field exists) plus the import; every assertion is unchanged. Same for `kimi_vl_tests.rs`.

### Verification

- [x] All stated requirements implemented (every acceptance criterion in #847 except the two it explicitly defers)
- [x] No placeholder/mock code remaining (no TODO / FIXME / `todo!` / `unimplemented!` in the diff; `pbd` appears only in doc comments marking it out of scope)
- [x] Integrated into project code flow (detection arm, `for_each_model_registration!`, `LoadedModel`, `VlmRuntimeRef`, loader route, `vlm_runtime` dispatch, `generate_vlm` summary, TP arch string, `docs/supported-models.md`; the only sites where an analogous VLM is registered and LocateAnything is not are the two video paths in `server/startup.rs` and `server/model_worker.rs`, which is correct since LocateAnything is image-only, and InternVL is likewise absent there)
- [x] Existing modules reused where applicable (Kimi-VL MoonViT encoder, `models::Qwen2Model`, `vision::merge::merge_llava`, `load_vlm_weights_common`, `reconcile_quantization_layout`)
- [ ] Tests pass (not run in this review; CUDA test execution is owned by the merge gate running concurrently)

inureyes:

Reviewed for security vulnerabilities and performance issues. Two fixes pushed (`9f65b9d4`, `b082ed26`). Details below, including what was checked and found safe, since that constrains the severity of what remains.

### `unsafe` block: SOUND

`src/loading/vlm_locateanything_quant.rs:152` is the standard cxx nullable-`biases` bridge pattern already used at 17+ call sites (for example `src/lib/mlxcel-core/src/layers.rs:514`).

- `b.as_ref()` correctly yields `None` for a null `UniquePtr`, so a missing `biases` becomes a null pointer rather than a dangling one. The C++ side (`mlx_cxx_bridge.cpp:2968`) dereferences only when non-null, copies into `std::optional<array>`, and never retains the pointer.
- The raw pointer derives from a shared borrow of `weights` that is still live at the call, and the map is not mutated until the `insert` after the block ends. No aliasing or use-after-free. A null `UniquePtr` in `w`/`s` panics in `Deref` rather than causing UB.
- Attacker-influenced values cannot reach it unvalidated: `reconcile_quantization_layout` runs first and validates `group_size`, `bits`, and the `mode` string, propagating `Err` before `dequantize` is called. A malformed checkpoint yields a clean error.

### Fixed

**Division by zero, `src/multimodal/locateanything_prompt.rs:137`.** `merge_kernel_size: [65536, 65536]` gives a `usize` product of exactly `2^32`; `.max(1)` left it and `as i32` truncated it to 0, so the next line divided by zero. That pair is square and non-zero, so it passed both existing guards. Confirmed by running it at `-O`. Now uses `saturating_mul` plus `clamp(1, i32::MAX)` before the cast, with the patch product widened to `i64`.

**Truncating `input_dim`, `src/loading/vlm_locateanything.rs:363`.** `(hidden_size * merge_h * merge_w) as i32` could overflow and truncate; with the released `hidden_size: 1152` and that same merge pair it produced `input_dim == 0`, reaching `reshape(image_features, &[-1, 0])` and throwing inside MLX, which aborts across the cxx boundary instead of unwinding. Now a `connector_input_dim` helper using `checked_mul` plus `i32::try_from`.

**Unbounded allocation from `patch_size` and `in_token_limit`.** The only ceiling on the resize was the 511-patch grid envelope, so the resized side scaled as `511 * patch_size` and the f32 patch buffer as `patch_size^2`. A declared `patch_size: 100000` asked `resize_exact` for a ~200000x200000 RGB buffer. The `in_token_limit` downscale could not save it, since `(w / p) * (h / p)` is 0 once `p` exceeds the image. Separately an unbounded `in_token_limit` from `preprocessor_config.json` left one image bounded only by that envelope, roughly 614 MB of f32 patch data. `patch_size` and `merge_kernel_size` are now rejected at `to_moonvit_config` rather than clamped, because both are consumed twice and clamping only the processor would desync the emitted patch grid from the MoonViT tower sized from the unclamped config. `in_token_limit` is clamped, since the tower never sees it.

**Quadratic tokenizer DoS, `src/tokenizer/mod.rs`.** The added-token loop registered one token per call, and each call rebuilds both Aho-Corasick automatons over everything accumulated so far. Measured release timings: 4000 tokens 2.51 s, 8000 10.3 s, 16000 42.9 s, extrapolating to about 28 minutes for 100k entries with no error and no timeout. This was also a regression against the `tokenizer.json` path on main, which batches: 10.2 ms versus 42.9 s at 16k. Now batches consecutive runs of the same `special` flag. Runs rather than a two-group split, because `add_tokens` assigns ids in slice order, so grouping all specials first reorders them (verified: a naive split maps declared 8 and 9 onto assigned 9 and 8). Id preservation was checked across 33 flag patterns including the 1038-token LocateAnything shape, all identical. Measured after: 1038 tokens 197.6 ms to 1.5 ms, 16000 tokens 44.05 s to 23.3 ms, with no regression in the alternating worst case.

**Silent token-id aliasing, `src/tokenizer/mod.rs`.** `base_vocab_size` came from `vocab.len()`, but `vocab.json` ids need not be dense and `tokenizers` also assigns from `len()`. With a 5-entry vocab whose max id is 5 plus an added token declaring id 5, the build returned `Ok` and `decode([5])` flipped from `"COLLIDE"` to `"<x>"`, with two tokens sharing id 5. Recomputing the base size does not help, since the library still assigns from `len()`, so a non-dense `vocab.json` is now rejected. All 65 real `vocab.json` files checked locally are dense.

### Checked and safe

- Path handling in the tokenizer is clean. Every `join` uses a hardcoded literal, so no path component originates from file content and `../` escape is not possible.
- No `unwrap`, `expect`, indexing, or `with_capacity` in the new tokenizer non-test code. All 16 malformed-input cases return `Err` rather than panicking. No UTF-8 boundary hazard, since the empty `continuing_subword_prefix` makes the one boundary-sensitive BPE slice a no-op.
- With a valid config, image-driven allocation is bounded at roughly 61 MB of f32 per image. The `in_token_limit` downscale and the grid clamp together make a 510x510 grid unreachable, so adversarial aspect ratios cannot escape the budget.
- The densification path is load-time only (single call site), engaged only when the three planes genuinely disagree on bits or group size, and bounded by model size. The decode hot path delegates straight to the Qwen2 backbone with no per-token allocation.

### Not fixed, for maintainer judgement

- **`language_weights_subset` (`src/loading/vlm_locateanything.rs:373`) copies every `language_model.*` tensor** via `mlxcel_core::copy`, which is a real array copy, transiently duplicating the whole text stack (roughly 1.8 GB for the 3B checkpoint) until the `drop` after `Qwen2Model::from_weights`. This stacks with the 190 MB densification. It is the identical pattern in `vlm_internvl.rs:247` and `vlm_lfm2_vl.rs:81`, so it is a family convention rather than a regression here. Since the language keys are disjoint from the vision and connector keys, draining instead of copying would remove the peak, but that is a shared-pattern change.
- **Unbounded `u32` ids from `vocab.json` reach `encode()` output.** The samplers bounds-check (`sampling.rs:222`, `:786`, `:1047`, `:1382`), but the prompt-embedding gather was not traced to a bounds check. Not new in kind: main's `tokenizer.json` path accepts the same range.
- **`tokenizer_config.json` is read and parsed three times per load** (roughly 1.3 s on a large file). Measured RSS amplification is 1.0x, so this is cost rather than risk.
- **An empty prompt combined with an image** skips the placeholder-count guard in `vlm_runtime.rs` and reaches `merge_llava` with feature rows but no matching positions. Niche, needs `--no-chat-template` with an empty prompt, and produces a confusing MLX-level failure rather than a clean error.

### Verification

Linux aarch64, CUDA feature, `--test-threads=1`, `MLX_ENABLE_TF32=0`. Re-run independently rather than taken from the fix runs: 51 passed across the four `locateanything` suites, 75 passed in `tokenizer` (5 ignored). `cargo clippy --features cuda -p mlxcel --lib --tests -- -D warnings` clean, `cargo fmt --check` clean. None of the 5 pre-existing failures on `origin/main` are affected.

One testing note worth carrying forward: the filter `--lib loading::vlm_locateanything` matches zero tests and prints `test result: ok`, a false green. The module path is `loading::vlm::locateanything`, declared under a `vlm` parent rather than by filename.

PR #1082, opened 2026-08-07
feat(models): support quantized Florence-2 checkpoints (4-bit / 8-bit / 6-bit / 3-bit)

### LINKED ISSUE #1072: feat(models): support quantized Florence-2 checkpoints (4-bit / 8-bit / 6-bit / 3-bit)
## Problem / Background

Epic #850 landed Florence-2 support across #852, #853, #854, #855, and #856, but only for bf16/f16 exports. `Florence2Model::load` (`src/models/florence2/model.rs`) rejects quantized checkpoints at load time with a named error:

```
Florence-2 quantized checkpoints are not supported yet: the BART text stack and the DaViT tower are built from dense layers. Use a bf16 or f16 export, for example mlx-community/Florence-2-base-ft-bf16.
```

That refusal was a deliberate decision made in #854. Neither the BART seq2seq stack from #852 nor the DaViT tower from #853 has a quantized code path, so a packed uint32 weight reaching MLX would abort the process rather than raise a catchable error, because MLX's eager throw cannot cross the FFI boundary. Refusing early with a named error was the right call. The gap is that the refusal is now permanent.

Practical impact: the following confirmed-available mlx-community conversions cannot be loaded at all.

- `mlx-community/Florence-2-large-ft-4bit`, `-8bit`, `-6bit`, `-3bit`
- `mlx-community/Florence-2-base-ft-4bit`, `-8bit`, `-6bit`, `-3bit`

Only the `-bf16` variants work. `docs/supported-models.md` already states this limitation plainly, so the documentation is accurate. This issue is about closing the gap, not correcting a false claim.

## Proposed Solution

Add quantized weight handling to Florence-2, following the quantized-path conventions already used by the other VLM loaders in this tree.

- BART encoder and decoder linear projections (`src/models/florence2/encoder.rs`, `decoder.rs`, `layers.rs`): route packed weight/scales/biases through the quantized linear path instead of dense `Linear`.
- DaViT tower projections (`src/models/florence2/` vision path, from #853): same treatment for the window/grid attention and MLP projections.
- Fusion and projection layers (`fusion.rs`) where the checkpoint quantizes them.
- Narrow the load-time refusal in `model.rs` so it only rejects what genuinely remains unsupported (for example a quantization group size or bit width the path does not implement), rather than rejecting every checkpoint that carries quantization metadata.

Validation: load `mlx-community/Florence-2-base-ft-4bit` and confirm that caption, OCR, and detection outputs stay consistent with the bf16 path within a stated tolerance. Record the tolerance in the issue closing comment.

The existing gated parity tests are the pattern to extend:

- `tests/florence2_text_parity.rs`
- `tests/florence2_vision_parity.rs`
- `tests/florence2_fusion_parity.rs`
- `tests/florence2_processor_parity.rs`
- `tests/florence2_postprocess_parity.rs`

## Acceptance Criteria

- [ ] A 4-bit checkpoint (`mlx-community/Florence-2-base-ft-4bit`) loads without error.
- [ ] The CLI produces correct caption output on a real image (`mlxcel generate -m <model> --image  -p '<CAPTION>'`).
- [ ] The CLI produces correct OCR output on a real image (`<OCR>` and `<OCR_WITH_REGION>`).
- [ ] The CLI produces correct detection output on a real image (`<OD>`), with boxes parsed into pixel space.
- [ ] Quantized output is consistent with the bf16 path within a tolerance stated in the issue, and the tolerance is justified.
- [ ] Parity tests extended to cover the quantized path, following the existing gated-test pattern.
- [ ] `docs/supported-models.md` updated to drop the bf16-only restriction and describe what is actually supported.
- [ ] The load-time refusal in `Florence2Model::load` is narrowed to whatever remains genuinely unsupported, not removed wholesale.
- [ ] At least one `-large-ft` quantized variant smoke-tested in addition to `-base-ft`.

## Technical Considerations

- Quantized models keep bf16 scales and biases as-is per the Apple Silicon precision policy; only non-quantized weights get the bf16 to f16 conversion in the loading path. Check that `convert_bf16_weights` is applied on the correct side of the quantized branch in `model.rs`.
- The four bit widths (3, 4, 6, 8) may not share a group size across the mlx-community conversions. Read the quantization metadata per checkpoint rather than assuming a default.
- Florence-2 generates through its own seq2seq CLI pipeline (cross-attention against cached encoder output) rather than the shared autoregressive loop, so the quantized path has to be validated through that pipeline specifically.
- `mlxcel-server` still refuses Florence-2 checkpoints at startup. That is a separate limitation and is out of scope here.

Refs #850, #852, #853, #854, #855, #856

inureyes:
## Summary

Florence-2 now loads the `mlx-community` quantized conversions (3-bit, 4-bit, 6-bit, 8-bit, both `base-ft` and `large-ft`) instead of refusing every checkpoint that carries quantization metadata. The BART projections, the LM head, the DaViT window and channel attention, the tower MLPs, the shared token table, both BART position tables, and both learned 2-D image position tables all route through `UnifiedLinear` / `UnifiedEmbedding`, which pick the quantized path per prefix from the presence of a `.scales` sibling and fall back to the dense layer otherwise, so a bf16 export loads through exactly the same code.

The #854 refusal is narrowed rather than removed: a checkpoint that packs one of the tensors this family genuinely consumes dense (`image_projection`, the cosine temporal buffer, the conv stack) is still rejected at load, now with the offending tensor named.

## Validation

Numerically pinned against upstream mlx-vlm running the *same* 4-bit checkpoint, which is the reference that can actually answer "is the graph right". Comparing 4-bit against bf16 can only answer "how lossy", because quantization is lossy by construction. On the deterministic synthetic input the dense Florence-2 parity tests already use, mlxcel matches upstream mlx-vlm on `Florence-2-base-ft-4bit` at the dense parity tests' own tolerances (max abs deviation 1e-2 on image features, 8e-2 on encoder hidden states, 5e-2 on step-0 logits), and reproduces its greedy ids `[0, 879, 27740, 868]` exactly.

The second half of the numeric claim is that the distance from dense is the checkpoint's lossiness and not something this port introduced. Measured on this box, 4-bit against bf16 through identical code:

| stage | mlxcel relative RMS | mlx-vlm relative RMS | mlxcel cosine | mlx-vlm cosine |
|---|---|---|---|---|
| image features | 0.25486 | 0.25540 | 0.967595 | 0.967458 |
| encoder hidden | 0.41641 | 0.41583 | 0.913923 | 0.914181 |
| step-0 logits | 0.11858 | 0.11861 | 0.992970 | 0.992966 |

The two implementations lose the same amount to the same weights: relative RMS agrees to 5.8e-4 and cosine to 2.6e-4 at every stage. `tests/florence2_quantized_parity.rs` asserts that agreement (tolerance 5e-3 on relative RMS, 2e-3 on cosine) rather than a hand-picked "close enough" threshold, because a threshold only says the result is not absurd while this says the port pays the reference implementation's cost. The image features are the loosest stage because the DaViT tower is pre-norm, so twelve blocks of 4-bit projections accumulate along an unnormalized residual stream before `image_proj_norm` renormalizes; the logits are the tightest because the BART decoder is post-norm and renormalizes after every sublayer. The greedy caption is identical between 4-bit and bf16 on this input.

## What changed

- `src/models/florence2/checkpoint.rs`: new `Florence2Quantization` (`group_size`, `bits`) parsed from the top-level `quantization` object with a range guard, `Florence2Quantization::config_is_quantized` for the bf16 gate, and `reject_unsupported_quantized_tensors`, the narrowed replacement for #854's blanket refusal. `sanitize`'s `model.shared` fill now carries `.scales` and `.biases` alongside `.weight`, so a checkpoint that materializes only `embed_tokens` does not present a packed table as a dense one.
- `src/models/florence2/layers.rs`: BART projections through `UnifiedLinear`; new `embedding_table_rows` adapter over the shared `validate_embedding_table` guard, whose quantized arm reconstructs the logical width from `scales` and the reconciled group size (a packed table's stored width is a function of bit depth, so the old `cols == d_model` comparison rejects every quantized checkpoint).
- `src/models/florence2/encoder.rs`, `decoder.rs`: `embed_positions` is an `nn.Embedding` upstream and is packed in a quantized export, so it loads through `UnifiedEmbedding` and the forward gathers rows `[offset + 2, offset + 2 + seq)` with `arange` instead of slicing the raw tensor. On a dense checkpoint the gather returns exactly the rows the slice returned.
- `src/models/florence2/fusion.rs`: both `image_pos_embed` tables through `UnifiedEmbedding`. The row/column split now comes from `embedding_dim` (`dim // 2` and the remainder, matching upstream) rather than from the loaded shapes, which are packed widths on a quantized checkpoint.
- `src/models/florence2/mod.rs`: shared token table and LM head through the unified layers, plus the activation-dtype fix below.
- `src/models/florence2/model.rs`, `src/vision/encoders/florence2_davit.rs`: the narrowed refusal runs after sanitize, and `convert_bf16_weights` is gated on the checkpoint being dense.
- `src/vision/encoders/florence2_davit_attention.rs`, `florence2_davit_blocks.rs`: tower qkv / proj / fc1 / fc2 through `UnifiedLinear`; `BlockParams` carries the packing so no extra parameter threads through three struct layers.
- `src/models/gpt2.rs`, `src/lib/mlxcel-core/src/layers.rs`: `// Used by:` comments updated for the two shared items this now calls (`validate_embedding_table`, `UnifiedEmbedding::quantized`). No behavior change.
- `docs/supported-models.md`: the bf16-only restriction is replaced by what actually works, including which tensors stay dense and why.
- Tests: `tests/florence2_quantized_parity.rs` (checkpoint-gated, upstream pins) and `src/models/florence2/florence2_quantized_tests.rs` (checkpoint-free: config parsing, the narrowed refusal, the three-plane shared fill).

The issue is a sketch; these are the places the checkpoints and the code disagreed with it.

- **The group size does not vary.** The issue cautions that the four bit widths "may not share a group size across the mlx-community conversions". All eight published conversions (`base-ft` and `large-ft` at 3, 4, 6, and 8 bits) declare `group_size: 64`; only `bits` varies. The implementation still reads both from `config.json` rather than assuming, because `UnifiedLinear` treats the declared group size as trusted and re-derives the bit width from the packed shapes, so a wrong group size would mis-stride every dequantization silently.
- **The embedding tables are the larger half of the change, and the issue does not mention them.** Its proposed solution lists only the BART linear projections, the DaViT projections, and "fusion and projection layers". But `language_model.model.shared`, `language_model.lm_head`, both `embed_positions` tables, and both `image_pos_embed` tables are packed in every conversion. Implementing only the linear projections would have aborted the process at the first embedding lookup.
- **`image_projection` is not quantized, and neither is the temporal buffer.** The issue's "fusion and projection layers (`fusion.rs`) where the checkpoint quantizes them" reads as if `image_projection` might be packed. It is not, in any published conversion, because upstream registers it as a raw `nn.Parameter` applied as a right-hand matmul rather than an `nn.Linear`, and `nn.quantize` only walks modules. Same for `visual_temporal_embed.pos_idx_to_embed` and the conv stack. Those three are exactly what the narrowed refusal still rejects.
- **The activation dtype had to move.** Not mentioned in the issue. `Florence2TextModel::from_weights` took `dtype` from `shared.weight`, which is the packed `uint32` plane on a quantized checkpoint; the decoder builds its causal mask and the fusion path casts its image features to that dtype. It now comes off `scales` on the quantized arm.
- **`convert_bf16_weights` was wrong in three places, not one.** The issue points at `model.rs`. `Florence2TextModel::load` and `Florence2DaViT::load` called it unconditionally too. All three are now gated on the checkpoint declaring no quantization, matching `finish_vlm_weights_common`.
- **`config.json` differs between the two `base-ft` checkpoints by one key.** The 4-bit config carries `vision_config.hidden_size: 768`, which the bf16 config does not. Neither mlxcel nor upstream reads it (the tower width comes from `dim_embed`), so it is inert, but it is worth recording given that an earlier issue in this chain flagged a nonexistent `hidden_size`.
- **The issue asks for a `-large-ft` smoke test; this also covers a second bit width.** `Florence-2-large-ft-4bit` (d_model 1024, 12 BART layers, `dim_embed` up to 2048) and `Florence-2-base-ft-8bit` are both exercised by the bit-width smoke test, which loads each present variant and runs one decoder step so a wrong bit width surfaces as a shape error rather than as a silent load. See the `large-ft` finding below for what that smoke test does and does not claim.

## `large-ft` generates degenerate output, on this branch and on `main`, in mlxcel and in mlx-vlm

Worth stating plainly because the issue asks for a `large-ft` smoke test and this is what that test found. `Florence-2-large-ft-4bit` loads cleanly and its packing metadata is internally consistent (294 quantized tensors, every one a 4-bit / group-64 fit against its own `scales` width, and the only dense tensors are the four `convs.*.proj`, the 48 depthwise `dw`, `image_projection`, and the temporal buffer). It then generates a run of BOS for every task.

That is not this change. Four measurements, all on the same COCO image or the same synthetic tensor:

- `Florence-2-large-ft-bf16` on the **pre-change `main` binary** produces the same BOS run for `<CAPTION>`, while `<OD>` and `<OCR>` work. So `large-ft` was already partially broken before this PR.
- Upstream mlx-vlm on `Florence-2-large-ft-4bit` produces `greedy_ids = [0, 0, 0, ...]` on the synthetic input, the same degeneracy mlxcel produces.
- Upstream mlx-vlm on `Florence-2-large-ft-bf16` produces `[0, 0, 0, ...]` on that input too.
- `Florence-2-base-ft` is correct at bf16, 4-bit, and 8-bit in both runtimes, on the same code paths.

So the `large-ft` release behaves this way in the reference implementation as well, and the quantized `large-ft` path here is not doing anything the dense one is not. Chasing it belongs in its own issue against the `large-ft` family rather than in a quantization change; `docs/supported-models.md` is left claiming `large-ft` because that claim predates this PR and correcting it needs the investigation that issue would do. The smoke test keeps `large-ft-4bit` in its list because loading it still proves the packing path is not `base-ft`-shaped, and it asserts only load plus finite logits.

## Observed upstream

`Florence2Encoder.__call__` in mlx-vlm ([language.py](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/florence2/language.py)) falls back to `self.embed_tokens(input_ids)` when `inputs_embeds` is None, but `Florence2Encoder.__init__` never defines `embed_tokens` (the shared table lives on the parent `Florence2Model`). The `input_ids` branch therefore raises `AttributeError` for any caller that uses it. mlxcel's encoder only accepts embeddings, so it is unaffected; noting it because the reference script for these pins has to pass `inputs_embeds=` by keyword.

## Review follow-ups

Two commits after the initial one, both hardening rather than feature work. The numeric and CLI results above were re-measured on the final tree and are unchanged.

`a470c03f` (security review), five findings under one threat model: `config.json` and the weight map are untrusted, an MLX throw crosses the cxx bridge as an uncatchable abort, and MLX does not range-check a positive gather index.

- **Bound the sequence length in the `LanguageModel::forward` trait impl.** This is the one finding the change itself introduced. Moving the position lookup from a slice on the raw table to a gather turned an over-long input from an out-of-range slice into an unchecked read past the position table, and past the packed weight, scales, and biases planes when quantized. Every other caller bounds its own sequence; this impl took whatever it was handed. The trait signature returns an array and cannot report an error, so the input is truncated with a warning, which is acceptable because the impl exists for warmup and tooling rather than as a generation path.
- **LayerNorm weights joined the refusal denylist.** Every Florence-2 norm is read as a raw weight and handed to `fast::layer_norm`, so a packed one aborts exactly as a packed conv would. Matching on the last dot-separated segment containing `norm` rather than on a suffix is load-bearing, because `layernorm_embedding` ends in `embedding`.
- **`lm_head` row count validated against `vocab_size`.** The shared table gained a row guard but the head did not, and `argmax_last_position` feeds the head's output straight back into a gather on that table, so a head wider than the table made an out-of-bounds read reachable.
- **A present-but-unreadable `quantization` field is now an error.** It previously fell back to the dense default, which is the one wrong value the unified layers cannot detect later.
- **Stale descriptions fixed**, including `ModelType::metadata()`, which still advertised "bf16/f16 only" and which `mlxcel arch` prints verbatim.

`0bb4051a`, two low-severity gaps where one entry point was weaker than another for no reason: `from_model_config` read the quantization block from one of the four locations `config_is_quantized` accepts, so a checkpoint nesting it elsewhere would keep its bf16 scales correctly and then dequantize at the dense fallback stride silently; and `Florence2TextModel::load` skipped both `sanitize` and the refusal that the other two entry points run, which matters now that the refusal covers normalization weights.

## Test plan

- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_quantized_parity`
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --test florence2_fusion_parity` (dense path unchanged)
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib models::florence2`
- [x] `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings`
- [x] CLI against a real image on the 4-bit checkpoint: `<CAPTION>`, `<OCR>`, `<OCR_WITH_REGION>`, `<OD>`

## CLI validation

COCO `000000039769` (the two-cats image the epic used) for caption and detection, and a deterministic 640x200 PNG rendering `HELLO MLXCEL` for OCR, since the cats image has no text and `<OCR>` correctly returns `-` on it.

| task | `base-ft-bf16` | `base-ft-4bit` | `base-ft-8bit` |
|---|---|---|---|
| `<CAPTION>` | Two cats are sleeping on a pink blanket. | Two cats laying on a pink blanket next to remotes. | Two cats are sleeping on a pink blanket. |
| `<OD>` | cat, cat, couch, remote, remote | cat, cat, couch, remote, remote | cat, cat, couch, remote, remote |
| `<OCR>` | HELLO MLXCEL | HELLO MLXCEL | HELLO MLXCEL |
| `<OCR_WITH_REGION>` | HELLO MLXCEL: [43.2, 71.1, 599.4, 71.1, 599.4, 127.7, 43.2, 127.7] | HELLO MLXCEL: [41.9, 70.9, 599.4, 70.9, 599.4, 127.7, 41.9, 127.7] | (not run) |

8-bit reproduces the dense caption word for word and the dense boxes to within 0.7 px. 4-bit keeps every detection label and every box to within about 1 px, reads the text exactly, and rephrases the caption, which is what 4-bit weights are expected to do to a free-form generation and not what a wiring defect does to one.

Closes #1072


inureyes:

Reviewed against the load-time threat model this tree uses: `config.json` and the safetensors weight map are untrusted, an MLX throw crosses the cxx bridge as an uncatchable `std::terminate` rather than a Rust `Err`, and MLX does not range-check a positive gather index. Five findings, all fixed in a470c03f.

### Verified sound as submitted

**`group_size` / `bits` bounds.** `Florence2Quantization::from_model_config` bounds `group_size` to `1..=4096` and `bits` to `1..=32`, both tighter than or equal to `mlxcel_core::layers::validate_quantization_params`, which every `UnifiedLinear` / `UnifiedEmbedding` re-runs through `reconcile_quantization_layout`. No path reaches a division by zero (`bits == 0` is refused twice), and the width reconstruction multiplies in `usize::checked_mul` and `i64` on the two sides of `validate_quantized_packing`. Nothing here sizes a host allocation.

**The two `image_pos_embed` tables.** `LearnedPositionEmbedding2D::from_weights` reads the row count back out of the loaded table into `num_pos` and `forward` refuses `height > num_pos || width > num_pos`, so the `arange` gather is bounded by the table itself rather than by a config field. The row and column tables are additionally required to agree.

**`LearnedPositionEmbedding2D` deriving `row_dim` / `column_dim` from the config.** A config/checkpoint mismatch is caught at load, on both arms. Dense goes through `dim_eq(cols, row_dim)`; quantized goes through `validate_quantized_packing`, which reconstructs `scales.shape(-1) * effective_group_size` and compares it to `row_dim`, so a table honestly packed for a different width is rejected rather than reaching a reshape or broadcast. `image_projection`'s own `[image_dim, d_model]` check independently pins `image_dim`.

**`convert_bf16_weights` gating.** Skipping the conversion wholesale for a quantized export matches `loading/vlm.rs::finish_vlm_weights_common`, which is the established convention. Checked the real checkpoints: `lm_head.scales` is F16 in all three quantized conversions, so the text stack activation dtype is F16 and the fix that takes the dtype from `scales` instead of the packed `uint32` plane lands on the right type.

### Fixed

1. **Unbounded sequence reaches the position-table gather (MEDIUM, memory safety).** Moving the position lookup from `mlxcel_core::slice` to a gather changed the failure mode for an over-long sequence from an out-of-range slice into an unchecked read past the table. `Florence2Model::encode_fused` and both `generate_greedy` loops bound it; the `LanguageModel::forward` trait impl in `runtime.rs` did not, and passed arbitrary length into both `encode_tokens` and `decode`. The same length also sized the O(n^2) host buffer in `additive_causal_mask`. Now truncated to `max_position_embeddings` with a warning, and the precondition doc comments on `encode_tokens` / `encode_embeds` / `encode_embeds_with_mask` / `decode` restated as memory-safety preconditions, with `decode` naming the offset term it actually carries.

2. **`reject_unsupported_quantized_tensors` missed every LayerNorm (MEDIUM).** The denylist covered `image_projection`, the temporal buffer, `convs.*proj`, and `*.dw`, but not the normalization weights, which are read as raw `{prefix}.weight` pairs and handed to `fast::layer_norm`. A packed one presents a `uint32` weight to that kernel and aborts the process, which is exactly what the guard exists to prevent. New arm refuses a `.scales` stem whose last dot-separated segment contains `norm`. Segment matching rather than suffix matching is load bearing, because `layernorm_embedding` ends in `embedding`.

3. **`lm_head` row count unvalidated against `vocab_size` (MEDIUM).** The new `shared` guard proves the token table has *at least* `vocab_size` rows, but nothing bounded the head, and `argmax_last_position` over the head output feeds straight back into `embed_tokens`, which is a gather into that table. A head wider than the table made the out-of-bounds read reachable. Now required to agree exactly.

4. **A present-but-malformed `quantization.group_size` silently became 64 (LOW).** `config_i32` maps a float, a string, or an out-of-i32 integer to `None`, and the caller turned that into the dense default. That is the one failure mode this type cannot detect later, since the unified layers take the declared group size on trust; the struct's own doc comment says as much. Absent still defaults, present-but-unreadable is now a named error.

5. **Stale descriptions (LOW).** The rustdoc on `Florence2VlmModel::load` still said quantized checkpoints are rejected, and `ModelType::metadata()` still advertised "bf16/f16 only", which `mlxcel list` prints verbatim.

### No-false-positive evidence for finding 2

Simulated the new refusal rule against the safetensors headers of every locally available conversion. It refuses zero real tensors: 198 packed tensors in `base-ft-4bit`, 198 in `base-ft-8bit`, 294 in `large-ft-4bit`, 0 in both bf16 exports. The packed tail segments across all three are exactly `column_embeddings, embed_positions, fc1, fc2, k_proj, lm_head, out_proj, proj, q_proj, qkv, row_embeddings, shared, v_proj`, none containing `norm`. Also confirmed `lm_head.weight` is `[51289, *]` against `vocab_size: 51289` in all five, so finding 3's exact-equality guard cannot false-reject a published conversion.

### Performance

The slice to gather change is not a meaningful per-step cost.

- Encoder, once per request: an `arange` plus a gather of about 587 rows at `d_model` 768, roughly a 0.9 MB copy where the slice was a lazy view. Against six encoder layers of `[587, 768] x [768, 768]` projections and a `[12, 587, 587]` attention, this is under 0.1 percent.
- Decoder, per step: `seq` is 1, so the gather returns a single 768-wide row. Two extra graph nodes against the roughly 150 to 200 the step already builds across 12 layers. On the quantized arm `quantized_embedding` dequantizes that one row, which is the same order.

Nothing else in the change adds per-token work. `UnifiedLinear::Regular` is an enum dispatch over the same `Linear::forward` a dense checkpoint used before, the causal mask and the argmax path are untouched, and the truncation added in finding 1 is a comparison on a path the CLI and server never take. The one measurable change is that a quantized export now keeps its stored dtypes throughout rather than being converted, which is the same behavior every other quantized VLM in this tree has.

### Validation

- `cargo fmt --all -- --check`, clean.
- `cargo clippy --profile test-fast --features metal,accelerate --lib -- -D warnings`, clean.
- `cargo test --profile test-fast --features metal,accelerate --lib models::florence2`, 132 passed.
- `cargo test --profile test-fast --features metal,accelerate --test florence2_quantized_parity`, 4 passed, on the real checkpoints. This is the one that matters for the hardening: `florence2_quantized_checkpoint_loads`, `florence2_other_quantized_bit_widths_load`, `florence2_quantized_matches_mlx_vlm_reference`, and `florence2_quantization_cost_matches_mlx_vlm` all still pass, so the three new load-time guards do not reject a published conversion and the numerics are unchanged.

No unresolved CRITICAL or HIGH findings.

inureyes:

### Intent
> Load the eight `mlx-community` quantized Florence-2 conversions (3/4/6/8-bit, `base-ft` and `large-ft`) by routing every packed prefix through `UnifiedLinear` / `UnifiedEmbedding`, and narrow #854's blanket refusal to the tensors this family still consumes dense.

### What I verified directly

**Quantized wiring is complete.** I parsed the safetensors headers of `models/Florence-2-base-ft-4bit` (198 packed tensors) and `models/Florence-2-large-ft-4bit` (294) and enumerated every distinct packed stem. All thirty map to a `UnifiedLinear` or `UnifiedEmbedding` construction: the BART q/k/v/out projections and fc1/fc2 on both stacks and on `encoder_attn`, `model.shared`, both `embed_positions`, `lm_head`, both `image_pos_embed` tables, and the DaViT `window_attn.fn` / `channel_attn.fn` qkv+proj and `ffn.fn.net.fc1/fc2`. Grep confirms zero remaining `Linear::from_weights` / `Embedding::from_weights` call sites anywhere reachable from `Florence2Model::load`, `Florence2TextModel::load`, or `Florence2DaViT::load`. The remaining dense tensors in both checkpoints are exactly the layer norms, `convs.N.proj`, the depthwise `dw`, `image_projection`, `image_proj_norm`, and `visual_temporal_embed.pos_idx_to_embed`, all of which are consumed dense.

**The narrowed refusal had one false negative.** `image_projection`, `visual_temporal_embed.*`, `convs`/`proj`, and `.dw` were covered; LayerNorm was not, even though `docs/supported-models.md` in this PR already listed the layer norms among the tensors the guard covers. Every Florence-2 norm is read as a raw weight/bias pair and handed to `fast::layer_norm`, so a packed one is the same uncatchable abort the conv arm exists to prevent. Fixed below. No false positives in either direction: `contains(\"convs\") && ends_with(\"proj\")` does not match `blocks.*.window_attn.fn.proj`, and no packed stem's last segment contains `norm`.

**`convert_bf16_weights` gating is correct in all three entry points**, and nothing else in the Florence-2 path promotes scales or biases. The activation dtype correctly comes off `scales` on the quantized arm.

**CLI integration is real.** `detection.rs:336` keys on `model_type == \"florence2\"` independent of quantization, `loading/mod.rs:222` routes to `load_florence2_vlm`, and `generate.rs:2134` routes `LoadedModel::Florence2VLM` to the seq2seq pipeline. The old config-level refusal is gone from `Florence2Model::load`.

### Findings Addressed
- [x] `ModelType::metadata()` still advertised Florence-2 as `bf16/f16 only`, and `mlxcel arch` / `mlxcel list` print that string verbatim (HIGH)
- [x] `reject_unsupported_quantized_tensors` had no LayerNorm arm, contradicting the doc text this PR added and leaving an uncatchable-abort path open (MEDIUM)
- [x] `Florence2VlmModel::load` rustdoc, the entry point the CLI loader calls, still said quantized checkpoints are rejected (MEDIUM)
- [x] `convert_bf16_weights` shared-function `Used by:` list did not name Florence-2's three direct call sites (LOW)

Also co-carried in `7b584e00` from a concurrent hardening pass on the same load path: an `lm_head.weight` row check against `vocab_size` (verified non-breaking, all five local checkpoints are 51289), strict parsing of a present-but-unreadable `quantization` field, and an input bound on the `LanguageModel::forward` trait impl.

### Remaining Items
- [ ] `Florence2Quantization::from_model_config` reads only the top-level `quantization` object, while `config_is_quantized` also answers true for `quantization_config` and `text_config.quantization` (LOW) — a checkpoint declaring packing in either of those two would skip the bf16 conversion but parse as `DENSE`, and `reconcile_quantization_layout` trusts `group_size`. No published Florence-2 conversion does this, so it is a latent divergence rather than a defect.
- [ ] `Florence2TextModel::load` runs neither `sanitize()` nor `reject_unsupported_quantized_tensors()`, unlike the other two entry points (LOW) — harmless today because every published conversion materializes `model.shared` and no dense-only tensor lives under `language_model.`, but the asymmetry is a trap for a future export.
- [ ] Adding a public `quantization` field to `Florence2TextConfig` / `Florence2VisionConfig` is source-breaking for external struct-literal construction (LOW, informational).

### Verification
- [x] Tests pass

`cargo check --profile test-fast --features metal,accelerate --lib --tests` clean, `cargo clippy ... --lib --tests -- -D warnings` clean, `cargo fmt --all -- --check` clean, `--lib florence2` 155 passed, `--test florence2_quantized_parity` 4 passed and `--test florence2_fusion_parity` 2 passed against the real checkpoints after the fixes.

PR #1096, opened 2026-08-10
fix: make autotune profiler tests deterministic

### LINKED ISSUE #1079: fix(test): two autotune tests decide on microsecond thread::sleep accuracy and flake under host load
## Problem / Background

`autotune::autotune_tests::profile_drops_failing_candidates_without_aborting` and `autotune::autotune_tests::resolve_profiles_and_persists_under_tune_mode` failed together in one serialized run of the `mlxcel-core` lib suite, and an identical re-run of the same commit passed 1416 / 0.

```
test autotune::autotune_tests::profile_drops_failing_candidates_without_aborting ... FAILED
test autotune::autotune_tests::resolve_profiles_and_persists_under_tune_mode ... FAILED
panicked at src/lib/mlxcel-core/src/autotune/autotune_tests.rs:301
panicked at src/lib/mlxcel-core/src/autotune/autotune_tests.rs:765
test result: FAILED. 1414 passed; 2 failed
```

Observed on an M1 Ultra (macOS 26.5.2, `--features metal,accelerate`, `test-fast` profile) while validating an unrelated C++ change, at a host load average of roughly 40.

## Why this is a test defect and not a product one

The module says so itself, at `autotune_tests.rs:17`:

> Every test here is CPU-only: the `FakeOp` double implements `TunableOp` by sleeping for a per-tactic duration and overrides `sync` to a no-op, so the harness, the cache, and the precedence chain are exercised without a GPU.

So no kernel, no MLX call, and no GPU is involved. `FakeOp::run` is `std::thread::sleep(Duration::from_micros(micros))` (`autotune_tests.rs:142`), and the profiler then measures the elapsed time and picks the fastest candidate.

The two failing tests separate their candidates by **microseconds**:

- `profile_drops_failing_candidates_without_aborting` uses `vec![(2, 800), (4, 80), (8, 400)]` and asserts 8 wins over 2, a 400 microsecond gap.
- `resolve_profiles_and_persists_under_tune_mode` uses `vec![(2, 800), (8, 80)]` and asserts 8 wins, a 720 microsecond gap.

`thread::sleep` only ever overshoots, and on a loaded host the overshoot is routinely hundreds of microseconds to milliseconds, which swamps both gaps. That the 720 microsecond gap also flipped is consistent with an overshoot well above 700 microseconds.

The file already anticipates the hazard in `fast_cfg`'s doc comment at `autotune_tests.rs:150-153`:

> These tests assert selection logic, not timing methodology, and a real budget would make their call counts depend on how accurately the host honors `thread::sleep`.

`fast_cfg` removes the wall-clock *budgets* for that reason, but the candidate *ordering* these two tests assert still rests on sleep accuracy. The neighbouring `profile_returns_the_fastest_candidate` at `autotune_tests.rs:285` shows the author was aware and widened its own margins deliberately ("exaggerated so that `thread::sleep` jitter cannot decide the assertion"); these two were not given the same treatment.

## Evidence it is load and not the change under test

- Both pass 55 / 0 when the `autotune::` filter is run alone.
- An identical re-run of the full suite on the same commit passed 1416 / 0.
- The commit under validation touched only `template_args` lists in C++ kernel launchers, which a CPU-only sleep-based double cannot reach.

## Proposed Solution

Make the assertions independent of how accurately the host honors `thread::sleep`. Options, roughly in order of preference:

- **Inject the clock.** Give `FakeOp` a deterministic time source the profiler reads instead of sleeping, so the test asserts selection logic on exact durations. This matches what the tests claim to be testing and removes the hazard rather than widening it.
- **Widen the gaps** the way `profile_returns_the_fastest_candidate` already does, so jitter cannot decide the ordering. Cheapest, but it only moves the threshold rather than removing the dependency, and it makes the suite slower.
- **Scale the durations up** (milliseconds rather than microseconds) so relative overshoot shrinks. Same objection, plus a larger wall-clock cost.

Whichever is chosen, the fix should cover every test in the module whose assertion depends on measured ordering, not only the two that flaked here.

## Acceptance Criteria

- [x] The two named tests no longer depend on `thread::sleep` accuracy for their assertions, or the dependency is documented as bounded with a stated margin.
- [x] The module is swept for other assertions resting on measured ordering, and they get the same treatment.
- [x] The tests still fail if the selection logic they exercise is broken. A fix that widens tolerances until nothing can fail is not acceptable; verify by deliberately inverting a candidate's cost and confirming a red.
- [ ] `cargo test --profile test-fast --features metal,accelerate -p mlxcel-core --lib autotune:: -- --test-threads=1` passes, and the full lib suite passes on a loaded host. Not available on this Linux host; the PR reports the unavailable macOS Metal and loaded-host full-lib gates explicitly.

## Technical Considerations

CI is exposed to the same hazard. A GitHub-hosted runner has no load guarantee, and `nightly-verify.yml` runs the full `make verify` on the shared self-hosted Apple Silicon runner, which is exactly where a co-scheduled job would produce this failure and file a spurious red-main issue.

Related: #997 covers a different set of concurrent-load flakes that also pass in isolation. Same class, different mechanism.

inureyes:
## Summary

Make autotune profiler ordering tests independent of host scheduler sleep accuracy by adding an internal deterministic timer seam for tests while leaving production profiling on real `Instant` timing.

## What changed

- Added an internal `ProfileTimer` path used by `profile()` with `RealTimer` in production and by test-only helpers with a manual microsecond clock.
- Routed ordering-sensitive fake-op profile and resolve tests through the manual timer, including failing-candidate, metadata/spread, persistence, retune, and memoization coverage.
- Kept the adaptive wall-clock budget test on real `thread::sleep` because it verifies repetition scaling, not candidate ordering, and added an inverted-cost selection proof.
- Bounded caller-supplied warmup repetitions by the normalized `max_reps` ceiling and clarified the public profiler and `mlxcel tune` descriptions of that cap.

## Test plan

- [x] `cargo fmt --check -p mlxcel-core`
- [x] `cargo fmt --check -p mlxcel`
- [x] `cargo test --profile test-fast -p mlxcel-core --lib autotune:: -- --test-threads=1` (57 passed / 1355 filtered out)
- [x] `cargo test --profile test-fast -p mlxcel-core --lib autotune::autotune_tests::profile_selection_changes_when_synthetic_costs_are_inverted -- --test-threads=1`
- [x] `cargo test --profile test-fast -p mlxcel-core --lib autotune::autotune_tests::profile_config_bounds_warmup_by_the_rep_ceiling -- --test-threads=1`
- [x] Temporary mutation proof: forcing the inverted-cost test to expect `p=8` for the faster-default case failed with `left: p=2`, then the assertion was restored and the test passed.
- [ ] `cargo test --profile test-fast --features metal,accelerate -p mlxcel-core --lib autotune:: -- --test-threads=1` and loaded-host full lib suite: unavailable on this Linux host; not claimed as passing. A current-host broad run reached 1,411 tests but aborted in an unrelated CUDA-backed cache test because no CUDA backend is exposed.

Closes #1079

### DIFF
--- a/TECHNICAL_REPORTS/1096-deterministic-autotune-profiling-20260810.en.md
+++ b/TECHNICAL_REPORTS/1096-deterministic-autotune-profiling-20260810.en.md
@@ -1,0 +1,71 @@
+# Technical Report: PR #1096 - Deterministic Autotune Profiling Tests
+
+**Date**: 2026-08-10
+**Status**: Completed with platform qualification pending
+**Languages**: Rust
+**Risk Level**: Medium
+
+## Executive Summary
+
+PR #1096 removes host-scheduler timing jitter from autotune tests that assert candidate ordering. Production profiling continues to measure real `TunableOp::run` and `sync` latency with `Instant`, while tests exercise the same warmup, interleaving, selection, persistence, and memoization logic through an internal manual microsecond clock.
+
+Review also closed a production resource-bound gap: caller-supplied warmup repetitions are now capped by normalized `max_reps`, with CLI and API documentation updated to match.
+
+## 1. Problem Statement
+
+Two CPU-only autotune tests intermittently selected the wrong synthetic tactic under host load because `FakeOp` modeled costs with microsecond `thread::sleep` calls. Scheduler overshoot could exceed the 400-720 microsecond gaps being compared, so tests intended to verify deterministic selection instead measured operating-system wakeup behavior.
+
+The same test double powered other profile and resolve assertions, so fixing only the two observed failures would have left equivalent latent flakes elsewhere in the module.
+
+## 2. Technical Decisions
+
+### 2.1 Inject time at the profiler boundary
+
+An internal `ProfileTimer` abstraction supplies marks and elapsed microseconds to the existing profiling algorithm. `profile()` instantiates `RealTimer`, preserving the production path, while test-only helpers use `ManualTimer` and advance it by the selected tactic's declared cost. The seam remains internal and the public `TunableOp` contract is unchanged.
+
+### 2.2 Retain real sleep only for a one-candidate budget test
+
+`timed_repetitions_scale_inversely_with_launch_cost` still sleeps because its subject is adaptive wall-clock repetition budgeting. Each profile in that test has one candidate, so scheduler jitter can change the realized repetition count within documented bounds but cannot decide candidate ordering.
+
+### 2.3 Prove selection sensitivity and cap caller work
+
+The regression suite evaluates both cost directions: a faster tuned candidate must replace the default, while an inverted cost map must retain the faster default. A temporary wrong expectation was observed red before restoration. Separately, sanitized `warmup` is capped after `max_reps` is normalized above the timed repetition floor, preventing pathological CLI/API input from forcing an oversized warmup loop.
+
+## 3. Change Summary
+
+| Item | Value |
+|------|-------|
+| Files changed | 4 |
+| Lines added | 210 |
+| Lines deleted | 56 |
+| Focused tests | 57 passed |
+
+- Added internal real/manual timer implementations and routed the profiler's warmup and timed phases through them.
+- Migrated all ordering-sensitive fake-op profile and resolver tests to deterministic synthetic time.
+- Added inverted-cost and warmup-ceiling regression coverage.
+- Added a profiling closure seam for test-only resolver execution without changing public APIs.
+- Clarified `ProfileConfig` and `mlxcel tune` help/output to state that warmup is capped by `max_reps`.
+
+## 4. Review Findings
+
+| Finding | Severity | Resolution |
+|---------|----------|------------|
+| Caller-supplied `warmup` could exceed the documented `max_reps` ceiling | Medium | Clamp warmup after normalizing `max_reps`; add a seven-call regression proof |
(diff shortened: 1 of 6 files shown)


inureyes:
Security/performance review summary for PR #1096:

- MEDIUM fixed: caller-supplied ProfileConfig could set warmup above max_reps, so the profiler documentation claimed a repetition ceiling the warmup loop did not enforce. Commit 26b20c15 clamps sanitized warmup to the same max_reps ceiling and adds a deterministic timer regression.
- No CRITICAL/HIGH findings in the timer seam: production profile still uses RealTimer backed by Instant; deterministic timer helpers stay cfg(test) or pub(super) inside autotune; no public API is added; resolve remains inert unless mode permits cache/profile work.
- No additional denial-of-service or unbounded-loop issue found after the fix: candidate count comes from bounded candidate lists, timed reps are clamped to [reps, max_reps], and warmup is now capped too.

Validation:
- cargo fmt --check -p mlxcel-core
- cargo test --profile test-fast -p mlxcel-core --lib autotune:: -- --test-threads=1 (57 passed / 1355 filtered out)
- cargo test --profile test-fast -p mlxcel-core --lib autotune::autotune_tests::profile_config_bounds_warmup_by_the_rep_ceiling -- --test-threads=1
- GitHub CI on 26b20c15: Detect changes, crate versions, kernel dtype keys, cross-repo refs, cargo-fmt, cargo-deny, license/cla all passed; MLX pin extraction skipped.

No merge performed.

PR #1135, opened 2026-08-13
feat(vlm): add vision feature caching for LFM2-VL

MrFelt0385:
# feat(vlm): add vision feature caching for LFM2-VL

## Summary

LFM2-VL currently **skips** the opportunistic vision feature cache (the code
says `// LFM2-VL runs every image in one preparation pass; skip the
opportunistic vision cache for this first integration.`). This PR implements
it, giving a **2.12× speedup** on repeated images — the main win for
multi-turn conversations and continuous screen analysis.

## Changes

- **`src/vision/lfm2_vl.rs`** — add `get_input_embeddings_with_cache`,
  mirroring the existing Qwen2.5-VL pattern:
  - Cache key: SHA-256 of the request image bytes + soft-token budget
    (`image_hash_from_bytes_with_soft_tokens` on the server side)
  - **HIT**: reuse the concatenated post-connector features, skip the vision
    tower + connector entirely
  - **MISS**: compute features, `eval` + deep-`copy` (decouple from the
    deferred graph), store snapshot in the cache
- **`src/multimodal/vlm_runtime.rs`** — wire the `Lfm2Vl` runtime to the
  single-feature cache (`ModelVisionCaches::single`), deriving the key from
  the concatenated pixel tensor (same approach as the Qwen path)

## Benchmarks

Measured on a Mac mini M1, `LFM2.5-VL-3B-MLX-4bit`, same image across
requests (`zone_hud.png`, 120×70):

| Request | Time | Notes |
|---|---|---|
| 1st (MISS) | 1158 ms | vision tower encodes the image |
| 2nd (HIT, same image) | **547 ms** | **2.12× faster** |
| 3rd (HIT, same image, different text) | **550 ms** | **2.10× faster** |

**Finesse preserved**: HIT responses are byte-identical to MISS responses
(vision features are deterministic; only the tower output is cached).

## Notes

- The connector output shape depends on the image's native patch count,
  which is folded into the key by the soft-token budget — two resolutions of
  the same image cannot collide.
- Cache lifetime is tied to the model registry entry (`clear_all` on unload),
  same as the existing Qwen/Gemma/Granite caches.

### DIFF
--- a/src/multimodal/vlm_runtime.rs
+++ b/src/multimodal/vlm_runtime.rs
@@ -1256,13 +1256,25 @@ where
                 total_image_tokens: stats.total_image_tokens,
             });
 
-            // LFM2-VL runs every image in one preparation pass; skip the
-            // opportunistic vision cache for this first integration.
-            let _ = active_caches;
-            let _ = image_cache_keys;
+            // LFM2-VL runs every image in one preparation pass. The cache key
+            // covers the concatenated pixel bytes for all images in the
+            // request; multi-turn conversations that revisit the same image
+            // set skip the vision tower + connector on subsequent turns.
+            let lfm2_cache_key = if active_caches.is_some() {
+                first_explicit_key(image_cache_keys)
+                    .or_else(|| Some(CacheKey::from_hash(image_hash_from_pixels(&pixel_values))))
+            } else {
+                None
+            };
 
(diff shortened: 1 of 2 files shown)


inureyes:
@MrFelt0385 Thank you for your contribution! To test and merge, please sign the CLA.

PR #1157, opened 2026-08-14
test: skip pinned muse_glimmer checkpoint tests and fix MLXCEL_BACKEND race

### LINKED ISSUE #1155: test: muse_glimmer tests panic without the pinned checkpoint and race on MLXCEL_BACKEND
## Problem / Background

Three muse_glimmer tests fail in a full `cargo test --release --features metal,accelerate --lib` run on a machine that does not have the pinned checkpoint. All three were introduced with #1116. Verified on 2026-08-15 on main `a206f089`: the failures reproduce identically before and after the epic #1148 merge window, and none of the epic commits touch muse files, so these are pre-existing defects.

**(a) Checkpoint-dependent tests panic instead of skipping.** `loading::vlm::muse_glimmer::tests::pinned_weight_index_classifies_each_source_weight_once` and `vision::encoders::muse_glimmer_fusion::tests::pinned_post_tower_weight_roots_and_shapes_match_published_contract` panic when `models/mlx/muse-glimmer-30b/model.safetensors.index.json` is absent ("Failed to read ... No such file or directory" and "Muse Glimmer pinned checkpoint index is required for this shape contract"). Environment-dependent tests elsewhere in the tree either use `#[ignore]` with a reason or skip gracefully with an `eprintln` (see `tests/prompt_cache_e2e.rs`).

**(b) Env-var race under parallel execution.** `server::startup::muse_glimmer_startup_guard_tests::muse_glimmer_startup_allows_baseline_and_keeps_video_disabled` fails in the full suite but passes in isolation. A sibling test in the same module sets `MLXCEL_BACKEND=xla` process-wide (`src/server/muse_glimmer_startup_guard_tests.rs` around lines 144-157, set_var/restore pattern), and the baseline test can observe the transient value.

## Proposed Solution

- For (a): make the two pinned-checkpoint tests skip gracefully when the index file is absent (eprintln + early return, matching `tests/prompt_cache_e2e.rs`), or mark them `#[ignore]` with a reason explaining the checkpoint requirement.
- For (b): one of:
  - a process-level env-var mutex shared by all tests that mutate `MLXCEL_BACKEND` (check whether the repo already has this pattern),
  - `serial_test` on the affected module,
  - or refactor the startup guard to take the backend as a parameter instead of reading the env var directly, so tests never need to mutate process state.

## Acceptance Criteria

- [ ] `cargo test --release --features metal,accelerate --lib` is green on a machine without `models/mlx/muse-glimmer-30b/`
- [ ] The two pinned-checkpoint tests skip (or are `#[ignore]`d) with a clear reason when the index file is absent, and still run their full contract check when it is present
- [ ] `muse_glimmer_startup_allows_baseline_and_keeps_video_disabled` passes reliably under default parallel test execution (no dependence on test ordering)
- [ ] No other test can observe a transient `MLXCEL_BACKEND` value set by the muse_glimmer startup guard tests

## Technical Considerations

Discovered during the epic #1148 integration verification; the env-race test passes standalone, which confirms the failure is ordering-dependent rather than logic-dependent. Refs #1116.

inureyes:
## Summary

Fixes three pre-existing muse_glimmer test defects (introduced with #1116, unrelated to epic #1148) that prevented `cargo test --release --features metal,accelerate --lib` from being green on a machine without the pinned `models/mlx/muse-glimmer-30b` checkpoint.

## What changed

- `src/loading/vlm_muse_glimmer_tests.rs`: `pinned_weight_index_classifies_each_source_weight_once` now checks whether `model.safetensors.index.json` exists before calling `read_muse_weight_inventory_from_index`, and skips with an `eprintln` reason instead of panicking when the pinned checkpoint is absent, matching the pattern already used in `tests/prompt_cache_e2e.rs`.
- `src/vision/encoders/muse_glimmer_fusion_tests.rs`: `pinned_post_tower_weight_roots_and_shapes_match_published_contract` skips the same way instead of asserting on the missing index file.
- `src/server/muse_glimmer_startup_guard_tests.rs`: `muse_glimmer_startup_allows_baseline_and_keeps_video_disabled` now acquires the crate-wide `crate::test_support::env_lock::env_lock()` guard before calling `validate_muse_glimmer_unsupported_startup`, which reads `MLXCEL_BACKEND`. The sibling test `muse_glimmer_startup_rejects_xla_backend_selection` in the same module already sets `MLXCEL_BACKEND=xla` under that lock; without the baseline test also taking it, the baseline test could intermittently observe the transient value under default parallel test execution and fail. A grep of `MLXCEL_BACKEND` across `src/` confirmed no other test module mutates it, so no other module needs to share this lock.

Both pinned-checkpoint tests still run their full contract check when the index file is present; that arm could not be exercised on this machine (no `models/mlx/` directory) and is verified by code inspection only, as noted in the commit body.

## Test plan

- [x] `cargo test --release --features metal,accelerate --lib muse_glimmer` (80 passed, 0 failed, on a machine without `models/mlx/muse-glimmer-30b`)
- [x] `cargo test --release --features metal,accelerate --lib muse_glimmer_startup_guard`, repeated 5 times under default parallelism (5 passed each run, no flakes observed)
- [x] `cargo clippy --release --lib --tests --features metal,accelerate -- -D warnings` (clean)

Closes #1155

### DIFF
--- a/TECHNICAL_REPORTS/1157-muse-glimmer-test-hygiene-20260815.en.md
+++ b/TECHNICAL_REPORTS/1157-muse-glimmer-test-hygiene-20260815.en.md
@@ -1,0 +1,171 @@
+# Technical Report: PR #1157 - test: skip pinned muse_glimmer checkpoint tests and fix MLXCEL_BACKEND race
+
+**Date**: 2026-08-15
+**Author**: AI Code Reviewer
+**Status**: Completed
+**Languages**: Rust
+**Risk Level**: Low
+
+---
+
+## Executive Summary
+
+Three muse_glimmer tests introduced with #1116 broke the full `cargo test --release --features metal,accelerate --lib` run on any machine without the pinned `models/mlx/muse-glimmer-30b` checkpoint: two panicked on the missing index file and one failed intermittently because a sibling test mutates `MLXCEL_BACKEND` process-wide. This PR makes the two checkpoint-pinned tests skip gracefully when the index is absent and serializes the racing test through the crate-wide `test_support::env_lock`, restoring a green suite (5621 passed / 0 failed) on checkpoint-less machines. Test-only change; no production code is touched.
+
+---
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+Issue #1155 was filed during the epic #1148 integration verification. A two-arm measurement on main `a206f089` and the pre-epic baseline showed the three failures reproduce identically on both arms and no epic commit touches muse files, so they are pre-existing defects from #1116, not an epic regression.
+
+### 1.2 Existing Issues
+
+- **Issue 1**: `loading::vlm::muse_glimmer::tests::pinned_weight_index_classifies_each_source_weight_once` panicked with "Failed to read ... No such file or directory" when `models/mlx/muse-glimmer-30b/model.safetensors.index.json` is absent.
+- **Issue 2**: `vision::encoders::muse_glimmer_fusion::tests::pinned_post_tower_weight_roots_and_shapes_match_published_contract` asserted on the same missing index ("pinned checkpoint index is required for this shape contract").
+- **Issue 3**: `server::startup::muse_glimmer_startup_guard_tests::muse_glimmer_startup_allows_baseline_and_keeps_video_disabled` failed under default parallel execution but passed in isolation: the sibling `muse_glimmer_startup_rejects_xla_backend_selection` sets `MLXCEL_BACKEND=xla` process-wide (set_var/restore), and the baseline test's validator reads the env var and can observe the transient value.
+
+### 1.3 Risk Assessment
+
+| Risk | Impact | Likelihood |
+|------|--------|------------|
+| Red full suite on contributor machines masks real regressions | Medium | High (any machine without the 30B checkpoint) |
+| Ordering-dependent flake erodes trust in the suite | Medium | Medium (scheduler-dependent) |
+
+---
+
+## 2. Technical Review
+
+### 2.1 Security
+
+No security surface: all three changed files are `#[cfg(test)]` test modules; nothing ships in production binaries. The security pass confirmed the index-path TOCTOU is inert (hardcoded literal path, no attacker input; worst case is the pre-PR panic) and found no lock-ordering cycle with the existing per-module MLX guards.
+
+**Issues Found:**
+| Issue | Severity | Status |
+|-------|----------|--------|
+| Unlocked `MLXCEL_BACKEND` reader in `backend::tests` under opt-in `xla-backend`/`experimental-backend` features | Medium | Open (pre-existing from #1116, unreachable in default and CI feature sets, left as follow-up to keep the PR scoped) |
+| Fusion test skip guard checks only the index; a partially materialized checkpoint still panics on `config.json`/shard `unwrap()`s | Low | Open (repo-wide skip-guard convention; follow-up candidate) |
+
+### 2.2 Performance
+
+The added mutex acquisition serializes exactly two tests in one module; the effect on suite wall time is unmeasurable (module runs in well under a second).
+
+### 2.3 Compatibility & Dependencies
+
+- **Breaking Changes**: none
+- **New Dependencies**: none (reuses the existing crate-internal `test_support::env_lock`)
+- **Compatibility**: behavior is unchanged on machines that do have the pinned checkpoint; both contract tests still run their full assertions there
+
+### 2.4 Code Quality
+
+- **Test Coverage**: unchanged in count; effective coverage improves because the suite is now runnable end-to-end on checkpoint-less machines
+- **Code Complexity**: two early-return guards and one lock acquisition; trivially reviewable
+- **Technical Debt**: decreased (removes an ordering dependency; aligns with the established skip convention in `tests/prompt_cache_e2e.rs`)
+
+---
+
+## 3. Technical Decisions
+
+### 3.1 Graceful skip (`eprintln` + early return) over `#[ignore]`
+
+**Context:** The two checkpoint-pinned contract tests must still run automatically on the machine that owns the checkpoint, but must not fail elsewhere.
+
+**Alternatives Considered:**
+
+| Option | Pros | Cons |
+|--------|------|------|
+| `#[ignore = "..."]` | Visible as ignored in test output | Never runs automatically, even where the checkpoint exists; needs `--ignored` |
+| **Chosen: guard + eprintln + return** | Runs fully when the checkpoint is present, skips with a printed reason otherwise; matches `tests/prompt_cache_e2e.rs` | Skipped runs report `ok`, so coverage erosion is silent (noted as a follow-up idea: an opt-in `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1`) |
+
+**Rationale:** Issue #1155 explicitly prefers the repo's existing skip convention, and the checkpoint-owning box keeps its contract coverage with zero workflow change.
+
+### 3.2 Crate-wide `env_lock` over per-module mutex, `serial_test`, or a parameter refactor
+
+**Context:** The race is between one mutator and one reader of `MLXCEL_BACKEND` in the same module, but env vars are process-wide state.
+
+**Rationale:** `crate::test_support::env_lock` (declared at `src/lib.rs:52`) is the established pattern with 18 acquisition sites and documented poison recovery (`unwrap_or_else(|p| p.into_inner())`); the mutating sibling already held it, so adding the acquisition on the reader side closes the race with one line. A `serial_test` dependency or a startup-guard signature refactor would be strictly larger changes for the same guarantee. The review pass proved the lock is load-bearing differentially: with the lock removed the module failed 25/25 stress runs; with it, 10/10 passed.
+
+---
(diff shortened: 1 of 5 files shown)


inureyes:

### Intent
> Make `cargo test --release --features metal,accelerate --lib` green on a box without the pinned `models/mlx/muse-glimmer-30b` checkpoint, by fixing three pre-existing test defects from #1116 (two panics on a missing checkpoint index, one `MLXCEL_BACKEND` race).

### Findings Addressed
None. No CRITICAL or HIGH findings; no fixes were required.

### Verification performed by this review

**Skip guards (deliverable a).** Both guards test exactly `models/mlx/muse-glimmer-30b/model.safetensors.index.json`, which is the same file `read_muse_weight_inventory_from_index` reads and the same file the fusion test parses, so the guard condition matches the real precondition with no false-skip window. The checkpoint-present arms are byte-identical to their pre-PR bodies: the loading test still asserts the full `MuseWeightInventory { 626, 806, 2, 1, 1, 1436 }` equality, and the fusion test still runs config validation, the sorted post-tower key list, and the three per-shard safetensors shape assertions. No assertion was weakened or removed. The `eprintln!` skip wording matches the established repo pattern (`tests/prompt_cache_e2e.rs:201`, plus roughly 25 other call sites under `tests/`), including the `Skipping <test_name>: <reason>` form. Confirmed with `--nocapture` that both messages actually print on this machine (`models/mlx` does not exist) and both tests report `ok`.

**Env lock (deliverable b).** The sibling `muse_glimmer_startup_rejects_xla_backend_selection` already took `crate::test_support::env_lock::env_lock()` at line 148 and holds it across both the `set_var` and the restore, so with the baseline test now taking the same guard, both sides of the race hold the same lock. `env_lock()` recovers from poisoning via `unwrap_or_else(|p| p.into_inner())`, so a panic in one env test cannot cascade into spurious failures elsewhere. `crate::test_support::env_lock` is the crate's established helper (declared `#[cfg(test)] pub(crate) mod test_support` at `src/lib.rs:52`, 18 acquisition sites across the crate), not a new parallel invention.

**Differential proof that the lock is load-bearing.** Removing only the added `env_lock()` line and rerunning `cargo test --release --features metal,accelerate --lib muse_glimmer_startup_guard`: **25/25 runs FAILED**. With the line restored: **10/10 runs passed**. The two arms are unambiguously distinguished; this is a real fix, not a coincidental reordering.

**Criterion 4 (no other test can observe the transient value), verified independently rather than taking the PR body's mutator-only argument.** Under `--features metal,accelerate` the only `MLXCEL_BACKEND` reader reachable from a lib test is `xla_backend_requested_from_env()` (`src/server/startup.rs:772`), called from `validate_muse_glimmer_unsupported_startup`. Its other caller, `start_server` (`src/server/startup.rs:1715`), is not unit-tested. The three `assert_rejected` siblings in the module never reach the env read, because the XLA `ensure!` is the **last** check in the function and each of their mutated fields trips an earlier `ensure!` that returns first. `backend::select_backend`'s env-reading form is compiled only under `experimental-backend` / `xla-backend`, and `model_provider.rs:302` only under `xla-iree`; none are in the CI matrix. So the criterion holds for the stated feature set.

**No production code touched.** All three changed files are `_tests.rs`. `cargo fmt --check` clean on all three. Commit message follows `commit-conventions` (`test:` prefix, unwrapped body, no AI attribution) and honestly records that the checkpoint-present arm was verified by inspection only.

### Remaining Items
- [ ] (LOW, pre-existing, not fixed here) Under the opt-in `xla-backend` / `experimental-backend` features, `backend::tests::mlx_session_threads_the_token_bias_through` calls `select_backend()`, which reads `MLXCEL_BACKEND` without the env lock, and carries an `unreachable!("select_backend defaults to MLX without MLXCEL_BACKEND=xla")` arm that would fire if it observed the transient value. Neither feature is in the CI matrix and neither is in this issue's scope, so fixing it here would widen a test-hygiene PR. Worth a follow-up if an XLA build is ever added to CI.
- [ ] (LOW, accepted tradeoff) The `eprintln!` skip pattern reports a skipped test as `ok`, so a partially-downloaded checkpoint (directory present, index missing) would silently skip rather than fail. This is the repo-wide convention and issue #1155 explicitly preferred it over `#[ignore]`.

### Verification
- [x] Integrated into project code flow (test-only change; both fixed tests run in the default lib suite)
- [x] Tests pass

inureyes:

No CRITICAL or HIGH findings. Test-only change, no production code touched, nothing auto-fixed.

Re-verified on this box: `cargo test --release --features metal,accelerate --lib muse_glimmer` (80 passed, 0 failed, 0.33s) and `--lib muse_glimmer_startup_guard` repeated 3x (5 passed each, no hang).

### Deadlock analysis: clear

- `env_lock()` is a non-reentrant `std::sync::Mutex`, but nothing called under the guard can re-acquire it. `test_support` is `#[cfg(test)]`-gated at the crate root, so `validate_muse_glimmer_unsupported_startup`, `detect_model_media_support`, `get_model_type`, and `tempfile` structurally cannot reach it.
- No lock-ordering cycle. The only other static test mutexes are the per-module MLX serialization guards in `models/gemma3n_helpers_tests.rs`, `models/llama4_helpers_tests.rs`, and `models/cohere2_moe_tests.rs`. None acquires `env_lock` while held, and none is acquired while `env_lock` is held.
- Poison recovery (`unwrap_or_else(|p| p.into_inner())`) prevents a panicking test from wedging every later env test.
- Drop order is correct: `_env_guard` is declared first, so it releases after the `TempDir`.

### MEDIUM: acceptance criterion 4 is not fully met under `--features xla-backend`

The commit's grep covered *mutators* of `MLXCEL_BACKEND`, but the hazard class is *readers*. `src/backend/tests.rs:103` (`mlx_session_threads_the_token_bias_through`) calls `select_backend()`, which under `xla-backend` or `experimental-backend` reads `MLXCEL_BACKEND` at `src/backend/mod.rs:316` without the env lock. If it observes the sibling's transient `xla` it takes `Session::Xla` and hits the `unreachable!` at line 122.

Pre-existing (the setter came in with #1116), and not reachable in the default or CI feature sets since no workflow enables `xla-backend`. The one-line fix is `let _env_guard = crate::test_support::env_lock::env_lock();` at the top of that test. Left unfixed here to keep this PR scoped; worth a follow-up so #1155's "no other test can observe a transient value" holds for feature-gated builds too.

### LOW

1. **Incomplete skip guard**, `src/vision/encoders/muse_glimmer_fusion_tests.rs:203`. The guard checks only `model.safetensors.index.json`, then `unwrap()`s the `config.json` read and every shard open in `safetensors_shape`. A partially materialized checkpoint (interrupted `mlxcel download` or `hf download`, since files land incrementally) panics with an opaque unwrap instead of skipping. The sibling loading test has no such gap because `read_muse_weight_inventory_from_index` reads the index and nothing else.
2. **Silent coverage erosion.** On the machine that owns the pinned checkpoint, a moved or renamed model dir turns both contract tests into green no-ops, and libtest captures the `eprintln` unless the run uses `--nocapture`. An opt-in `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` that fails instead of skipping would keep the skip honest. Same weakness exists in the `tests/prompt_cache_e2e.rs` pattern this follows.
3. **Ordering fragility.** The other reject tests in the module avoid the unlocked env read only because `ensure!(!xla_backend_requested_from_env())` is the last check in `validate_muse_glimmer_unsupported_startup` (`src/server/startup.rs:827`) and each of them trips an earlier `ensure!`. Move that check earlier, or add a check after it, and they silently regain the race.

### Not findings

- **TOCTOU** between `exists()` and the read is present but inert: hardcoded relative path, no attacker-controlled input, no privileged context, and the worst case is the panic the tests already produced before this PR.
- **Path handling.** `model_dir` is a literal; the shard name comes from the pinned index JSON, which is a trusted local artifact read the same way by the production loader. No injection surface.
- **Resource use.** No leaks. The guard is held across one tempdir create and two small config reads; the added serialization against the other `env_lock` users is not measurable (0.00s across three runs).

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: no new tests expected (change is test-only). Verified the three modified files against their respective conventions: the two checkpoint-skip guards match the existing `eprintln`-and-return pattern already used in `tests/prompt_cache_e2e.rs`, and confirmed by `git log` that both were introduced by #1116 (not this PR's baseline). Ran `cargo test --release --features metal,accelerate --lib muse_glimmer` (80 passed, 0 failed) and `cargo test --release --features metal,accelerate --lib muse_glimmer_startup_guard` 5 times under default parallelism (5 passed each run, no flakes). Grepped `MLXCEL_BACKEND` across `src/`; confirmed only `muse_glimmer_startup_guard_tests.rs` mutates it, so the added `env_lock()` guard is the correct scope and no other module needs to share it.
- **Documentation**: none required (test-hygiene fix, no user-facing or architectural change). PR body's claim that the checkpoint-present arm is verified by code inspection only, not execution, is accurate and left as-is.
- **Lint/Format**: `cargo fmt --check` on the three changed files is clean. `cargo clippy --release --lib --tests --features metal,accelerate -- -D warnings` ran against the existing build cache (no C++ recompile triggered) and is clean.

No changes were needed; the branch is unchanged from commit 79bc61a4. Ready for merge.

PR #1158, opened 2026-08-14
test: derive prompt_cache_e2e slack from the paged block size

### LINKED ISSUE #1156: test: prompt_cache_e2e dense multi-turn assertion under-allows for block flooring with thinking models
## Problem / Background

`prompt_cache_e2e::multi_turn_chat_reports_cached_tokens_and_lowers_prefill_latency` (an `#[ignore]`d test run manually with local qwen3-0.6b-4bit weights) fails deterministically at turn 3: `cached_tokens` is 64 against a previous prompt length of 73, but the assertion at `tests/prompt_cache_e2e.rs:359` allows only 4 tokens of slack.

Attribution is settled by a two-arm measurement: the pre-epic baseline `9c154ff3` fails with token-for-token identical values (cached 0/48/64/96/112 across 5 turns), so this is pre-existing and not a regression from epic #1148. Reproduced 2026-08-15 on main `a206f089` and on the baseline with identical numbers; the full turn table is in epic #1148's summary comment.

**Mechanism.** Dense trie adoption is block-floored to multiples of 16 (all observed cached values are multiples of 16). With `max_tokens=16`, the qwen3 thinking model produces empty assistant content (all 16 tokens go to the unterminated think block), so the history re-render diverges right at the previous prompt boundary, and flooring can cost up to 15 tokens, which exceeds the 4-token slack.

## Proposed Solution

One of:

- Raise the slack to one block minus 1 (15 tokens).
- Derive the slack from the paged block size constant instead of hardcoding 4, so the assertion stays correct if the block size changes.
- Make the fixture avoid the empty-content case: use a non-thinking model, or a larger `max_tokens` so assistant content is non-empty and the re-rendered history does not diverge at the previous prompt boundary.

The other assertions in the test (`cached > 0`, monotonic growth) are sound and should stay strict.

## Acceptance Criteria

- [ ] The multi-turn test passes deterministically with the local qwen3-0.6b-4bit fixture across all 5 turns
- [ ] The slack in the under-allowance assertion is either derived from the block size constant or documented with the block-flooring rationale
- [ ] The strict assertions (cached_tokens > 0 from turn 2, monotonic growth) remain unchanged

## Technical Considerations

Refs epic #1148 (integration verification that surfaced this, full turn table in its summary comment). The deterministic token-identical failure on both arms rules out cache-policy changes from the epic as a cause.

inureyes:
## Summary

Fixes the deterministic failure in the ignored `prompt_cache_e2e::multi_turn_chat_reports_cached_tokens_and_lowers_prefill_latency` test, where the turn-3 under-allowance assertion allowed only 4 tokens of slack between `cached_tokens` and the previous turn's prompt length. APC lookup can only credit whole verified blocks, so `cached_tokens` is floored to a multiple of the APC block size (16 by default), which legitimately costs up to 15 tokens when a re-render diverges just past a block boundary.

## What changed

- `tests/prompt_cache_e2e.rs`: the turn-3+ under-allowance assertion changes from a hardcoded `cached + 4 >= prev_prompt_len` to `cached + (APC_BLOCK_SIZE - 1) >= prev_prompt_len`, where `APC_BLOCK_SIZE` is `mlxcel::server::prompt_cache::DEFAULT_APC_BLOCK_SIZE` imported directly, so the assertion tracks the constant rather than a copied literal.
- `tests/prompt_cache_e2e.rs`: expanded comments record where the flooring actually comes from (`apc_consistent_prefix_len` in `src/server/prompt_cache/apc_lookup.rs` returns `consistent_blocks * block_size`; the dense adopt path in `Scheduler::try_adopt_cached_prefix` then truncates to exactly that length), why 16 is the right constant for this test (`--apc-enabled` defaults to `true`, and `--batch-size 1` resolves the decode backend to Dense), and that the paged backend floors to the coarser 32-token pool block so raising `--batch-size` here would require widening the slack.
- `tests/prompt_cache_e2e.rs`: `spawn_server` now scrubs the `APC_BLOCK_SIZE` / `APC_ENABLED` env vars before spawning `mlxcel-server`, so an exported `APC_BLOCK_SIZE` in the invoking shell cannot silently change the block size the server uses out from under the `DEFAULT_APC_BLOCK_SIZE`-derived slack assertion, which would otherwise break the bound in either direction.
- The strict assertions (`cached_tokens > 0` from turn 2 onward, monotonic growth across turns) are unchanged.
- This is a test-only change; no production code paths are touched.

## Test plan

- [x] `cargo fmt --check -- tests/prompt_cache_e2e.rs`
- [x] `cargo test --release --features metal,accelerate --test prompt_cache_e2e multi_turn -- --ignored --nocapture` run five times against the local `qwen3-0.6b-4bit` checkpoint (twice before the import fix, twice after, once more after the `spawn_server` env-var scrub); every run passed with identical `cached_tokens` across all 5 turns (0/48/64/96/112 against prompts 50/73/96/119/142), confirming the assertion fix, the constant import, and the env-var hardening are all deterministic.

Closes #1156

### DIFF
--- a/TECHNICAL_REPORTS/1158-prompt-cache-e2e-block-slack-20260815.en.md
+++ b/TECHNICAL_REPORTS/1158-prompt-cache-e2e-block-slack-20260815.en.md
@@ -1,0 +1,171 @@
+# Technical Report: PR #1158 - test: derive prompt_cache_e2e slack from the paged block size
+
+**Date**: 2026-08-15
+**Author**: AI Code Reviewer
+**Status**: Completed
+**Languages**: Rust
+**Risk Level**: Low
+
+---
+
+## Executive Summary
+
+The manual multi-turn prompt-cache e2e test (`tests/prompt_cache_e2e.rs`, `#[ignore]`d, run against local qwen3-0.6b-4bit weights) failed deterministically at turn 3 because its under-allowance assertion granted only 4 tokens of slack while the APC lookup path floors cache credit to whole 16-token blocks, which can legitimately cost up to 15 tokens. This PR derives the slack from the server's own `DEFAULT_APC_BLOCK_SIZE` constant (slack = block size - 1 = 15), documents the flooring mechanism accurately, and scrubs the `APC_BLOCK_SIZE`/`APC_ENABLED` env-var fallbacks when spawning the server so the assertion's premise holds regardless of the invoking shell. Test-only change; six recorded runs with the real fixture produce identical per-turn values.
+
+---
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+Issue #1156 was filed from epic #1148's integration verification. A two-arm measurement settled attribution: the pre-epic baseline `9c154ff3` fails with token-for-token identical values (cached 0/48/64/96/112 across 5 turns), so the defect is pre-existing in the test, not an epic regression.
+
+### 1.2 Existing Issues
+
+- **Issue 1**: At turn 3, `cached_tokens` is 64 against a previous prompt length of 73, and the assertion `cached + 4 >= prev_prompt_len` fails: the 4-token slack does not cover block flooring.
+- **Issue 2**: The mechanism: APC lookup credits only whole verified blocks (`apc_consistent_prefix_len` returns `consistent_blocks * block_size`, `src/server/prompt_cache/apc_lookup.rs`), and the dense adopt path truncates to exactly that. With `max_tokens=16` the qwen3 thinking model emits empty assistant content (all 16 tokens go to the unterminated think block), so the history re-render diverges right at the previous prompt boundary and the flooring loss is fully exposed: worst case `block_size - 1 = 15` tokens, which exceeds 4.
+- **Issue 3** (found during review): the spawned server inherited the parent environment, and `mlxcel-server` falls back to `APC_BLOCK_SIZE`/`APC_ENABLED` env vars when the flags are absent — an exported `APC_BLOCK_SIZE` would silently change the server's block size while the assertion kept using the compile-time constant.
+
+### 1.3 Risk Assessment
+
+| Risk | Impact | Likelihood |
+|------|--------|------------|
+| Deterministic false failure hides real cache regressions behind a known-red test | Medium | High (fails every run at turn 3) |
+| Hardcoded slack silently diverges if the block size constant changes | Low | Low |
+
+---
+
+## 2. Technical Review
+
+### 2.1 Security
+
+No security surface (test-only). The security pass verified 15 is the tight bound, arithmetic is overflow-safe (const-evaluated; `cached + 15` wrap requires `cached` near `u64::MAX` where the assertion passes either way), and surfaced the env-var inheritance finding (MEDIUM) that the orchestrator fixed with `.env_remove` in `spawn_server` (precedent: `tests/cli_help_consistency.rs`).
+
+**Issues Found:**
+| Issue | Severity | Status |
+|-------|----------|--------|
+| False justification comment: claimed the crate constant was unimportable (root package "has no lib target") — it does, and a sibling test already imports the exact path | High | Fixed (`203d1118`) |
+| Mechanism misattribution in the comment: credited "dense-trie flooring"/"whole-block donation"; donation stores exact tokens and the trie returns an exact LCP — the floor is at APC lookup | Medium | Fixed (`203d1118`) |
+| Inherited `APC_BLOCK_SIZE`/`APC_ENABLED` env vars could break the assertion's premise both ways (32 → false failure; 8 → over-permissive) | Medium | Fixed (`45d8757c`) |
+| Pre-existing doc bug in `apc_lookup.rs:46` ("input matched_len is preserved" — false, return is always block-floored) | Low | Open (production doc, out of scope; flagged on the PR) |
+
+### 2.2 Performance
+
+None; assertion arithmetic only. The prefill-latency assertion continued to pass in every recorded run (ratios 0.37-0.50 against a 1.3 upper bound).
+
+### 2.3 Compatibility & Dependencies
+
+- **Breaking Changes**: none
+- **New Dependencies**: none — the lib import adds no build surface (18 other files under `tests/` already import from the `mlxcel` lib)
+- **Compatibility**: valid only for the Dense decode backend (`--batch-size 1`); the paged backend applies a coarser 32-token floor (`DEFAULT_PAGED_BLOCK_SIZE`), now documented on the constant
+
+### 2.4 Code Quality
+
+- **Test Coverage**: the previously always-red manual test is now a usable regression gate
+- **Code Complexity**: one imported constant, one assertion change, two env removals
+- **Technical Debt**: decreased (magic number replaced with the source-of-truth constant; env-dependent premise closed)
+
+---
+
+## 3. Technical Decisions
+
+### 3.1 Derive the slack from the constant instead of widening the literal or changing the fixture
+
+**Alternatives Considered:**
+
+| Option | Pros | Cons |
+|--------|------|------|
+| Raise the literal to 15 | Minimal diff | Silently diverges if the block size changes |
+| Change the fixture (non-thinking model or larger `max_tokens`) | Avoids the empty-content divergence | Hides rather than accommodates the mechanism; the flooring loss still exists for any divergence at a non-aligned boundary |
+| **Chosen: import `DEFAULT_APC_BLOCK_SIZE`, slack = block size - 1** | Assertion tracks the source of truth; 15 is provably the tight bound | Requires the (correct) importability of the crate constant |
+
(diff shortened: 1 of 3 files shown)


inureyes:

### Intent
> Widen the under-allowing turn-3+ slack in the ignored `prompt_cache_e2e` multi-turn test so it covers one block of prefix-cache flooring, derived from the block size constant rather than hardcoded (issue #1156).

### Findings Addressed
- [x] **The "constant is not importable" premise is false** (HIGH). The local `const APC_BLOCK_SIZE: u64 = 16;` was justified by a comment claiming the test crate has no `[lib]` edge and that the root `mlxcel` package "carries no `[lib]` target either, only `[[bin]]`s". The root package does have an auto-discovered lib target at `src/lib.rs`, `pub mod server;` is public, `src/server/prompt_cache/mod.rs` re-exports `DEFAULT_APC_BLOCK_SIZE`, and sibling integration tests already import from that exact path (`tests/paged_prefix_share_parity.rs:64-67`). Since the fallback's precondition did not hold, the issue's preferred deliverable (derive from the constant) was available. Fixed: `use mlxcel::server::prompt_cache::DEFAULT_APC_BLOCK_SIZE;` with `const APC_BLOCK_SIZE: u64 = DEFAULT_APC_BLOCK_SIZE as u64;`.
- [x] **Rationale comment misattributed the mechanism** (MEDIUM). It credited "dense-trie flooring" and said "paged prefix-cache adoption only donates back whole blocks". Both are wrong: donation stores the exact token vector (`scheduler.rs:2819-2821`) and the radix trie returns an exact LCP with no rounding (`prompt_cache/lookup.rs:103`). The floor is `apc_consistent_prefix_len` returning `consistent_blocks * block_size` (`prompt_cache/apc_lookup.rs:82`), after which the dense adopt path truncates to exactly that length (`scheduler.rs:2113-2143`). Comment rewritten to name the real call sites; the assert message now says "APC block-flooring slack".
- [x] **Missing precondition for why 16 is the right constant** (MEDIUM). 16 applies here only because `--apc-enabled` defaults to `true` (`src/bin/mlx_server.rs:1208`, note `ApcConfig::default().enabled` is `false`, so the CLI default is what makes APC live) and `--batch-size 1` resolves the decode backend to Dense. On the paged backend the adopt path floors to the coarser `DEFAULT_PAGED_BLOCK_SIZE` = 32 (`scheduler.rs:2028`, `:2157`), so raising `--batch-size` in this test would require widening the slack. Both facts are now documented on the constant.

### Remaining Items
- [ ] `apc_lookup.rs:46` doc comment claims "If every covered block agrees, the input `matched_len` is preserved", which is false: the return is always `consistent_blocks * block_size`. The unit test that appears to confirm it (`apc_lookup_tests.rs:28-42`) only passes because it uses `matched_len = 64`, already block-aligned. Pre-existing production doc bug, out of scope for this test-only PR.

### Verification
- [x] All stated requirements implemented (slack now derived from the constant; block-flooring rationale documented)
- [x] Integrated into project code flow (test-only change by design; no production path touched)
- [x] Project conventions followed (`cargo fmt --check` clean; import style matches sibling integration tests; no `references/` paths, no cross-repo bare `#NNN`)
- [x] Existing modules reused where applicable (the crate constant is now imported instead of duplicated)
- [x] No unintended structural changes (single file, `tests/prompt_cache_e2e.rs`)
- [x] Tests pass. Slack magnitude confirmed correct: `APC_BLOCK_SIZE - 1` = 15 is exactly the maximum flooring loss, not an over-allowance. Observed values are exactly `floor(prev_prompt_len / 16) * 16` at every turn (50→48, 73→64, 96→96, 119→112). The strict assertions (`cached_tokens > 0` from turn 2, `cached >= prev_cached`) are byte-identical to `main`. Ran the ignored e2e test twice more against the local `qwen3-0.6b-4bit` fixture (four runs total with the implementer's two); every run produced identical per-turn values 0/48/64/96/112.

inureyes:
## Security / performance review

Test-only, one file. No CRITICAL or HIGH findings, so nothing was auto-fixed and no commits were pushed. One MEDIUM worth addressing before merge.

### Verified: 15 is the tight bound, it does not over-allow

`apc_consistent_prefix_len` (`src/server/prompt_cache/apc_lookup.rs:49`) returns `consistent_blocks * block_size`, and the dense arm of `try_adopt_cached_prefix` truncates to exactly that `matched_len` (`src/server/batch/scheduler.rs`, `let mut adopted_len = matched_len;`). So when the previous turn's prompt of length `P` is a genuine prefix, the credited value is `floor(P/16)*16`, whose worst case is `P - 15`. Anything below 15 can false-fail; anything above 15 starts admitting real losses.

It cannot mask a representable regression either. Losing one whole APC block gives `cached <= floor(P/16)*16 - 16 <= P - 16`, so `cached + 15 <= P - 1` and the assertion fires. A full block is the smallest loss the dense APC path can express, so the band admits exactly the flooring artifact and nothing else.

The recorded runs confirm the model exactly: cached `48/64/96/112` against previous prompts `50/73/96/119` is `floor(P/16)*16` in all four cases, with observed slack 2/9/0/7 against the 15 allowed.

### MEDIUM: the assertion's block size is the compile-time default, not the value the spawned server uses

`mlxcel-server` reads `APC_BLOCK_SIZE` as a fallback when `--apc-block-size` is absent (`env_fallback_apc_block_size`, `src/server/cli_input.rs:1354`), and `spawn_server` (`tests/prompt_cache_e2e.rs:103`) inherits the parent environment. The test passes no `--apc-block-size`, so an exported `APC_BLOCK_SIZE` silently changes the server's block size while the assertion keeps using `DEFAULT_APC_BLOCK_SIZE`.

That makes the new doc comment's claim ("this test passes no `--apc-block-size`, so this is the block size the server under test uses") incorrect in that case, and it breaks the bound in both directions: `APC_BLOCK_SIZE=32` gives a real bound of 31 against 15 allowed (false failure), and `APC_BLOCK_SIZE=8` gives a real bound of 7 against 15 allowed (over-permissive, an 8-token single-block regression passes).

Suggested fix, in the same shape the test already uses to pin the decode backend with `--batch-size 1`: pass `--apc-block-size` built from the constant, so the flag beats the env var and the assertion and the server agree by construction. Scrubbing `APC_BLOCK_SIZE` and `APC_ENABLED` via `Command::env_remove` in `spawn_server` also works; there is precedent in `tests/cli_help_consistency.rs:120`.

`APC_ENABLED=false` is benign here: it only removes the floor, which raises `cached` and keeps the assertion passing.

### LOW: arithmetic is safe

`APC_BLOCK_SIZE - 1` is const-evaluated to 15. A hypothetical zero block size would be rejected by const eval at compile time rather than wrapping at runtime, so there is no `u64` underflow path. `cached + 15` could in principle wrap in release only if a malformed `cached_tokens` came back near `u64::MAX`, and the assertion passes either way in that case, so there is no masking difference. No change needed. If the line is touched again, the floor form `cached >= (prev_prompt_len / APC_BLOCK_SIZE) * APC_BLOCK_SIZE` is equivalent given the credited value is a block multiple, and is overflow-free without the extra reasoning.

### Note on the comment's scope

`try_adopt_cached_prefix` runs the snapshot-reuse path (`lookup_snapshot_outcome`) ahead of the APC path when `model.supports_snapshot_reuse()`, and that path returns an unfloored `matched_len`. It can only make `cached` larger, so it cannot false-fail this assertion, but "cached_tokens is floored to a multiple of APC_BLOCK_SIZE" describes the APC path rather than every path that can produce the number.

Nothing in the changed lines touches process state or external resources: the diff is a `use` of a compile-time constant plus an arithmetic change. The lib import adds no build or link surface, since the test target already depends on the `mlxcel` lib and 18 other files under `tests/` import from it.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: no new tests needed for this test-only PR; reran `cargo test --release --features metal,accelerate --test prompt_cache_e2e multi_turn -- --ignored --nocapture` against the local `qwen3-0.6b-4bit` fixture as a fifth confirmation of the final three-commit state, and it reproduced the same `cached_tokens` sequence (0/48/64/96/112 against prompts 50/73/96/119/142) as the four prior runs.
- **Documentation**: no `docs/` updates were expected for this test-only change; the PR body's "What changed" section was missing the `spawn_server` env-var scrub from 45d8757c, so I added one bullet describing it and updated the test-plan line to five runs so the body matches the final diff.
- **Lint/Format**: `cargo fmt --check -- tests/prompt_cache_e2e.rs` passes clean; traced the doc comment through all three commits and confirmed it now reads coherently (constant source, `--apc-enabled`/`--batch-size 1` preconditions, and the env-var scrub note are all consistent with the code and with each other).

Verified `DEFAULT_APC_BLOCK_SIZE` is genuinely importable at `mlxcel::server::prompt_cache::DEFAULT_APC_BLOCK_SIZE` (defined in `src/server/prompt_cache/block_hash.rs`, re-exported via `src/server/prompt_cache/mod.rs`), and that `APC_BLOCK_SIZE` / `APC_ENABLED` are the exact env var names the server's CLI fallback reads (`src/server/cli_input.rs`), so both prior review fixes are sound in the final state.

No source changes were required; only the PR body was edited via `gh pr edit`.

PR #1170, opened 2026-08-15
fix(cli): reject a DFlash drafter before the offline standalone load

### LINKED ISSUE #1168: fix(cli): mlxcel generate --draft-model loads a DFlash drafter as a standalone model and dies on "Weight not found: model.embed_tokens.weight"
## Summary

`mlxcel generate --draft-model <dflash-drafter>` always fails with

```
Error: Weight not found: model.embed_tokens.weight
```

because the offline CLI loads the drafter directory as a **full standalone `LoadedModel`** instead of through `mlxcel_core::drafter::load_drafter`. A DFlash checkpoint is not a loadable standalone model, so the generic loader dies on its first weight lookup.

The same drafter and target work fine on `mlxcel-server`. This is a CLI-path defect, not a model, size, or shape problem.

Secondarily, `docs/supported-models.md:523` advertises DFlash for "Qwen 3.5 text/VLM paths" without noting that the offline CLI supports it for neither.

## Reproduction (2026-08-15, `main` @ `6e781fcf`)

Fails identically with and without `--draft-kind`:

```
$ ./target/release/mlxcel generate -m models/qwen3.5-27b-4bit \
    --draft-model models/qwen3.5-27b-dflash -p "hi" -n 20 --temp 0
Model loaded in 0.220s
Generating...
Loading draft model from "models/qwen3.5-27b-dflash"...
Converting 58 bf16 weight tensors to f16 for Apple Silicon fp16 optimization.
Error: Weight not found: model.embed_tokens.weight

$ ... --draft-kind dflash          # same error
```

Target `models/qwen3.8-27b-4bit` reproduces it identically, so it is not generation-specific. The target loads successfully first; the failure is entirely on the drafter side.

## Root cause

`src/commands/generate.rs:1440-1441` loads the drafter as a normal model:

```rust
println!("Loading draft model from {:?}...", draft_model_path);
let (draft_model, _draft_tokenizer) = select_backend().load_model(draft_model_path)?;
```

`load_drafter` is reached **only** for explicit `--draft-kind mtp` (gate `should_route_offline_mtp`, `generate.rs:1579-1584`, then `run_offline_mtp` at `:1646` and `load_drafter(..., Some(DrafterKind::Mtp))` at `:1688`). Every other kind, DFlash included, goes through `load_model`.

From there:

1. `get_model_type` (`src/models/detection.rs:97`) reads the drafter's `config.json`. The published DFlash checkpoint declares **`"model_type": "qwen3"`** (its `architectures: ["DFlashDraftModel"]` and `auto_map` are never consulted). `detection.rs:128`: `"qwen3" => Ok(ModelType::Qwen3)`.
2. `src/model_metadata.rs:101` routes `Qwen3` to `models::Qwen3Model::load`.
3. `src/models/qwen3.rs:801-803` does its first lookup:
   ```rust
   let embed_tokens =
       UnifiedEmbedding::from_weights(weights, "model.embed_tokens", group_size, bits)?;
   ```
4. `src/lib/mlxcel-core/src/layers.rs:205-209` produces the message.

The DFlash checkpoint's 58 tensors carry **no `model.` prefix at all**: `fc.weight`, `hidden_norm.weight`, `norm.weight`, `layers.{i}.input_layernorm.weight`, `layers.{i}.mlp.{gate,up,down}_proj.weight`, `layers.{i}.post_attention_layernorm.weight`, `layers.{i}.self_attn.{q,k,v,o}_proj.weight`, `layers.{i}.self_attn.{q,k}_norm.weight`. There is no `embed_tokens` and no `lm_head` by design: the drafter borrows both from the target at bind time.

So this fails for **every** DFlash drafter regardless of target, including the 4B pairing used in tests.

### The correct loader exists and is not reached

`DFlashDrafter::load` (`src/lib/mlxcel-core/src/drafter/dflash/drafter.rs:98-136`) handles exactly this shape, and `DFlashDraftModel::from_weights` (`.../dflash/model.rs:136-184`) treats the missing modules as optional tombstones:

```rust
let has_embed_tokens = weights.contains_key("embed_tokens.weight")
    || weights.contains_key("embed_tokens.scales");
let embed_tokens = if has_embed_tokens { Some(UnifiedEmbedding::from_weights(...)?) } else { None };
...
let lm_head = if weights.contains_key("lm_head.weight") {
    LmHead::Own(Linear::from_weights(weights, "lm_head")?)
} else if config.tie_word_embeddings { LmHead::Tied } else { LmHead::TargetBound(None) };
```

`Drafter::bind` (`.../dflash/drafter.rs:210-249`) then takes the target's modules directly, not by weight-map key:

```rust
if self.model.needs_embed_binding() {
    let embed = target.embed_tokens_module().ok_or_else(|| DrafterError::BindFailed { ... })?;
    self.model.bind_target_embedding(embed);
}
```

`embed_tokens_module()` / `lm_head_module()` return shared-buffer clones (`src/models/qwen3_5.rs:2646-2655`; the VLM wrapper delegates at `src/vision/qwen3_5_vl.rs:473-479`, documented precisely so "a DFlash drafter can lazy-bind even when the target is a VLM-wrapped Qwen 3.5 checkpoint"). Quantized target embeddings are covered by `UnifiedEmbedding::clone_shared` (`src/lib/mlxcel-core/src/layers.rs:383-392`).

### Second wall behind the first

Even if the load were fixed, `src/commands/generate.rs:1475-1482` rejects an explicit `--draft-kind dflash` offline:

> `--draft-kind dflash is plumbed end-to-end but the offline mlxcel generate path does not yet construct the kind-specific DFlashGenerator round loop on this target model.`

This is intentional and documented (`docs/speculative-acceptance.md:274-276`). The bug is that a user hits the confusing `Weight not found` message *before* ever reaching this clear one, and that omitting `--draft-kind` gives them the confusing message with no clear one behind it at all.

## The server path works

Same 27B pair, `mlxcel-server`:

```
$ ./target/release/mlxcel-server -m models/qwen3.5-27b-4bit \
    --draft-model models/qwen3.5-27b-dflash --draft-kind dflash --port 8241

INFO Lazy-loading drafter from models/qwen3.5-27b-dflash (kind=Some(Dflash))
INFO Drafter loaded (kind=dflash, 0 ms)
```

A 120-token completion produced real speculative work:

| metric | value |
|---|---|
| rounds | 24 |
| proposed_tokens | 347 |
| accepted_tokens | 95 |
| acceptance_rate | 0.274 |
| emitted_per_verify | 4.96 |

The server reaches `load_drafter` correctly (`src/server/batch/speculative_burst.rs:283`, inside `DrafterSlot::ensure_loaded`), and its target-variant gate accepts `Qwen35 | Qwen35Moe | Qwen35VLM | Qwen35MoeVLM` (`speculative_burst.rs:1351-1364`) with VLM wrappers dispatched into `run_dflash_on_qwen35` (`:1406-1430`, batched twin at `:2439-2490`).

This also settles a question the docs leave open: DFlash **does** work against a VLM-shaped Qwen 3.5 target. The 27B checkpoint here is `Qwen3_5ForConditionalGeneration` with a `vision_config`, so it loads as `ModelType::Qwen35VLM`.

## Test coverage gap

- No test anywhere invokes `mlxcel generate --draft-model`. Grepping `tests/`, the flag appears only in `tests/cli_help_consistency.rs` help-text assertions. That is why a path that cannot work for any drafter has stayed green.
- The one real-model DFlash test, `greedy_parity_dflash_qwen35_4b` (`tests/speculative_parity.rs:609-706`, pairing at `:378-384`), is `#[ignore]`d, self-skips when the checkpoints are absent, and drives **`mlxcel-server` subprocesses**, not the CLI.
- That test also does not pin the target's shape: `speculative_parity.rs:641-648` accepts `Qwen35 | Qwen35Moe | Qwen35VLM | Qwen35MoeVLM`, with a comment at `:634-640` noting mlx-community publishes even text-only 4B as `Qwen3_5ForConditionalGeneration`. So it can neither confirm nor deny VLM coverage. Same caveat for MTP at `tests/qwen35_mtp_verify_parity.rs:101-113`.
- `tests/speculative_dispatch.rs:130-360` covers DFlash config resolution with synthetic fixtures and never loads weights. `src/server/batch/speculative_burst_tests.rs:209-218` is a gate-predicate unit test with no model.

## Proposed work

1. **Route the drafter through `load_drafter` on the CLI.** `src/commands/generate.rs:1441` should resolve the drafter kind first and call `mlxcel_core::drafter::load_drafter` for DFlash, as the server does, rather than `load_model`. If the classic `SpeculativeGenerator` fallback genuinely needs a full `LoadedModel`, then a DFlash checkpoint must be rejected before that load with a named error.
2. **Fail with a message that names the problem.** Whatever the resolution, `Weight not found: model.embed_tokens.weight` must not be what a user sees for a structurally valid drafter. It should say that the directory is a DFlash drafter and that the offline path does not construct the DFlash round loop, mirroring the existing `--draft-kind` message.
3. **Do not let the drafter's `model_type: "qwen3"` route it to a full model loader.** Detection should recognize `architectures: ["DFlashDraftModel"]` (or the presence of `dflash_config`) and refuse to treat the directory as a standalone model, which would also give a better error for `mlxcel generate -m <dflash-dir>`.
4. **Correct the docs.** `docs/supported-models.md:523` should state that DFlash is server-only today. `docs/speculative-acceptance.md:274-276` already says the right thing; the supported-models table contradicts it by omission.
5. **Add a regression test** that runs `mlxcel generate --draft-model` against a synthetic DFlash-shaped fixture and asserts the named error rather than `Weight not found`. This is the missing coverage that let the defect persist.

## Scope note

Found while qualifying Qwen3.8-27B (#1163), but pre-existing and unrelated to Qwen3.8: the failure reproduces identically on `qwen3.5-27b-4bit`, and by inspection applies to every DFlash drafter including the 4B one. Not a regression from any recent change.

Whether the offline `mlxcel generate` path should gain a real DFlash round loop is a separate, larger question already tracked by the message at `generate.rs:1478`; this issue is only about the load failing in the wrong place with the wrong message, and about the docs overstating CLI support.

## Acceptance criteria

- [x] `mlxcel generate --draft-model <dflash-dir>` no longer reports `Weight not found: model.embed_tokens.weight`, with or without `--draft-kind`
- [x] The resulting behavior is either a working classic/DFlash path or a named error explaining that the offline path is server-only
- [x] A DFlash directory passed to `mlxcel generate -m` is rejected as not a standalone model
- [x] Regression test on a synthetic DFlash fixture covering both the `--draft-model` and `-m` cases
- [x] `docs/supported-models.md:523` states the server-only limitation; the DFlash row no longer contradicts `docs/speculative-acceptance.md`
- [ ] The measured server-side numbers above recorded in the closing comment as the working reference (27B VLM-shaped target, 24 rounds, 27.4% acceptance)

The five checked boxes are delivered in #1170: a structural DFlash probe in `mlxcel-core`, a `get_model_type` rejection arm for the `-m` case, a `reject_dflash_drafter_offline` precheck for the `--draft-model` case, synthetic-fixture tests at four levels, and the docs correction. That PR's Test plan section records exactly which verification commands were executed on the branch and which are queued for the broader suite. The last box stays open until the closing comment is posted at merge.

inureyes:
## Summary

`mlxcel generate --draft-model <dflash-drafter>` failed with `Error: Weight not found: model.embed_tokens.weight`, with and without `--draft-kind`, for every DFlash drafter and every target. The offline path calls `load_model` on the drafter directory because the classic `SpeculativeGenerator` it falls through to drives a full `LoadedModel`, and a DFlash checkpoint is not one: it ships no `embed_tokens` and no `lm_head` because it borrows both from the target when it binds, and its tensors carry no `model.` prefix. Its `config.json` nevertheless declares an ordinary `"model_type": "qwen3"`, so `get_model_type` classified it `ModelType::Qwen3`, `model_metadata` routed it to `Qwen3Model::load`, and the first `UnifiedEmbedding::from_weights` lookup failed with a message naming a tensor rather than the problem. This PR makes both offline entry points fail where it makes sense, with a message that names the real problem, and corrects the docs that overstated CLI support.

This deliberately does not build an offline DFlash round loop. That remains out of scope and is still described by the existing `--draft-kind` message in `generate.rs` and by `docs/speculative-acceptance.md`.

## The discriminator is structural, not the resolved drafter kind

`DEFAULT_DRAFTER_KIND` is `DrafterKind::Dflash` and `drafter_kind_by_model_type()` maps only `gemma4_assistant` and `gemma4_unified_assistant`, so **every** drafter that is not a Gemma 4 assistant auto-resolves to `Dflash`, including an ordinary small full model (a Qwen3 0.6B, say) used as a classic drafter. That pairing loads and runs fine today on the classic `SpeculativeGenerator` path, so rejecting on the resolved kind would have broken a working workflow. The new probe keys on the checkpoint's own DFlash markers instead: a nested `dflash_config` object (what `DFlashConfig::from_json` itself reads) and/or `architectures: ["DFlashDraftModel"]` (what HuggingFace `AutoModel` dispatch reads). Every test added here carries an ordinary-full-model control asserting the classic path still resolves.

## The server path is unaffected

`DFlashDrafter::load` reads `config.json` into `DFlashConfig::from_json` directly and `resolve_drafter_kind` peeks `model_type` through its own reader, so neither reaches `get_model_type`. Re-confirmed on this branch that every `get_model_type` call site in the tree takes a target or model path and never a drafter path (`src/loading/mod.rs`, `src/distributed/pipeline/stage_executor/mod.rs`, `src/distributed/tensor_parallel/inference.rs`, `src/server/chat_template.rs`, `src/server/startup.rs`, `src/server/model_provider.rs`, `src/server/batch/xla_worker.rs`, `src/multimodal/host_preprocessor.rs`, and the three target-path uses in `src/commands/generate.rs`). The drafter reaches `load_drafter` through `SpeculativeDispatch` and `DrafterSlot::ensure_loaded`, neither of which touches detection. `src/bin/speculative_bench.rs` likewise passes only the target to `load_model`. A new test in `tests/speculative_dispatch.rs` pins that `SpeculativeDispatch::resolve` still returns the DFlash variant for the same fixture directory that detection now rejects, with and without an explicit `--draft-kind`.

For the record, the working reference this issue measured on `main` @ `6e781fcf` (Qwen 3.5 27B 4-bit target, which loads as `ModelType::Qwen35VLM`, plus the 27B DFlash drafter, 120-token completion): rounds 24, proposed_tokens 347, accepted_tokens 95, acceptance_rate 0.274, emitted_per_verify 4.96. That also backs the "VLM paths" claim the supported-models table makes for DFlash, which is why this PR keeps it.

## What changed

- `src/lib/mlxcel-core/src/drafter/dflash/config.rs`: new `DFLASH_DRAFT_ARCHITECTURE` constant, `is_dflash_drafter_config(&Value)`, and `is_dflash_drafter_dir(&Path)`, placed next to the parser that consumes `dflash_config` so the structural definition lives with the loader that keys on it. The doc comment records why `model_type` and the resolved `DrafterKind` are both wrong discriminators. An unreadable or missing `config.json` returns `false` so the general loader keeps reporting its own I/O and parse diagnostics.
- `src/lib/mlxcel-core/src/drafter/dflash/mod.rs`: re-exports the three new items.
- `src/models/detection.rs`: `get_model_type` rejects a structurally-DFlash directory before any `model_type` dispatch, via a shared `dflash_drafter_not_standalone_error`. This is the single source of truth for the `-m` case and covers server startup and the distributed stage loaders at the same time.
- `src/commands/generate.rs`: new `reject_dflash_drafter_offline`, called in `run_generation_mode` immediately before `load_model(draft_model_path)`. The error says the directory is a DFlash drafter, that the offline path does not construct the `DFlashGenerator` round loop, and that `mlxcel-server` with `--draft-kind dflash` does. It fires with no `--draft-kind` and with `--draft-kind dflash`; an explicit `--draft-kind mtp` is a different request that keeps its own targeted errors from `run_offline_mtp`.
- `docs/supported-models.md`: the DFlash row now reads `mlxcel-server` only and explains that offline `mlxcel generate` rejects a DFlash drafter with a named error, so it no longer contradicts `docs/speculative-acceptance.md` by omission. The MTP row gained the matching detail that it does work offline under `--draft-kind mtp`. A following paragraph explains why a DFlash checkpoint is not a standalone model.
- `docs/speculative-acceptance.md`: the "note on the drafter-kind line" section now distinguishes the auto-detected `dflash` fallback applied to an ordinary full model (which still runs the classic path) from a real DFlash drafter (which is now refused up front), and records the structural discriminator.
- `src/commands/generate_tests.rs` (finalization pass, security finding M3): `drafter_fixture_dir` and the identical inline `empty`-fixture pattern built a predictable path under `env::temp_dir()` with `create_dir_all` and cleaned up only on the success path (CWE-377/379 class, CI-relevant on shared-`TMPDIR` Linux runners). Both now use `tempfile::tempdir()`, matching the pattern the `mlxcel-core` tests added in this PR already use.
- `CHANGELOG.md`, `TECHNICAL_REPORTS/1170-dflash-drafter-cli-load-20260816.en.md` (finalization pass): the `## [Unreleased]` entry and technical report this repository's convention expects for every merged PR.

## Test plan

New coverage, all on synthetic fixtures. No test previously invoked `mlxcel generate --draft-model` at any level, which is why a path that could never work for any DFlash drafter stayed green. All of the below has now actually been executed (by the orchestrator and by review), not just written.

- [x] `cargo test --release --workspace --features metal,accelerate --no-fail-fast`: every target passes except `-p mlxcel --lib`, which reports 5621 passed / 3 failed. All 3 failures are `multimodal::video::tests::*` and are unrelated to this PR: this branch touches 0 lines of `src/multimodal/video.rs`, and a PATH-controlled experiment on the same binary gives 34 passed / 0 failed with ffmpeg hidden from `PATH` versus 31 passed / 3 failed with ffmpeg 9.0.1 present. Root cause is `src/multimodal/video.rs:1123` passing `-vsync`, an argument ffmpeg 9 removed. Filed as #1172.
- [x] Gotcha for the next person: the four new `commands::generate::tests::*` precheck tests compile into `--bin mlxcel`, not `--lib`, so they were **not** part of the `--lib` run above. They were run separately with `cargo test --release --bin mlxcel --features metal,accelerate commands::generate`: 4 passed.
- [x] `cargo clippy -- -D warnings` clean on both crates (`mlxcel`, `mlxcel-core`).
- [x] `cargo fmt --check` clean.
- [x] Classic-path non-regression, executed rather than reasoned about: `-m qwen3-0.6b-4bit --draft-model qwen3-0.6b-4bit` still runs `SpeculativeGenerator` end to end, 24 tokens, acceptance_rate 1.0000, 137.54 tok/s.
- [x] Mutation test confirming the new detection tests are real assertions, not tautologies: guarding the new arm as `if false && ...` makes `models::detection` and `commands::generate` fail with the pre-fix classification, on both markers.

Not verified here: an end-to-end run of `mlxcel generate --draft-model models/qwen3.5-27b-dflash` against a real 27B target. The change is a pre-load guard, so the reproduction is a one-line error rather than a load, but that confirmation has not been re-run on this branch.

## Known limitations

Reported and deliberately not fixed in this PR:

- The precheck fires only after the target model finishes loading, so an operator with a large target waits through the full load before seeing the error. Moving the check earlier would also intercept `--draft-kind mtp`, which is deliberately routed elsewhere and must keep its own targeted errors from `run_offline_mtp`.
- `--draft-model <dir>/config.json` (pointing directly at the config file rather than the directory) bypasses the precheck and falls through to the detection error with `-m` phrasing, because the precheck does not apply `resolve_model_dir` the way the rest of the loading path does.
- The probe uses raw `serde_json::from_slice` on `config.json`, while `get_model_type` applies `sanitize_config_json` first. A config containing bare `NaN`/`Infinity` literals (invalid JSON, but occasionally emitted by some tooling) would therefore fail the precheck open. No current checkpoint is affected, and the sanitizer lives in the root crate while the probe lives in `mlxcel-core`, so closing this gap is a cross-crate change left for later.

Closes #1168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Fixed
+
+- **`mlxcel generate --draft-model <dflash-drafter>` now fails with a named error instead of a misleading `Weight not found: model.embed_tokens.weight`** (#1168). A DFlash drafter borrows `embed_tokens` and `lm_head` from its target when it binds, so it ships neither, but its `config.json` still declares an ordinary `model_type` (`qwen3`, say), so the offline path classified it as a standalone model and drove it through the full `LoadedModel` loader, which failed on the first missing tensor with a message that named a tensor rather than the problem. Detection now rejects a directory that is structurally a DFlash drafter (a nested `dflash_config` object and/or `architectures: ["DFlashDraftModel"]`) before dispatch, both for `-m/--model` (`get_model_type`, which also covers server startup and the distributed stage loaders) and for offline `--draft-model` (a new pre-load check in `mlxcel generate`), and points at `mlxcel-server --draft-kind dflash` instead. The discriminator is the checkpoint's own markers, not the resolved `DrafterKind`: an ordinary small full model still auto-resolves to `DrafterKind::Dflash` by default (`DEFAULT_DRAFTER_KIND`) and keeps loading through the classic `SpeculativeGenerator` path unaffected, confirmed by a non-regression run (`qwen3-0.6b-4bit` as both target and drafter, acceptance_rate 1.0000, 137.54 tok/s). `docs/supported-models.md` and `docs/speculative-acceptance.md` are corrected to match: DFlash's offline entry is `mlxcel-server` only.
+
 ## [v0.5.1] - 2026-08-15
 
 ### Added
--- a/TECHNICAL_REPORTS/1170-dflash-drafter-cli-load-20260816.en.md
+++ b/TECHNICAL_REPORTS/1170-dflash-drafter-cli-load-20260816.en.md
@@ -1,0 +1,195 @@
+# Technical Report: PR #1170 - fix(cli): reject a DFlash drafter before the offline standalone load
+
+**Date**: 2026-08-16
+**Author**: AI Code Reviewer
+**Status**: Completed
+**Languages**: Rust
+**Risk Level**: Low
+
+---
+
+## Executive Summary
+
+`mlxcel generate --draft-model <dflash-drafter>` failed for every DFlash drafter and every target with `Error: Weight not found: model.embed_tokens.weight`, a message that names a tensor instead of the actual problem. A DFlash checkpoint borrows `embed_tokens` and `lm_head` from its target when it binds, so it ships neither, but its `config.json` still declares an ordinary `model_type` (`qwen3`), which routed it through the full standalone `LoadedModel` loader. This PR adds a structural probe, keyed on the checkpoint's own DFlash markers (a nested `dflash_config` object and/or `architectures: ["DFlashDraftModel"]`) rather than on the resolved `DrafterKind` (which defaults to `Dflash` for any non-Gemma-4-assistant drafter, including ordinary full models used as classic drafters), and rejects a DFlash drafter before both the `-m/--model` detection path and the offline `--draft-model` load, with a message that names the real problem and points at `mlxcel-server --draft-kind dflash`. The server path, which never reaches `get_model_type` for a drafter, is unaffected and pinned by a non-regression test. The finalization pass additionally fixed a predictable-temp-directory test hygiene finding (M3) in the new test fixtures.
+
+---
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+Issue #1168: offline speculative decoding with a DFlash drafter could never succeed, with or without `--draft-kind`, because the offline entry point has no `DFlashGenerator` round loop and instead drives the drafter through the same loader used for a full standalone model.
+
+### 1.2 Existing Issues
+
+- **Issue 1**: `get_model_type` classified a DFlash drafter's `config.json` as `ModelType::Qwen3` (or whatever ordinary `model_type` the checkpoint's config declares), so `model_metadata` routed it to the matching family loader, which failed on the first missing tensor (`UnifiedEmbedding::from_weights` looking for `model.embed_tokens.weight`) rather than reporting that the directory is a drafter.
+- **Issue 2**: The obvious fix, rejecting on the resolved `DrafterKind`, would have broken a working workflow: `DEFAULT_DRAFTER_KIND` is `Dflash` and `drafter_kind_by_model_type()` maps only the two Gemma 4 assistant model types, so an ordinary small full model used as a classic `--draft-model` drafter (a Qwen3 0.6B, for example) also auto-resolves to `Dflash` and runs fine today on the classic `SpeculativeGenerator` path.
+- **Issue 3** (found during this finalization pass, security finding M3): `drafter_fixture_dir` in the new `src/commands/generate_tests.rs` tests built a predictable directory name under `env::temp_dir()` and created it with `create_dir_all`, which succeeds against a pre-existing directory or symlink there (CWE-377/379 class), and cleaned up only on the success path, leaking the fixture on an assertion failure. CI-relevant on Linux, where `TMPDIR` is shared across jobs.
+
+### 1.3 Risk Assessment
+
+| Risk | Impact | Likelihood |
+|------|--------|------------|
+| Offline DFlash `--draft-model` stays permanently broken with a misleading error | Medium (every DFlash drafter, every target) | Certain before this fix |
+| Rejecting on the wrong discriminator breaks the classic small-model drafter workflow | High if mis-keyed | Avoided by keying on checkpoint structure, confirmed by a dedicated control test |
+| Predictable temp-dir fixture is squatted or symlinked in a shared `TMPDIR` (M3) | Low (test-only, no production surface) | Low, but real on shared Linux CI runners |
+
+---
+
+## 2. Technical Review
+
+### 2.1 Security
+
+Review and security passes were completed by the orchestrator and reviewer before finalization: no CRITICAL, no HIGH findings. One MEDIUM finding (M3) was raised against the test-only fixture helper added in this PR.
+
+**Issues Found:**
+
+| Issue | Severity | Status |
+|-------|----------|--------|
+| `drafter_fixture_dir` built a predictable path under `env::temp_dir()` with `create_dir_all` (CWE-377/379) and leaked on assertion failure | Medium | Fixed (`37136aa6`) |
+
+The fix replaces the helper, and the identical inline pattern in the missing-config fixture, with `tempfile::tempdir()`, which mints a randomly-named, securely-permissioned directory and removes it on drop regardless of how the test exits. This is the same pattern the `mlxcel-core` `drafter::dflash::config` tests already use, so the PR is now internally consistent on this point.
+
+### 2.2 Performance
+
+None. This is a pre-load guard on the config-inspection path (a `serde_json` parse of `config.json`), not a hot inference path. No benchmark was required or run.
+
+### 2.3 Compatibility & Dependencies
+
+- **Breaking Changes**: an offline `--draft-model <dflash-drafter>` invocation that previously failed with a misleading `Weight not found` error now fails earlier with an explanatory one. No previously-working invocation is affected: no test or documented workflow constructed the offline DFlash round loop, since it does not exist.
+- **New Dependencies**: none in the fix itself. `tempfile` (already a dev-dependency, used elsewhere in this same PR) backs the corrected test fixtures.
+- **Compatibility**: the server path (`DFlashDrafter::load`, `resolve_drafter_kind`) never calls `get_model_type` for a drafter path, confirmed by re-checking every `get_model_type` call site in the tree; a new integration test pins that `SpeculativeDispatch::resolve` still returns the DFlash variant for the same fixture directory that detection now rejects.
+
+### 2.4 Code Quality
+
+- **Test Coverage**: new tests across four locations (`src/commands/generate_tests.rs`, `src/models/detection_tests.rs`, `src/lib/mlxcel-core/src/drafter/dflash/config.rs`, `tests/speculative_dispatch.rs`), each carrying an ordinary-full-model control that must keep resolving through the classic path. A mutation test (guarding the new detection arm as `if false && ...`) confirmed the new tests are real assertions rather than tautologies: both the `-m` and `--draft-model` rejection tests failed with the pre-fix classification.
+- **Code Complexity**: the structural probe is two small pure functions (`is_dflash_drafter_config`, `is_dflash_drafter_dir`) plus one shared error constructor; no control flow changes to any existing load path.
+- **Technical Debt**: decreased on both counts addressed here — the CI-relevant test hygiene inconsistency (M3) is closed, and the offline entry point no longer silently misroutes a class of checkpoint it was never built to run.
+
+---
+
+## 3. Technical Decisions
+
+### 3.1 Key on checkpoint structure, not on the resolved `DrafterKind`
+
+**Alternatives Considered:**
+
+| Option | Pros | Cons |
+|--------|------|------|
+| Reject when `DrafterKind::Dflash` is resolved | Minimal code, reuses existing resolution | Breaks the classic small-full-model drafter workflow, which also resolves to `Dflash` by default |
+| **Chosen: probe `config.json` for `dflash_config` and/or `architectures: ["DFlashDraftModel"]`** | Matches exactly what `DFlashConfig::from_json` and HuggingFace `AutoModel` dispatch each read; an ordinary full model has neither marker | Two independent markers to check instead of one resolved enum value |
+
+**Rationale:** the resolved kind is a fallback default, not a structural fact about the checkpoint. The checkpoint's own markers are what the DFlash loader and the HuggingFace ecosystem both already treat as authoritative, so keying on them cannot misclassify an ordinary drafter.
+
+### 3.2 Fix M3 by adopting `tempfile::tempdir()` rather than tightening the ad hoc path
+
+**Rationale:** the PR already establishes the `tempfile::tempdir()` pattern in `mlxcel-core`'s own new tests. Hardening the ad hoc `env::temp_dir()` path (unique suffix, permission checks, explicit cleanup on every exit path including panics) would have reimplemented what `tempfile` already provides as an RAII guard, at higher risk of missing an exit path than reusing the proven crate.
+
+---
+
+## 4. Implementation Details
+
+### 4.1 Detection (`src/models/detection.rs`, `src/lib/mlxcel-core/src/drafter/dflash/config.rs`)
+
+`get_model_type` now rejects a structurally-DFlash directory before any `model_type` dispatch, through a shared `dflash_drafter_not_standalone_error`, covering the `-m` case, server startup, and the distributed stage loaders from one call site.
+
+### 4.2 Offline CLI (`src/commands/generate.rs`)
+
+New `reject_dflash_drafter_offline`, called in `run_generation_mode` immediately before `load_model(draft_model_path)`. Fires with no `--draft-kind` and with `--draft-kind dflash`; `--draft-kind mtp` is a separate, already-handled request routed to `run_offline_mtp`.
+
+### 4.3 Test fixture hygiene fix (`src/commands/generate_tests.rs`, commit `37136aa6`)
+
+```rust
+// Before
+fn drafter_fixture_dir(name: &str, config: &str) -> PathBuf {
+    let dir = std::env::temp_dir().join(format!(
+        "mlxcel_generate_drafter_{name}_{}",
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_nanos()
+    ));
+    fs::create_dir_all(&dir).unwrap();
+    fs::write(dir.join("config.json"), config).unwrap();
+    dir
+}
+// ... each call site manually `fs::remove_dir_all(dir).unwrap()`s on the success path only
+
+// After
+fn drafter_fixture_dir(config: &str) -> tempfile::TempDir {
+    let dir = tempfile::tempdir().expect("temp dir");
+    fs::write(dir.path().join("config.json"), config).unwrap();
+    dir
+}
+// call sites use dir.path(); TempDir removes the directory on drop, on every exit path
+```
+
+The `name` parameter was dropped: it existed only to build a unique-ish path segment, which `tempfile::tempdir()` already guarantees by construction. The identical inline pattern in `offline_draft_model_check_defers_missing_or_broken_configs_to_the_loader`'s `empty` fixture was fixed the same way.
+
+---
+
+## 7. Change Summary
+
+### Statistics
+
+| Item | Value |
+|------|-------|
+| Files changed (feature commit `82096193`) | 9 |
+| Files changed (M3 fix commit `37136aa6`) | 1 |
+| Lines added / removed (feature) | +645 / -5 |
+| Lines added / removed (M3 fix) | +19 / -34 |
+| Tests added | 4 (`commands::generate`) + 3 (`models::detection`) + 3 (`drafter::dflash::config`, plus existing coverage extended) + 2 (`speculative_dispatch`) |
+
+### Changes by Category
+
+| Category | Count | Summary |
+|----------|-------|---------|
+| Detection / loading | 3 | `is_dflash_drafter_config`, `is_dflash_drafter_dir`, `get_model_type` rejection, shared error text |
+| CLI | 1 | `reject_dflash_drafter_offline` pre-load guard in `run_generation_mode` |
+| Docs | 2 | `docs/supported-models.md`, `docs/speculative-acceptance.md` corrected to state the offline limitation |
+| Test hygiene (this finalization pass) | 1 | `drafter_fixture_dir` and the inline `empty` fixture moved to `tempfile::tempdir()` |
+
+### Related Commits
+
+| Hash | Type | Message |
+|------|------|---------|
+| `82096193` | fix | reject a DFlash drafter before the offline standalone load |
+| `37136aa6` | fix | back drafter test fixtures with `tempfile::tempdir()` |
+
+---
+
+## 8. Follow-up Actions
+
+### Required
+
+- [ ] None; the security and review passes reported no CRITICAL or HIGH findings, and the one MEDIUM finding (M3) is fixed in this pass.
+
+### Future Improvements (recorded as known limitations, not fixed here)
+
+- The precheck fires only after the target model finishes loading, so an operator with a large target waits through the full load before seeing the error. Moving it earlier would also intercept `--draft-kind mtp`, which is deliberately routed elsewhere.
+- `--draft-model <dir>/config.json` (pointing at the config file rather than the directory) bypasses the precheck and falls through to the detection error with `-m` phrasing, because the precheck does not apply `resolve_model_dir`.
+- The probe uses raw `serde_json::from_slice`, while `get_model_type` applies `sanitize_config_json` first; a config with bare `NaN`/`Infinity` literals would fail the precheck open. No current checkpoint is affected, and closing this gap is a cross-crate change (the sanitizer lives in the root crate, the probe in `mlxcel-core`).
+- Building the actual offline `DFlashGenerator` round loop remains out of scope, as stated in the PR summary; the CLI directs an operator who needs this today at `mlxcel-server --draft-kind dflash`.
+
+---
+
+## Appendix
+
+### A. Test Results
+
+- `cargo test --release --workspace --features metal,accelerate --no-fail-fast`: every target passes except `-p mlxcel --lib`, which reports 5621 passed / 3 failed. All 3 failures are `multimodal::video::tests::*`, proven unrelated to this PR (0 lines touched in `src/multimodal/video.rs`; a PATH-controlled experiment on the same binary gives 34 passed / 0 failed with ffmpeg hidden versus 31 passed / 3 failed with ffmpeg 9.0.1 present; root cause is `src/multimodal/video.rs:1123` passing `-vsync`, which ffmpeg 9 removed). Filed separately as #1172.
+- The four `commands::generate::tests::offline_draft_model_*` precheck tests compile into `--bin mlxcel`, not `--lib`, so they were not part of the `--lib` run above. Run separately (`cargo test --release --bin mlxcel --features metal,accelerate commands::generate`): 82 passed, including all four.
+- `cargo test --release -p mlxcel-core --lib --features metal,accelerate drafter::dflash::config`: 16 passed.
+- `cargo test --release --test speculative_dispatch --features metal,accelerate`: 22 passed.
+- `cargo test --release --lib --features metal,accelerate models::detection`: 30 passed.
+- `cargo clippy --release --workspace --features metal,accelerate --tests -- -D warnings`: clean.
+- `cargo fmt --check`: clean.
+- Classic-path non-regression, executed against a real checkpoint: `-m qwen3-0.6b-4bit --draft-model qwen3-0.6b-4bit` runs `SpeculativeGenerator`, 24 tokens, acceptance_rate 1.0000, 137.54 tok/s.
+- Mutation test: guarding the new detection arm as `if false && ...` makes both the `-m` and `--draft-model` rejection tests fail with the pre-fix classification, on both markers.
+
+### B. References
+
+- Issue #1168 (specification)
+- Issue #1172 (pre-existing, unrelated ffmpeg 9 `-vsync` regression surfaced by the full workspace test run)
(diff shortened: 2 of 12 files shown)


inureyes:

### Intent

> Make `mlxcel generate` fail where it makes sense, with a message that names the real problem, when a DFlash drafter checkpoint is passed to `--draft-model` or `-m`, and fix the docs that overstated offline CLI support.

### Findings Addressed

None requiring a fix. No CRITICAL or HIGH findings. The four items below are MEDIUM/LOW and are reported rather than fixed, so the branch stays at the already-validated `82096193`.

### Remaining Items

- [ ] The `--draft-model` precheck runs after the target model has fully loaded (MEDIUM). Empirically the operator sees `Loading model from ... / Model loaded ... / Generating...` and the echoed prompt before the error, so with the real 27B target they pay the whole target load first. `run_generate_once` has a pre-load validation block (`src/commands/generate.rs:1982-1990`) that would be the natural home. Not a one-liner: an early call would also intercept `--draft-kind mtp`, which is deliberately routed to `run_offline_mtp`'s own targeted errors, so it needs the parsed explicit kind threaded up.
- [ ] `reject_dflash_drafter_offline` does not apply `resolve_model_dir` (LOW). `--draft-model <dflash-dir>/model.safetensors` skips the CLI-specific message and falls through to the detection-level error, which phrases the remedy as "Pass a full model to -m" even though the user passed `--draft-model`. `resolve_model_dir` is private at `src/loading/mod.rs:75`, so this is not a one-liner. Marginal, since `--draft-model` takes a directory.
- [ ] `is_dflash_drafter_config` treats an empty `"dflash_config": {}` as a marker (LOW). Upstream `mlx-vlm`'s `_has_config` excludes `{}` while its `get_model_and_args` does not, so upstream is internally inconsistent and no published checkpoint ships an empty block. Informational only.
- [ ] The probe's doc comment argues the marker choice from first principles but does not cite that it matches upstream (LOW). `https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/utils.py` uses `is_dflash = config.get("dflash_config", None) is not None`, and `mlx-vlm` `docs/usage.md` states "DFlash checkpoints are detected from `dflash_config`". Adding that URL would follow the "Referencing upstream source" convention. Not applied because a comment-only edit in `mlxcel-core` invalidates the whole downstream build cache and would stale the completed full-suite validation.

Also LOW, style only: the `src/commands/generate_tests.rs` fixture helper uses `std::env::temp_dir()` plus a manual `fs::remove_dir_all` after the assertions, so a failing assertion leaks the directory. It matches an existing pattern in the same file, but `tempfile::tempdir()` is the dominant convention in `src/**_tests.rs` and is what this PR's own sibling fixtures in `config.rs` and `tests/speculative_dispatch.rs` use.

### Blast radius of the detection change

All 24 non-test `get_model_type` call sites take a target or model path, never a drafter path: `src/loading/mod.rs:528,537,689,706`, `src/distributed/pipeline/stage_executor/mod.rs:346`, `src/distributed/tensor_parallel/inference.rs:83`, `src/server/chat_template.rs:1338`, `src/server/startup.rs:733,767,1076,1696,2149`, `src/server/model_provider.rs:131`, `src/server/batch/xla_worker.rs:250`, `src/multimodal/host_preprocessor.rs:170`, `src/commands/generate.rs:564,797,1264`. A grep of `draft_model_path` across the root crate finds no detection call on a drafter path. Three of the server-startup uses swallow `Err` (`.ok()?`, `is_ok_and`, `match { Err => default }`), so even a hypothetical false positive degrades to a feature-off default rather than a failure.

The separation is structural, not just current practice: `get_model_type` lives in the root `mlxcel` crate, and `mlxcel-core`, which owns `load_drafter`, `resolve_drafter_kind`, and `DFlashDrafter::load`, cannot call it. The server chain was re-read end to end: `WorkerDrafterSlot::ensure_loaded` (`src/server/batch/speculative_burst.rs:270`) to `load_drafter` (`src/lib/mlxcel-core/src/drafter/mod.rs:810`) to `resolve_drafter_kind` (`:415`, own `peek_drafter_model_type` reader) to `DFlashDrafter::load` (`src/lib/mlxcel-core/src/drafter/dflash/drafter.rs:98`, reads `config.json` straight into `DFlashConfig::from_json`). The strongest confirmation is `tests/speculative_parity.rs`, the `#[ignore]`-gated real-model DFlash test: it calls `load_model` on the target only and `resolve_drafter_kind` / `load_drafter` on the drafter (lines 632, 659, 669), so the one test that drives the whole DFlash pipeline in process never routes the drafter through detection.

False-positive surface: every `config.json` under `models/` on the development box was re-scanned (11 directories). Only `qwen3.5-27b-dflash` carries a marker. The MTP drafter `qwen3.8-27b-mtp-bf16` has no `architectures` key and no `dflash_config`, so the offline MTP path is untouched. Nothing in the tree writes or reads `dflash_config` outside the DFlash config parser. `dflash_config` is also upstream `mlx-vlm`'s own discriminator, so a standalone model carrying it would misbehave upstream too. No script or CI job keys on the old `Weight not found` text.

### Verification performed

Beyond the queued suites, all on `82096193` with `DEVELOPER_DIR` set and `--release --features metal,accelerate`:

**Mutation test (are the new tests real?).** Guarding the detection arm as `if false && is_dflash_drafter_config(&v)` and rebuilding makes both `dflash_drafter_is_rejected_as_a_standalone_model` and `dflash_drafter_is_rejected_on_either_marker_alone` fail with `a DFlash drafter is not a standalone model: Qwen3`, which is exactly the pre-fix classification. Reverted with `git checkout HEAD -- src/models/detection.rs`. Coverage runs both ways: a probe stuck at true is caught by four ordinary-full-model controls, a probe stuck at false by five rejection tests. The controls are not vacuous, `ordinary_qwen3_full_model_still_detects_as_qwen3` asserts the concrete `ModelType::Qwen3`, and `ordinary_full_model_is_not_mistaken_for_a_dflash_drafter` asserts both `resolve_drafter_kind == Dflash` and `!is_dflash_drafter_dir`, which is precisely the trap the structural discriminator exists to avoid.

**End-to-end against the real published checkpoint** (`models/qwen3.5-27b-dflash`, 58 tensors), using `models/qwen3-0.6b-4bit` as a cheap target so the drafter guard is reached without a 27B load:

| Case | Result |
|------|--------|
| `-m <dflash-dir>` | New detection error, no `Weight not found` |
| `--draft-model <dflash-dir>`, no `--draft-kind` | New CLI precheck error, no `Weight not found` |
| `--draft-model <dflash-dir> --draft-kind dflash` | Same new CLI precheck error |
| `--draft-model <dflash-dir> --draft-kind mtp` | Clean MTP-capability error from `run_offline_mtp`, no `Weight not found` |
| `--draft-model <ordinary 0.6B>`, no `--draft-kind` | Classic path intact: drafter loads, `Resolved drafter kind: dflash (block_size = 16, default)`, `SpeculativeGenerator` runs, 24 tokens, `acceptance_rate=1.0000`, 137.54 tok/s |
| `--draft-model <ordinary 0.6B> --draft-kind dflash` | Pre-existing `generate.rs:1482` error preserved, so the `speculative-acceptance.md:274-276` claim stays true |

**Targeted suites, all green:** `models::detection` 30 passed, `drafter::dflash::config` 16 passed (6 new), `commands::generate::tests::offline_draft` 4 passed, `--test speculative_dispatch` 22 passed (5 new).

**Lints:** `cargo fmt --check` clean. `cargo clippy --release -p mlxcel-core --lib --tests -- -D warnings` clean. `cargo clippy --release -p mlxcel --lib --bins --tests -- -D warnings` clean. `python3 scripts/ci/check_cross_repo_refs.py` clean, every `#1168` reference is same-repo. That closes the two boxes the PR description left unchecked.

Note for the record: `commands::generate::tests::*` compile into the `--bin mlxcel` target, not `--lib`, so a `-p mlxcel --lib` run does not include the four new `--draft-model` precheck tests. They were run separately.

**Not measured:** `--draft-kind mtp` against a genuinely Gemma 4 MTP-capable target combined with a DFlash drafter directory. That combination bypasses the new precheck by design and lands in `load_drafter(.., Some(Mtp))`. No Gemma 4 target is on this box. On the Qwen target that is available, the equivalent case errors cleanly at the MTP-capability gate.

### Verification

- [x] Tests pass

inureyes:

**No CRITICAL and no HIGH findings. Nothing auto-fixed, no commits pushed.** Three MEDIUM and three LOW items below, all advisory.

### 1. Path handling in `is_dflash_drafter_dir`

No panic path. `std::fs::read` returns `Err` on any I/O failure; `serde_json::from_slice` returns `Err` on invalid UTF-8 content and on malformed JSON rather than panicking; serde_json 1.0.151 is pinned with neither the `unbounded_depth` feature enabled nor `disable_recursion_limit()` called anywhere in the tree, so the default 128-frame nesting limit holds and an adversarially nested config is an `Err`, not a stack overflow. `Value::get`, `as_array`, and `as_str` are all non-panicking. Verified by grep across the workspace and `Cargo.lock`.

Symlink following is required behavior here, not a defect: on the development box `models/qwen3.5-27b-dflash` is itself a symlink, so a probe that refused to traverse links would break the operator's own layout. The path is an operator-supplied CLI argument, not a value crossing a trust boundary.

TOCTOU against the later real load exists but buys nothing. Winning the race only downgrades the message back to the pre-PR one, and `load_model` re-reads `config.json` through `get_model_type`, which now carries the same structural rejection. The two guards cover each other.

### 2. Information disclosure in the two error messages

Both echo `path.display()`, which matches the established convention in this tree (`src/models/gpt_oss.rs:1182`, `src/models/gemma3.rs:951`, `src/loading/vlm_muse_glimmer.rs:258`, `src/models/sanitize.rs:1195`, `src/multimodal/host_preprocessor.rs:170`). The path is the operator's own argument, and both messages are otherwise free of tensor names, environment values, and internal identifiers.

Neither message reaches an HTTP response body. The two server-side `get_model_type` sites that could carry an error toward a client discard it instead: `src/server/model_provider.rs:131` uses `is_ok_and`, and `src/server/chat_template.rs:1338` uses `.ok()?`. The remaining server sites fail at startup into logs.

### 3. Denial of service and resource use

The `get_model_type` arm adds **zero** I/O. It runs on the `Value` that function already parsed, so the cost is one map lookup plus a scan of a one-element `architectures` array.

`get_model_type` is not on any per-request path. All 18 call sites were traced to their callers: every one is startup, `ModelProvider`/worker construction, or model load. The three whose results feed request-time decisions are memoized, notably `ModelProvider.single_stream_queue_admission: Arc<AtomicBool>` (constructed at `model_provider.rs:432`, `:746`, `:850`; read per request at `:944`), the `AppState` chat template `Arc<ChatTemplateProcessor>`, and `XlaServeWorker.phi4mm_media_preprocessor`. One near-miss worth a comment guard rather than a change: inlining `uses_single_stream_queue_admission` into `model_provider.rs:944` would turn it into a `config.json` read per request.

The only added I/O is the single `fs::read` in `reject_dflash_drafter_offline`, once per CLI process, immediately before a load that reads the same file anyway.

### 4. Correctness of the failure-open choice

**Confirmed not a bypass, and the right call.** The guard is a usability guard, not a trust boundary. It grants no capability, and no privilege, authentication, or authorization decision sits downstream of it. Failing open falls through to exactly the code path that ran before this PR, and that path now contains the detection-level rejection, so the worst an attacker gains by forcing `false` is the pre-PR error message on their own local invocation.

---

## MEDIUM

**M1. The precheck does not normalize a file path to its directory.** `reject_dflash_drafter_offline` joins `config.json` onto the raw `--draft-model` value, but `loading::load_model` calls `resolve_model_dir` (`src/loading/mod.rs:75-81`) first, which maps a file to its parent. So `--draft-model models/qwen3.5-27b-dflash/config.json` skips the precheck entirely. Contained, not broken: `load_model` then hits the `get_model_type` rejection, so the user still gets a named error. It is the `-m` phrasing though, which does not name `mlxcel-server --draft-kind dflash`. Mirroring `resolve_model_dir` in the probe, or accepting a path that already points at `config.json`, closes it.

**M2. Parse divergence between the two entry points.** `is_dflash_drafter_dir` calls `serde_json::from_slice` on raw bytes, while `get_model_type` runs `sanitize_config_json` first, which rewrites bare `NaN`, `Infinity`, and `-Infinity` literals (`src/models/sanitize.rs:1923-1928`). A drafter config carrying those literals, which is the reason that sanitizer exists, makes the CLI precheck fail open. Same containment as M1: detection still catches it. I confirmed the published `qwen3.5-27b-dflash` config carries none of those literals, so no current checkpoint is affected. Note the layering constraint before fixing: `sanitize_config_json` lives in the root `mlxcel` crate and the probe lives in `mlxcel-core`, so reuse means moving the three-line sanitizer down rather than adding a dependency. A doc-comment note on the divergence is an acceptable alternative.

**M3. Test temp-directory hygiene in `src/commands/generate_tests.rs`.** `drafter_fixture_dir` and the inline `empty` directory build a predictable name under `std::env::temp_dir()` and create it with `create_dir_all`, which succeeds on a pre-existing directory or a symlink to one, then clean up only on the success path. On Linux CI, where `TMPDIR` is the world-writable `/tmp`, that is the CWE-377/CWE-379 pattern: an attacker who pre-creates the name as a symlink gets `config.json` written through it and the target's contents removed by `remove_dir_all`. On macOS `env::temp_dir()` is the per-user `/var/folders/.../T`, so the practical exposure is CI only. Separately, an assertion failure leaks the directory, and the two-case loop in `offline_draft_model_rejects_a_dflash_drafter_on_either_marker_alone` leaks on the first failure. This matches a pre-existing pattern in the same file and in `detection_tests.rs::temp_path`, but `tempfile` is already a dev-dependency and the `mlxcel-core` tests added by this same PR use `tempfile::tempdir()`, so the PR is internally inconsistent. Test-only, no production impact.

## LOW

**L1.** The CLI now reads the drafter's `config.json` three times per invocation: `resolve_drafter_kind` via `peek_drafter_model_type`, then `is_dflash_drafter_dir`, then `load_model` via `get_model_type`. Microseconds once per process, so no action needed. Noted only because having `peek_drafter_model_type` hand back the parsed `Value` would collapse two of them.

**L2.** `fs::read` is unbounded. Consistent with `get_model_type`'s `read_to_string` on the same file, and the input is operator-supplied, so this is not new exposure. A bounded read would still be strictly better for a probe whose input is always a few KB.

**L3.** `--draft-kind mtp` pointed at a DFlash drafter directory still reaches `Gemma4AssistantDraftModel::from_path` and surfaces a weight-level error. The PR scopes this out deliberately and says so. Flagged only so it reads as a known remainder rather than an oversight.

## What the review confirmed as correct

Keying on checkpoint structure rather than the resolved `DrafterKind` is the right discriminator, and it is pinned by an ordinary-full-model control at all four test levels. Scanning every `config.json` under `models/` reproduced the PR's claim independently: only `qwen3.5-27b-dflash` carries a marker, and the MTP target (`model_type: qwen3_5_mtp`, no `architectures`) and both Qwen 3.5/3.8 targets are untouched. The two guards form a genuine belt and braces, which is what makes the failure-open choice in M1 and M2 safe rather than merely tolerable.

inureyes:
## Orchestrator verification against real checkpoints

Run on the development box at `27654b29` with freshly built release binaries, `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer`.

### The reported symptom is gone, in all three entry shapes

All three now produce a named error, and `Weight not found: model.embed_tokens.weight` appears in none of them.

| invocation | result |
|---|---|
| `generate -m qwen3-0.6b-4bit --draft-model qwen3.5-27b-dflash` | named DFlash-drafter error naming `mlxcel-server` as the way to use it |
| same plus `--draft-kind dflash` | same named error |
| `generate -m qwen3.5-27b-dflash` | named "not a standalone model" error |

### The server path is not regressed, measured rather than reasoned

This was the highest-risk item, since the fix adds a rejection inside `get_model_type`. Same 27B pair, after the change:

```
INFO Lazy-loading drafter from models/qwen3.5-27b-dflash (kind=Some(Dflash))
INFO Drafter loaded (kind=dflash, 1 ms)
```

An 80-token completion drove real speculative work: rounds 15, proposed_tokens 185, accepted_tokens 65, acceptance_rate 0.351, emitted_per_verify 5.27. Comparable to the pre-change reference recorded in #1168 (24 rounds, 0.274 acceptance on a 120-token completion); the two differ in completion length, not in behavior.

### Suite

`cargo test --release --workspace --features metal,accelerate --no-fail-fast`: every target passes except `-p mlxcel --lib` at 5621 passed / 3 failed.

All 3 failures are `multimodal::video::tests::*` and are **not attributable to this PR**. Two independent lines of evidence:

1. This PR touches **0 lines** of `src/multimodal/video.rs`.
2. A PATH-controlled experiment on the same compiled binary, same commit, varying only whether `ffmpeg` is visible:

| arm | `ffmpeg` on PATH | result |
|---|---|---|
| A | hidden | **34 passed, 0 failed** |
| B | ffmpeg 9.0.1 | **31 passed, 3 failed** |

Root cause is `src/multimodal/video.rs:1123` passing `-vsync`, which ffmpeg 9 removed. Filed as #1172. Worth noting the second-order finding recorded there: those tests **pass by silently skipping** when ffmpeg is absent, which is the same shape of gap this PR closes on the CLI side, where no test invoked `mlxcel generate --draft-model` at all.

`cargo test --release --bin mlxcel --features metal,accelerate commands::generate`: 82 passed, 0 failed, after the `tempfile::tempdir()` change.

PR #1171, opened 2026-08-15
docs: correct apc_consistent_prefix_len doc claim that matched_len is preserved

### LINKED ISSUE #1160: docs: correct apc_consistent_prefix_len doc claim that matched_len is preserved
## Problem / Background

The doc comment on `apc_consistent_prefix_len` (`src/server/prompt_cache/apc_lookup.rs:46`) claims: "If every covered block agrees, the input `matched_len` is preserved." This is false. The function always returns `consistent_blocks * block_size` (`src/server/prompt_cache/apc_lookup.rs:82`), i.e. a block-floored length. A non-block-aligned `matched_len` is never returned as-is even when every covered block agrees; the tail past the last full block boundary is dropped.

The unit test that appears to pin the documented behavior, `matching_chains_preserve_matched_len` (`src/server/prompt_cache/apc_lookup_tests.rs:27-43`), only passes because it uses an already block-aligned input (`matched_len = 64` with `BLOCK = 16`), so it cannot distinguish "preserved" from "floored to block granularity".

## Proposed Solution

Doc and test change only; no behavior change.

- Correct the doc comment to state that the return value is always block-floored: `min(consistent_blocks, covered blocks) * block_size`, never the raw input `matched_len`.
- Strengthen (or add) a unit test with a non-aligned `matched_len`, e.g. `matched_len = 73` with block size 16 and fully agreeing chains, expecting 64. This pins the actual flooring behavior so the doc claim stays honest.

## Acceptance Criteria

- [ ] Doc comment on `apc_consistent_prefix_len` describes the always-floored return and no longer claims `matched_len` is preserved.
- [ ] A unit test exercises a non-block-aligned `matched_len` with fully agreeing chains and asserts the block-floored result.
- [ ] No production behavior change; existing prompt-cache tests remain green.

## Technical Considerations

Surfaced by the review comment on PR #1158 (implementation of #1156), which documented this finding and noted that the flooring behavior is what the end-to-end slack in `tests/prompt_cache_e2e.rs` now depends on. The flooring itself is correct and intentional (block-granular verification of the Merkle-DAG chain); only the documentation misstates it, so the fix must not change the return value.

inureyes:
## Summary

Fixes a false doc claim on `apc_consistent_prefix_len` (`src/server/prompt_cache/apc_lookup.rs`) that said the input `matched_len` is preserved when every covered block agrees. It never is: the function always returns a block-floored length, `consistent_blocks * block_size`. This is a doc-and-test-only change; the return value, bounding logic, and function signature are unchanged.

## What changed

- Corrected the doc comment on `apc_consistent_prefix_len` to state the return is always a multiple of `block_size`, explain how `consistent_blocks` is capped (by `floor(matched_len / block_size)`, by the candidate chain's length, and by the request's own recomputed chain length), and note explicitly that the flooring is deliberate so it is not mistaken for a bug later.
- Renamed the existing test `matching_chains_preserve_matched_len` to `aligned_matching_chains_return_the_block_floored_len` because its old name restated the false claim; its `matched_len` of 64 against a block size of 16 is already block-aligned, so it could never distinguish "preserved" from "floored". The test body and assertions are unchanged, only the name and its comment.
- Added `non_aligned_matched_len_floors_to_covered_blocks`, a new unit test in `src/server/prompt_cache/apc_lookup_tests.rs` that uses 73 tokens against a 16-token block size with fully agreeing chains. The candidate chain has `ceil(73/16) = 5` hashes, the 5th being a 9-token partial trailing block, but `coverable_blocks = 73/16 = 4` caps the comparison before that partial block is ever reached, so the result floors to `4 * 16 = 64`, not 73. The test asserts `64` explicitly and also asserts the result is not equal to the raw token count, so it genuinely pins the flooring behavior instead of restating the input.

## Test plan

- [x] `cargo test --lib server::prompt_cache::apc_lookup` (10/10 passing, including the new and renamed tests)
- [x] `cargo fmt -- --check` on the two touched files

Closes #1160

### DIFF
--- a/TECHNICAL_REPORTS/1171-apc-prefix-len-doc-20260816.en.md
+++ b/TECHNICAL_REPORTS/1171-apc-prefix-len-doc-20260816.en.md
@@ -1,0 +1,204 @@
+# Technical Report: PR #1171 - docs: correct apc_consistent_prefix_len doc claim that matched_len is preserved
+
+**Date**: 2026-08-16
+**Author**: AI Code Reviewer
+**Status**: Completed
+**Languages**: Rust
+**Risk Level**: Low
+
+---
+
+## Executive Summary
+
+The doc comment on `apc_consistent_prefix_len` (`src/server/prompt_cache/apc_lookup.rs`) claimed that "if every covered block agrees, the input `matched_len` is preserved". That was false: the function always returns `consistent_blocks * block_size`, so a non-block-aligned `matched_len` loses its tail past the last full block boundary. The unit test that appeared to pin the documented behavior could not detect the difference, because its `matched_len` of 64 was already block-aligned to a block size of 16. This PR corrects the doc, renames the misleading test, and adds a test with a non-aligned `matched_len` that genuinely distinguishes flooring from preservation.
+
+No production behavior changed. Every changed line in `apc_lookup.rs` begins with `///`, verified mechanically, which matters because `tests/prompt_cache_e2e.rs` now derives its cache-slack assertion from exactly this flooring.
+
+---
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+The defect was recorded as a follow-up in the technical report for PR #1158 (implementation of #1156). That PR fixed a deterministically failing e2e assertion by deriving its slack from `DEFAULT_APC_BLOCK_SIZE`, and in doing so had to reason carefully about the flooring. The false doc line was noticed then, flagged on the PR as a pre-existing production doc bug, and left out of scope.
+
+### 1.2 Existing Issues
+
+- **Issue 1**: The doc claimed `matched_len` is preserved on full agreement. The function has exactly two kinds of return, three early exits returning literal `0` and the final `consistent_blocks * block_size`. No path returns `matched_len`.
+- **Issue 2**: `matching_chains_preserve_matched_len` asserted the false claim in its very name, and its input (`matched_len = 64`, `BLOCK = 16`) was already block-aligned, so it passed identically under either contract and could never catch the discrepancy.
+- **Issue 3** (found during review): the first replacement wording, "never the raw input `matched_len`", was itself imprecise. With an already block-aligned `matched_len` and full agreement, the return does equal `matched_len` numerically. It is computed rather than passed through, and for a PR whose entire purpose is to kill that confusion, the distinction had to be explicit.
+
+### 1.3 Risk Assessment
+
+| Risk | Impact | Likelihood |
+|------|--------|------------|
+| A future implementer "fixes" the flooring to match the doc, breaking the APC safety property and the e2e slack derivation | Medium | Low |
+| A reader trusts the doc when reasoning about cache credit and miscomputes expected `cached_tokens` | Low | Medium |
+
+---
+
+## 2. Technical Review
+
+### 2.1 Security
+
+No security surface is added: the diff is doc comments plus tests, with the function body untouched. The security pass returned zero findings at every severity and confirmed the invariant that actually matters here, that a candidate is never adopted past a verified divergence. The described behavior only ever removes tokens, since the result is bounded above by `matched_len`, and the corrected text explicitly warns against turning it into a pass-through. The dense branch of `try_adopt_cached_prefix` truncates to exactly the APC-floored `matched_len` and the paged branch floors further to the pool block size, so no path adopts KV past a verified boundary.
+
+Unbounded-work was examined as the one place a genuine finding could plausibly live and dismissed on evidence: `matched_len` is attacker-influenced only through the prompt, and it is bounded by `tokens.len()` at both selection tiers (`match_depth.min(dl.token_len)` in `lookup.rs`, and `common_prefix_len` in `select_best_by_scan`), while `cap_tokens` is additionally min'd with `request_tokens.len()` so the slice cannot panic. The work is one linear hash pass over at most `matched_len` tokens. This is pre-existing and untouched.
+
+**Issues Found:**
+| Issue | Severity | Status |
+|-------|----------|--------|
+| Replacement wording "never the raw input `matched_len`" imprecise for the block-aligned case | Low | Fixed (`daf4a69a`) |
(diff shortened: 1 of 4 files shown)


inureyes:

### Intent

> Replace the false doc claim that `apc_consistent_prefix_len` preserves the input `matched_len` with an accurate description of the always block-floored return, and pin the real behavior with a test that can actually tell "floored" from "preserved". Doc and test only, no behavior change.

### Verification performed

Every clause of the new doc paragraph was re-derived independently against the function body (`apc_lookup.rs:57-91`) and found true: the result is `consistent_blocks * block_size` on the main path and a literal `0` on all three early-return paths, so it is always a multiple of `block_size`; `consistent_blocks` is bounded by `coverable_blocks = matched_len / block_size` (integer division, so `floor`), by `.min(candidate_hashes.len())`, and by `upper = coverable_blocks.min(request_chain.hashes.len())`; and the `take_while` count can only shrink it further. The two surviving pre-existing lines are still correct: the Merkle-DAG rationale matches `parent = next` chaining in `BlockHashChain::compute`, and the short-circuit line matches `matched_len < block_size` returning `0`.

The new test's arithmetic was re-derived from scratch and 64 is right: `73.div_ceil(16) = 5` candidate hashes, `coverable_blocks = 73 / 16 = 4` then `.min(5) = 4`, `cap_tokens = min(64, 73) = 64`, the request chain over `tokens[..64]` has exactly 4 hashes, `upper = 4`, all 4 agree, so `4 * 16 = 64`. The 5th, 9-token partial candidate block is never reached, exactly as the test comment says. Because a preserving implementation would return 73, both `assert_eq!(consistent, 64)` and `assert_ne!(consistent, tokens.len())` genuinely fail under the false contract, so the test distinguishes the two cases rather than restating its input.

No behavior change was confirmed mechanically: every added and removed line in `apc_lookup.rs` starts with `///`, verified by filtering the diff for any changed line that is not a doc comment (zero results). This matters because `tests/prompt_cache_e2e.rs:384-392` derives its under-allowance slack from precisely this flooring.

The old test name `matching_chains_preserve_matched_len` has zero remaining references anywhere in the tree, and the cross-reference inside the renamed test points at the new test's exact name.

### Findings Addressed

- [x] The phrase "never the raw input `matched_len`" was open to a reading under which it is itself false (LOW). With an already block-aligned `matched_len` and full agreement, the return does equal `matched_len` numerically, it is simply computed rather than passed through, which is exactly the case the renamed 64-token test covers. Since this PR exists to remove that specific confusion, the wording is now "never a pass-through of the raw input `matched_len`" plus an explicit sentence that an aligned `matched_len` comes back equal only because the arithmetic lines up.
- [x] The `matched_len < block_size` short-circuit sentence ran on directly from the new nine-line paragraph with no `///` separator, so rustdoc rendered them as one merged block (LOW). It is now its own paragraph.

Both fixes are doc-comment only, pushed as `daf4a69a`. rustfmt does not reflow doc comments under the default config and this repo has no `rustfmt.toml`, so formatting is unaffected.

### Remaining Items

None. The pre-existing em dashes elsewhere in `apc_lookup.rs` (lines 15, 24, 30, 43, 66, 98, 105) are untouched context and a repo-wide condition (371 files under `src/`); cleaning them here would push the diff outside the issue's scope. No line added by this PR contains one.

### Verification

- [x] Integrated into project code flow (the test module is wired via `#[cfg(test)] #[path = "apc_lookup_tests.rs"] mod tests;` at `apc_lookup.rs:117-119`)
- [x] Project conventions followed (no AI attribution and no em dashes in any commit message, PR body, or added line)
- [x] Existing modules reused where applicable (the new test reuses `BLOCK`, `empty_extra()`, and `BlockHashChain::compute` exactly as the sibling tests do)
- [x] No unintended structural changes (doc-only in `apc_lookup.rs`, additive plus one rename in the test file, nothing else touched)
- [x] Tests pass (`cargo test --lib server::prompt_cache::apc_lookup` 10 passed 0 failed, `cargo fmt --check` clean)

inureyes:

No findings. Nothing to fix, no commits pushed.

The diff is doc-comment-only in `src/server/prompt_cache/apc_lookup.rs` (every changed line is `///`, function body untouched) plus one test rename and one new test. No behavior, API, or dependency change, so there is no new attack or performance surface.

Checks that were actually load-bearing here:

- **Doc accuracy vs the body.** Every clause of the new text holds: the only returns are `0` and `consistent_blocks * block_size`, and the three caps named in the doc are real (`matched_len / block_size` at the `coverable_blocks` binding, `.min(candidate_hashes.len())`, and `coverable_blocks.min(request_chain.hashes.len())`). Nothing returns `matched_len`.
- **Same flooring the e2e relies on.** `tests/prompt_cache_e2e.rs` derives its slack from `DEFAULT_APC_BLOCK_SIZE` and documents that this is correct because `--batch-size 1` resolves to the Dense adopt path, which truncates to exactly the APC-floored `matched_len`. That is the flooring the new doc describes.
- **Safety invariant is reinforced, not weakened.** The new wording only ever removes tokens (result is bounded above by `matched_len`), and it explicitly warns against "fixing" the flooring to pass `matched_len` through. Downstream, the dense adopt truncates to `matched_len` and the paged adopt floors further to the pool block size, so no path adopts KV past a verified block boundary.
- **The new test can fail.** With 73 tokens and `BLOCK = 16`, `assert_eq!(consistent, 64)` fails at 73 if the flooring were replaced by a pass-through, and fails at 80 if `coverable_blocks` were changed from floor to ceil. The ceil variant is the dangerous one: it would compare the candidate's 9-token partial trailing block as if it were full and credit 16 tokens for it. The test pins that boundary.
- **No DoS surface.** `matched_len` is attacker-influenced only through the prompt, and it is bounded by `tokens.len()` at both selection tiers (`match_depth.min(dl.token_len)` in the trie tier, `common_prefix_len` in the scan tier). The recompute is one linear hash pass over at most `matched_len` tokens with a `Vec` of `matched_len / block_size` 32-byte digests, and `request_tokens[..cap_tokens]` cannot panic because `cap_tokens` is min'd with `request_tokens.len()`. Unbounded work is not reachable.

Verdict: approve.

PR #1173, opened 2026-08-15
fix(test): skip partial Muse Glimmer checkpoints instead of panicking

### LINKED ISSUE #1161: test: muse_glimmer fusion skip guard panics on a partially materialized checkpoint
## Problem / Background

`pinned_post_tower_weight_roots_and_shapes_match_published_contract` (`src/vision/encoders/muse_glimmer_fusion_tests.rs:203`) skips gracefully when `models/mlx/muse-glimmer-30b/model.safetensors.index.json` is absent (guard at lines 206-213, added in PR #1157). But when the index IS present, the test proceeds to `unwrap()` everything else it reads: `config.json` (lines 215-217), the index contents (lines 218-219), and every referenced safetensors shard via `safetensors_shape` (lines 252-254, with further unwraps at lines 262-268). A partially materialized checkpoint, index present but config or shards missing (for example an interrupted `hf download` or `mlxcel download`), panics with an opaque unwrap message instead of skipping with a reason.

The sibling loading test does not share this gap: `pinned_weight_index_classifies_each_source_weight_once` (`src/loading/vlm_muse_glimmer_tests.rs:473`) calls `read_muse_weight_inventory_from_index` (`src/loading/vlm_muse_glimmer.rs:253`), which reads only the index file itself, so index-present-shards-missing cannot trip it.

## Proposed Solution

Either option is acceptable; the implementer chooses:

1. Extend the skip guard to check every file the test will read (config.json plus each shard named in the index's `weight_map` for the pinned keys) before proceeding, skipping with a reason naming the missing file.
2. Convert the unwraps into skip-with-reason: on any read/parse failure, print a skip message naming the file and return, mirroring the existing index-absent guard.

Optional nice-to-have (a separate checkbox, not required for closing): reviewers on PR #1157 suggested an opt-in `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` env var that turns silent skips into hard failures on the machine that owns the pinned checkpoint, so a corrupted checkpoint cannot quietly disable the contract test forever there.

## Acceptance Criteria

- [x] With the index present but `config.json` or any referenced shard missing, the test skips with a message naming what is missing instead of panicking.
- [x] With the full pinned checkpoint present, the test still validates the published weight-root and shape contract unchanged.
- [x] With the index absent, the existing skip behavior is unchanged.
- [x] (Optional) `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` turns skips into failures for checkpoint-owning machines.

## Technical Considerations

Surfaced by the security review LOW findings on PR #1157 (implementation of #1155), which added the index-absent guard but deliberately left the partial-checkpoint case out of scope. Test-only change; no production loading path is affected. If option 1 is chosen, note that reading the index to enumerate shards for the pre-check already requires parsing it, so the guard and the assertion body should share that parse rather than reading the index twice.

---

## Resolution note (added 2026-08-16, after merge)

Resolved by PR #1173, squash-merged as `0adb53c5`. The problem statement above is preserved as filed; this note only reconciles it with the current tree so the references still lead somewhere.

**The code moved.** Every `src/vision/encoders/muse_glimmer_fusion_tests.rs` line reference above describes the pre-fix layout and no longer resolves. The pinned contract check and its `safetensors_shape` helper now live in `src/vision/encoders/muse_glimmer_fusion_pinned_tests.rs`, attached as a second `#[cfg(test)]` module from `src/vision/encoders/muse_glimmer_fusion.rs`. `muse_glimmer_fusion_tests.rs` retains only the fusion-math tests.

**What was implemented.** The hybrid of the two proposed options: `load_pinned_checkpoint` parses the index once and shares it with the assertion body, checks that `config.json` and every shard named for the pinned keys is present and readable, and skips with a message naming the offending file. `safetensors_shape` returns a `Result`, so a truncated or corrupt header also skips rather than panicking, and it bounds the declared header length against both a 256 MiB absolute ceiling and the file size before allocating.

**The correctness boundary.** Only availability and readability problems skip. Contract violations still fail, and that is proven executably by four `#[should_panic]` tests over complete-but-wrong synthetic checkpoints rather than by inspection.

**Acceptance criteria.** All four are met, including the optional `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` gate. Criterion 2 was proven with that gate: because it converts every skip into a hard failure, the pinned test passing under it is positive evidence that the real checkpoint was read and asserted rather than skipped.

**Follow-ups.** #1177 extends the env gate to the loading-side sibling guard named above, which does not yet honor it, and documents the variable in `docs/environment-variables.md`. #1178 covers making the repository's file-size guidance discoverable from contributor-facing docs.

**One caveat for anyone reviewing from the commit.** The squash commit body on `main` justifies the file split with a "500-line limit" that is not this repository's convention; the actual guideline is 800 lines. PR #1173's body has been corrected, but commit messages cannot be edited after merge, so the commit retains the original wording. Where they disagree, the PR body is accurate.

inureyes:
> **Post-merge correction (2026-08-16).** This body was corrected after merge. As originally written it justified the file split with a "500-line limit" that is not this repository's convention, and its test counts and run results described only the first of the four commits on this branch. The squash commit on `main` (`0adb53c5`) still carries the original wording, since commit messages cannot be edited after merge; where the two disagree, this body is the accurate one. The threshold question is tracked in #1178, and the incorrect rationale is also recorded in `TECHNICAL_REPORTS/1173-pinned-checkpoint-skip-guard-20260816.en.md`.

## Summary

`pinned_post_tower_weight_roots_and_shapes_match_published_contract` skipped gracefully only when `model.safetensors.index.json` was absent. With the index present but anything else it reads missing or corrupt, the shape an interrupted `hf download` or `mlxcel download` leaves behind, it unwrapped its way to an opaque panic instead of skipping with a reason. This adds an availability pre-check that names the offending file, while leaving every contract assertion able to fail exactly as before.

## Implementation choice

The issue offered two options; this is the hybrid the brief recommended. Option 1 alone (a pure existence pre-check) cannot detect a shard that exists but whose header was truncated mid-download, and option 2 alone (converting each unwrap into a skip) reads the index twice, which the issue explicitly asked to avoid. The hybrid gets both: `load_pinned_checkpoint` parses the index once and hands the parsed `weight_map` to the assertion body, and `safetensors_shape` returns a `Result` so header corruption is covered too.

## How the correctness boundary is kept

The skip is deliberately narrow, and the split is enforced by construction rather than by convention.

- Only **availability and readability** may skip: an absent file, an unreadable file, JSON that does not parse, a truncated or unparseable safetensors header. `load_pinned_checkpoint` returns `Err` for exactly these, and `read_checkpoint_file` distinguishes "not present" from "could not be read" so the message names the path either way.
- **Contract violations still fail.** The `assert_eq!` calls are untouched and are not wrapped in any skip path. `load_pinned_checkpoint` is structured so it cannot swallow one: a `weight_map` that is absent or non-object yields an empty map (which drives the weight-root assertion to fail, not to skip), and a pinned key absent from an otherwise well-formed `weight_map` is skipped over by the shard loop with a comment saying why, so the weight-root assertion is what reports it.
- `config.json` is parsed as plain JSON only in the pre-check, so a truncated config still skips, while deserializing into `MuseGlimmerConfig` happens next to `validate()` in the assertion body, where a schema mismatch panics. A config that parses but does not fit the schema means the pinned directory holds a wrong or drifted model, which is a contract violation rather than an availability problem. This was corrected in review; the first commit deserialized in the pre-check and therefore skipped on schema drift.
- Inside `safetensors_shape`, a header that parses but does not describe the key as a shaped tensor panics rather than returning `Err`, following the issue's "if unsure whether it is missing or wrong, it is wrong" rule.
- The contract assertions moved into `assert_post_tower_contract`, which takes an already-validated `PinnedCheckpoint`. It holds one skip branch only, for a shard header that cannot be read, which is scoped to availability by construction because those are the only conditions under which `safetensors_shape` returns `Err`. That split is what makes the four `should_panic` tests below possible.

## What changed

- `src/vision/encoders/muse_glimmer_fusion_pinned_tests.rs` (new, 603 lines): the pinned-checkpoint contract check, the `load_pinned_checkpoint` availability pre-check, a `Result`-returning `safetensors_shape`, and synthetic partial-checkpoint coverage. `safetensors_shape` bounds the declared header length twice before allocating: against an absolute `MAX_SAFETENSORS_HEADER_BYTES` ceiling of 256 MiB, mirroring the guard the production reader already applies in `src/lib/mlxcel-core/src/weights.rs:196`, and against the file's own size. The absolute ceiling was added in review: the file-relative bound alone permits a declared header as large as the whole file, and the pinned shards are 50 GB and 9.6 GB, so a corrupt eight-byte prefix could have reached an uncatchable allocation abort.
- `src/vision/encoders/muse_glimmer_fusion_tests.rs`: the pinned block and its now-unused imports removed, leaving 196 lines of fusion-math tests.
- `src/vision/encoders/muse_glimmer_fusion.rs`: attaches `pinned_tests` as a second `#[cfg(test)]` module, the same pattern `src/vision/mod.rs` already uses. The split separates disk-reading contract checks from fusion math and follows the established `#[cfg(test)] #[path]` sibling-module convention. Both files sit under the repository's actual 800-line single-file guideline (`.claude/skills/mlxcel-code-structure/SKILL.md`), at 196 and 603 lines.

Test-only change. No production loading path is touched, and `src/loading/vlm_muse_glimmer_tests.rs` is untouched as the issue required.

## Synthetic-checkpoint tests

The real pinned checkpoint is either complete or absent, so the partial cases cannot be exercised against it, and nothing under `models/` may be moved or truncated to fake one. All seventeen new tests build throwaway checkpoints in `tempfile::tempdir()` (already a dev-dependency, no new dependency added). The pinned module now holds 18 tests in total, the eighteenth being the pre-existing contract check that moved here.

Pre-check names the specific missing or unreadable file: `pinned_precheck_names_an_absent_index`, `pinned_precheck_names_a_missing_config`, `pinned_precheck_names_a_missing_shard` (index present naming a shard that does not exist), `pinned_precheck_names_an_unparseable_index`, `pinned_precheck_names_an_unparseable_config`, and `pinned_precheck_accepts_a_complete_synthetic_checkpoint` as the negative control.

Pre-check refuses to swallow contract violations: `pinned_precheck_leaves_a_missing_weight_root_to_the_contract_assertion`, `pinned_precheck_leaves_an_absent_weight_map_to_the_contract_assertion`, and `pinned_precheck_leaves_a_schema_mismatched_config_to_the_contract_assertion` each assert the pre-check returns `Ok`, so the violation reaches the assertion.

Contract violations still fail, proven executably rather than by reading the diff. Four `#[should_panic]` tests run `assert_post_tower_contract` over complete-but-wrong synthetic checkpoints: `contract_assertion_still_fails_on_a_missing_weight_root`, `contract_assertion_still_fails_on_an_absent_weight_map`, `contract_assertion_still_fails_on_a_schema_mismatched_config`, and `contract_assertion_still_fails_on_a_wrong_recorded_shape`. The last one expects the `failed: ` prefix that `assert_eq!` emits before its custom message, so it can only pass on the shape assertion itself; a bare key name would also match the weight-root assertion and the "must name a shard" panic, letting it pass for the wrong reason.

Header handling: `safetensors_shape_reads_a_well_formed_header`, `safetensors_shape_reports_a_truncated_header`, `safetensors_shape_reports_an_unparseable_header`, and `safetensors_shape_refuses_a_header_over_the_allocation_ceiling`, which declares `u64::MAX` in a tiny temp file so the test itself allocates nothing large.

## Optional env var

`MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` is implemented. It is read inside `skip_or_fail_pinned_checkpoint` under `crate::test_support::env_lock::env_lock()`, held for the duration of the read and dropped before the assertion, because this crate is edition 2024 and `src/server/muse_glimmer_startup_guard_tests.rs` mutates the process environment with `unsafe set_var` under that same lock; an unlocked read is the same class of race being fixed elsewhere in this chain. Only the exact value `1` enables it.

It also turned out to be the cleanest way to prove acceptance criterion 2 on this machine: with skips disabled, the pinned test passing is positive evidence that it read the real checkpoint rather than silently skipping.

Follow-up #1177 tracks extending this gate to the loading-side sibling guard, which does not yet honor it, and documenting the variable in `docs/environment-variables.md`.

## Test plan

Results below are for the merged branch tip, all four commits.

- [x] `cargo test --lib vision::encoders::muse_glimmer_fusion` on a machine holding the real pinned checkpoint: `23 passed; 0 failed` (18 in the pinned module, 5 fusion-math).
- [x] `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1 cargo test --lib vision::encoders::muse_glimmer_fusion`: `23 passed; 0 failed`. Since any skip is a hard failure under this variable, this confirms the full pinned checkpoint still validates the weight-root and shape contract for real (criterion 2).
- [x] Test binary run from a directory containing no `models/`: skips with `Skipping pinned_post_tower_weight_roots_and_shapes_match_published_contract: models/mlx/muse-glimmer-30b/model.safetensors.index.json is not present` (criterion 3, behavior unchanged).
- [x] Same run with `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1`: fails with `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1 but the pinned Muse Glimmer checkpoint is not usable: ... is not present` (criterion 4).
- [x] `cargo clippy --lib --tests -- -D warnings`: clean.
- [x] `cargo check --lib --tests`: clean.
- [x] Nothing under `models/` or `/home/inureyes/models` was moved, renamed, deleted, truncated, or written to; both 50 GB and 9.6 GB shards verified present afterwards with their original timestamps.

Closes #1161

### DIFF
--- a/TECHNICAL_REPORTS/1173-pinned-checkpoint-skip-guard-20260816.en.md
+++ b/TECHNICAL_REPORTS/1173-pinned-checkpoint-skip-guard-20260816.en.md
@@ -1,0 +1,203 @@
+# Technical Report: PR #1173 - fix(test): skip partial Muse Glimmer checkpoints instead of panicking
+
+**Date**: 2026-08-16
+**Author**: AI Code Reviewer
+**Status**: Completed
+**Languages**: Rust
+**Risk Level**: Low
+
+---
+
+## Executive Summary
+
+`pinned_post_tower_weight_roots_and_shapes_match_published_contract` skipped gracefully only when the pinned checkpoint's index file was absent. With the index present but anything else missing or corrupt, an interrupted `hf download` or `mlxcel download`, it panicked on an opaque `unwrap` instead of skipping with a reason. This PR adds an availability pre-check that skips with a message naming the offending file, while keeping every genuine contract violation a hard failure.
+
+The governing constraint was that turning a wrong checkpoint into a silent skip would be strictly worse than the panic being fixed, because it would permanently disable the contract this test exists to enforce. The implementation separates the two categories structurally rather than by convention: `load_pinned_checkpoint` may only veto on availability, and `assert_post_tower_contract` holds the assertions with no skip path except for an unreadable shard header.
+
+Two defects were found in review and fixed before merge, both of them cases where the first implementation put something on the wrong side of that line or left a hazard behind it.
+
+---
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+Issue #1161 came from the security review LOW findings on PR #1157 (implementation of #1155). That PR added the index-absent guard and deliberately left the partial-checkpoint case out of scope. The sibling loading test `pinned_weight_index_classifies_each_source_weight_once` does not share the gap, because it reads only the index file itself.
+
+### 1.2 Existing Issues
+
+- **Issue 1**: past the index-existence guard, the test unwrapped every read: `config.json`, the index contents, `index["weight_map"].as_object()`, and every referenced shard through `safetensors_shape`, which itself unwrapped `File::open`, two `read_exact` calls, and the header parse.
+- **Issue 2** (found in review): the availability pre-check deserialized `config.json` straight into `MuseGlimmerConfig`, so a config that was valid JSON but did not fit the schema was reported as unavailable and skipped. Schema drift in the pinned config, or a different model dropped into the pinned directory, is precisely the divergence this contract test exists to catch, so it belongs on the failing side.
+- **Issue 3** (found in the security pass): the new header-length guard was file-relative only. `header_len > file_len.saturating_sub(8)` permits a declared header as large as the whole file, and the pinned shards are 50 GB and 9.6 GB, so a corrupt eight-byte prefix passed the check and reached `vec![0u8; header_len as usize]`. The outcome is either a `handle_alloc_error` abort, which is uncatchable and prints no reason, or gigabytes of tensor payload pulled into host memory, on the one machine that owns the checkpoint. An abort is strictly worse than the opaque unwrap this PR removes.
+
+### 1.3 Risk Assessment
+
+| Risk | Impact | Likelihood |
+|------|--------|------------|
+| Opaque panic on a partially materialized checkpoint, misread as a code defect | Low | Medium (any interrupted download) |
+| A wrong checkpoint silently skipping, disabling the contract check permanently | Medium | Low (closed by the boundary design and by issue 2's fix) |
+| Uncatchable allocation abort from a corrupt header on the checkpoint-owning machine | Medium | Low (closed by the absolute ceiling) |
+
+---
+
+## 2. Technical Review
+
+### 2.1 Security
+
+The security pass filed one MEDIUM (issue 3 above) and fixed it by mirroring the guard the production reader already uses: `read_safetensors_header_bytes` (`src/lib/mlxcel-core/src/weights.rs:196`) caps the identical eight-byte prefix at `MAX_HEADER_BYTES = 256 * 1024 * 1024` with the rationale "reject absurdly large headers to avoid OOM". This test reader was the only header reader in the tree without such a cap. The mirrored constant is checked before the file-relative bound, and `safetensors_shape_refuses_a_header_over_the_allocation_ceiling` declares `u64::MAX` in a tiny temp file so the test itself allocates nothing large. The real pinned shard records an 11,720-byte header against a 9.6 GB file, roughly four orders of magnitude of margin, so no genuine checkpoint can be rejected.
+
+Everything else was verified clean. The boundary holds on every traced path: an absent or non-object `weight_map` becomes an empty map that fails the weight-root assertion; a pinned key missing from a well-formed map is skipped past in the loop and reported by that same assertion; a key mapped to a non-string hits the "must name a shard" panic; a header that parses without a shaped entry for the key panics rather than returning `Err`. The env-var read holds the crate-wide env lock in a scoped block and drops it before the `assert!`, so a failing assertion cannot poison it, and nothing reachable from that function re-acquires the non-reentrant mutex. Only file headers are read, never tensor payloads, and every `TempDir` is bound to a live local so the `should_panic` tests still clean up on unwind.
+
+**Issues Found:**
+| Issue | Severity | Status |
+|-------|----------|--------|
+| Schema-mismatched `config.json` skipped instead of failing | Medium | Fixed (`3336732c`) |
+| Header allocation bounded only by file size, so a corrupt length in a 50 GB shard could abort the process | Medium | Fixed (`3c61dd21`) |
+| `contract_assertion_still_fails_on_a_wrong_recorded_shape` expected a bare key name that three different panics emit | Low | Fixed (`3336732c`) |
+| `pinned_precheck_leaves_an_absent_weight_map_...` had no `should_panic` companion, unlike its missing-weight-root sibling | Low | Fixed (`3336732c`) |
+
+### 2.2 Performance
+
+No production path is touched. The pinned check reads only file headers, so it stays fast despite the 60 GB checkpoint: the full module runs in about 0.01 s.
+
+### 2.3 Compatibility & Dependencies
+
+- **Breaking Changes**: none
+- **New Dependencies**: none; `tempfile` was already a dev-dependency (`Cargo.toml:308`)
+- **Compatibility**: no production loading path is affected
+
+### 2.4 Code Quality
+
+- **Test Coverage**: 5 tests became 23 in the module, adding synthetic coverage for every skip branch plus executable proof that contract violations still fail
+- **Code Complexity**: the pinned check is now three named functions with one responsibility each instead of one unwrap-laden body
+- **Technical Debt**: decreased
+
+---
+
+## 3. Technical Decisions
+
+### 3.1 Hybrid pre-check rather than either option the issue offered
+
+**Alternatives Considered:**
+
+| Option | Pros | Cons |
+|--------|------|------|
+| Option 1: existence pre-check only | Simple | Cannot see a shard that exists but whose header was truncated mid-download |
+| Option 2: convert every unwrap into skip-with-reason | Covers corruption | Reads the index twice, which the issue explicitly ruled out |
+| **Chosen: hybrid** | Single index parse shared with the assertions, and corruption coverage | Slightly more structure |
+
+**Rationale:** `load_pinned_checkpoint` parses the index once and hands the parsed `weight_map` to the assertion body, and `safetensors_shape` returns `Result` so header corruption is covered too.
+
+### 3.2 Separate availability from contract structurally, not by convention
+
+**Rationale:** The assertions live in `assert_post_tower_contract`, which takes an already-validated `PinnedCheckpoint`. Reaching it means every file it needs was confirmed readable, so anything failing there is the checkpoint disagreeing with the contract. The one skip inside it, an unreadable shard header, is scoped to availability by construction: the only `Err` producers in `safetensors_shape` are open failure, metadata failure, truncation, an over-ceiling or over-file declared length, and a non-JSON header. A header that parses but records no shape, or a non-integer dimension, panics.
+
+### 3.3 Defer `MuseGlimmerConfig` deserialization into the assertion body
+
+**Rationale:** This is the fix for issue 2. The pre-check now parses `config.json` as plain JSON only, so a truncated config, which is a syntax error and a real availability problem, still skips, while a config that parses and does not fit the schema reaches `serde_json::from_value::<MuseGlimmerConfig>` in the assertion body and panics. `PinnedCheckpoint` carries the raw `Value` to make the split explicit. The fix has teeth rather than passing vacuously: `MuseGlimmerConfig.text_config` and its inner fields carry no serde defaults (`src/models/muse_glimmer_config.rs:26`), so a mismatched config genuinely fails to deserialize.
+
+### 3.4 Implement the optional env gate
+
+**Rationale:** `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` turns every skip into a failure, so a corrupted checkpoint cannot quietly disable the contract check forever on the machine positioned to enforce it. It is read under the crate-wide env lock, matching the convention `muse_glimmer_startup_guard_tests` established, because this crate is edition 2024 and a sibling test mutates the environment with `unsafe set_var` under that same lock. It also turned out to be the cleanest way to prove the happy path: see the appendix.
+
+### 3.5 Split into a separate file
+
+**Rationale:** the pinned check plus its synthetic coverage is a distinct concern from the fusion math tests, and the `#[cfg(test)] #[path = "..."] mod` sibling pattern is already used throughout `src/vision/`. Note that the PR and commit bodies justify the split with a "500-line limit"; the repo's own code-structure guidance actually puts the threshold at 800 lines, so that stated reason is wrong even though the split itself is sound.
+
+---
+
+## 4. Implementation Details
+
+### 4.1 Key Code Changes
+
+**File: `src/vision/encoders/muse_glimmer_fusion_pinned_tests.rs`** (new)
+```rust
+// The availability veto, which may never see a contract violation.
+fn load_pinned_checkpoint(model_dir: &Path) -> Result<PinnedCheckpoint, String> {
+    let index_text = read_checkpoint_file(&index_path)?;
+    let index: Value = serde_json::from_str(&index_text)
+        .map_err(|err| format!("{} does not parse as JSON: {err}", index_path.display()))?;
+    // ... config parsed as plain JSON only ...
+    // A missing or non-object `weight_map` is a malformed index rather than a
+    // missing file: hand the caller an empty map so the weight-root assertion
+    // reports it as the contract violation it is.
+    let weight_map = index["weight_map"].as_object().cloned().unwrap_or_default();
+    // ... each referenced shard checked for presence, naming any that is missing ...
+}
+```
+
+```rust
+// Bound the declared header before allocating.
+if header_len > MAX_SAFETENSORS_HEADER_BYTES { /* Err */ }
+if header_len > file_len.saturating_sub(8) { /* Err */ }
+let mut header = vec![0u8; header_len as usize];
+```
+
+**Reason for change:** the original `safetensors_shape` did `vec![0u8; header_len]` straight from an unvalidated eight-byte little-endian read.
+
+---
+
+## 7. Change Summary
+
+### Statistics
+| Item | Value |
+|------|-------|
+| Files changed | 3 |
+| Lines added | +609 |
+| Lines deleted | -79 |
+| Tests added | 18 (module went from 5 to 23) |
+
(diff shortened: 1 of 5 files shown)


inureyes:

### Intent

> Make `pinned_post_tower_weight_roots_and_shapes_match_published_contract` skip with a named reason on a partially materialized checkpoint instead of panicking on an opaque `unwrap`, without letting any genuine contract violation become a skip.

### Correctness boundary audit

Every path in `load_pinned_checkpoint`, `assert_post_tower_contract`, and `safetensors_shape` was traced against the fail/skip split. One ambiguous case, everything else is correctly classified.

- `unwrap_or_default()` on `weight_map`: correct. A non-object or absent `weight_map` yields an empty map, `actual_keys` comes out empty, and `assert_eq!(actual_keys, PINNED_POST_TOWER_WEIGHT_KEYS)` fails. A non-object `index` (array, scalar) also lands here because `Value`'s string index returns `Null` rather than panicking, so a wholly malformed index fails rather than skips. That is the safe direction.
- `continue` on a pinned key absent from a parsed `weight_map`: correct, and executably proven by `contract_assertion_still_fails_on_a_missing_weight_root`.
- A pinned key present but mapped to a non-string value: the pre-check `continue`s, the weight-root assertion passes (the key exists), and the loop's `unwrap_or_else(|| panic!("{key} must name a shard ..."))` fires. Fails, does not skip. Correct.
- `config.validate()` stays in `assert_post_tower_contract` as `.unwrap()`, so a config `validate()` rejects still panics. Correct.
- `safetensors_shape` panics (does not return `Err`) when a header parses but records no shape or a non-integer dimension for the key. Correct per the "if unsure, it is wrong" rule.
- The `Err` skip branch at lines 185-191 is scoped to availability only: open failure, metadata failure, truncation before or inside the header, a declared header length exceeding the file, and a header that is not JSON. No contract signal can reach it.

### Findings

**MEDIUM: a `config.json` that parses as JSON but does not deserialize into `MuseGlimmerConfig` skips rather than fails** (`muse_glimmer_fusion_pinned_tests.rs:74-79`). This is the one place where something genuinely wrong can go silent. Schema drift in the pinned checkpoint's config (a renamed or dropped `text_config` field, a wrong model placed in the pinned directory) is exactly the kind of divergence this contract test exists to catch, and the brief's tie-breaker says an ambiguous case is "wrong" and must fail. An interrupted download leaves a truncated `config.json`, which is a JSON syntax error and would still skip correctly under a tighter split. Suggested shape: parse `config.json` to `Value` in the pre-check (syntax error skips), carry the `Value` in `PinnedCheckpoint`, and do `serde_json::from_value::<MuseGlimmerConfig>(..).unwrap_or_else(|err| panic!(..))` inside `assert_post_tower_contract` next to `validate()`. `pinned_precheck_names_an_unparseable_config` would move to asserting "does not parse as JSON", with a new `#[should_panic]` test for the schema case. Note that the doc comment on `load_pinned_checkpoint`, the PR body, and the commit body all describe the skip set as "JSON that does not parse", which is narrower than what the code actually skips on, so at minimum the documented boundary and the implemented boundary should be reconciled.

**LOW: the mid-loop skip can mask a later contract violation** (`:185-191`). `expected` is a `BTreeMap`, so keys are visited in sorted order. If `fc1`'s shard is unreadable, the function returns before checking `fc2` and `vision_projection`, and a wrong recorded shape on those keys goes unreported. Previously the same checkpoint produced a loud failure. Inherent to any skip design (an unreadable shard cannot be asserted on) and bounded by `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1`, but asserting the keys that are readable and skipping only at the end would be strictly better.

**LOW: the header-length bound is file-relative only** (`:221`). The arithmetic is right: for a valid safetensors file `header_len = file_len - 8 - payload_len <= file_len - 8`, so the check cannot reject a valid header, `saturating_sub` cannot underflow (an 8-byte `read_exact` already succeeded), and the `as usize` cast is lossless on this 64-bit target. What remains is that a corrupt length field inside a genuinely large shard (the real one is 50 GB) can still request a multi-GB allocation. An absolute sanity cap in addition to the file-size bound would close that.

**LOW: `contract_assertion_still_fails_on_a_wrong_recorded_shape`'s `expected` is not unique to the shape assertion** (`:454`). `"model.vision_adapter.fc1.weight"` also appears in the weight-root `assert_eq!` output and in the "must name a shard" panic, so a future refactor could make this test pass for the wrong reason. `expected = "failed: model.vision_adapter.fc1.weight"` (the custom-message suffix `assert_eq!` emits) or a substring of the left value pins it to the shape assertion. `contract_assertion_still_fails_on_a_missing_weight_root` is fine as written: the only other reachable panic names `fc1`, not `vision_projection`.

**LOW: the empty-`weight_map` case is not proven to fail end to end.** `pinned_precheck_leaves_an_absent_weight_map_to_the_contract_assertion` proves the pre-check returns `Ok` with an empty map, but no test runs `assert_post_tower_contract` over it. The sibling missing-root case does have that `#[should_panic]` companion. A four-line addition closes the gap.

**LOW: body inaccuracies that become the squash commit body.** The PR body says "eleven new tests"; there are 13 new (14 in the file, 19 in the module overall). Both the PR body and the commit body justify the file split by a "500-line limit" that is not this repo's convention: the repo's own `mlxcel-code-structure` skill doc puts the threshold at 800 lines and 394 files under `src/` already exceed 500. The split itself is sound on its own merits (disk-reading contract checks separated from fusion math, using the established `#[cfg(test)] #[path]` sibling-module pattern that `src/vision/mod.rs` already uses); only the stated rationale is wrong.

**INFO: only one of the two pinned guards is hardened.** `pinned_weight_index_classifies_each_source_weight_once` in `src/loading/vlm_muse_glimmer_tests.rs` does not honor `MLXCEL_REQUIRE_PINNED_CHECKPOINTS`, so on the checkpoint-owning machine one pinned guard can still skip silently. Out of scope for this issue by explicit instruction; worth a follow-up.

### Env lock

No deadlock is possible. `env_lock()` is a non-reentrant `std::sync::Mutex` behind a `OnceLock` that recovers from poisoning, `skip_or_fail_pinned_checkpoint` acquires it in a nested block that contains only the `std::env::var` read, and the guard is dropped before the `assert!`, so a failing assertion cannot even poison it. The function is called from exactly two sites, neither of which already holds the lock, so there is no nesting. The backend seam tests from #1169 and `muse_glimmer_startup_guard_tests` share the same static in the same lib test binary and can only serialize briefly against this read. Worth noting that `test_support::env_lock`'s module doc still says it is used by tests that call `set_var`/`remove_var`; this is the first read-only user.

### Scope

Proportionate, with room to trim. Of the 499 lines, roughly half is the pinned test and helper that merely moved, and the synthetic coverage is what the issue explicitly asked for. The doc comments are longer than the surrounding code's norm and could lose about 30 lines without losing meaning, but they document the correctness boundary, which is the substance of the issue. Not worth churn.

### Verification

- [x] All stated requirements implemented (criteria 1, 2, 3, and the optional 4)
- [x] Integrated into project code flow (second `#[cfg(test)] #[path]` module, all 14 tests run, none orphaned, no production path touched)
- [x] Project conventions followed (license header, sibling test-module pattern, `tempfile` already a dev-dependency, no new dependency)
- [x] Existing modules reused where applicable (`crate::test_support::env_lock`, `MuseGlimmerConfig`)
- [x] No unintended structural changes (removal from the original file is exactly the pinned test, its helper, and the now-unused imports)
- [x] Tests pass (19 passed, and 19 passed again under `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` against the real checkpoint)
- [x] No AI attribution and no em dashes in the diff, commit body, or PR body; `Closes #1161` present

No changes were made by this review; all findings are reported for the author to accept or decline.

inureyes:

Reviewed at `3336732c`, fix pushed as `3c61dd21`. Test-only change, no production path touched, so the review focused on the binary parsing, the resource handling, the env-lock interaction, and the availability-versus-violation boundary.

### Fixed: unbounded header allocation against a multi-gigabyte shard (MEDIUM)

`safetensors_shape` bounded the declared header length against the file size only. That bound is file-relative, and the pinned shards are 50 GB and 9.6 GB, so a corrupt eight-byte prefix could declare a header of several gigabytes, pass the check, and reach `vec![0u8; header_len as usize]` followed by `read_exact`. The outcome is either an allocation-failure abort or gigabytes of tensor payload read into memory, on the one machine that owns the checkpoint. An abort carries no reason at all, which is a worse failure mode than the opaque `unwrap` this PR set out to replace.

`3c61dd21` adds a 256 MiB ceiling checked before the file-relative bound, the same value and rationale as the guard `read_safetensors_header_bytes` in `mlxcel-core` already applies to this identical eight-byte prefix. The real pinned shard records an 11,720-byte header, so the margin is roughly four orders of magnitude and no genuine checkpoint can be rejected. Covered by `safetensors_shape_refuses_a_header_over_the_allocation_ceiling`, which declares `u64::MAX` in a tiny temp file so the test allocates nothing large either.

### Verified clean

- **Correctness boundary.** Every `Err` path in `load_pinned_checkpoint` and `safetensors_shape` is an availability or readability problem. Everything wrong-but-present reaches an assertion: an absent or non-object `weight_map` becomes an empty map that fails the weight-root assertion, a pinned key missing from a well-formed map is skipped by the shard loop and reported by that same assertion, a key mapped to a non-string value hits the "must name a shard" panic, and a header that parses without a shaped entry for the key panics rather than returning `Err`. No path was found where something wrong skips.
- **The `3336732c` follow-up is correct and complete.** `PinnedCheckpoint.config` is a raw `Value`, the pre-check verifies JSON syntax only, and `from_value::<MuseGlimmerConfig>` plus `validate()` both sit in `assert_post_tower_contract` where they panic. `text_config` and its inner fields have no serde defaults, so a config that is valid JSON but is not this model genuinely fails deserialization rather than passing by default.
- **Env lock.** `env_lock()` recovers from poisoning via `unwrap_or_else(|p| p.into_inner())`, the guard is scoped and dropped before the `assert!`, and nothing reachable from `skip_or_fail_pinned_checkpoint` re-acquires it. No deadlock against the four `src/backend/tests.rs` holders or against `muse_glimmer_startup_guard_tests`, and no poisoning hazard in either direction.
- **Resources.** The shard `File` is dropped at function scope on every path including the early `Err` returns, only the header is ever read, and every `TempDir` is bound to a live local, so the `should_panic` tests still clean up on unwind.

### Observation, not blocking

`contract_assertion_still_fails_on_an_absent_weight_map` and `contract_assertion_still_fails_on_a_missing_weight_root` expect a bare key name that also appears in the loop's "must name a shard in the pinned weight index" panic, so they do not pin the weight-root assertion the way the wrong-shape test pins the shape assertion with its `failed: ` prefix. Both alternatives are failures rather than skips, so the invariant the tests exist to protect still holds either way. Left as is to avoid depending on more of `assert_eq!`'s output formatting.

### Verification

`cargo test --lib vision::encoders::muse_glimmer_fusion` 23 passed, the same run under `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` against the real checkpoint 23 passed (so the new ceiling does not turn the contract check into a skip), `cargo clippy --lib --tests -- -D warnings` clean, `cargo fmt --check` clean. Nothing under `models/` was written, moved, or truncated; both shards verified present and unmodified afterwards.

Issue #1166, opened 2026-08-15
feat(models): video input for the Qwen VL families (Qwen3.5-VL / Qwen3.8, then the rest)


inureyes:
## Summary

`--video` is rejected for every Qwen VL family with `"--video input is currently only supported by Gemma 4 and Kimi-VL VLMs"` (`src/commands/generate_vlm.rs:447`), even though the checkpoints ship a full video contract and upstream mlx-vlm supports video on this path end to end. Add video input for the Qwen VL families, starting with `qwen3_5` (Qwen3.5-VL / Qwen3.8-27B).

The rejection is honest and fails closed, so this is a missing capability, not a correctness bug.

## Evidence that the checkpoints expect video

`mlx-community/Qwen3.8-27B-4bit` (revision `3e6447f0`) ships all of:

- `config.json`: `video_token_id: 248057`, `vision_start_token_id: 248053`, `vision_end_token_id: 248054`
- `video_preprocessor_config.json`: `video_processor_type: "Qwen3VLVideoProcessor"`, `patch_size 16`, `temporal_patch_size 2`, `merge_size 2`, `size.longest_edge 25165824`
- `processor_config.json`: `fps: 2.0`, `min_frames: 4`, `max_frames: 768`
- `chat_template.jinja`: emits `<|vision_start|><|video_pad|><|vision_end|>` for `item.type == 'video'`

The Qwen3.8-27B model card advertises *"Native support for image and video understanding, from STEM diagrams and documents to hour-scale videos."*

mlxcel already reads `video_token_id` for these families (`src/loading/vlm.rs:611-626`, `src/loading/vlm_qwen.rs`), so the id plumbing is half there; nothing consumes it.

## What upstream implements

`mlx_vlm/models/qwen3_5/vision.py` is five lines subclassing `qwen3_vl`'s vision model, so all video behavior comes from the shared Qwen3-VL stack:

- **Token merge treats image and video alike**: `special_image_mask = (input_ids == image_token_index) | (input_ids == video_token_index)`, then `masked_scatter`. `get_input_embeddings` falls back to `pixel_values_videos` when `pixel_values` is None and picks `grid_thw = image_grid_thw if image_grid_thw is not None else video_grid_thw`.
[...]

Issue #1165, opened 2026-08-15
feat(inference): support the qwen3_5_mtp drafter for Qwen3.5/3.6/3.8 MTP speculative decoding


inureyes:
## Summary

Add support for the `qwen3_5_mtp` drafter so Qwen3.5 / 3.6 / 3.8 targets can run MTP speculative decoding. Today mlxcel's MTP round loop exists but only Gemma 4 implements `MtpTarget`, and `qwen3_5_mtp` is not a known drafter `model_type`. Both failure paths are named and fail closed, so this is a missing capability rather than a bug.

## Where the MTP weights actually live

This was the first thing to establish, because the naming suggests the weights ride along inside the model.

**They do not, in any MLX build.** Verified against the safetensors index of every mlx-community Qwen3.8-27B variant, and against the downloaded shard headers directly:

| repository | tensors | `mtp.*` tensors |
|---|---|---|
| `mlx-community/Qwen3.8-27B-4bit` | 2180 | **0** |
| `mlx-community/Qwen3.8-27B-8bit` | 2180 | **0** |
| `mlx-community/Qwen3.8-27B-bf16` | 1184 | **0** |
| `mlx-community/Qwen3.8-27B-mxfp8` | 1682 | **0** |
| `mlx-community/Qwen3.8-27B-nvfp4` | 1682 | **0** |
| `Qwen/Qwen3.8-27B` (upstream HF bf16) | 1199 | **15** |

The upstream base checkpoint does ship MTP, as a sibling of `model.`, not nested under it:

```
mtp.fc.weight
mtp.norm.weight
mtp.pre_fc_norm_embedding.weight
mtp.pre_fc_norm_hidden.weight
mtp.layers.0.input_layernorm.weight
mtp.layers.0.post_attention_layernorm.weight
mtp.layers.0.mlp.{gate,up,down}_proj.weight
mtp.layers.0.self_attn.{q,k,v,o}_proj.weight
mtp.layers.0.self_attn.{q,k}_norm.weight
```

The MLX conversion pipeline drops them, and mlx-community republishes them split out as a standalone drafter:

- **`mlx-community/Qwen3.8-27B-MTP-bf16`** — `model_type: "qwen3_5_mtp"`, `block_size: 3`, single 849 MB `model.safetensors`, full `text_config` mirroring the target (`hidden_size` 5120, `head_dim` 256, `mtp_num_hidden_layers: 1`, `mtp_use_dedicated_embeddings: false`).

[...]

Issue #1040, opened 2026-08-05
perf(models): decide whether chunked GLA prefill becomes the bailing_moe_linear default


inureyes:
## Summary

PR #1039 (`bailing_moe_linear`, the Ling / Ring linear-attention MoE family) ships two evaluations of the same GLA recurrence. The chunked one is 2.6x to 4x faster on prefill and, in isolation, the more accurate of the two, but it is opt-in. This issue is the data-backed decision on whether it becomes the default.

## Background

Two paths, same recurrence, both in `src/models/bailing_moe_linear.rs`:

- `gla_sequential` is upstream mlx-lm's per-token loop transcribed step for step. It is the **default**.
- `gla_chunked` is the closed form of that recurrence evaluated over 64-token chunks. It is opt-in behind `MLXCEL_BAILING_LINEAR_CHUNKED_PREFILL=1`.

Measured on `mlx-community/Ring-mini-linear-2.0-4bit` on an M1 Ultra, prefill only. Decode is identical on both paths, since both take the same single-step recurrence at `L == 1`.

| prompt tokens | sequential | chunked |
|---|---|---|
| 101 | 0.29s | 0.07s |
| 512 | 0.75s | 0.28s |
| 2048 | 2.81s | 1.07s |

Chunking is also the more accurate path in isolation, because the intra-chunk sum lands in a matmul accumulator instead of a bf16 running state.

**Why it is not already the default.** Reassociating the sum changes results by well under one bf16 ulp per layer, but this stack amplifies that. On a 101-token prompt, layer 0's mean absolute activation moves by 0.04%; layer 3 then multiplies its input magnitude by roughly 23x; and the 256-expert top-8 router converts a sub-ulp score difference into a different selected expert. The two paths therefore decode different (not worse, just different) continuations from the same checkpoint. Rather than assume the faster path is fine, PR #1039 shipped the reference's own arithmetic by default and left the promotion to a measurement.

## What to measure

- **Perplexity** on a held-out corpus for both paths on `mlx-community/Ring-mini-linear-2.0-4bit`, over enough tokens that the difference is outside run-to-run noise. If chunked is equal or better, the accuracy objection is answered.
- **Larger sibling.** Repeat on `inclusionAI/Ring-flash-linear-2.0` if it is reachable, since the amplification behaviour may scale with depth.
- **Prefill throughput** at 512 / 2048 / 8192 tokens: median of repeated runs with dispersion reported. This box carries background load, so a single measurement is not trustworthy.
[...]

PR #1174, opened 2026-08-15
feat(models): qualify Qwen3.8-27B on the qwen3_5 path

### LINKED ISSUE #1163: feat(models): qualify and register Qwen3.8 (Qwen3.8-27B) on the existing qwen3_5 path
## Summary

Qualify `mlx-community/Qwen3.8-27B-4bit` as a supported checkpoint, register the Qwen3.8 generation in the docs and detection tables, and turn the config keys mlxcel currently ignores into explicit reads or explicit assertions.

This is **not** a new model family port. Qwen3.8-27B ships `model_type: "qwen3_5"` / `Qwen3_5ForConditionalGeneration` and is architecturally identical to Qwen3.5-27B. It already loads and runs on `main` with zero code changes. The deliverable is qualification, registration, and closing the silent-ignore gaps that make that "works today" accidental rather than guaranteed.

## Research baseline (2026-08-15)

- Repository: `mlx-community/Qwen3.8-27B-4bit`, revision `3e6447f082e89cc7f0bc6e5441afd38dfce760ff`
- License: Apache-2.0, public, ungated. Converted from `Qwen/Qwen3.8-27B` with mlx-vlm 0.6.8
- Pipeline: `image-text-to-text`. 3 shards, 16.05 GB, 2180 tensors
- mlxcel revision under test: `main` @ `6e781fcf7e70a8509c9b4617e6bacb48692a2d5a` (v0.5.1), release binaries built 2026-08-14
- Hardware: Apple Silicon, Metal backend, load average 3.4-4.1 during measurement

Upstream's own framing (Qwen3.8-27B model card): *"Built on the architectural foundation of Qwen3.5."*

### Architectural identity with Qwen3.5-27B

The weight map is identical. Normalizing layer indices and diffing `model.safetensors.index.json` against the local `qwen3.5-27b-4bit`:

```
count35 2180  count38 2180
--- only in 3.5 ---   (none)
--- only in 3.8 ---   (none)
```

Same 64 layers, `hidden_size` 5120, `intermediate_size` 17408, `head_dim` 256, 24 Q heads / 4 KV heads, 48/16 linear value/key heads, `full_attention_interval` 4, vocab 248320, and the same Qwen3-VL vision tower (depth 27, hidden 1152, patch 16, merge 2, `deepstack_visual_indexes: []`).

Upstream agrees: mlx-vlm's `mlx_vlm/models/qwen3_5/config.py` and `vision.py` are byte-identical between the 2026-07-11 tree and current `main` (v0.6.14). The open upstream PR [Blaizzy/mlx-vlm#1899](https://github.com/Blaizzy/mlx-vlm/pull/1899) "Add Qwen3.8-27B support" contains **no architectural change** either: it is FP8 block-dequant, ModelOpt NVFP4/MXFP4 adaptation, and MTP splitting. None of it is needed for the plain `{"group_size": 64, "bits": 4, "mode": "affine"}` mlx-community 4-bit build.

### Config delta vs. Qwen3.5-27B

Only metadata moved. Diff of the two `config.json` files, sorted:

| Key | Qwen3.5-27B | Qwen3.8-27B | mlxcel today |
|---|---|---|---|
| `text_config.output_gate_type` | absent | `"swish"` | ignored |
| `text_config.partial_rotary_factor` | absent (nested only) | `0.25` (also nested) | ignored at top level, read from `rope_parameters` |
| `text_config.mlp_only_layers` | `[]` | absent | `#[serde(default)]` → `[]` |
| `text_config.tie_word_embeddings` | absent | `false` | parsed |
| `language_model_only` | absent | `false` | ignored |
| `generation_config` (nested in config.json) | absent | present | ignored |
| `temperature` / `top_k` / `top_p` / `do_sample` at top level | absent | present | ignored |
| `generation_config.json` `temperature` | 0.6 | 1.0 | read |
| `processor_config.json` `image_processor_type` | `Qwen2VLImageProcessorFast` | `Qwen3VLImageProcessor` | ignored |
| `processor_config.json` sizing | `size.{shortest,longest}_edge` | flat `min_pixels`/`max_pixels` (same values) | `preprocessor_config.json` is byte-identical and is what the loader reads |
| `tokenizer_config.json` `tokenizer_class` | `TokenizersBackend` | `Qwen2Tokenizer` | irrelevant, `tokenizer.json` is used |

`Qwen35Config` (`src/models/qwen3_5.rs:50-105`) carries no `deny_unknown_fields`, so every one of the new keys is silently dropped. The six fields without a serde default (`model_type`, `hidden_size`, `num_hidden_layers`, `num_attention_heads`, `num_key_value_heads`, `vocab_size`) are all present. That is why the checkpoint loads.

## Verification performed (2026-08-15, `main` @ `6e781fcf`)

Everything below was run against the real checkpoint on the release binaries, not synthetically.

**Static preconditions**, each of which could have made the load silently wrong:

- `layer_types` is ignored by mlxcel; the linear/full schedule is computed positionally as `!(i+1).is_multiple_of(full_attention_interval)` (`src/models/qwen3_5.rs:174-176`). Checked all 64 entries of the checkpoint's `layer_types` against that rule: **0 mismatches**.
- `mlp_only_layers` is absent, which would make every layer MoE if `num_experts > 0` (`qwen3_5.rs:178-182`). The checkpoint has no `num_experts`: dense, so the branch is unreachable.
- `PatchEmbed::from_weights` (`src/vision/encoders/qwen3_vl.rs:118-157`) has no quantized branch. The checkpoint's vision tower carries **0** `scales`/`biases` tensors; `vision_tower.patch_embed.proj.{weight,bias}` are plain. Safe.
- `deepstack_visual_indexes` is `[]`, which is the only value upstream accepts for qwen3_5 and the case mlxcel is written for (`src/vision/qwen3_5_vl.rs:35`).

**CLI text.** `mlxcel generate -m models/qwen3.8-27b-4bit -p "What is 17 times 23?" -n 300 --temp 0` → `391`. Correct.

**CLI vision.** 224x224 synthetic PNG (red circle + blue square), prompt "Describe the shapes and their colors" → `Inserted 1 Qwen VL image token blocks (49 total image tokens)`, answer names the red circle correctly.

**Server text.** `POST /v1/chat/completions`, temp 0 → `content: "391"`, `reasoning_content` populated and correctly separated from `content`, `finish_reason: "stop"`, `usage.prompt_tokens: 66`.

**Server vision.** base64 `image_url` data URI + text part → `"The image contains a red circle and a blue square."` Both shapes correct. `usage.prompt_tokens: 102`, which includes the 49 image tokens (no repeat of the Florence-2 undercount in #1086).

**Server multi-turn prompt cache.** Snapshot donation from epic #1148 carries to this checkpoint:

```
turn1: prompt=48  cached=0
turn2: prompt=77  cached=61
turn3: prompt=107 cached=90
```

**Decode throughput**, 3 repeats, same prompt, `--temp 0`, load average 3.8:

| run | tok/s |
|---|---|
| 1 | 18.20 |
| 2 | 18.21 |
| 3 | 18.24 |

Median 18.21 tok/s, spread 0.04. This box carries background load, so treat the absolute number as an order-of-magnitude baseline, not a benchmark record.

### Cache footprint: at the architectural minimum, and the architecture is expensive

Measured by fitting `snapshot_bytes` per entry against prompt length (four points, distinct `prompt_cache_key` per point, 24 GiB snapshot budget so nothing is evicted):

| prompt_tokens | bytes/entry |
|---|---|
| 1,610 | 195,887,104 |
| 6,410 | 514,654,208 |
| 14,410 | 1,034,747,904 |
| 25,610 | 1,772,945,408 |

Fit: `bytes/entry = 91,251,006 + 65,637.6 x tokens`, i.e. **~87 MiB length-independent state plus 64.1 KiB/token**.

The per-token term matches the theoretical f16 minimum for this architecture exactly: 16 full-attention layers x 4 KV heads x 256 head_dim x 2 (K and V) x 2 bytes = **65,536 B/token**. So mlxcel is **not** over-allocating: KV is stored only for the 16 full-attention layers, not for all 64. The 48 gated-delta layers contribute only the length-independent ~87 MiB recurrent + conv state. `paged_blocks_allocated` and `paged_bytes_reserved` are both 0 (this family owns its cache), so the startup line `Paged KV block budget: 2825536 blocks (64 layers, 32-token blocks)` is a cap, not a reservation.

The cost is intrinsic to the checkpoint: `head_dim` 256 is unusually large, so even at a 3:1 linear:full layer ratio the KV is 64 KiB/token. At the native 262,144-token context that is **16 GiB of KV state** for a single sequence, on top of the 15 GiB of weights. Worth stating in the docs.

Two flags that look like mitigations are not. `--kv-bits 8` and `--kv-cache-mode int8` each leave the stored footprint **byte-for-byte identical** to the fp16 default (514,654,208 at 6,410 tokens; 1,772,945,408 at 25,610 tokens under all three settings), and no startup line reports which KV mode is actually active for this family. `qwen3_5` is on the symmetric-Turbo allowlist (`src/lib/mlxcel-core/src/cache/turbo/allowlist.rs:102`), so a Turbo mode may behave differently; that was not measured. Whether the mode is ignored for model-owned caches or the snapshot stores dequantized state is left to the snapshot-capacity follow-up issue.

### 1. The checkpoint is nowhere in the docs or the test set

`docs/supported-models.md` mentions this family in exactly three lines (`:22`, `:96`, `:523`) and names no validated checkpoint. There is no Qwen3.5 or Qwen3.8 paragraph, and no pinned 27B checkpoint outside `tests/tensor_parallel_real_models.rs:410-418`.

### 2. Three config keys are load-bearing upstream and silently ignored here

- **`output_gate_type: "swish"`** originates in vLLM (`qwen_gdn_linear_attn.py`), which selects the gated-delta output-gate activation from `{silu, swish, sigmoid}` and treats `swish` as an alias for `silu`. mlxcel hardcodes SiLU in `RMSNormGated::forward` (`src/models/gated_delta.rs:700-733`), so it is **accidentally correct** for this checkpoint. A future checkpoint declaring `sigmoid` would load and produce silently wrong output. mlx-vlm has the same hole (zero occurrences of the key anywhere upstream).
- **`language_model_only`** is `false` here so nothing changes, but upstream has an open bug ([Blaizzy/mlx-vlm#1812](https://github.com/Blaizzy/mlx-vlm/issues/1812)) where a vision-stripped checkpoint sets it `true` and fails to load; the fix PR #1855 was closed unmerged. If a `language_model_only: true` Qwen3.8 build appears, mlxcel will look for a vision tower that is not there.
- **`rope_parameters.mrope_interleaved: true`** is never read (`InterleavedMRoPE` is hardcoded at `src/models/qwen3_next.rs:556-557`). It matches by construction today.

### 3. Constants that must not be defaulted

mlx-vlm's dataclass defaults disagree with this checkpoint on two token ids, and mlxcel has its own `qwen35_vlm_token_defaults()` (`src/loading/vlm.rs:628-634`):

| Field | mlxcel/mlx-vlm default | Qwen3.8-27B actual |
|---|---|---|
| `vision_start_token_id` | 248045 | **248053** |
| `vision_end_token_id` | 248046 | **248054** |
| `image_token_id` | 248056 | 248056 |
| `video_token_id` | 248057 | 248057 |

The config supplies both explicitly and `qwen_vl_token_ids` (`src/loading/vlm.rs:611-626`) reads `vision_start_token_id`, so the working path is correct. But `vision_end_token_id` is not read at all, and a fallback to 248045 would silently mis-segment vision spans in MRoPE position assignment. Likewise `rope_theta` here is **1e7**, not the 1e5 default, and the checkpoint spells the type key `rope_type` rather than `type`.

### 4. Two upstream fixes landed after mlxcel's reference pin (PREMISE CORRECTED: both were already in the tree)

`references/mlx-vlm` is pinned at `8e2638b` (2026-07-11), which predates this model. Two post-pin commits have a Rust analogue:

- [`24910756`](https://github.com/Blaizzy/mlx-vlm/commit/24910756) (#1741) chunked MRoPE: slice `position_ids` to the current chunk when `position_ids.shape[-1] > seq_length`. Relevant to chunked prefill.
- [`a2482f29`](https://github.com/Blaizzy/mlx-vlm/commit/a2482f29) (#1805) structured output on a padded vocab: the docstring names this exact shape, *"Qwen3.5 pads 248077 real tokens up to 248320 rows"*. A grammar/JSON-schema token mask sized to the real vocab will be wrong against the padded `lm_head`.

The other two post-pin commits (`162e7af8` threadgroup probe, `224f92a4` gated_delta vendoring) are MLX-Python specific with no Rust counterpart.

**Correction, verified before implementation.** Neither fix needed porting. Both behaviors were already present on `main` @ `865bede2`, so the deliverable became "confirm, pin with a regression test, and correct this issue" rather than "port".

- **#1805 (padded-vocab structured-output mask): already correct.** `apply_structured_mask_to_logits` (`src/server/structured.rs:736-788`) already sizes the bias from `vocab_size_hint`, the model's logits axis, and masks every position in `[matcher_vocab, vocab_size_hint)`. Its doc comment at `:718-735` already names this exact case. What landed instead is `apply_mask_covers_the_qwen3_8_padded_lm_head` in `tests/structured_outputs.rs`, which pins the 248077-into-248320 shape at its production magnitudes over a synthetic 248,077-entry byte-level tokenizer, so a refactor that starts sizing the bias from the matcher fails with the checkpoint's own numbers.
- **#1741 (chunked MRoPE position slice): already correct, and stronger.** `Qwen35Model::forward_with_mrope_state` (`src/models/qwen3_5.rs`) already sliced the stored `position_ids` to `[cache_offset, cache_offset + seq_len)` and additionally validated `pos_shape[1] == batch` before reuse, which is a superset of #1741. It was untestable without a checkpoint, so the reuse decision moved into a pure `mrope_position_source` helper covered by four unit tests (chunked walk, overrun, batch mismatch and wrong rank, nothing stored). Behavior is unchanged.

## Implementation plan

1. **Detection and metadata.** Confirm `qwen3_5` + `vision_config` routes to `ModelType::Qwen35VLM` (it does, `src/models/detection.rs:132-141`) and add a config-shape test pinned to this checkpoint's exact `config.json`, so a future serde change that starts rejecting `output_gate_type` or `language_model_only` fails a test rather than a user's load.
2. **Assert the ignored keys instead of dropping them.** Read `output_gate_type` and reject any value other than `swish`/`silu` with a named error; read `mrope_interleaved` and reject `false`; read `language_model_only` and reject `true` until the vision-less path exists. Cheap insurance, and it converts three silent-wrong modes into named failures.
3. **Remove the token-id fallbacks that can be wrong.** Require `vision_start_token_id` from the config for this family and read `vision_end_token_id`; do not default either to the Qwen3.5-era constants.
4. ~~**Port the two upstream fixes**~~ Both were already implemented. Confirm with evidence, pin each with a regression test, and correct the issue (see the correction under "Two upstream fixes" above).
5. **Docs.** Add Qwen3.8 to `docs/supported-models.md` alongside Qwen3.5-VL, name `mlx-community/Qwen3.8-27B-4bit` as the validated checkpoint with the measured numbers, and state the three unsupported surfaces explicitly: no MTP speculative decoding, no video input, no audio (there is no audio tower).
6. **Test set.** Add the checkpoint to the skip-if-missing real-model tests next to the existing `qwen3.5-27b-4bit` entry.

## Explicitly out of scope, tracked separately

Each of these is a property of the `qwen3_5` family or of mlxcel generally, not of Qwen3.8, and each can ship as its own PR:

- **`reasoning_effort` request surface.** Qwen3.8's chat template is the first shipped template in this tree that *validates* its kwargs, which exposes two defects in the server's template handling. Filed separately.
- **MTP speculative decoding.** `Qwen/Qwen3.8-27B` ships 15 `mtp.*` tensors but every mlx-community MLX conversion drops them, and mlx-community publishes the drafter separately as `Qwen3.8-27B-MTP-bf16` (`model_type: qwen3_5_mtp`). mlxcel has no `qwen3_5_mtp` drafter and no `MtpTarget` for this family. Filed separately.
- **Video input.** Upstream qwen3_5 supports video end to end; mlxcel's Qwen VL path is image-only and rejects `--video` with a named error. Filed separately.
- **Snapshot capacity default.** At an agent-scale prompt this model's multi-turn cache hit rate is 0% under the 512 MiB default. Filed separately.

## Non-goals

- **Not omni.** Despite the family naming, Qwen3.8-27B has no audio tower. `Qwen/Qwen3.8-27B/config.json` top-level keys are exactly `architectures, image_token_id, language_model_only, model_type, text_config, tie_word_embeddings, transformers_version, video_token_id, vision_config, vision_end_token_id, vision_start_token_id` — no `audio_config`, `audio_tower`, `thinker_config`, or `talker_config`, and the weight map has only `model.language_model.*`, `model.visual.*`, `mtp.*`, `lm_head.weight`. It is a vision-only VLM (images and video). `modeltype:omni` does not apply.
- **Not the FP8/NVFP4 builds.** `Qwen/Qwen3.8-27B-FP8` and the ModelOpt NVFP4 variants need the block-dequant plumbing from the still-open upstream PR #1899. Out of scope; the plain mlx-community affine 4-bit build is the target.
- **Not Qwen3.8-2.4T-A95B.** The MoE sibling is a separate size class (`modelsize:xlarge`) and would route through `qwen3_5_moe`.

## Acceptance criteria

- [x] `mlx-community/Qwen3.8-27B-4bit` named in `docs/supported-models.md` with the validated surfaces and the three unsupported ones
- [x] Config-shape test pinned to the checkpoint's `config.json`, covering `output_gate_type`, `language_model_only`, `mrope_interleaved`, absent `mlp_only_layers`, and the nested `generation_config`
- [x] `output_gate_type != swish|silu`, `mrope_interleaved: false`, and `language_model_only: true` each fail with a named error rather than loading
- [x] `vision_start_token_id` required from config for the `qwen3_5` family with no fallback to 248045 (`vision_end_token_id` deliberately not read: nothing in the tree consumes it, so adding the field would be dead code). Read from config with no silent fallback to 248045/248046
- [x] ~~#1741 chunked-MRoPE slice and #1805 padded-vocab mask ported with tests~~ Premise corrected: both were already implemented on `main`. Confirmed with file evidence and pinned with regression tests instead
- [x] Real-checkpoint smoke test added next to the existing `qwen3.5-27b-4bit` entries, skip-if-missing
- [x] CLI text and CLI image re-verified against the checkpoint after the change; decode throughput within noise of the 18.2 tok/s baseline

inureyes:
## Summary

Qualifies `mlx-community/Qwen3.8-27B-4bit` on the existing `qwen3_5` path. Qwen3.8-27B declares `model_type: "qwen3_5"` / `Qwen3_5ForConditionalGeneration` and is architecturally identical to Qwen3.5-27B down to a byte-identical weight-map key set, so it already loaded and ran on `main` with zero code changes. That was accidental rather than guaranteed: `Qwen35Config` carries no `deny_unknown_fields`, so every key the generation added was dropped silently, three of which are load-bearing upstream. This PR registers the checkpoint in the docs and the real-model tests and converts those silent drops into explicit reads or named errors. No new model family, no inference-math change.

## Premise correction: the two upstream ports were already implemented

The issue asked to port two post-pin mlx-vlm fixes. Both were verified present on `main` @ `865bede2` before any code was written, so per CLAUDE.md ("Verify the premise before building. If it does not hold, correcting the premise with evidence *is* the deliverable") the deliverable became confirm, pin, and correct the issue. Neither was re-implemented, and neither was skipped. Both pins were re-verified by mutation testing during finalization (see below): forcing the mask helper to size from the matcher's vocabulary instead of the model's logits axis aborts the process on an MLX broadcast-shape mismatch, and dropping the MRoPE reuse condition's batch check fails its own regression test.

**Blaizzy/mlx-vlm#1805, padded-vocab structured-output mask: already correct.** `apply_structured_mask_to_logits` at `src/server/structured.rs:736-788` already pins the bias length to `vocab_size_hint`, the model's logits axis, not to the matcher's vocabulary (`let vocab_size = vocab_size_hint;` at `:747`), and already masks every position in `[matcher_vocab, vocab_size_hint)` via `let allowed = i < mask_len && constraint.mask_buf[i];` at `:780`. Its doc comment at `:718-735` names this exact case ("the model has padded its embedding table beyond the tokenizer's natural vocabulary"). The scheduler call site at `src/server/batch/scheduler.rs:2988` passes the hint through unchanged. What this PR adds instead is `apply_mask_covers_the_qwen3_8_padded_lm_head` in `tests/structured_outputs.rs`, which builds a synthetic 248,077-entry byte-level tokenizer and pins the real 248077-into-248320 shape: the bias must come back 248320 long, and all 243 unnameable rows must be `-inf`. The pre-existing generic test used a 16-row gap; this one fails with the checkpoint's own numbers if a refactor starts sizing the bias from the matcher. It runs in 0.48s.

**Blaizzy/mlx-vlm#1741, chunked MRoPE position slice: already correct, and stronger.** `Qwen35Model::forward_with_mrope_state` in `src/models/qwen3_5.rs` already sliced the stored `position_ids` to `[cache_offset, cache_offset + seq_len)` during chunked prefill, already validated `pos_shape[1] == batch` before reuse (Blaizzy/mlx-vlm#1040, which #1741 does not do), and already fell back to delta-based compute when the stored tensor no longer covered the window. The comment block already cited the upstream lineage. It was untestable without a 16 GB checkpoint, so the reuse decision moved into a pure `mrope_position_source(stored_shape, batch, seq_len, cache_offset) -> MRopePositionSource` helper covered by four unit tests. Behavior is byte-for-byte unchanged; the bare `#1048` / `#1040` cross-repo refs in that comment were also qualified per CLAUDE.md.

## What changed

- `src/models/qwen3_5.rs`: adds `Qwen35Config::output_gate_type` and `mrope_interleaved()`, a `Qwen35UnsupportedConfig` error enum, `Qwen35Config::validate_supported()` for the text-config keys, `validate_qwen35_wrapper_config()` for the top-level `language_model_only`, and the extracted `mrope_position_source` / `MRopePositionSource` pair that `forward_with_mrope_state` now calls.
- Validation is wired into every Qwen3.5-family parse site, all six of them: `Qwen35Model::load`, the pipeline-stage loader, `src/loading/vlm_qwen.rs` (VLM), `src/loading/special.rs` (from-weights special path), `src/distributed/tensor_parallel/llama_runtime.rs` (TP), and (added in finalization) `src/loading/vlm_special.rs`'s MiniCPM-V 4.6 text-backbone loader. MiniCPM-V 4.6 reuses `Qwen35Config` as its language backbone, and the gated-delta / MRoPE invariants `validate_supported` enforces belong to `Qwen35Model` itself, not to the `qwen3_5` model_type string, so excluding this site was a review finding (S-M1) rather than a deliberate design choice. Both local MiniCPM-V 4.6 checkpoints (bf16 and mxfp4) were checked before wiring the call in: neither declares `output_gate_type` or `mrope_interleaved`, so no working load turns into a startup failure.
- `src/loading/vlm.rs`: replaces `qwen35_vlm_token_defaults()` with `qwen35_vl_token_ids()`, which requires `vision_start_token_id` from the config for this family and keeps the image/video defaults. The stale 248045 fallback is gone. All three vision token ids are now range-checked against the checkpoint's own `vocab_size` (added in finalization; see S-M3 below). The shared `qwen_vl_token_ids` is untouched, so Qwen2-VL, Qwen2.5-VL, Qwen3-VL, Qwen3-VL MoE, and GLM-4V keep their existing defaults. `vision_end_token_id` is deliberately not read: nothing in the tree consumes it, so a field would be dead code.
- `src/models/qwen3_5_tests.rs`: a config-shape fixture pinned to the published `config.json` (inline, so it runs without the checkpoint) asserting that `output_gate_type`, `attn_output_gate`, `mamba_ssm_dtype`, `mtp_num_hidden_layers`, the duplicated top-level `partial_rotary_factor`, the absent `mlp_only_layers`, and the nested `generation_config` all still parse; plus accept and reject cases for each of the three keys, and four `mrope_position_source` tests.
- `src/loading/vlm_tests.rs`: a non-regression test walking the five families that still use `qwen_vl_token_ids`, and three tests for the new Qwen3.5 resolver.
- `tests/tensor_parallel_real_models.rs`: `qwen3_8_27b_tp4_matches_single_rank_stepwise`, next to the existing `qwen3.5-27b-4bit` entry, using the same `repo_model_dir` + skip-if-absent pattern.
- `docs/supported-models.md`: names `mlx-community/Qwen3.8-27B-4bit` as the validated checkpoint with the measured decode and cache-footprint numbers, states the three unsupported surfaces (MTP #1165, video #1166, no audio tower at all), and records that despite the family naming this is a vision-only VLM, not an omni model.

## Finalization (post-review hardening)

Review and security passes reported no CRITICAL or HIGH findings. This pass closed five code findings and corrected three documentation findings; commits `4e2fab1e` (code), `298b511e` (docs + `CHANGELOG.md`), `7a9c6d87` (technical reports).

- **S-M1** (Medium): `validate_supported()` was missing from the sixth `Qwen35Config` parse site, the MiniCPM-V 4.6 text backbone (`src/loading/vlm_special.rs`). Fixed by wiring the call in, right after `text_config` is parsed and before any weight is read. Confirmed safe against both local MiniCPM-V 4.6 checkpoints first (see above).
- **S-M2** (Medium): `validate_qwen35_wrapper_config` and `Qwen35Config::mrope_interleaved()` used `.as_bool()`, so a present-but-wrong-typed value (`language_model_only: "true"`/`1`/`{}`, `mrope_interleaved: "false"`/`0`) silently read as absent and loaded instead of hitting the named error. Fixed with two new `Qwen35UnsupportedConfig` variants (`LanguageModelOnlyWrongType`, `MropeInterleavedWrongType`) and a private strict accessor used only by `validate_supported`.
- **S-M3** (Medium): `qwen35_vl_token_ids` cast `vision_start_token_id`/`image_token_id`/`video_token_id` from `i64` to `i32` with an unchecked truncating `as i32`; `vision_start_token_id: 1099511627776` (2^40) used to truncate to 0 and `-5` loaded unchanged, both with no diagnostic. Fixed with `i32::try_from` plus a range check against the checkpoint's own `vocab_size`, shared by all three ids via a new `qwen35_vl_token_id_in_range` helper.
- **R-M2** (Medium): `output_gate_type` matching was exact and case-sensitive against `{"silu", "swish"}`. Fixed with `eq_ignore_ascii_case`.
- **S-L2** (Low): The `OutputGateType` error echoed the offending config value uncapped; a 1,002,990-byte value reproduced as 1,000,568 bytes of process output. Fixed with a shared `truncate_for_error` helper (64-char cap) used by all three wrong-value error variants.
- **R-M1** (doc): Qwen3.5-VL and Qwen3.8-VL appeared in two consecutive `docs/supported-models.md` bullets. Collapsed to the one detailed bullet.
- **R-L1** (doc): The KV-footprint sentence called the measured 64.1 KiB/token figure "exactly" the 65,536 B/token (64.0 KiB) architectural minimum; they differ by about 0.16%. Reworded to state both numbers as the distinct things they are.
- **R-L2** (doc): The weight-map key list in that paragraph described the upstream `Qwen/Qwen3.8-27B` repo, not the `mlx-community/Qwen3.8-27B-4bit` conversion named as validated in the same sentence. Split the list and attributed each half to the repo it actually describes, verified against both checkpoints' `model.safetensors.index.json`.
- `CHANGELOG.md` gained the `## [Unreleased]` entry for #1163, including the upgrade note that making `vision_start_token_id` mandatory can turn a previously-working deployment into a startup failure if a checkpoint's config omits the key.

Finalization test plan (all commands `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer`, `--release --features metal,accelerate`):

- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate qwen3_5` — 25 passed, 11 ignored (pre-existing serial-MLX gates), 0 failed.
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate loading::vlm` — 178 passed, 0 failed, including a new end-to-end wiring test (`load_minicpmv4_6_vlm_rejects_an_unsupported_output_gate_type`) that drives the real `load_minicpmv4_6_vlm` entry point rather than only the shared config helper.
- [x] `cargo clippy --release --lib --tests --features metal,accelerate -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] Mutation-verified both premise-correction pins, then reverted: forcing `apply_structured_mask_to_logits` to size from the matcher vocabulary aborts `apply_mask_covers_the_qwen3_8_padded_lm_head` on an MLX broadcast-shape mismatch; dropping the `shape[1] == batch` check in `mrope_position_source` fails `mrope_position_source_rejects_a_batch_mismatch_and_a_wrong_rank`.

## Test plan

Every command ran with `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer` and `--release --features metal,accelerate` (the debug profile cannot build on this box).

- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate qwen3_5_tests::` — 11 passed, 11 ignored (pre-existing serial-MLX gates), 0 failed.
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate loading::vlm` — 173 passed, 0 failed.
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate qwen35` — 10 passed, 4 ignored (need CUDA), 0 failed.
- [x] `cargo test --release --features metal,accelerate --test structured_outputs` — 16 passed, 2 ignored (need local weights), 0 failed, including the new padded-lm_head test.
- [x] `cargo clippy --release --features metal,accelerate -p mlxcel --lib --tests -- -D warnings` — clean.
- [x] `cargo fmt` on every touched file; `python3 scripts/ci/check_cross_repo_refs.py` — OK.
- [x] Real checkpoint, CLI text: `mlxcel generate -m models/qwen3.8-27b-4bit -p "What is 17 times 23?" -n 300 --temp 0` answers `391` at 18.81 tok/s, within noise of the 18.21 tok/s baseline recorded in the issue.
- [x] Real checkpoint, CLI image: 224x224 synthetic PNG (red circle, blue square), `--image ... -p "Describe the shapes and their colors" -n 400 --temp 0` inserts 49 image tokens and names both shapes with the correct colors and positions.
- [x] Shared-path non-regression, other Qwen VL families: `models/qwen3.5-27b-4bit` (49 image tokens, both shapes correct), `models/qwen3-vl-4b-4bit` (49 image tokens, both shapes correct), `models/qwen2.5-vl-3b-4bit` (64 image tokens, loads and answers; the answer quality is that 3B model's own, and the code path is untouched by this PR).
- [x] All four rejection paths driven end to end against config variants of the real 16 GB checkpoint (weights symlinked, `config.json` patched): `language_model_only: true`, `output_gate_type: "sigmoid"`, `mrope_interleaved: false`, and a removed `vision_start_token_id` each abort the load with their named error instead of loading.

Not verified here: the server surface (text and image over `/v1/chat/completions`) was not re-run, since the change is confined to config validation at load time and to token-id resolution, both of which the CLI exercises on the same code path. The TP real-model test is `#[ignore]`d in line with its Qwen3.5 sibling and was not run.

Closes #1163


inureyes:

### Intent

Qualify `mlx-community/Qwen3.8-27B-4bit` on the existing `qwen3_5` path: register it in docs and tests, convert three silently dropped config keys into named errors, stop defaulting `vision_start_token_id` for this family, and pin (rather than re-implement) the two upstream fixes that were already present.

### Findings Addressed

None. No CRITICAL or HIGH finding was raised, so no auto-fix commit was made and the branch is unchanged at `34f14df3`.

### Verification performed

Shared VLM path, regression risk:

- `qwen_vl_token_ids` is byte-for-byte identical between `main` and `HEAD` (compared as text, not only as a diff hunk).
- Exactly one call site moved to the strict resolver: `load_qwen3_5_vlm_with_variant`. The other seven `qwen_vl_token_ids` call sites (Qwen2-VL, Qwen2.5-VL, Qwen3-VL, Qwen3-VL MoE, GLM-4V and the remaining two) are untouched.
- The strict resolver is reachable only when `vision_config` is present, because `detect_text_or_vlm` routes on that key. A text-only `qwen3_5` checkpoint never reaches it. Locally, `qwen3.5-0.8b-optiq-4bit` is exactly that shape and is unaffected.
- All 11 local `qwen3_5` / `qwen3_5_moe` checkpoints declare `vision_start_token_id: 248053`, not the seven cited in the PR body.
- No production code in the tree writes a `config.json`, and `sanitize_config_json` only rewrites `Infinity` / `NaN`, so there is no in-tree path that could strip the now-required key.
- MiniCPM-V 4.6 loads through `src/loading/vlm_special.rs` with its own MiniCPMO processor and never calls either token-id resolver, so reusing `Qwen35Config` as a text backbone does not put it on the strict path.

Rejection scoping:

- `output_gate_type`: accepts `silu`, `swish`, and absent; rejects everything else. Accept and reject cases both tested.
- `mrope_interleaved`: accepts `true` and absent; rejects only explicit `false`. Accept and reject cases both tested.
- `language_model_only`: `validate_qwen35_wrapper_config` is called only from the VLM loader, so the text-only path is unaffected. Accept (`false` and absent) and reject (`true`) cases tested.
- `validate_supported()` is wired into five of the six production `Qwen35Config` parse sites; the sixth is MiniCPM-V 4.6, left out deliberately and documented.

Premise-correction pinning tests, checked by mutation:

- Mutating `mrope_position_source` to drop the `shape[1] == batch` check and force the slice window to `[0, seq_len)` fails `mrope_position_source_rejects_a_batch_mismatch_and_a_wrong_rank` and `mrope_position_source_slices_each_chunk_of_a_chunked_prefill`. Both pins are real.
- Mutating `apply_structured_mask_to_logits` to size the bias from the matcher vocabulary instead of `vocab_size_hint` fails `apply_mask_covers_the_qwen3_8_padded_lm_head` with the checkpoint's own numbers (`Shapes (1,248320) and (1,248077) cannot be broadcast`). The pin is real.

Documentation, checked against the artifacts:

- 2180 weight-map keys, byte-identical to `qwen3.5-27b-4bit`: confirmed.
- `mtp.*` present in the HF bf16 repo (15 tensors) and dropped by the mlx-community 4-bit conversion (0): confirmed.
- No `audio_config` / `audio_tower` / `thinker_config` / `talker_config`: confirmed.
- Tokenizer 248,044 vocab plus 33 added tokens equals the 248,077 stated, against `vocab_size: 248320`: confirmed.
- 16 x 4 x 256 x 2 x 2 = 65,536 B/token, and 262,144 tokens x 65,536 B = 16 GiB: confirmed.

### Remaining Items

- [ ] `docs/supported-models.md`: Qwen3.5-VL and Qwen3.8-VL now appear in two consecutive bullets, the summary line and the new detail bullet (MEDIUM). Every other family in that list appears once.
- [ ] `output_gate_type` matching is exact and case-sensitive against `{"silu", "swish"}` (MEDIUM). A checkpoint spelling it `"SiLU"` would hard-fail a load that the implemented path handles correctly. No published checkpoint does this; `eq_ignore_ascii_case` would close it.
- [ ] `docs/supported-models.md`: "64.1 KiB per token, which is exactly ... 65,536 B/token" (LOW). 65,536 B is 64.0 KiB. The measured fit and the architectural minimum differ by about 0.16 percent, so "exactly" overstates it.
- [ ] `docs/supported-models.md`: the weight-map list (`model.language_model.*`, `model.visual.*`, `mtp.*`, `lm_head.weight`) describes the upstream `Qwen/Qwen3.8-27B` repo, while the named validated checkpoint in the same paragraph is the mlx-community 4-bit conversion, whose map is `language_model.*` / `vision_tower.*` with `mtp.*` dropped (LOW).
- [ ] `mrope_position_source` is pinned at the helper level only (LOW). A regression that reverted `forward_with_mrope_state` to a naive head slice, or dropped the helper call, would leave all four unit tests green. Acknowledged in the PR body as a checkpoint-size constraint.
- [ ] Two `.expect()` calls carry the `SliceStored` invariant in `forward_with_mrope_state` (LOW). Provably unreachable and documented, but `expect` in library code is against the project Rust standards.
- [ ] MiniCPM-V 4.6 keeps the old silent-drop behavior for `output_gate_type` and `mrope_interleaved` (LOW). Deliberate and documented; no such checkpoint exists.

### Verification

- [x] Tests pass

inureyes:

No CRITICAL and no HIGH findings, so nothing was auto-fixed and no commits were pushed. Three MEDIUM and three LOW items below, plus the negative results, which are the more useful half of this review.

### Verified clean

**No panic on adversarial config.** 14 malformed `config.json` variants were run through a release binary built from `34f14df3`: `output_gate_type` as number, object, array, bool, and null; `rope_parameters` as string, number, and array; `mrope_interleaved` as string and integer; and 5,000-level nesting. Every one produced a clean `Error:` line. Deep nesting is caught by serde_json's 128-level recursion limit (`parse_vlm_config` uses `from_str`, `src/loading/vlm.rs:166`) and reports `failed to parse config.json` with no stack overflow.

**No terminal escape or log injection.** The only new error interpolating config content is `OutputGateType`, and it uses `{0:?}`. Feeding it a value containing ESC, BEL, CRLF, and a forged log line produced `output_gate_type="\u{1b}[2J\u{1b}]0;pwned\u{7}sigmoid\r\n2026-08-16 ERROR forged log line"` with zero raw ESC bytes in the process output. `Debug for str` routes through `char::escape_debug`, which escapes CR, LF, quote, backslash and unicode-escapes every non-printable, so both the terminal-control and the forged-log-line vectors are closed. Nothing sensitive is echoed: the messages carry a config key, a config value, and an operator-supplied model directory path.

**The two `.expect()` calls are genuinely unreachable.** `stored_shape` is `Some` iff `entry.position_ids` is `Some`, and `mrope_position_source` returns `SliceStored` only for a `Some` shape, so both hold by construction inside a 12-line block. The slice bounds are safe too: `cache_offset` comes from `kv.offset` (non-negative) and `shape[2] >= cache_offset + seq_len` is checked before slicing. No crafted config or cache state reaches them.

**Performance is neutral.** All three validators run once per model load, from the five loader entry points, never per request or per token. `mrope_position_source` is allocation-neutral in the decode path: before, one `Vec<i32>` from `array_shape` inside the `if let Some` arm; after, the same single `Vec<i32>` via `.map()`, still only when `position_ids` is `Some`, plus a free `as_deref()`. Same `&MlxArray` handed to `slice`, no extra copy. `[profile.release]` carries `lto = true` and `codegen-units = 1`, so the extracted free function inlines. `cargo clippy --release --features metal,accelerate --lib --all-targets -- -D warnings` exits 0, and the 11 new `qwen3_5_tests` pass.

### MEDIUM

**M1. `validate_supported()` is missing from the sixth `Qwen35Config` call site.** `src/loading/vlm_special.rs:510` parses a `Qwen35Config` for MiniCPM-V 4.6 and hands it to `models::Qwen35Model::from_weights` at `:556`, which builds the same `RMSNormGated` hardcoded-SiLU gated-delta layers. The PR added the call at the other five sites (`qwen3_5.rs:1519`, `qwen3_5.rs:2250`, `vlm_qwen.rs:401`, `special.rs:118`, `llama_runtime.rs:318`) but not this one, so a MiniCPM-V 4.6 checkpoint declaring `output_gate_type: "sigmoid"` still gets exactly the silently-wrong-output mode this PR set out to close. The invariant belongs to `Qwen35Model`, not to the `qwen3_5` model_type string. One line after the parse.

**M2. Two of the three new guards are bypassed by a type mismatch, silently.** `validate_qwen35_wrapper_config` and `mrope_interleaved()` both use `.as_bool()`, so a present-but-wrong-typed value reads as absent. Confirmed on the release binary: `language_model_only: "true"`, `language_model_only: 1`, `language_model_only: {}`, `mrope_interleaved: "false"`, and `mrope_interleaved: 0` all pass validation and proceed to load. `output_gate_type` behaves differently because it is a typed serde field and correctly rejects a non-string. Three guards documented as equivalent have two different strictness policies, and `mrope_interleaved: "false"` is precisely the "produce silently wrong output" case the error text names.

**M3. `vision_start_token_id` is checked for presence but not for range.** `qwen35_vl_token_ids` does `.as_i64()...? as i32`, an unchecked truncating cast. Verified against the real 16 GB checkpoint with the weights symlinked: `vision_start_token_id: 1099511627776` (2^40, truncates to 0) and `vision_start_token_id: -5` both load and generate with no diagnostic at all. The justification for making the key mandatory is in its own error string, "a wrong start id silently mis-segments vision spans in MRoPE position assignment instead of failing", and an out-of-range value reproduces that failure mode while passing the new guard. `i32::try_from` plus a `0..vocab_size` check would close it. Same pattern on `image_token_id` and `video_token_id` in the same function.

### LOW

**L1. Availability blast radius of the mandatory `vision_start_token_id`.** The requirement lands on the whole family, both `Qwen35VlmVariant::Dense` and `::Moe`, not just Qwen3.8. Any VLM checkpoint routing through `load_qwen3_5_vlm_with_variant` whose top-level config omits the key now fails to start where it previously started on the stale 248045 default. All three local qwen3_5-family VLM checkpoints supply `vision_start_token_id: 248053` at the top level. Text-only `qwen3_5` checkpoints have no `vision_config` and route to the text path, so they are unaffected, and `qwen3_5_mtp` is a different model_type. The change is correctly fail-closed, since a wrong start id degrades silently and that is worse, the message names the exact fix, and the failure is at startup rather than mid-serving. Two residual notes: the lookup is top-level only, so a checkpoint nesting the key under `text_config` fails (no local checkpoint does, and the shared `qwen_vl_token_ids` makes the same assumption), and this is worth a release-note line because it can turn a working server deployment into a startup failure on upgrade.

**L2. Unbounded config value echoed into the error.** Measured: a config with a 1,002,990-byte `output_gate_type` produced 1,000,568 bytes of process output, a 1:1 amplification of config content into logs. Not a DoS vector, because model load is startup-only and the path comes from the CLI rather than a request (`model_path` at `src/server/model_worker.rs:194` is set at worker construction, and there is no request-driven model-load surface). Truncating the quoted value to something like 64 chars would close it.

**L3. The new `.expect()` pair re-derives an invariant the types could carry.** Unreachable as analyzed above, so this is style rather than risk, but it is the one new `expect` the PR adds and the tree's Rust convention is to avoid it in library code. Hoisting the match inside `if let Some(stored_pos) = entry.position_ids.as_ref()` and passing the shape down removes both with no behavior change.

inureyes:
## Orchestrator verification

Full workspace suite re-run after the finalization commits, `cargo test --release --workspace --features metal,accelerate --no-fail-fast` with `DEVELOPER_DIR` set to Xcode 26.6.0.

**Totals: 6059 passed, 4 failed.** The mlxcel lib went 5621 (pre-PR baseline) to 5665, so this PR net-added 44 tests.

Both failure groups are pre-existing and attributed, neither is a regression:

1. **Three `multimodal::video::tests::*`** — ffmpeg 9 removed `-vsync`, which `src/multimodal/video.rs:1123` still passes. Filed as #1172. This PR touches 0 lines of that file. A PATH-controlled experiment on the same binary gives 34 passed / 0 failed with ffmpeg hidden versus 31 passed / 3 failed with ffmpeg 9.0.1 present.
2. **One `hunyuan_vl_parity::text_only_forward_produces_finite_logits`** — this is #997, which names this exact file, line, and assertion (`tests/hunyuan_vl_parity.rs:105`, "text-only logits must be finite") and predates this work. Attribution checked rather than assumed: hunyuan loads through `src/loading/vlm_hunyuan_vl.rs`, untouched here, and five standalone repeats of that single test **on this branch** gave 4 passed / 1 failed, so the flake is present independently of the whole-suite context. New evidence contributed back to #997: the hunyuan case now reproduces in isolation at roughly 20 percent, which contradicts the "passes every time in isolation" premise currently in that issue body.

### Real-checkpoint validation

Server surface re-verified on this branch against `models/qwen3.8-27b-4bit`, since the implementer's report listed it as not run:

| request | result | `usage.prompt_tokens` |
|---|---|---|
| text, temperature 0 | `391` | 66 |
| image (base64 data URI) + text | "The image contains a red circle and a blue square." | 102 |

Both `prompt_tokens` values are byte-identical to the pre-change baseline measured on `main` before this PR, which is the useful part: the token-id resolution change did not shift prompt construction.

### On the premise correction

Worth restating for the record, because it changed what this PR is. Two of the issue's six deliverables asked to port Blaizzy/mlx-vlm#1805 and #1741. Both were **already implemented in this tree**, and #1741's equivalent here is strictly stronger than upstream (it also validates the batch dimension before reusing stored positions). The deliverable became confirm-and-pin, and both pins were mutation-verified twice, independently, by the reviewer and the finalizer.

That correction is the orchestrator's error to own: the issue was written without checking the code closely enough. Recording it here so the next person reads the pins as deliberate rather than redundant.

Issue #1172, opened 2026-08-15
fix(video): every video path is broken on ffmpeg 8+ because -vsync was removed, and the tests hide it by passing when ffmpeg is absent


inureyes:
## Summary

Every video input path in mlxcel is broken on ffmpeg 8 and later. `src/multimodal/video.rs:1123` passes `-vsync vfr` to the `ffmpeg` subprocess; ffmpeg removed the long-deprecated `-vsync` option, so the process exits immediately with

```
Unrecognized option 'vsync'.
Error splitting the argument list: Option not found
```

and every `load_video` / `load_video_source` call fails. This affects the shipped Gemma 4, Gemma 4 Unified, and Kimi-VL video paths, which are the only video-capable families today.

The reason nobody noticed: **the video tests pass by silently skipping when `ffmpeg` is not on PATH**, and CI hosts apparently have no ffmpeg. So the suite has been green against a code path that cannot work on a current ffmpeg.

## Evidence (2026-08-16)

Host: Apple Silicon, ffmpeg 9.0.1 installed from Homebrew this session. Branch under test carried an unrelated change and touches **0 lines** of `src/multimodal/video.rs`.

A PATH-controlled experiment on the same compiled test binary, same commit, varying only whether `ffmpeg` is visible:

| arm | `ffmpeg` on PATH | `cargo test --release -p mlxcel --lib ... multimodal::video::tests` |
|---|---|---|
| A | hidden (`PATH` without Homebrew) | **34 passed, 0 failed** |
| B | ffmpeg 9.0.1 | **31 passed, 3 failed** |

The three failures:

- `multimodal::video::tests::load_video_single_pass_produces_correct_frame_count` (`src/multimodal/video_tests.rs:365`)
- `multimodal::video::tests::load_video_source_fd_variant_decodes_via_dev_fd_n` (`video_tests.rs:611`)
- `multimodal::video::tests::load_video_source_fd_variant_matches_path_variant_frame_count` (`video_tests.rs:655`)

All three carry the identical error payload:

```
Extract { path: "...mlxcel-test-video-64x64-10fps-10s.mp4",
          message: "Unrecognized option 'vsync'.\nError splitting the argument list: Option not found\n" }
```

Direct confirmation against the binary, independent of mlxcel:

```
$ ffmpeg -hide_banner -loglevel error -f lavfi -i "color=c=red:s=32x32:d=1" -vsync 0 -f null -
Unrecognized option 'vsync'.
[...]

PR #1175, opened 2026-08-15
fix(server): fail requests a template refuses, map reasoning_effort

### LINKED ISSUE #1164: fix(server): a chat template that rejects its own kwargs degrades to a silent fallback prompt, and top-level reasoning_effort is ignored
## Summary

Two defects in the server's `reasoning_effort` handling, both found while qualifying `mlx-community/Qwen3.8-27B-4bit` on `main` @ `6e781fcf`:

1. When a chat template rejects one of its own kwargs via `raise_exception`, mlxcel swallows the failure into `render_simple_fallback` and returns **HTTP 200 with a stripped prompt**. The client receives a plausible answer generated from a prompt that has no chat framing, no system message, and no tool declarations. Only a server-side `WARN` records it.
2. The OpenAI-standard top-level `reasoning_effort` request field is accepted and **silently ignored**. It is not mapped to the template kwarg and not rejected.

Together these mean the single most likely client call, `reasoning_effort: "high"` (OpenAI's standard value, which Qwen3.8's template rejects), either does nothing or silently degrades the prompt depending on where the caller puts it.

## Why this surfaces now

Qwen3.8's chat template is the first shipped checkpoint template in this tree that **validates** its kwargs:

```jinja
{%- set resolved_reasoning_effort = reasoning_effort|default('xhigh') %}
{%- if resolved_reasoning_effort not in ('xhigh', 'medium', 'low') %}
    {{- raise_exception('Unexpected reasoning effort ' ~ reasoning_effort ~ '. Supported types are xhigh (default), medium, and low.') }}
{%- endif %}
```

Note the accepted set is `{xhigh, medium, low}`. OpenAI's `reasoning_effort` values are `{minimal, low, medium, high}`. **`high` is valid OpenAI and invalid Qwen3.8**, and `xhigh` is the reverse. Any OpenAI-compatible client that sets the field the standard way hits this.

## Measurements (2026-08-15, `main` @ `6e781fcf`)

Server: `mlxcel-server -m models/qwen3.8-27b-4bit`. Identical request each time, only the kwarg varies, `max_tokens: 1`, message `"hi"`. `prompt_tokens` is the discriminator: the template injects a reasoning-instruction system message whose length is a direct function of the resolved effort, so it can only move if the template actually consumed the value.

| passed as | value | `prompt_tokens` | interpretation |
|---|---|---|---|
| (none) | | 53 | `xhigh` default, instruction injected |
| `chat_template_kwargs` | `xhigh` | 53 | honored |
| `chat_template_kwargs` | `low` | 41 | honored (shorter instruction) |
| `chat_template_kwargs` | `medium` | 11 | honored (template sets no instruction for `medium`) |
| `chat_template_kwargs` | `high` | **7** | **template raised; silent fallback prompt** |
| `chat_template_kwargs` | `HIGH` | **7** | same |
| top-level `reasoning_effort` | `low` | **53** | **ignored** |
| top-level `reasoning_effort` | `high` | **53** | **ignored** |

The 7-token responses are HTTP 200 with a normal-looking completion. The only trace is a server log line:

```
WARN mlxcel::server::chat_request: Chat template render failed, using fallback:
  Failed to render chat template: invalid operation: Unexpected reasoning effort high.
  Supported types are xhigh (default), medium, and low. (in chat:49)
```

`grep -rn "reasoning_effort" src docs` returns **zero matches** at `6e781fcf`, confirming defect 2 is an absence rather than a mis-mapping.

## Defect 1: `raise_exception` should not degrade into a silent fallback

`render_simple_fallback` exists for checkpoints whose template mlxcel cannot render at all, and dropping to a bare prompt is the right call there. It is the wrong call when the template rendered fine and deliberately rejected a caller-supplied value: that is a 400-class user error, and the fallback converts it into a wrong answer the caller cannot detect.

The two cases are already distinguishable. A `raise_exception` from the template is an explicit, intentional signal, and `src/server/chat_template.rs:2162` already documents it as its own category. The fix is to route template-raised errors to an HTTP 400 carrying the template's own message, while leaving the fallback in place for genuine render failures (unsupported filters, malformed templates).

The blast radius is not Qwen3.8-only. Any template that validates `enable_thinking`, a tool-choice value, or a role it does not support has the same silent-degradation path today.

## Defect 2: map or reject the top-level `reasoning_effort`

`ChatCompletionRequest` has no `reasoning_effort` field, so the value is dropped by serde before anything can act on it. Two defensible resolutions:

- **Map it.** Treat top-level `reasoning_effort` as a lower-precedence source for the `reasoning_effort` template kwarg, behind explicit `chat_template_kwargs.reasoning_effort`, and only when the template actually references `reasoning_effort`. This matches the existing precedence chain in `src/server/chat_request.rs:195-202` and the `chat_template_kwargs` design note at `src/server/chat_template_kwargs.rs:17-28`.
- **Reject it.** Return 400 when the field is set and the template does not consume it, so the caller learns the parameter had no effect.

Mapping is preferred: it is what an OpenAI-compatible client expects, and it composes with defect 1's fix, which then turns `high` into an actionable 400 instead of a silent nothing. Do **not** silently translate `high` to `xhigh`; the value sets the model's reasoning budget, and guessing on the caller's behalf is the kind of silent rewrite `resolve_drafter_kind` explicitly refuses to do (`src/lib/mlxcel-core/src/drafter/mod.rs:401-412`).

## Scope note

`preserve_thinking`, the other new Qwen3.8 template kwarg, already works: the prompt-cache-driven `preserve_thinking=true` default (`src/server/chat_request.rs:570-594`) is compatible with the template's `preserve_thinking is undefined or preserve_thinking is true` guard, and the CLI help already documents it. No change needed there.

## Acceptance criteria

- [x] A template-raised `raise_exception` returns HTTP 400 with the template's message, not a 200 with a fallback prompt
- [x] Genuine render failures (unsupported construct, malformed template) still fall back as today, with a test proving the two paths diverge
- [x] Top-level `reasoning_effort` reaches the template as a kwarg, at lower precedence than `chat_template_kwargs.reasoning_effort`, and only when the template references it
- [x] No value translation: `high` against a template that rejects `high` produces a 400 naming the accepted set, never a silent remap to `xhigh`
- [x] Regression test pinned to the Qwen3.8 chat template covering all of `xhigh` / `medium` / `low` / `high` / unset, asserting the rendered prompt length changes for the valid values and 400s for the invalid one
- [x] Streaming path returns the same 400 before the first chunk, not mid-stream

inureyes:
## Summary

Two defects in the server's `reasoning_effort` handling, both reachable from an ordinary OpenAI-compatible client. A chat template that refuses a caller-supplied kwarg through Jinja's `raise_exception` was swallowed into `render_simple_fallback` and answered with HTTP 200 from a stripped prompt; that is now a 400 carrying the template's own message. And the OpenAI-standard top-level `reasoning_effort` field was accepted and silently dropped; it now reaches the template as a kwarg, without value translation.

## What changed

**`src/server/chat_template.rs`** adds a private `TemplateRejection` sentinel that `raise_exception` attaches as the `source` of the `minijinja::Error` it raises, plus `template_rejection_message(&anyhow::Error)` which recovers it by walking the error chain. This is the discriminator between "the template deliberately refused this input" and "mlxcel could not render this template". `ErrorKind::InvalidOperation` is too coarse to serve as one, because minijinja raises that same kind for genuine engine problems and keying on it would convert real render failures into 400s. Message-substring matching is worse, because the template author writes that text. The sentinel is exact and needs neither. It also adds `template_mentions(name)`, the same shape of substring probe as `supports_image_content`, and generalizes the existing log-key truncation into `truncate_chars`, which bounds the partly client-controlled rejection message at 512 chars before it reaches the HTTP body.

**`src/server/chat_request.rs`** adds `reject_if_template_rejection`, called at both fallback sites in `prepare_chat_request_with_cache`: the typed path at what was line 296 and the structurally identical raw/multimodal path at what was line 283. Both needed it; fixing one would have left the defect reachable through any request carrying tool calls, a prior-turn `reasoning` field, or typed media parts. The error propagates out of `prepare_chat_request_with_cache`, which the chat, Responses, and Anthropic routes already map to a 400 before generation starts, so no route wiring changed and the streaming path returns the 400 in place of the SSE stream rather than mid-stream. It also adds `resolve_effective_kwargs`, now the single source of truth for the kwargs map, and `map_reasoning_effort_kwarg`, which applies the new field under three guards: an explicit `chat_template_kwargs.reasoning_effort` wins, the template must mention the name, and the value is never translated.

**`src/server/types/request.rs`** adds `reasoning_effort: Option<String>` to `ChatCompletionRequest` and `resolve_reasoning_effort()`, which follows the same three-tier chain as the existing `resolve_prompt_cache_key` and `resolve_user`: top-level field, flattened OpenAI-SDK `extra_body`, nested `extra_body`.

**`src/server/routes/chat.rs`** has `build_prompt_cache_request_context` call `resolve_effective_kwargs` instead of re-deriving the merge, so a mapped `reasoning_effort` participates in `template_sig` and two requests differing only in effort do not share a cache bucket. The prompt-cache `preserve_thinking` defaulting is deliberately left out of the shared helper: folding it in would change every existing deployment's `template_sig` on upgrade, which this change has no reason to cause.

**`src/server/chat_template_kwargs.rs`** gains a generic `set(key, value)` next to `set_preserve_thinking`, and its precedence documentation gains the new source.

**`src/server/reasoning_effort_tests.rs`** is new: 14 cases against the Qwen3.8 chat template, vendored as `tests/fixtures/qwen3_8/chat_template.jinja` and pinned by SHA-256 the way the Muse Glimmer fixture already is. Docs updated in `docs/supported-models.md` and `docs/responses-api.md`.

## On not translating values

OpenAI's `reasoning_effort` vocabulary is `minimal` / `low` / `medium` / `high`. Qwen3.8's template accepts `xhigh` / `medium` / `low`. So `high` is valid OpenAI and invalid there, and `xhigh` is the reverse; `high` is also the single most likely value a client actually sends. Remapping it to `xhigh` would silently set a reasoning budget the caller did not ask for and could not detect, which is the same silent rewrite `resolve_drafter_kind` explicitly refuses to perform. The value goes through verbatim, the template decides, and a refusal surfaces as a 400 naming the set the template accepts.

## Real-checkpoint validation

`models/qwen3.8-27b-4bit`, identical request each time, only the effort varies, `max_tokens: 1`, message `"hi"`. `prompt_tokens` is the discriminator: the template injects a reasoning-instruction system message whose length is a direct function of the resolved effort, so it can only move if the template actually consumed the value. "Before" is the pre-change release binary from `main` @ `0de77ced`; "after" is this branch. Both were measured on the same box in the same session.

| passed as | value | before | after |
|---|---|---|---|
| (none) | | 200 / 53 | 200 / 53 |
| `chat_template_kwargs` | `xhigh` | 200 / 53 | 200 / 53 |
| `chat_template_kwargs` | `low` | 200 / 41 | 200 / 41 |
| `chat_template_kwargs` | `medium` | 200 / 11 | 200 / 11 |
| `chat_template_kwargs` | `high` | 200 / 7 | **400** |
| `chat_template_kwargs` | `HIGH` | 200 / 7 | **400** |
| top-level `reasoning_effort` | `low` | 200 / 53 | **200 / 41** |
| top-level `reasoning_effort` | `medium` | 200 / 53 | **200 / 11** |
| top-level `reasoning_effort` | `xhigh` | 200 / 53 | 200 / 53 |
| top-level `reasoning_effort` | `high` | 200 / 53 | **400** |
| both: kwargs `low`, top-level `medium` | | 200 / 41 | 200 / 41 |
| streaming, top-level `high` | | 200, SSE stream | **400, no stream opened** |

The 400 body is `{"error":{"message":"the model's chat template rejected this request: Unexpected reasoning effort high. Supported types are xhigh (default), medium, and low.","type":"invalid_request_error","code":null}}`. The server log line changed from `WARN ... Chat template render failed, using fallback: ...` to `INFO ... Chat template rejected the request: ...`, and no fallback warning remains for any of these cases.

Control: `models/qwen3-0.6b-4bit`, whose template never mentions `reasoning_effort`, returns 200 with 9 prompt tokens for all twelve cases above, byte-identical before and after. Nothing about a checkpoint that ignores the field changed.

## What I did not change

The Responses API's `reasoning.effort` is still advisory and is not mapped onto the template kwarg, so a Responses client sending `effort: "high"` at Qwen3.8 still gets a silent 200 rather than the 400 the chat-completions field now produces. Mapping it is a one-line change in `responses_translator.rs`, but it changes documented behavior on a surface this issue did not scope, so it is called out in `docs/responses-api.md` instead and left for a follow-up decision.

The CLI has the same silent-degradation shape at `src/commands/generate.rs::apply_user_chat_template` (`.unwrap_or_else(|_| user_prompt.to_string())`). It is a lower-stakes surface, since the operator sees the output rather than an HTTP client that cannot tell, and it is out of this issue's server scope.

## Finalization (post-review hardening)

Review and security passes reported two HIGH findings, both fixed before this finalization pass: a warm-up-kwargs divergence where `render_next_turn_history` derived its chat-template kwargs by hand instead of through `resolve_effective_kwargs`, so it missed a mapped top-level `reasoning_effort` and could file a warmed prefill vector under a prompt-cache bucket whose `template_sig` did not match it (`8dac53ea`); and a missing `CHANGELOG.md` entry (`2bd259b6`). This pass found and fixed one further defect this PR itself introduced, closed one security finding, and corrected two documentation findings; commits `3c880883`, `e355d4c2`, `18ed6fe4` (code + docs), `cb6f6ae4` (technical reports).

- **Availability regression** (Medium-High, treated as top priority): `tests/fixtures/muse_glimmer/chat_template.jinja`'s `render_atem` macro raises when `tool_call.function.arguments` is not a mapping. `normalize_tool_call_arguments` only rewrote a wire-format `arguments` string into an object when it parsed as valid JSON *and* that JSON was an object, so `arguments: ""` (the common zero-argument spelling agentic clients echo back) and `arguments: "null"` stayed strings and hard-failed the request with a 400 the caller could not act on, sticky across every later turn in the same conversation. Fixed: both spellings now normalize to `{}`. `"[1,2]"`, bare scalars, and malformed/truncated JSON are deliberately left untouched, since there is no safe reading of them as "no arguments". New unit tests in `chat_request_tests.rs` (empty, whitespace-only, `"null"`, malformed-still-a-string) and an end-to-end regression test in `muse_atem_roundtrip_tests.rs` that replays an empty-arguments tool call through the real Muse Glimmer fixture template.
- **Log-injection sink** (Medium): `reject_if_template_rejection` logs up to 512 chars of partly caller-controlled text into a single-line plaintext `tracing_subscriber::fmt` record; a newline could forge a fake log line and an ANSI escape could rewrite terminal/log-viewer state in `--log-file` output. Not a regression (the pre-PR `WARN ... using fallback: {:#}` line carried the same text untruncated), but `truncate_chars` already owns the length cap this PR added, so the fix folds a control-character filter into that same helper, closing both of its call sites (`TemplateRejection::new`, `truncate_key_for_log`) at once. New tests pin the 512-char cap, a UTF-8 boundary case where the 512th character starts a run of 3-byte characters, and the control-character filter on both raw `truncate_chars` output and a `TemplateRejection` message.
- **Doc accuracy** (Low x2): `MAX_MESSAGE_CHARS`'s doc comment claimed the server-side log line was unchanged by the cap; on the rejection path it is not; the `WARN` line is now unreachable there and the message logs truncated at `INFO` instead. `docs/supported-models.md` called `TemplateRejection` a "private sentinel"; the type is `pub` and nameable outside the crate, only its constructor and field are private, which is the privacy that actually matters.
- The CHANGELOG upgrade note for #1164 named "an unsupported role, non-alternating roles, a tool declaration it refuses, an unknown kwarg value" as the traffic that turns from 200 into 400. For the four Qwen3.8/Qwen3.5-27B checkpoints in this tree the two highest-traffic new 400s are neither: a `system` message at any index other than 0 (`'System message must be at the beginning.'`, fires on a mid-conversation system-reminder injection or any second system message) and a conversation with no `user` message at all (`'No user query found in messages.'`, covers assistant-prefill and system-only requests, also fires when every user turn is `<tool_response>`-wrapped). Both are sticky per conversation. Also names that `qwen3.5-27b-4bit` carries eight of the Qwen3.8 template's nine `raise_exception` sites but not its `reasoning_effort` guard, so it gains the role/ordering 400s without gaining the effort one.

**Explicitly not fixed here:** `src/server/router_front.rs:657` maps a template rejection to 500 rather than 400 on the disaggregated router-front surface. Real (the OpenAI SDK retries 5xx and not 4xx, so one bad request becomes three plus a false server-error alert), but pre-existing and on a different surface than this PR touches. Tracked as #1176.

Finalization test plan (all commands `DEVELOPER_DIR=/Applications/Xcode-26.6.0.app/Contents/Developer`, `--release --features metal,accelerate`):

- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::chat_request` — 87 passed, 0 failed.
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::muse` — 17 passed, 0 failed, including the new `atem_replay_with_empty_string_tool_call_arguments_does_not_raise`.
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::chat_template` — 104 passed, 1 ignored (local-models audit), 0 failed.
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::reasoning_effort_tests` — 19 passed, 0 failed.
- [x] `cargo clippy --release -p mlxcel --lib --features metal,accelerate --tests -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] `python3 scripts/ci/check_cross_repo_refs.py` — advisory-only pass, all flagged refs are same-repo `lablup/mlxcel`.

## Test plan

- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::reasoning_effort_tests`: 14 passed, 0 failed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::chat_request`: 84 passed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::chat_template`: 100 passed, 1 ignored
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::routes`: 118 passed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::prompt_cache`: 168 passed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::types`: 69 passed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::media`: 36 passed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::muse`: 16 passed
- [x] `cargo test --release -p mlxcel --lib --features metal,accelerate server::request_options`: 39 passed
- [x] `cargo build --release --features metal,accelerate --bin mlxcel-server`: clean
- [x] `cargo fmt`
- [x] Real-checkpoint validation against `models/qwen3.8-27b-4bit` and `models/qwen3-0.6b-4bit`, table above

Closes #1164


inureyes:

### Intent

> A chat template that refuses a caller-supplied value must fail the request with 400 instead of answering 200 from a stripped prompt, and the OpenAI-standard top-level `reasoning_effort` field must reach the template as a kwarg without value translation.

### Focus: can a genuine engine failure now produce a 400?

No. Verified against the pinned minijinja 2.21.0 source rather than accepted on the description:

- `Error::source()` returns whatever `with_source` attached, and `TemplateRejection` is constructible only inside `chat_template.rs` (private `new`, private field), so nothing but `raise_exception` can put one in a chain.
- `process_err` (`vm/mod.rs:1142`) annotates a propagating error **in place** (`set_filename_and_span` / `set_filename_and_line`); it never rebuilds it, so the source survives ordinary propagation.
- The three wrapping sites are `BadInclude` (`vm/mod.rs:890`), `EvalBlock`/`super()` (`vm/mod.rs:950`), and loop recursion (`vm/mod.rs:1068`). All three use `.with_source(err)`, so the chain walk reaches the sentinel. Two of them are unreachable here anyway: `ChatTemplateProcessor` builds a fresh single-template environment with only `"chat"` registered, so `{% include %}` and `super()` have no target.
- Both `apply*` paths wrap with `.with_context(...)`, which preserves `source()` links, so `anyhow::Error::chain()` reaches the sentinel from the outermost error.
- Unimplemented filter, malformed template, and fuel exhaustion all carry no sentinel and still fall back. The first two were already pinned; fuel exhaustion is now pinned too (see below).

### Findings addressed

- [x] **The next-turn prompt-cache warm-up rendered with different kwargs than the request it mirrors** (HIGH, `8dac53ea`). `render_next_turn_history` still derived its kwargs by hand with `extract_request_kwargs` + `merge_server_and_request`, so it never saw the mapped `reasoning_effort`, while the bucket its probes are filed under (`build_prompt_cache_request_context`) does include it in `template_sig`. A request setting top-level `reasoning_effort: low` against Qwen3.8 therefore prefilled and stored a vector rendered at the template's `xhigh` default under a bucket that said `low`, and the next turn matched nothing past the shared head. That is the exact disagreement the function's own comment three lines below already guards against for `preserve_thinking`. Now routed through `resolve_effective_kwargs`, with the `preserve_thinking` default left where it was so no existing deployment's `template_sig` moves on upgrade. Independently flagged by Codex as P2.
- [x] **Missing CHANGELOG entry** (HIGH, `2bd259b6`). A client-visible 200 to 400 change plus a new request field, with no `[Unreleased]` bullet. The two merges immediately preceding this one both carried theirs in the PR (`0de77ced` for #1163, `865bede2` for #1168). Added under the existing `### Fixed` heading with an `**Upgrade note:**` for deployments whose template validates caller input.

Five regression tests were added with the first fix (`server::reasoning_effort_tests` 14 to 19):

- `next_turn_warmup_renders_with_the_mapped_reasoning_effort` plus an unset-effort negative control. Both assertions fail on the pre-fix code, in opposite directions.
- `nested_raise_exception_is_still_a_template_rejection`: the sentinel survives a `{% for %}` body and a `{% macro %}` body. The macro shape is not hypothetical, the Qwen3.8 fixture's own `render_content` raises that way.
- `fuel_exhaustion_is_not_a_template_rejection`, asserting `ErrorKind::OutOfFuel` as well as `is_none()` so a parse error cannot pass it.
- `template_sig_separates_two_top_level_efforts`, with the equality control against a template that never mentions the name.

### Remaining items

- [ ] **Disaggregated router front reports a template rejection as 500, not 400** (MEDIUM). `router_front.rs::route_chat` does `prepare_chat_request_with_cache(...).map_err(|e| anyhow::anyhow!("{e}"))?` at line 657, and the caller maps every `Err` to `INTERNAL_SERVER_ERROR`. This is a pre-existing generic mapping that already swallowed image-fetch and `image_soft_tokens` client errors the same way; this PR adds one more class to it rather than introducing it. Fixing it means reworking that surface's error mapping, which is outside this issue.
- [ ] **Vendored `tests/fixtures/qwen3_8/chat_template.jinja` carries no upstream derivation notice or `NOTICE` entry** (MEDIUM, raised by Codex as P1). Left deliberately: the tree's actual precedent is against it. `tests/fixtures/muse_glimmer/chat_template.jinja` (Meta), `tests/fixtures/phi4mm_audio_config.json` (Microsoft), and `tests/fixtures/wikitext2_excerpt.txt` are all vendored third-party checkpoint metadata with no `NOTICE` entry, and CLAUDE.md's third-party rule is written for ported or adapted **source code**. Attributing only this one file would leave the tree inconsistent. This wants a maintainer decision covering all four fixtures at once.
- [ ] **`configure_environment` has no `// Used by:` annotation** (LOW, raised by Codex as P3). Declined. `docs/code-guidelines.md` scopes that convention to functions "that multiple models depend on" (attention, normalization, KV cache, activations) under `src/models` and `mlxcel-core`. This is a file-private helper with two in-file callers that predates the PR.
- [ ] **`TemplateRejection` is documented as a "private sentinel" but is publicly reachable** (LOW). `src/server/chat_template.rs` is `pub mod` under `pub mod server`, so `mlxcel::server::chat_template::TemplateRejection` is in the library crate's API surface. The privacy that matters holds (private constructor and field, so no external code can forge one); only the wording is loose.
- [ ] **`MAX_MESSAGE_CHARS` doc comment is stale** (LOW). It says "the server-side log line is unchanged", but a rejection now logs the truncated message at INFO instead of the previous untruncated `WARN`. The behavior is better than documented, the comment just no longer describes it.
- [ ] **Empty-string asymmetry** (LOW). `resolve_reasoning_effort` skips empty strings, so top-level `reasoning_effort: ""` is silently ignored, while `chat_template_kwargs: {"reasoning_effort": ""}` reaches the template and 400s. Deliberate and documented, but the two spellings of "empty" behave differently.
- [ ] **No route-level test for the streaming 400** (LOW). Structurally guaranteed (both `stream_chat_completion` and the Responses/Anthropic streaming handlers return `ErrorResponse::new(...).into_response()` before any SSE construction) and validated on the real checkpoint, but not pinned by a test. A route test needs a loaded model, so this is a real cost rather than an oversight.
- [ ] **`resolve_reasoning_effort`'s `extra_body_fields` tier is unreachable through deserialization** (LOW). `reasoning_effort` is now a named field, and `#[serde(flatten)]` only collects unmatched keys, so that branch can only fire for a hand-built request. Harmless, and identical to the pre-existing `resolve_user` / `resolve_prompt_cache_key` shape it deliberately copies.

### Verification

- [x] All stated requirements implemented (all six acceptance criteria from the brief)
- [x] Integrated into project code flow: both fallback sites in `prepare_chat_request_with_cache` (typed and raw/multimodal), and no third server-side `render_simple_fallback` site exists. All four route families map `Err` to 400 before generation starts: chat non-streaming and streaming, Responses non-streaming and streaming, Anthropic messages non-streaming, streaming, and `count_tokens`.
- [x] Precedence and gating correct: `chat_template_kwargs.reasoning_effort` wins over the top-level field, which wins over the server-wide `--chat-template-kwargs` default; the mapping is gated on `template_mentions("reasoning_effort")`, and `reasoning_effort` is not in `RESERVED_KEYS` so it reaches the Jinja context.
- [x] No value translation: `high` produces the 400 naming the accepted set.
- [x] Prompt-cache interaction: the mapped effort reaches `template_sig` through `canonicalize_map(kwargs.as_map())`, now pinned by test. `preserve_thinking` is genuinely unchanged, `build_prompt_cache_request_context` did not apply the prompt-cache default before this PR and still does not, so `template_sig` is byte-identical for any request that sets no `reasoning_effort`.
- [x] Project conventions followed: `// Used by:` annotations on the new shared functions, tests beside the code, fixture pinned by digest like `muse_glimmer`, docs updated, CHANGELOG entry added.
- [x] Existing modules reused where applicable (`extract_request_kwargs`, `merge_server_and_request`, `template_sig`, the `resolve_user` / `resolve_prompt_cache_key` three-tier pattern, the `supports_image_content` gating shape)
- [x] Tests pass: `server::reasoning_effort_tests` 19 passed, `server::chat_request` 84 passed, `cargo fmt --check` clean. Workspace suite 6073 passed with 3 pre-existing `multimodal::video` failures tracked in #1172.

inureyes:

Reviewed at `2bd259b6`. **No CRITICAL or HIGH findings, so no code was changed.** Six MEDIUM/LOW items below, all reported rather than fixed.

### Verified clean

**Client-controlled input surface.** `reasoning_effort` is bounded only by axum 0.7's default 2 MiB body limit (`create_app` raises the limit on the audio sub-router only, and never disables it), which is the same bound message content already has. `set_fuel(Some(50_000_000))` bounds VM instruction count, not per-instruction string size, so it does not bound this value; what does is that a large value is a single concat, not a loop. Worst case against the Qwen3.8 guard is roughly 3x2 MiB transient per in-flight request (the value, the concatenated message, the `minijinja::Error` copy), all dropped before the 400 is built. No amplification over a 2 MiB user message. `template_sig` hashing is O(n) blake3 over a value that is already in memory.

**Cache-key poisoning.** `template_sig` folds the canonicalized kwargs map, so two efforts cannot share a bucket, confirmed by `template_sig_separates_two_top_level_efforts`. The dangerous direction (same sig, different rendered prompt) is closed: all five `prepare_chat_request_with_cache` call sites, `build_prompt_cache_request_context`, and `render_next_turn_history` pass `&state.chat_template`, so the `template_mentions` gate cannot disagree between the sig and the render. A false-positive `template_mentions` yields a distinct sig for an identical prompt, which is a wasted bucket and not a collision. `template_sig` only ever refines the key, so it cannot weaken the `session_key` (`prompt_cache_key` then `user` then anonymous sentinel) separation. A rejected request 400s before generation and therefore never inserts.

**Information disclosure in the 400 body.** The Jinja context is exactly `messages`, `bos_token`, `eos_token`, `add_generation_prompt`, `tools`, `enable_thinking`, the merged kwargs, and the derived `thinking` / `thinking_mode` aliases. No filesystem path, no config, no other request. The only non-caller data a template could echo into a rejection is the operator's own `--chat-template-kwargs` / `LLAMA_ARG_CHAT_TEMPLATE_KWARGS` defaults, which a template can already route into the prompt anyway. `reject_if_template_rejection` rebuilds the error with `anyhow::bail!` from the sentinel message alone, so the `minijinja` Display and its `(in chat:NN)` line annotation do not reach the client. `truncate_chars` uses `chars().take(n).collect()`, so the 512-cap cannot panic on a UTF-8 boundary. Terminal escapes cannot ride the HTTP body: `serde_json` escapes every scalar below 0x20, including ESC.

**Performance.** `map_reasoning_effort_kwarg` early-returns on `resolve_reasoning_effort()` before touching the template, so a request without the field pays nothing. When the field is present, the added work is one `str::contains` per `resolve_effective_kwargs` call, up to three per request, over a template source `template_sig` already blake3-hashes on the same path. No new clone: `resolve_effective_kwargs` runs the same `extract_request_kwargs` + `merge_server_and_request` the three call sites each ran inline before, and `merged_extra_body()` still returns `None` without allocating when neither `extra_body` shape is present.

### Findings

**MEDIUM 1: the disaggregated router front answers a template refusal with 500.** `src/server/router_front.rs:657` maps every `prepare_chat_request_with_cache` error through `anyhow::anyhow!` into `router_chat_completions`' `INTERNAL_SERVER_ERROR` arm. The CHANGELOG already calls this out as deliberate surface consistency, and it is, but the consequence changed with this PR: the OpenAI SDK retries 5xx and does not retry 4xx, so `reasoning_effort: "high"` against a disaggregated deployment now costs three requests instead of one and raises a server-error alert for what is a client error. Consider giving `route_chat` the same `ErrorResponse::new(..., "invalid_request_error")` arm the other three surfaces use, gated on `template_rejection_message`.

**MEDIUM 2: the Muse Glimmer ATEM raise is an internal contract, not a caller-input refusal.** `tests/fixtures/muse_glimmer/chat_template.jinja` (compiled into the binary at `src/server/muse_atem_stream_support.rs:11`) raises `'Onyx ATEM chat template requires tool_call.function.arguments to be a dict (mapping)...'` when `args is not mapping`. `normalize_tool_call_arguments` (`src/server/chat_request.rs:1078`) only rewrites a string that parses to a JSON **object**, so `arguments: ""` (the common zero-argument spelling), `"null"`, `"[1,2]"`, and a truncated `"{\"a\": 1"` all stay strings and now 400. This one raise is not the caller refusing anything: the client sent valid OpenAI wire format and mlxcel's own normalizer left it in a shape mlxcel's own template refuses, so the client reads an unactionable message about the HF jinja sandbox. It is also sticky, since the offending `tool_calls` entry lives in the conversation history and every later turn replays it. mlxcel never emits this itself (`atem.rs::parse_parameters` serializes an empty map to `"{}"`), so it takes a client that builds `tool_calls` by hand. Narrow fix: normalize an empty or whitespace-only `arguments` string to `{}` in `normalize_tool_call_arguments`.

**MEDIUM 3: the CHANGELOG upgrade note does not name the two cases operators will actually hit.** It lists "an unsupported role, non-alternating roles, a tool declaration it refuses, an unknown kwarg value". For the four Qwen3.8/Qwen3.5-27B checkpoints in this tree the two highest-traffic new 400s are neither: a `system` message at any index other than 0 (`'System message must be at the beginning.'`, which fires on the common mid-conversation system-reminder injection and on any second system message), and a conversation with no `user` message at all (`'No user query found in messages.'`, which covers assistant-prefill and system-only requests, and also fires when every user turn is `<tool_response>`-wrapped). Both are sticky per conversation. Note also that `qwen3.5-27b-4bit` carries eight of the nine raise sites but not the `reasoning_effort` guard, so it gains the role/ordering 400s without gaining the effort one. Worth naming these explicitly in the upgrade note.

**MEDIUM 4: the new INFO log line is a log-injection sink.** `reject_if_template_rejection` logs up to 512 chars of partly caller-controlled text into a single-line plaintext `tracing_subscriber::fmt` record, and newlines or ANSI escapes in a `reasoning_effort` value ride through into `--log-file` output verbatim. This is not a regression, since the pre-PR `WARN ... using fallback: {:#}` line carried the same text untruncated, and the PR narrows it. But the PR is already touching this exact helper, so folding a control-character filter into `truncate_chars` alongside the length cap would close it for both call sites at once. Precedent for the filter exists at `src/server/florence2_worker.rs:134`.

**LOW 5: `MAX_MESSAGE_CHARS`'s doc comment is stale.** It says "The unbounded original still reaches the `minijinja::Error` itself, so the server-side log line is unchanged." On the rejection path there is no longer a log line carrying it: `reject_if_template_rejection` logs the truncated message and bails, and the `WARN ... using fallback` line is now unreachable for rejections. The untruncated message survives only in the discarded error value and in the `warmup: this-turn history render failed` DEBUG line.

**LOW 6: no test pins the 512-char cap.** `reasoning_effort_tests.rs` has 14 good cases but none exercises `TemplateRejection::new` at the boundary. The `chars().take()` construction makes a UTF-8 panic impossible today, but nothing stops a later refactor to byte slicing, and nothing asserts the cap applies to the HTTP body while the log keeps its own. Two cases would cover it: a >512-char value, and a multi-byte value whose 512th char straddles a boundary.

### Verification

- `cargo test --release --features metal,accelerate -p mlxcel --lib` over `server::reasoning_effort_tests`, `server::chat_request`, `server::chat_template`, `server::routes::chat`, `server::prompt_cache::key`: 277 passed, 0 failed, 1 ignored.
- `cargo clippy --release --features metal,accelerate -p mlxcel --lib --all-targets`: clean.
- Full-tree survey of every file containing `raise_exception`, including `models/` symlink targets and the `references/` upstream clones. Upstream `mlx-lm` and `mlx-vlm` contain zero raising templates; the reachable set is the four Qwen3.8/3.5-27B checkpoints and Muse Glimmer.

inureyes:
## Orchestrator verification

Suite at `cb6f6ae4`, `cargo test --release --workspace --features metal,accelerate --no-fail-fast`: **6042 passed, 3 failed**. All 3 are `multimodal::video::tests::*`, pre-existing, ffmpeg 9 removed `-vsync` (#1172). This PR touches 0 lines of `src/multimodal/video.rs`. No other failure, and the known #997 hunyuan flake did not fire this run.

### End-to-end against the real checkpoint

`mlxcel-server -m models/qwen3.8-27b-4bit`, built from `cb6f6ae4`. The `prompt_tokens` column is the discriminator: the template's reasoning-instruction system message length is a direct function of the resolved effort, so it can only move if the template actually consumed the value.

| request | before (from #1164) | after |
|---|---|---|
| `chat_template_kwargs.reasoning_effort: xhigh` | 200, pt 53 | 200, pt 53 |
| `chat_template_kwargs.reasoning_effort: low` | 200, pt 41 | 200, pt 41 |
| `chat_template_kwargs.reasoning_effort: medium` | 200, pt 11 | 200, pt 11 |
| `chat_template_kwargs.reasoning_effort: high` | **200, pt 7 (silent fallback)** | **400**, message names the accepted set |
| top-level `reasoning_effort: low` | **200, pt 53 (ignored)** | **200, pt 41** |
| top-level `reasoning_effort: high` | **200, pt 53 (ignored)** | **400** |

Both defects in the issue are closed, and the valid values are byte-identical to the pre-change baseline.

**Streaming**, which mattered because a 400 emitted mid-stream would be worse than the original bug:

```
HTTP/1.1 400 Bad Request
content-type: application/json
```

Body is a JSON error object, and the response contains **zero** `data:` lines, so the rejection lands before any SSE frame.

**Control, a model whose template never mentions `reasoning_effort`** (`models/qwen3-0.6b-4bit`): top-level `low`, `high`, and `xhigh` all return 200 with `prompt_tokens` 9, unchanged. The gate does not leak a kwarg into templates that do not ask for one.

### Two findings worth carrying forward

**The reviewer found a live prompt-cache correctness bug, not a style issue.** `render_next_turn_history` was a third kwargs-derivation site that the first implementation missed, so a request setting `reasoning_effort: low` warmed a vector rendered at the template's `xhigh` default into a bucket whose `template_sig` said `low`. The next turn then matched nothing past the shared head. Fixed in `8dac53ea` with regression tests that fail in opposite directions on the pre-fix code.

**The security review found an availability regression this PR introduced**, fixed in `3c880883`. The Muse Glimmer ATEM template raises when `tool_call.function.arguments` is not a mapping, and `normalize_tool_call_arguments` only rewrote strings that parse to a JSON object. So `arguments: ""`, the common zero-argument spelling, went from a degraded fallback answer to a hard 400 on input that is valid OpenAI wire format, stickily, since the offending entry replays from conversation history every turn. That raise is an internal contract rather than a caller refusal, so it should never have reached the client. Empty, whitespace-only, and `"null"` now normalize to `{}`; `"[1,2]"`, scalars, and malformed JSON deliberately stay strings.

### Deliberately out of scope, tracked

`src/server/router_front.rs:657` still maps a template rejection to 500 on the disaggregated surface, which matters because the OpenAI SDK retries 5xx and not 4xx. Pre-existing generic mapping, recorded on #1176 alongside the `/v1/responses` and offline-CLI gaps that swallow the same rejection.

One item left for a maintainer rather than decided here: the vendored `tests/fixtures/qwen3_8/chat_template.jinja` carries no derivation notice, matching `muse_glimmer/chat_template.jinja`, `phi4mm_audio_config.json`, and `wikitext2_excerpt.txt`, which are all unattributed the same way. CLAUDE.md's third-party rule is written for ported source code rather than vendored checkpoint metadata. Attributing only the new one would leave the tree inconsistent, so this wants one decision covering all four fixtures.

PR #1180, opened 2026-08-16
fix(test): gate the loading-side pinned Muse Glimmer guard on MLXCEL_REQUIRE_PINNED_CHECKPOINTS

### LINKED ISSUE #1177: fix(test): gate the loading-side pinned Muse Glimmer guard on MLXCEL_REQUIRE_PINNED_CHECKPOINTS
## Problem / Background

PR #1173 (issue #1161, merged as `0adb53c5`) hardened the vision-side pinned checkpoint test. It introduced an opt-in environment gate, `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1`, which turns that test's graceful skips into hard failures, so a half-downloaded or corrupted checkpoint cannot quietly disable the contract check forever on the one machine that actually owns the checkpoint.

The gate currently covers only one of the two pinned Muse Glimmer guards in the tree. The sibling on the loading side was left untouched, so on the checkpoint-owning machine the protection is only half applied.

Verified against `main` at `0adb53c5`:

- The gate lives in `src/vision/encoders/muse_glimmer_fusion_pinned_tests.rs`, in `skip_or_fail_pinned_checkpoint` (around lines 122-143). It reads the variable under `crate::test_support::env_lock::env_lock()`, holds the lock only for the read and drops it before the assertion, and treats only the exact value `1` as enabling.
- `grep -rn "MLXCEL_REQUIRE_PINNED_CHECKPOINTS" src/ docs/` matches exactly one file, `muse_glimmer_fusion_pinned_tests.rs` (3 occurrences). Nothing else in the tree honors the variable and nothing documents it.
- The uncovered sibling is `pinned_weight_index_classifies_each_source_weight_once` at `src/loading/vlm_muse_glimmer_tests.rs:473`. It still uses the original unconditional skip: it checks `index_path.exists()`, and when the index is absent it prints `"Skipping pinned_weight_index_classifies_each_source_weight_once: pinned Muse Glimmer checkpoint index not present at {}"` and returns. There is no env gate on that path.
- That test asserts a real contract: `read_muse_weight_inventory_from_index(model_dir)` must yield `MuseWeightInventory { language_model: 626, vision_tower: 806, vision_adapter: 2, vision_projection: 1, lm_head: 1, total: 1436 }`. A silent skip therefore does hide something worth catching.
- One asymmetry makes this test's exposure smaller than the vision-side one, but not zero: it reads only the index file itself via `read_muse_weight_inventory_from_index` (`src/loading/vlm_muse_glimmer.rs:253`), so an "index present but shards missing" checkpoint cannot trip it. Its gap is purely that an absent or unreadable index skips unconditionally, even on a machine that is supposed to own the checkpoint.
- `MLXCEL_REQUIRE_PINNED_CHECKPOINTS` is not documented in `docs/environment-variables.md`. That document already covers test and diagnostic oriented knobs (for example `MLXCEL_DIFFUSION_DEBUG_CANVAS` for cross-implementation parity testing, and `MLXCEL_ENABLE_MTP_BATCH` for parity and debug testing), so a test gate of this kind has a natural home there.

## Proposed Solution

1. Make `pinned_weight_index_classifies_each_source_weight_once` honor `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1`, so its skip becomes a failure on the checkpoint-owning machine.
2. Factor the gate so both call sites share it instead of duplicating the env read and its env-lock discipline. The existing `skip_or_fail_pinned_checkpoint` is private to the vision test module, so extracting it into a shared test-support helper (for example under `src/test_support/`) is the natural move. Confirm the placement fits the crate's `#[cfg(test)]` gating; `test_support` is already `cfg(test)`-gated at `src/lib.rs:51-52`.
3. Document `MLXCEL_REQUIRE_PINNED_CHECKPOINTS` in `docs/environment-variables.md`, stating that it is a test-only gate, that only the exact value `1` enables it, and what it does.

## Implementation Constraints

**Preserve the availability versus contract boundary that PR #1173 established.** Only availability and readability problems (an absent file, an unreadable file, JSON that does not parse) may skip. A genuine contract violation must always fail. This gate converts skips into failures; it must not be used to convert a failure into a skip anywhere.

**Keep the env lock on any environment read.** If the shared helper reads the environment, it must continue to take `crate::test_support::env_lock::env_lock()` for that read. This crate is edition 2024, and `src/server/muse_glimmer_startup_guard_tests.rs` mutates the process environment with `unsafe set_var` under that same lock, so an unsynchronized read is unsound.

## Acceptance Criteria

- [x] `pinned_weight_index_classifies_each_source_weight_once` fails rather than skips when `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` and the pinned checkpoint is unusable.
- [x] With the variable unset, its skip behavior is unchanged.
- [x] With the full pinned checkpoint present, it still validates the weight inventory contract unchanged.
- [x] The gate logic is shared between both pinned guards rather than duplicated.
- [x] `MLXCEL_REQUIRE_PINNED_CHECKPOINTS` is documented in `docs/environment-variables.md`.
- [x] The documentation states when and how to actually turn the gate on: that it is test-only, that it belongs on a machine that owns the pinned checkpoint, a single-run invocation example plus a persistent-export suggestion, and its role in turning an ambiguous `ok` into positive evidence (added per the comment below, which supersedes the body where they differ).

## Verification Notes

The pinned checkpoint is fully materialized on the maintainer's GB10 machine at `models/mlx/muse-glimmer-30b` (`models` is a gitignored symlink to `/home/inureyes/models`), so every criterion above is directly testable there.

Nothing under `models/` may be moved, renamed, deleted, or truncated in order to simulate a partial checkpoint. Use a temp directory, as PR #1173 did.

inureyes:
## Summary

Extends the pinned-checkpoint gate PR #1173 introduced to the loading-side sibling test, shares the gate logic between both call sites instead of duplicating it, and documents the variable including when and how to actually turn it on, per the scope the issue's comment added.

## What changed

`src/test_support/pinned_checkpoint.rs` (new) holds a single `skip_or_fail_pinned_checkpoint(test_name, reason)` helper, declared from `src/test_support/mod.rs` alongside the existing `env_lock` module; it reads `MLXCEL_REQUIRE_PINNED_CHECKPOINTS` under `crate::test_support::env_lock::env_lock()`, drops the guard before asserting, and treats only the exact value `1` as enabling.
`src/vision/encoders/muse_glimmer_fusion_pinned_tests.rs` drops its private copy of the gate and calls the shared helper at both of its call sites, passing its own test name so the skip message is byte-identical to before and the failure message still names that test.
`src/loading/vlm_muse_glimmer_tests.rs` routes `pinned_weight_index_classifies_each_source_weight_once`'s previously unconditional skip through the same shared helper, so it now fails rather than skips when the gate is on and the pinned index is absent; the existing `panic!` on `read_muse_weight_inventory_from_index` returning `Err` is untouched, since that is a genuine contract violation rather than an availability problem.
`docs/environment-variables.md` adds a row for `MLXCEL_REQUIRE_PINNED_CHECKPOINTS` to the existing "Test and CI variables" table, whose `Variable | Purpose` shape carries the value and default semantics inline, stating that only the exact string `1` enables it, plus prose covering that it is test-only, that it belongs on a machine that owns the pinned checkpoint, a single-run invocation example and a persistent-export suggestion, and its role in turning an ambiguous `ok` into positive evidence that the checkpoint was actually read (how acceptance criterion 2 of #1161 was proven).

## Test plan

- [x] `cargo test --lib vision::encoders::muse_glimmer_fusion` (23 passed, real pinned checkpoint present)
- [x] `cargo test --lib loading::vlm::muse_glimmer::` (14 passed, real pinned checkpoint present)
- [x] Compiled test binary run from a temp directory with no `models/`, variable unset: both pinned guards skip, each still naming itself
- [x] Compiled test binary run from a temp directory with no `models/`, `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1`: both pinned guards fail, each still naming itself, exit 101

Closes #1177


inureyes:

### Intent

> Extend `MLXCEL_REQUIRE_PINNED_CHECKPOINTS=1` to the loading-side pinned Muse Glimmer guard, share the gate between both call sites, and document the variable including when and how to actually turn it on (issue #1177 body plus its scope-adding comment).

### Findings addressed

- [x] PR body described the doc change as a `Variable | Values | Default | Notes` row, which stopped being true after `d399363` folded the entry into the section's existing two-column `Variable | Purpose` table. Corrected, since this body becomes the squash-merge commit body. (LOW)
- [x] PR body claimed the vision-side "skip or failure message is unchanged". The skip message is byte-identical, but the assertion message changed from "the pinned Muse Glimmer checkpoint is not usable" to "the pinned checkpoint `<test_name>` needs is not usable". Reworded to say what is actually true. (LOW)

### Remaining items

- [ ] `src/lib.rs:47-49` still describes `test_support` as providing "the single shared `ENV_LOCK`", which is no longer exhaustive now that the module also holds the pinned-checkpoint gate. Its opening sentence ("Crate-wide helpers for `#[cfg(test)]` paths") still covers the module, so this is left alone rather than pulling the crate root into a test-only PR. (LOW, not fixed by design)
- [ ] The loading-side call site passes its test name as an inline literal while the vision side hoists a `TEST_NAME` const. Both are correct and each file has the number of call sites that justifies its choice. (LOW, style only)
- [ ] `docs/environment-variables.md:38` contains a pre-existing em dash, unrelated to this PR and one of six docs files in that state. Not touched here. (LOW, pre-existing)

### Verification

- [x] All stated requirements implemented, including the comment's fourth deliverable
- [x] No placeholder or mock code
- [x] Integrated: `pinned_checkpoint` is declared from `test_support/mod.rs` and called from both guards, no orphan
- [x] Correctness boundary intact: the helper only ever turns a skip into a failure, the loading-side `panic!` on `read_muse_weight_inventory_from_index` returning `Err` is untouched, `safetensors_shape` still panics rather than returning `Err` when a parsed header does not describe the key, and all four `#[should_panic]` synthetic contract tests carry `expected =` strings the gate's panic message cannot match, so the gate cannot let a contract regression pass for the wrong reason
- [x] Env-lock discipline correct: guard scoped to the read, dropped before the assertion, nothing inside the scope re-acquires the non-reentrant `ENV_LOCK`, and no caller of the helper holds that lock, so there is no deadlock against `src/backend/tests.rs` or `src/server/muse_glimmer_startup_guard_tests.rs`
- [x] Doc claims check out: nothing under `.github/` or `scripts/` references the variable, no production path reads it (`test_support` is `cfg(test)` at the crate root), and #1161's acceptance criterion 2 was indeed proven with this gate
- [x] Conventions: no AI attribution, no em dashes in any added line, `Closes #1177` present, table row well formed at two columns like its neighbors
- [x] Tests pass (orchestrator-verified: 37 passed both with and without the gate against the real checkpoint, plus gate-off skip and gate-on failure from a directory with no `models/`)

inureyes:

Reviewed at `d2b4f419` across the five dimensions this diff can actually carry risk in. No CRITICAL, HIGH, or MEDIUM findings. One LOW documentation-accuracy finding, fixed in `758fe64c`.

### Environment handling under Rust 2024

Sound. `skip_or_fail_pinned_checkpoint` binds the read as `let required = { let _env_guard = env_lock(); ... }`, so the guard is a named binding dropped at the end of the block rather than a temporary dropped at the end of the statement, and the `assert!` runs with the lock released.

- No deadlock. `env_lock` is a non-reentrant `std::sync::Mutex`, but none of the three call sites (`muse_glimmer_fusion_pinned_tests.rs:133`, `:185`, `vlm_muse_glimmer_tests.rs:477`) holds it; neither module references `env_lock` anywhere else, and no `env_lock` holder in the crate reaches the helper, so there is no lock-order cycle.
- No poisoning hazard, twice over: the guard drops before the assertion, and `env_lock()` already recovers with `unwrap_or_else(|p| p.into_inner())`.
- No read/write race. Every env mutation in this crate's test binary takes the same lock. The three files that mutate the environment without it are `src/multimodal/video.rs` (a doc comment only, no mutation), `src/cli/turbo_args.rs` (production `apply_to_environment`, whose tests do take the lock), and `src/loading/mod.rs` (`maybe_disable_cuda_graphs_for_model`, reachable only from real model-load paths; `src/backend/tests.rs` exercises the seam with a nonexistent dir, where `get_model_type(...)?` returns before the write, and it holds the lock anyway).

### Can the gate hide a failure

No. Its only effect when set is to panic; when unset it is byte-for-byte the prior `eprintln!` plus return. There is no path where setting or unsetting it lets a contract violation pass:

- `load_pinned_checkpoint` returns `Err` for availability and readability only. An absent `weight_map`, an omitted pinned key, a config that parses but does not fit the schema, and a wrong recorded shape all flow past it to assertions.
- `safetensors_shape` returns `Err` only for open/truncation/ceiling/parse problems, and panics rather than returning `Err` when a parsed header does not describe the key.
- On the loading side the gate covers only the `!index_path.exists()` branch, upstream of every contract check; the `panic!` on `read_muse_weight_inventory_from_index` returning `Err` is untouched.
- None of the four `#[should_panic]` tests can be satisfied by the gate's panic text. Their `expected =` strings are weight-key names, `"does not deserialize into a Muse Glimmer config"`, and `"failed: model.vision_adapter.fc1.weight"`; the gate message carries only the variable name, the test name, and a `reason` that names paths, never keys. Independently, all four fail before reaching the gate, since their synthetic shards are well formed.

### Surface

Not widened beyond the test binary. `src/lib.rs:51-52` is a plain `#[cfg(test)] pub(crate) mod test_support;`, the new module is `pub(crate)` inside it, there is no re-export, and the only three references in the tree are the test call sites. Unreachable from production code, from the binaries, and from `tests/`.

### Documentation (LOW, fixed)

The advice to export the variable persistently is safe: nothing outside the test files reads it, no code enumerates `std::env::vars()`, and there is no unknown-`MLXCEL_*` rejection, so a persistent export cannot affect a running `mlxcel` or `mlxcel-server`. The entry also does not overstate the gate's scope; it names the two tests it covers.

The last paragraph was inaccurate. It said there was no availability pre-check before PR #1173. PR #1157 (`92d6d4c8`) had already replaced the vision-side `assert!(index_path.exists(), ...)` with a silent skip and added the same index-absent skip on the loading side, which issue #1161 also records in its third acceptance criterion. As written it implied every partially materialized state was loud before #1173, overstating what that PR gave up and what the gate buys back. `758fe64c` restates the sequence: #1157 made the index-absent case quiet, #1173 made the index-present-but-incomplete case quiet, and the gate restores loudness for both.

### Performance

Nil. The lock is taken only on the skip or failure path, held for exactly one `std::env::var` call, and released before both the assertion and any I/O. In `assert_post_tower_contract` the shard header read happens outside the lock. On a machine with a complete checkpoint the helper is never called.

PR #1183, opened 2026-08-16
fix(video): replace the removed -vsync with -fps_mode and stop the tests hiding it

### LINKED ISSUE #1172: fix(video): every video path is broken on ffmpeg 8+ because -vsync was removed, and the tests hide it by passing when ffmpeg is absent

inureyes:
## Summary

Every video input path was broken on ffmpeg 8 and newer, and the tests reported green throughout. Frame extraction passed `-vsync vfr`, which ffmpeg 5.0 deprecated and ffmpeg 8 removed, so the subprocess exited at argument parsing with `Unrecognized option 'vsync'` before decoding anything and every `load_video` / `load_video_source` call failed. That covers the Gemma 4, Gemma 4 Unified, and Kimi-VL video paths, on both the CLI and the server.

## The fix, and the supported ffmpeg range

`-vsync vfr` becomes `-fps_mode vfr`. This is the straight substitution rather than a cached capability probe, because `-fps_mode` landed in ffmpeg 5.0 in the same release that deprecated `-vsync`: one spelling covers every ffmpeg a user is plausibly running, and there is no second branch to keep correct or to test. A probe would buy support for 4.x and older at the cost of a code path that only executes on machines nobody in this project has, which is how the original bug survived in the first place.

**The supported range is now stated as ffmpeg 5.0 (2022) or newer, with no upper bound** (9.x works unchanged), in the `video` module docs, `docs/supported-models.md`, `docs/installation.md` (a new "Video input and ffmpeg" section plus both platform prerequisite lists), and `docs/environment-variables.md`. Nothing else in the video path depends on a version-gated ffmpeg feature, so that one flag is the whole requirement.

## Making the skip visible

This is the half that let a total breakage sit in `main`. The ffmpeg-backed tests inspected `PATH` and returned early, and stable libtest counts an early return as a **pass**. Three of the frame-fidelity tests were worse: they matched on `load_video`'s error and skipped, which swallowed the exact `Extract` error this bug produces.

Stable libtest cannot report a skip decided at run time, so two mechanisms are needed and neither is sufficient alone:

- **The nine ffmpeg-dependent tests carry `#[ignore]`.** It is resolved at compile time and is the only skip that reaches the summary line, so a plain run now reports `28 passed; 0 failed; 9 ignored` rather than an all-green `34 passed`. A reader can tell "video not exercised" from "video verified" without reading the log.
- **`MLXCEL_TEST_VIDEO=1` turns a missing ffmpeg into a hard failure**, via a new `src/test_support/video_gate.rs` modelled on the existing `pinned_checkpoint.rs` gate. `#[ignore]` alone would let the opt-in run silently degrade into a no-op on a host that lost its ffmpeg, so the run that is supposed to exercise video now asserts that it can.

`make verify-test-video` is the combination, and `nightly-verify.yml` installs ffmpeg and runs it as its own reported step with its own row in the failure report. That closes the hole on the one gate that actually protects `main`. The wall-clock benchmark is skipped by name there: it is `#[ignore]` for an unrelated reason, and a timing bound does not belong in a correctness gate on a loaded runner.

The three tests that swallowed decode errors now fail on them. Only availability may skip; a video that will not decode is the defect the assertions exist to catch.

## What changed

- `src/multimodal/video.rs`: `-vsync` becomes `-fps_mode` via a new `FPS_MODE_ARG` constant; the argument list moves into a `frame_extraction_args` helper so the flag contract is testable without ffmpeg on PATH; module docs state the 5.0 floor and the new test-gating contract.
- `src/test_support/video_gate.rs` (new) and `src/test_support/mod.rs`: the `MLXCEL_TEST_VIDEO` gate, with `video_capability_available` and `skip_or_fail_video_fixture`.
- `src/multimodal/video_tests.rs`: nine tests gated with `#[ignore]` and routed through the gate; three swallowed `load_video` errors turned into failures; two new ffmpeg-free tests pinning the flag spelling and the PNG-pipe argument contract; a bare `matches!(err, ...)` statement that asserted nothing turned into a real assertion.
- `Makefile`: new `verify-test-video` target, deliberately not part of `verify`.
- `.github/workflows/nightly-verify.yml`: installs ffmpeg alongside cmake, runs `make verify-test-video` as a fourth step, and surfaces its outcome in the report table.
- `CONTRIBUTING.md`, `docs/installation.md`, `docs/supported-models.md`, `docs/environment-variables.md`: the ffmpeg range, the new target, and the `MLXCEL_TEST_VIDEO` variable.

## Test plan

Validated against **ffmpeg 9.0.1** on Apple Silicon (M-series, macOS 26).

Controlled before/after on one pre-fix binary at `115f6089`, varying only whether ffmpeg was on `PATH`:

| arm | ffmpeg | result |
|---|---|---|
| A | hidden | `ok. 34 passed; 0 failed; 1 ignored` |
| B | 9.0.1 | `FAILED. 31 passed; 3 failed; 1 ignored` |

All three failures carried `Extract { message: "Unrecognized option 'vsync'." }`. Five tests were silently skipping in arm A; the two cap tests passed anyway because caps are enforced after `ffprobe` and before any decode, so they never reached the broken argument list.

After this change, on the same host:

- [x] `make verify-test-video` reports `36 passed; 0 failed; 0 ignored` (was 3 failures).
- [x] Plain `cargo test -p mlxcel --lib multimodal::video` reports `28 passed; 0 failed; 9 ignored`, so the unexercised capability is visible in the summary.
- [x] Same binary, ffmpeg hidden, `MLXCEL_TEST_VIDEO=1`: all 8 gated correctness tests fail by name with `MLXCEL_TEST_VIDEO=1 but <test> cannot run: ffmpeg and ffprobe must both be invokable on PATH (ffmpeg 5.0 or newer)`.
- [x] Same binary, ffmpeg hidden, no variable: `28 passed; 0 failed; 9 ignored`, no false green.
- [x] `make verify-clippy` (the real `--workspace --all-targets --features metal,accelerate -- -D warnings` gate): clean.
- [x] `make verify-fmt`: clean.
- [x] `python3 scripts/ci/check_cross_repo_refs.py`: clean.

End-to-end against real checkpoints, which was not possible when the issue was filed:

- [x] `mlxcel generate -m models/gemma-4-12b-it-4bit --video red-then-blue-4s.mp4` on a clip that is solid red for two seconds then solid blue for two: "The video shows a solid red color that transitions to a solid blue color." Correct, and it requires both the first and last sampled frames to have decoded in the right temporal order.
- [x] Same checkpoint on a white disc translating left to right over black: "a white circle moves diagonally downward and to the right across a black background." The rightward direction is correct and is the ordering-sensitive part; the vertical component is a model embellishment, the disc stays vertically centred.
- [x] `models/gemma-4-e2b-it-4bit` decodes and describes both clips as well, exercising the plain Gemma 4 path alongside Gemma 4 Unified.
- [x] Independent of the model, the exact argument list the runtime now builds was run by hand against the red/blue clip and the PNG pipe split into 8 frames whose centre pixels are `fe0000, fe0000, fe0000, fe0000, 0000ff, 0000ff, 0000ff, 0000ff`: correct frame count, correct temporal order, correct RGB channel order.

Not verified: Kimi-VL's video path was not run (no local checkpoint exercised here), though it funnels through the same `load_videos` entry point. The ffmpeg 5.0 lower bound is documented from upstream release history rather than measured, since only 9.0.1 is installed on this box.

Closes #1172


inureyes:

### Intent

> Replace the `-vsync` flag ffmpeg 8 removed with `-fps_mode`, and stop the video tests from reporting an unexercised capability as a pass.

Reviewed against the consolidated requirements for #1172 (issue body plus thread), not the body alone.

### Verified independently

The two mechanisms were checked by running them, not by reading them:

| Run | Result |
|---|---|
| Plain `cargo test -p mlxcel --lib multimodal::video` | `28 passed; 0 failed; 9 ignored` |
| `make verify-test-video` (before the review fix) | `36 passed; 0 failed; 0 ignored` |
| Test binary, ffmpeg hidden from `PATH`, `MLXCEL_TEST_VIDEO=1` | `FAILED. 28 passed; 8 failed`, each named on the gate's `cannot run` message |
| `make verify-test-video` (after the review fix) | `37 passed; 0 failed; 0 ignored` |

The three claims the review was asked to confirm all hold. Nine tests on `main` carried the `if !ffmpeg_available() { SKIP; return }` shape, not five. Three of them (`extract_frames_preserves_color_increment_per_frame`, `extract_frames_preserves_channel_order`, `extract_frames_preserves_moving_square_position`) matched on `load_video`'s error and returned, so they swallowed exactly the `Extract` error this issue is about; all three now `.expect()`. `load_video_without_ffmpeg_returns_clear_error` did carry a bare `matches!(err, VideoError::FfmpegMissing);` that asserted nothing, and is now a real `assert!`.

`nightly-verify.yml` reads correctly. The ffmpeg install sits after the `export PATH=/opt/homebrew/bin:...` line and mirrors the existing cmake block, `$GITHUB_PATH` is appended after both so later steps resolve ffmpeg, the `video` step has no `continue-on-error` so a failure fails the job, `outputs.video` is wired to the report table, and the report job fires on `needs.verify.result == 'failure'`. The duration lookup's `startswith("fmt + clippy + test")` filter still matches the renamed job.

### Findings addressed

- [x] **A tenth ffmpeg-backed test was outside the mechanism** (HIGH). Fixed in b5168a0f.

  `src/vision/processors/gemma4_tests.rs::process_videos_pixel_values_match_input_color` decodes a synthetic clip through the same `crate::multimodal::video::load_video` entry point and asserts per-channel pixel content through `Gemma4Processor::process_videos`. It carried all three defects this PR exists to remove: a file-local duplicate `ffmpeg_available()` with an `eprintln!` and `return` and no `#[ignore]`; a `load_video` error arm that swallowed the failure as a skip; and no selection by `make verify-test-video`, whose filter was `multimodal::video` only.

  Reproduced before the fix: with ffmpeg hidden and `MLXCEL_TEST_VIDEO=1` set, that test printed `SKIP: ffmpeg not available` and reported `ok`. It was the last remaining false green on the video path, and Gemma 4 is one of the three video-capable families the issue names.

  It now carries the same `#[ignore]` reason string, gates on `video_capability_available`, routes fixture failures through `skip_or_fail_video_fixture`, and `.expect()`s the decode. `verify-test-video` passes a second positional libtest filter so the gate still selects it, and `CONTRIBUTING.md` and `docs/environment-variables.md` moved with it.

### Remaining items, reported not fixed

- [ ] `load_video_without_ffmpeg_returns_clear_error` still early-returns as a pass on any host that has ffmpeg (MEDIUM). The bare `matches!` is fixed, but the enclosing `if ffmpeg_available() { return; }` means the test asserts nothing on the nightly runner, which now installs ffmpeg, or on any developer machine that has it. It is the inverse gate, so `#[ignore]` is not the right answer, and `ffmpeg_available()` is `OnceLock`-cached so `PATH` cannot be manipulated in process. Left as is deliberately; noting it so the shape is not mistaken for coverage.
- [ ] `make verify-test-video` has no floor on how many tests it selects (MEDIUM). If a filter ever stops matching, for instance after a module move, libtest reports `0 passed` with everything filtered out and the target exits 0. That is the same failure mode this PR closes, relocated from the test body to the Makefile filter. Worth a minimum-count assertion or a canary if the target grows more filters.
- [ ] No runtime hint on an ffmpeg older than 5.0 (LOW). `ffmpeg_available()` only checks that the binary runs, so a 4.x host passes the availability check and then fails with a raw `Unrecognized option 'fps_mode'`. `docs/installation.md` documents that exact text, which covers the requirement, but the error path itself could name the version floor.
- [ ] `make verify`'s success line still claims it "matches the nightly-verify GitHub Actions job" (LOW). Nightly now has a fourth gate that `verify` deliberately omits, so the claim is looser than it was.
- [ ] The PR body's "Five tests were silently skipping in arm A" is imprecise (LOW). Eight tests printed the skip line in the ffmpeg-hidden arm. Five is the count that produced no signal in either arm, which is what the following clause explains, but the sentence reads as the former.

### Verification

- [x] All stated requirements implemented: flag fix, skip visibility, ffmpeg range in docs
- [x] Integrated into project code flow: `frame_extraction_args` is called from `extract_frames_single_pass`, the gate is called from every ffmpeg-backed test, the Makefile target exists, and the CI step and its failure-report row are wired
- [x] Project conventions followed: `video_gate.rs` mirrors `pinned_checkpoint.rs` including the env-lock read window and the "only ever turns a skip into a failure" rule
- [x] Existing modules reused where applicable, after the duplicate `ffmpeg_available()` was removed
- [x] `make verify-fmt`, `make verify-clippy` (`--workspace --all-targets`, forced past the warm cache), and `scripts/ci/check_cross_repo_refs.py` all clean

PR #1182, opened 2026-08-16
feat(inference): support the qwen3_5_mtp drafter for Qwen 3.5 MTP speculative decoding

inureyes:
## Summary

Adds MTP speculative decoding for the qwen3_5 family: a first-class `qwen3_5_mtp` drafter (the split-out MTP head, e.g. `mlx-community/Qwen3.8-27B-MTP-bf16`), an `MtpTarget` adapter for `Qwen35Model` / `Qwen35VLModel`, the server burst + tick-slice + offline CLI wiring, and a chain-parity gated-delta kernel that makes the temperature-0 exactness gate actually pass. Scope is B = 1; batched windows decline to classic with the existing named error, which the issue's own acceptance criteria allow.

## What changed

- `src/lib/mlxcel-core/src/drafter/qwen3_5_mtp/` (new: config, layer, model, tests): the fc-fused single full-attention decoder layer with the Qwen 3.5 gated-Q layout, drafter-owned per-layer KV cache, embeddings and LM head borrowed from the target at bind. Sanitize ports upstream's `mtp.`-prefix strip and applies the `+1.0` norm shift ONLY to raw-HF layouts; a test pins that the published pre-stripped layout is never shifted again. A strict weight-inventory gate rejects any directory that is not a standalone 15-tensor head, which is also what keeps a target checkpoint advertising `mtp_num_hidden_layers` without tensors from ever being treated as a drafter.
- `Drafter` trait: two new defaulted hooks, `prefill_from_target_hidden` and `accept_verified_tokens`, because the Qwen MTP drafter is stateful (its cache accumulates one entry per emitted position, mirroring upstream's prefill/accept/seed lifecycle); `MtpVerifyOutput` gains an optional `verify_hidden_full`; `MtpGenerator` calls the hooks when the target advertises the hidden. Gemma paths carry `None` plus the no-op defaults and are byte-untouched.
[...]

Issue #1186, opened 2026-08-16
fix(speculative): MTP temperature-0 byte-identity fails on Apple GPU generation 15+ and at block widths above the qmv batch limit


inureyes:
## Problem / Background

PR #1182 (issue #1165) shipped Qwen 3.5 MTP speculative decoding with a temperature-0 contract: the multi-token verify block emits the same tokens as the single-token decode chain. The gate protecting that contract is `mtp_capable_target` in `src/server/batch/speculative_burst.rs`, which checks `metal_is_available() && supports_chain_parity_kernel()`.

Both conditions are necessary and neither is sufficient. The contract also depends on which MLX kernel each quantized projection dispatches to at `M = block_size` versus `M = 1`, and that admits hosts where byte-identity does not hold.

### Apple M5 Max 128 GB (macOS 26.6.1), `main` at `115f6089`

- `block_verify_chain_matches_single_token_chain` FAILS at the shipped block width 3: 2 diverging positions over 241 tokens, first at chain index 48 (single-token argmax 11592, block argmax 5587).
- `adapter_chain_with_perfect_drafts_matches_classic_chain` and `adapter_chain_with_rejected_drafts_matches_classic_chain` also fail (1 position each, chain index 151).
- The production server path shows it too: same prompt, temperature 0, 150 tokens, one server per arm, token ids compared through logprobs. Token ids agree for 13 tokens, then diverge at index 13 (classic `' broken'`, MTP `' illustrated'`).

It is **not** the gated-delta chain-parity kernel. The synthetic bitwise pin `parity_kernel_block_is_bitwise_equal_to_single_token_chain` passes there, the kernel is provably dispatched (mutation-checked with `MLXCEL_GDN_CHAIN_PARITY=0`), and unset / `=1` / `=0` all produce the identical failure. It is also not the M5 neural-accelerator attention route, which was A/B'd with that path disabled.

### Apple M1 Ultra 128 GB (macOS 26.6.1), same commit

A new model-free per-op diagnostic (`tests/metal_block_vs_chain_op_parity.rs`) compares a `T = K` call against `K` single-position calls at Qwen3.8-27B shapes.

| op (T = 3) | M5 Max | M1 Ultra |
| --- | --- | --- |
| `quantized_matmul` 5120 to 17408 (affine 4-bit, group 64) | DIVERGES | equal |
| `quantized_matmul` 5120 to 5120 (affine 4-bit) | DIVERGES | equal |
[...]

PR #1190, opened 2026-08-16
test: pin which primitives keep block-vs-chain bitwise equality on Metal

inureyes:
## Summary

Adds `tests/metal_block_vs_chain_op_parity.rs`, a model-free diagnostic that asks, per primitive, whether one `T = K` call is bitwise equal to `K` single-token calls.

Speculative decoding rests on that equivalence: a `T = K` verify forward must reproduce what `K` separate `T = 1` decode steps would have produced. The gated-delta chain-parity kernel from #1182 fixed one violation of it. Nothing in the tree checked whether any other primitive violates it, so when the temperature-0 exactness gate failed on M5 Max there was no cheap way to attribute it.

This is the attribution tool. It loads no checkpoint and runs in about a second, so it can be run on any Apple Silicon generation and the results compared. Full context and the M5 Max measurement are in https://github.com/lablup/mlxcel/pull/1182#issuecomment-5307311278.

## What it found on M5 Max

Apple M5 Max 128 GB, macOS 26.6.1, `main` at `115f6089`, Qwen3.8-27B shapes (hidden 5120, intermediate 17408, head_dim 256, 24 q heads, 4 kv heads, affine 4-bit group 64):

| op | block vs single-token chain |
|---|---|
| `quantized_linear` 5120 -> 17408 (MLP gate/up) | **DIVERGES** (621, 1834, 1447 of 34816 bytes per position) |
| `quantized_linear` 5120 -> 5120 (o_proj shape) | **DIVERGES** (167, 501, 387 of 10240) |
| dense matmul f16 | **DIVERGES** (203, 204, 202 of 10240) |
| dense matmul f32 | **DIVERGES** (14625, 14771, 14746 of 20480) |
| `fast_rms_norm` | equal |
| `fast_rope` at matching per-position offsets | equal |
| SDPA, causal `T = K` block vs the `attend_per_position` form | equal |

Norm, RoPE and attention are position-independent. Only matmul moves, and matmul is every projection in the model.

The block-length sweep puts the break at `T = 2`:

```
    T   quantized 4-bit   dense f16
    1   equal             equal
    2   DIVERGES          DIVERGES
    3   DIVERGES          DIVERGES
    ...
   32   DIVERGES          DIVERGES
```

That rules out the `get_qmv_batch_limit` vector/matrix threshold (`mlx/backend/metal/quantized.cpp`), which computes to 10 or 12 for these operands and would have kept a `T = 3` block on the vector kernel.

It also rules out the M5 Neural Accelerator gemm as the sole cause. MLX gates that on `is_nax_available() && (enable_tf32() || dtype != float32)`, so float32 can turn it off with `MLX_ENABLE_TF32=0`. Under the toggle the block digest moves (`0f03cc5caa81a58b` to `08a556ebcff82dfb`) while the chain digest does not (`51df14d9f29f9f3d` both ways), which is the arm-attribution control, and the divergence survives: 72% of bytes differing drops to 37%. NAX roughly doubles the perturbation but is not the origin.

## What is asserted and what is not

`fast_rms_norm`, `fast_rope` and SDPA are **asserted** equal. The verify path depends on all three being position-independent, `Qwen3NextAttention::attend_per_position` exists specifically to hold the attention half of that, and a regression in any of them is a real defect worth failing on. Nothing pinned this before.

Matmul is **reported, not asserted**. It is a property of MLX kernel selection rather than something this repository can assert away, and failing on it would produce a permanently red suite that says nothing actionable. The sweep prints the curve instead, which is the part that transfers across hardware.

## Test plan

- [x] `cargo test --test metal_block_vs_chain_op_parity --release --features metal,accelerate -- --nocapture` passes on M5 Max (2 tests, 1.39 s, no checkpoint required)
- [x] `cargo clippy --release --features metal,accelerate --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] Skips cleanly when Metal is unavailable (`metal_is_available()` guard on both tests), so Linux/CUDA builds are unaffected

## Follow-up

The finding this tool produced is that byte-identity between the speculative verify path and classic decode is not reachable by fixing any single kernel on Metal. Deciding what the contract should say instead is tracked separately.

Refs #1165, #1182

### DIFF
--- a/tests/metal_block_vs_chain_op_parity.rs
+++ b/tests/metal_block_vs_chain_op_parity.rs
@@ -1,0 +1,402 @@
+//! Which primitive stops being bitwise equal between a `T = 1` chain and a
+//! `T = K` block on this GPU?
+//!
+//! ## Why this exists
+//!
+//! Speculative decoding rests on an equivalence: verifying `K` positions in
+//! one `T = K` forward must reproduce what `K` separate `T = 1` decode steps
+//! would have produced. The gated-delta chain-parity kernel (#1165) fixed one
+//! violation of that equivalence. This file asks which *other* primitives
+//! violate it, by comparing, per op, one `T = K` call against `K` single-token
+//! calls on the real Qwen3.8-27B shapes.
+//!
+//! It loads no model and takes about a second.
+//!
+//! ## What it pins and what it only reports
+//!
+//! `fast_rms_norm`, `fast_rope` and SDPA are **asserted** equal: the verify
+//! path depends on them being position-independent, and a regression there
+//! would be a real defect worth failing on.
+//!
+//! Matmul is **reported, not asserted**. On M5 Max every linear op measured
+//! here differs between `T = 1` and `T >= 2`, which is a property of MLX's
+//! kernel selection rather than something this repository can assert away;
+//! turning it into a failing test would only produce a permanently red suite.
+//! The sweep test prints where the break happens so the same command can be
+//! run on other Apple Silicon generations and compared.
+//!
+//! ## Run
+//!
+//! ```bash
+//! cargo test --test metal_block_vs_chain_op_parity --release --features metal,accelerate -- --nocapture
+//! ```
+
+use mlxcel::initialize_runtime;
+use mlxcel_core::{MlxArray, UniquePtr};
+
+const H: i32 = 5120;
+const INTER: i32 = 17408;
+const HEAD_DIM: i32 = 256;
+const N_Q_HEADS: i32 = 24;
+const N_KV_HEADS: i32 = 4;
+const GROUP_SIZE: i32 = 64;
+const BITS: i32 = 4;
+const BLOCK: i32 = 3;
+
+fn fill(n: usize, seed: f32, scale: f32) -> Vec<f32> {
+    (0..n)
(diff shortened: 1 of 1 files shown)


inureyes:
Superseded by #1189, closing.

#1189 carries `tests/metal_block_vs_chain_op_parity.rs` in a broader form than this PR does: it also covers mxfp4 (group 32) alongside affine 4-bit, which turns out to matter, since mxfp4 diverges on every GPU generation rather than only on generation 15 and newer. Both PRs add the same path, so keeping this one open only creates a conflict.

The M5 Max measurements that motivated this file are recorded on #1182 and folded into #1186. Nothing here is lost.

One correction to this PR's description, for anyone reading it from a link: it says the sweep result "rules out the `get_qmv_batch_limit` vector/matrix threshold". That is right for M5 Max and wrong as a general statement. The batch limit is exactly what produces the M1 Ultra cliff at block width 12. There are two separate switches, and this hardware only shows the first:

- `use_qmv_wide(mode, d)` is `mode != "affine" || arch_gen >= 15`, and an `M >= 2` call takes `qmv_wide` when it holds. That is the generation 15+ break at `T = 2`.
- `get_qmv_batch_limit` sends `M` at or above the limit to the matrix-matrix kernel. That is the break at 12 on generation 13/14.

#1186 states both correctly.

Issue #1065, opened 2026-08-07
fix(core): f32 matmul and single-query SDPA lose ~fp16 precision on M5, breaking four numeric tests


inureyes:
## Problem / Background

Four deterministic test failures occur on Apple M5 Max that do not occur on the M1 Ultra nightly runner. All four are numeric-agreement assertions on f32 computations. The observed error magnitudes cluster at fp16 epsilon scale (2^-10 = 9.77e-4), roughly three orders of magnitude larger than f32 epsilon, which indicates the f32 inputs are being computed at reduced internal precision on M5-class hardware.

### The four failing tests

1. `mlxcel-core` `layers::tests::chunked_causal_attention_matches_fast_causal` at `src/lib/mlxcel-core/src/layers.rs:6344`. Message: `q_len=4 k_len=4 chunk=1 diverged from do_causal SDPA`. Measured `max_abs_diff` = 0.0009678602 against a tolerance of 1e-5. Fails only at `chunk=1`; chunk 2 and 3 pass.
2. `mlxcel-core` `layers::tests::chunked_query_attention_matches_unchunked_with_mask` at `src/lib/mlxcel-core/src/layers.rs:6288`. Message: `chunk=1 diverged from unchunked SDPA`. Measured `max_abs_diff` = 0.0009678602 against a tolerance of 1e-5. Fails only at `chunk=1`; chunks 2, 4 and 5 pass.
3. `mlxcel-core` `mla::decode::decode_tests::absorbed_attention_respects_an_additive_causal_mask` at `src/lib/mlxcel-core/src/mla/decode_tests.rs:93`. Message: `masked absorbed attention drifted by 0.0014691837`, against `F32_TOL = 2e-5` (relative error).
4. `mlxcel-core` `mla::decode::decode_tests::expand_latent_reproduces_the_up_projection` at `src/lib/mlxcel-core/src/mla/decode_tests.rs:132`. Message: `k_nope[0][0][1] 0.30357933044433594 != 0.3037950135765861`, absolute tolerance 1e-4, observed absolute error 2.157e-4 (relative error about 7.1e-4).

## Why this is a precision problem and not a test problem

[...]

PR #1196, opened 2026-08-16
feat(core): add MLXCEL_METAL4_ATTENTION as a kill switch for the M5 NA route

inureyes:
## Summary

`MLXCEL_METAL4_ATTENTION=0` forces the M5 neural-accelerator fused attention route off on hardware that has it, so the ordinary SDPA path runs instead.

This is one of #1065's acceptance criteria, landed on its own because it is the criterion that unblocks the others: that issue's suggested investigation is "determine whether MLX selects a reduced-precision NAX kernel", and without a switch the only way to test it is to patch `should_use_metal4_attention` and rebuild. The M5 investigation on #1182 did exactly that, by hand, and reverted afterwards. The next person to need it will not know that happened.

## Behaviour

Off-switch only. The hardware test still gates the route, so:

| host | variable | route |
|---|---|---|
| NA-capable (`has_neural_accelerator && macos_supports_na`) | unset | on |
| NA-capable | `0` / `false` / `no` / `off` | **off** |
| not NA-capable | anything, including unset | off |

A stray value on an M1 is inert rather than routing it into an M5 kernel. Read once per process, matching `MLXCEL_GDN_CHAIN_PARITY`, `MLXCEL_SSM_CUDA_KERNEL` and the other kill switches in this tree.

## A drift the switch exposed

There were **two** copies of the hardware test. `layers::should_use_metal4_attention` had one, and `lib.rs`'s causal-attention dispatch carried its own inline `hw.has_neural_accelerator && hw.macos_supports_na` before calling `layers::metal4_causal_attention`. Adding the switch to only the first would have disabled the route in one place and left it running in the other.

For a diagnostic whose entire value is a clean A/B, a half-applied switch is worse than no switch: it produces a run that looks like the route is off and is not, and the resulting "the NA route is not the cause" conclusion would be wrong and hard to catch. `lib.rs` now calls the shared predicate.

## What this does not do

It does not diagnose #1065. It makes the diagnosis a one-command A/B instead of a patch-and-rebuild, on a machine this session does not have.

## Test plan

Apple M1 Ultra 128 GB (no neural accelerator), macOS 26.6.1, branched from `main` at `ffde948e`.

- [x] `cargo test -p mlxcel-core metal4_attention_switch`: 1 passed. The test asserts the off-switch property that can be checked from any host: without the hardware the predicate is false regardless of the variable. On NA hardware it returns early rather than pinning an answer to one machine's ambient environment.
- [x] `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings`: clean.

Not verified: the switch's *effect* is unobservable on this host, because the route it disables never engages here. Confirming that `MLXCEL_METAL4_ATTENTION=0` actually changes numerics needs an M5-class machine, which is the same machine #1065 needs anyway. Suggested check there: run the four tests #1065 names with and without the variable and report whether the failures move.

Refs #1065


inureyes:
The switch is in on an M5 Max (40 GPU cores, 128 GB, macOS 26.6.1, Xcode 26.5) and behaves correctly as a variable, but the test plan's open item is still open, and the obvious way to close it does not work.

The suggested check was to run the four tests #1065 names with and without the variable and report whether the failures move. Done, and they do not move: identical failures with `MLXCEL_METAL4_ATTENTION` unset, `=1` and `=0`, byte-identical values in all three. Full cross including `MLX_ENABLE_TF32` is on #1065.

That result says nothing about this switch, because those four tests never reach the route. It is consulted at `lib.rs:3207`, `layers::metal4_attention` and `layers::metal4_causal_attention`; the tests call `attention_dispatch`, `chunked_causal_attention`, `chunked_query_attention`, `ffi_fast_scaled_dot_product_attention_causal` and the MLA decode paths directly. Re-running test 1 with `MLXCEL_LOG_NA_ATTENTION=all` gives zero dispatch log lines with the switch on and zero with it off, which confirms it rather than inferring it.

So the "not verified" line in the test plan should stay, and the suggested check in it is worth amending: on this hardware it produces a clean-looking null result for the wrong reason. A real-checkpoint decode A/B, with `MLXCEL_LOG_NA_ATTENTION` used to prove the route actually engages in the on arm, is the check that can close it. Running that now and will report here.

Separately, this switch did its job for #1065 even though it turned out not to be the cause: the four-arm cross is what ruled the route out as a suspect, and before #1196 that would have needed a patch and a rebuild.

inureyes:
## Verified on M5 Max: the switch works, and the route it disables is not costing anything measurable

Apple M5 Max, 40 GPU cores, 128 GB, macOS 26.6.1, Xcode 26.5, `main` at `33951285`. Three real checkpoints.

### The switch does what it says

`MLXCEL_LOG_NA_ATTENTION=all`, counting dispatch log lines over a 32-token generation:

| checkpoint | route on | `MLXCEL_METAL4_ATTENTION=0` |
|---|---:|---:|
| `llama-3.2-1b-4bit` | 528 | **0** |
| `gemma-3-4b-it-4bit` | 1122 | **0** |
| `qwen2.5-7b-instruct-4bit` | 924 | **0** |

The route engages on all three and the switch takes it to zero on all three. That closes the test plan's open item on this hardware, including the drift this PR fixed: a switch applied to only one of the two hardware tests would have left one of these non-zero.

### Its effect, measured separately

96 tokens, `--temp 0`, arms alternated, two reps, **no** NA logging:

| checkpoint | on rep1 | on rep2 | off rep1 | off rep2 | generated text |
|---|---:|---:|---:|---:|---|
| `llama-3.2-1b-4bit` | 519.44 | 522.69 | 523.26 | 521.16 | **identical** |
| `gemma-3-4b-it-4bit` | 177.41 | 178.37 | 178.08 | 178.97 | **identical** |
| `qwen2.5-7b-instruct-4bit` | 119.93 | 119.06 | 119.86 | 119.53 | **identical** |

Byte-identical output on all three, and throughput differences under 1% with the sign flipping between reps, so within noise. On these checkpoints the NA attention route neither changes what is generated nor pays for itself.

### A trap worth writing down before someone hits it

My first pass measured route engagement and throughput in the **same** run and produced 360 tok/s with the route on against 577 with it off, a 60% gap that looks like a serious finding and is entirely an artifact: only the on arm emits log lines, 1820 of them, and `MLXCEL_LOG_NA_ATTENTION=all` costs roughly 40% of throughput by itself. The two measurements have to be separate runs. Discarded, and repeating it here so the number does not resurface as a quotable result.

A second one: `qwen3-0.6b-4bit` generates essentially empty output for a bare continuation prompt, so a byte-identity comparison on it passes without comparing anything. Checkpoints for this A/B need to be checked for non-empty output first.

### Scope

This says the route is numerically invisible at temperature 0 for 96 tokens on three 4-bit checkpoints, which is a statement about argmax flips, not about per-op precision. It does not measure prefill-heavy or long-context shapes, and it does not touch #1065, whose cause turned out to be `MLX_ENABLE_TF32`. Details on that are in https://github.com/lablup/mlxcel/issues/1065#issuecomment-5311690868.

Issue #1187, opened 2026-08-16
perf(speculative): recover MTP throughput on Apple GPU generation 15+ without giving up temperature-0 byte-identity


inureyes:
## Problem / Background

The exactness probe from #1186 fails closed on Apple GPU generation 15 and newer, which is exactly where MTP pays off.

Measured on an M5 Max 128 GB with the Qwen3.8-27B 4-bit target plus its bf16 MTP drafter: block 3 is 1.19x classic decode (39.37 against 32.96 tok/s, medians of 5 interleaved repetitions with 90 s cooldowns), and the target's own multi-token forward amortizes at T=4/T=1 = 1.29. On an M1 Ultra the same pairing is 0.70x.

So the honest position after the gate change in #1186 is that byte-identity holds where MTP is slow and fails where MTP is fast, and declining is a real throughput loss on the newer hardware. The decision recorded for this project is to keep the contract and research the speed, not to restate the contract to accommodate the slowness.

## Mechanism to Attack

The divergence on generation 15+ comes from `qmv_wide`, which MLX selects for `M >= 2` affine quantized matmuls when `arch_gen >= 15` (see https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/quantized.cpp). `qmv` and `qmv_wide` reduce over K with different lane counts, so their results differ in the last ulp. The verify block is the only place that runs `M >= 2`.

## Proposed Solution

Directions worth measuring, in rough order of cost:

1. **Measure what `qmv_wide` actually buys on the verify shapes.** If forcing the plain `qmv` reduction at `M = K` costs only a few percent on a memory-bound 27B target, exactness is recoverable almost for free on generation 15+. Note that `dispatch_qmv` already skips `qmv_wide` when a global scale is present, which is one existing lever, and an upstream switch to force `qmv` would be another. Report the measured verify-forward delta at block 3 and 4.
2. **If a forced-`qmv` verify is viable, wire it behind the probe.** Engage it only where the probe would otherwise decline, and re-run the probe to confirm the arms converge.
3. **Account for the drafter.** Independently of exactness, the drafter is now the binding constraint on this hardware (see #1185): about 10.5 ms per drafter step against a 30.3 ms full target decode step, roughly 149 GB/s effective against the target's 530 GB/s, and about 26% of the round is drafter overhead. Whatever recovers exactness should be measured against a round whose drafter cost has already been addressed, or the two efforts will each look smaller than they are.

## Acceptance Criteria

[...]

PR #1199, opened 2026-08-17
feat(speculative): let the MTP gate buy back exactness by dropping qmv_wide

inureyes:
> **Note added 2026-08-22 (see #1278):** the third row of the "Three paths, all verified live" table below, "`MLXCEL_MTP_ALLOW_INEXACT=1` engages on the fast kernel", is inaccurate for the code this PR merged. The retry runs before `allow_inexact()` is consulted, so on a generation 15+ host where the narrow retry passes, `MLXCEL_MTP_ALLOW_INEXACT=1` alone engages MTP on the narrow kernel with byte-identity kept: verified live on an M3 Ultra on 2026-08-22, where the override-alone run logs the retry's INFO line, produces output byte-identical to the default env, and measures the narrow arm's throughput (117 vs 139 tok/s against the pinned-wide recipe). Reaching the fast kernel requires `MLXCEL_QMV_WIDE=1` together with `MLXCEL_MTP_ALLOW_INEXACT=1`. The corrected recipes are in `docs/environment-variables.md` and `docs/benchmark_results/qmv-wide-pin-tax-m3ultra-2026-08-22.md`. The rest of this description is unchanged and reflects what was measured at the time.

## Summary

MTP was off on every Apple GPU generation 15 and later, and not because of a bug: `use_qmv_wide` in MLX sends `M >= 2` affine quantized matmuls to a kernel that reduces along K differently from the `qmv` that `M == 1` takes, so a verify block is never byte-equal to the single-token chain there. The #1189 gate is fail-closed, so it declined every time and fell back to classic decode.

Generations 13 and 14 take `qmv` on both sides and stay exact below `get_qmv_batch_limit`. This makes that path reachable on generation 15+: an MLX overlay adds an off-switch for `use_qmv_wide`, and the gate re-probes with it off before giving up.

Two commits: the overlay and switch, then the gate wiring.

## Why the numbers say to do this

Measured on Apple M5 Max (40 GPU cores, 128 GB, macOS 26.6.1), `qwen3.8-27b-4bit` with the `qwen3.8-27b-mtp-bf16` drafter, `--draft-block-size 3`, four alternating reps at `-n 300` compared per rep because the host drifts thermally across a session:

| | vs classic decode | temperature-0 byte-identity |
|---|---:|---|
| what ships today (gate declines) | 1.00x | kept |
| `MLXCEL_MTP_ALLOW_INEXACT=1` | 1.16x | **broken** |
| **this PR** | **1.04x** | **kept** |

It is strictly better than today at the same contract, and it recovers about a third of what forfeiting the contract would buy. `qmv` costs 17 to 20 percent on the verify forward and about 10 percent end to end against `qmv_wide`.

## The overlay

`src/lib/mlx-cpp/patches/mlx/backend/metal/quantized.cpp`, seeded from pin `9a795735` with one added predicate and one runtime setter. Upstream has no equivalent knob and `use_qmv_wide` is unchanged as of that pin, so it has to live here. The delta is two hunks, so a bump refreshes the file and re-applies them.

`MLXCEL_QMV_WIDE=0` seeds it from the environment; `mlxcel_core::set_qmv_wide` moves it at runtime, which is what the gate uses. The flag is a relaxed atomic: it selects between two kernels computing the same function to different last-ulp results, so a stale read costs precision, never correctness, and it sits on the dispatch path of every quantized matmul.

The bridge declaration is gated on `CARGO_FEATURE_METAL` rather than `__APPLE__`, because a macOS build without the `metal` feature sets `MLX_BUILD_METAL=OFF` and would link against a symbol that was never emitted.

## The switch does only what it says

Block-width sweep with `MLXCEL_QMV_WIDE=0`, from `tests/metal_block_vs_chain_op_parity`:

| shape | equal through | breaks at | `get_qmv_batch_limit` |
|---|---:|---:|---:|
| qwen q4 5120 -> 5120 | 12 | 13 | 13 |
| qwen q4 5120 -> 17408 | 12 | 13 | 13 |
| gemma q4 3840 -> 4096 | 24 | past 24 | 25 |
| gemma q8 3840 -> 15360 | 12 | 13 | 13 |
| **dense f16 5120x5120** | **1** | **2** | n/a |

Every quantized column turns equal exactly where the batch limit says it should, and the dense f16 control still breaks at `T = 2`. The switch moved the quantized path and nothing else.

## Three paths, all verified live

| environment | behavior |
|---|---|
| nothing set | probe fails under `qmv_wide`, passes without it, MTP engages, `INFO` says what was traded |
| `MLXCEL_QMV_WIDE=1` | retry skipped, MTP declines, and the message says the retry was skipped because the kernel is pinned |
| `MLXCEL_MTP_ALLOW_INEXACT=1` | engages on the fast kernel with the existing loud warning |

An explicit `MLXCEL_QMV_WIDE` wins over the retry on purpose: someone who pinned a kernel gets it, and learns from the decline that the contract was unreachable under it.

## Two decisions worth arguing with

**The switch is left off for the rest of the process, not restored.** Restoring it would break the very block the gate just approved. The exact arm stays correct for every other caller, so the cost is throughput, not correctness. But it is process-wide, and a server sharing the process pays it on batched decode, whose `M` lands in the same 2-to-limit window. Scoping it to the verify forward needs more than a global and is filed as follow-up rather than half-done here.

**Larger blocks do not rescue the cost.** With the switch on, the gate passes up to block 12, so the obvious idea is to amortize `qmv` over a wider block. Measured, acceptance collapses and throughput with it:

| block | acceptance | emitted/verify | tok/s |
|---:|---:|---:|---:|
| 3 | 0.65 | 2.3 | 21.30 |
| 4 | 0.49 | 2.47 | 16.24 |
| 6 | 0.35 | 2.72 | 12.12 |
| 8 | 0.25 | 2.74 | 9.12 |
| 12 | 0.17 | 2.82 | 5.80 |

(One thermally degraded sweep, comparable within itself only.) Block 3, the drafter's own configured value, stays the operating point.

## Test plan

- [x] `-p mlxcel --lib` fails the same **13** tests over three runs, matching the pre-change baseline exactly; `-p mlxcel-core --lib` is 1472 passed / 4 failed. All are the pre-existing M5 reduced-precision failures of #1065.
- [x] Two new gate tests, plus the eight existing ones, pass three runs in a row.
- [x] `cargo fmt --check` and `cargo clippy --all-targets --features metal,accelerate -- -D warnings` clean.
- [x] Turbo kernel launchers re-validated after the overlay landed: `sparse_v_kernel_threshold_zero_matches_graph`, and the two delegated-SDPA tests at RMS 1.7263e-4 and 1.5259e-4 against the 5e-3 contract.

The switch is indirected behind a thread-local under `cfg(test)`. Toggling the real one from a unit test perturbs the numeric tests libtest runs beside it, in the direction of making them pass, and that cost the suite its determinism (13 failures became 12 or 13 at random) while this was being built. The live path is covered by the CLI runs above, not by the unit tests.

## Not covered

CUDA is untouched and was not built here. The end-to-end byte-identity claim rests on the model-level probe, not on diffing generated text, because `mlxcel generate --draft-kind mtp` prints no output at all: it reports the token count and rate and emits nothing. That predates this branch and the MLX pin bump, and is filed separately.

Refs #1187, #1186

### DIFF
--- a/src/lib/mlx-cpp/CMakeLists.txt
+++ b/src/lib/mlx-cpp/CMakeLists.txt
@@ -29,9 +29,14 @@ function(mlx_apply_source_overlays mlx_source_dir)
   # The steel/gemm/mma.h overlay (issue #217: upstream safe-load fix
   # cherry-pick, PRs #3560/#3565) was retired when the pin moved to
   # 2026-06-11 upstream main, which carries the fix natively (issue #222).
+  # quantized.cpp carries one delta, the MLXCEL_QMV_WIDE off-switch on
+  # `use_qmv_wide` (issue #1187). Upstream has no equivalent knob and the
+  # predicate is unchanged as of pin 9a795735, so it has to be overlaid here.
+  # Refresh the file wholesale on a bump and re-apply that one hunk.
   set(_metal_patch_files
       "mlx/backend/metal/compiled.cpp"
-      "mlx/backend/metal/kernels/utils.h")
+      "mlx/backend/metal/kernels/utils.h"
+      "mlx/backend/metal/quantized.cpp")
   foreach(_patch_file ${_metal_patch_files})
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/patches/${_patch_file}")
       configure_file(
--- a/src/lib/mlx-cpp/patches/mlx/backend/metal/quantized.cpp
+++ b/src/lib/mlx-cpp/patches/mlx/backend/metal/quantized.cpp
@@ -1,0 +1,2161 @@
+// Copyright © 2023-2026 Apple Inc.
+// Patched by mlxcel: `use_qmv_wide` gains an off-switch, `MLXCEL_QMV_WIDE=0`.
+// Synced to upstream 9a795735; the only delta is `qmv_wide_enabled()` and the
+// one call it adds to `use_qmv_wide`. Everything else is upstream verbatim, so
+// a bump refreshes this file and re-applies those two hunks.
+//
+// Why it exists (lablup/mlxcel#1186, #1187). `use_qmv_wide` sends `M >= 2`
+// affine quantized matmuls down `qmv_wide` on GPU generation 15 and later
+// while `M == 1` keeps taking `qmv`. Those two kernels reduce along K in a
+// different order, so a speculative verify block and the single-token decode
+// chain it must match stop being bitwise equal at block width 2 on M5-class
+// hardware, which is what closes the #1189 exactness gate and disables MTP
+// there entirely. Generations 13 and 14 take `qmv` on both sides and stay
+// equal up to `get_qmv_batch_limit`, so forcing that path is a way to buy the
+// contract back. Upstream exposes no knob for this and `use_qmv_wide` is
+// unchanged as of 9a795735, so the switch has to live here.
+//
+// Off-switch only, matching MLXCEL_METAL4_ATTENTION: unset behaves exactly as
+// upstream, and the value is read once per process because this sits on the
+// dispatch path of every quantized matmul.
+
+#include <cstdlib>
+#include <cstring>
+
+#include "mlx/backend/common/quantized.h"
+#include "mlx/backend/common/broadcasting.h"
+#include "mlx/backend/common/compiled.h"
+#include "mlx/backend/gpu/copy.h"
+#include "mlx/backend/metal/device.h"
+#include "mlx/backend/metal/kernels.h"
+#include "mlx/backend/metal/reduce.h"
+#include "mlx/backend/metal/unary.h"
+#include "mlx/backend/metal/utils.h"
+#include "mlx/fast_primitives.h"
+#include "mlx/primitives.h"
+#include "mlx/utils.h"
+
+namespace mlx::core {
+
+namespace {
+
+template <typename... Args>
+auto get_quantized_kernel_wrapped(
+    metal::Device& d,
+    const std::string& name,
+    const std::string& func,
+    const std::string& mode,
+    const std::string& type,
+    int group_size,
+    int bits,
+    Args... args) {
+  std::string template_def;
+  std::string fname = ((mode == "affine") ? "affine_" : "fp_") + func;
+  template_def = get_template_definition(
+      name, fname, type, group_size, bits, std::forward<Args>(args)...);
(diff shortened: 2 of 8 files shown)


inureyes:
## Correction: this PR's contract claim was too strong

The PR body says the byte-identity contract is "kept". Measured end to end after merge, it is not. The probe passes and that part is real, but MTP's generated text still differs from classic decode's.

`qwen3.8-27b-4bit`, `--temp 0 -n 120 --show-reasoning`, M5 Max, `main` at `0199be00`. The generated span compared directly, banners and timing stripped:

| comparison | result |
|---|---|
| classic vs classic with `MLXCEL_QMV_WIDE=0` | identical |
| classic run twice | identical |
| MTP run twice | identical |
| **MTP vs classic** | **diverges at char 159 of 630** |
| MTP with the switch vs MTP with `qmv_wide` forced on | **identical** |
| MTP at `--draft-block-size 2` vs at 3 | **identical** |

Two of those rows are what matter.

**The switch does not change MTP's output.** So whatever breaks end-to-end parity is not the `qmv_wide` kernel split, and this PR does not fix it.

**The divergence is independent of block width.** Block 2 and block 3 produce byte-identical output and diverge from classic at the same character. A verify-block-versus-chain effect would scale with width; this does not, so the cause is not that mechanism at all.

Ruled out so far: `qmv_wide` (above), classic-side nondeterminism (classic is reproducible and unaffected by the switch), and the GDN chain-parity kernel (it is on by default and the shape-fallback warning does not fire on this checkpoint).

## What still stands

The probe genuinely went from failing to passing, the op-level sweep turns equal exactly at the batch limits, and the throughput numbers are unchanged. What is wrong is the inference I drew from a passing probe. `models::speculative_exactness`'s own module docs say a passing probe is evidence and not proof, and this is that caveat cashing out: the probe compares one verify block against one chain from one synthetic state, while the contract is a claim about a whole generation from the real state sequence.

So this PR should be read as "removes a real systematic divergence and recovers throughput on generation 15+", not as "restores the contract". #1187 was closed on the stronger reading and I am reopening the contract half as its own issue rather than leaving that claim standing.

Also worth recording: the check that caught this was the one #1200 said was impossible. It was not a defect. `mlxcel generate` was hiding the text behind the reasoning filter, because this checkpoint stays inside its thinking channel for the whole budget; `--show-reasoning` shows it. Closing #1200 accordingly.

PR #1246, opened 2026-08-19
refactor(server): remove unused build_failed_response helper

### LINKED ISSUE #1226: refactor(server): remove or wire up the unused build_failed_response helper in responses_translator.rs
## Problem

`src/server/responses_translator.rs` defines a public helper, `build_failed_response`, that is documented as a convenience entry point for the route layer but is never called from anywhere. It survives only because of an `#[allow(dead_code)]`.

## Evidence

src/server/responses_translator.rs:649:

```rust
/// Convenience entry point for the route layer: build a response object
/// representing an error that occurred mid-generation.
#[allow(dead_code)]
pub fn build_failed_response(
    id: String,
    model: String,
    created_at: f64,
    error_code: &str,
    error_message: &str,
    ...
```

A repo-wide grep for `build_failed_response` finds exactly one hit: the definition itself. No callers, no tests, no documentation references.

## Suggested fix

Pick one of two directions:

1. Delete the function and its doc comment, about 25 lines removed. This is the default choice if the route layer already builds failed Responses objects another way.
2. If it is intentionally part of the public surface, wire it up: call it from the route layer where mid-generation failures on the `/v1/responses` path are currently assembled by hand, add a unit test covering the produced failed-response shape (status, error code, error message), and drop the `#[allow(dead_code)]`.

## Acceptance criteria

- [ ] No `#[allow(dead_code)]` remains on this function
- [ ] Either the function is deleted, or it has at least one real caller in the route layer plus a unit test
- [ ] `cargo build` and the server test suite pass

DebayanSen96:
Fixes #1226.

`build_failed_response` in `src/server/responses_translator.rs` has zero callers, zero tests, and survives only behind an `#[allow(dead_code)]`. I checked the route layer before deleting: `routes/responses.rs` surfaces mid-generation failures as HTTP `ErrorResponse`s (`generation_error_to_response`), never as a failed `ResponseObject`, so wiring the helper in would mean changing the endpoint's error semantics — a behavioral decision I didn't want to make unilaterally. Deleted the function (option 1 in the issue). `OutboundContext`, `ResponseErrorBody`, `ResponseStatus` and `build_response_object` all remain in use elsewhere in the file.

Note: no Rust toolchain on this machine, so I couldn't run `cargo build`/tests locally; the change is a pure dead-code removal with no call sites.

### DIFF
--- a/src/server/responses_translator.rs
+++ b/src/server/responses_translator.rs
@@ -644,42 +644,6 @@ pub(crate) fn short_uuid() -> String {
     raw.chars().take(16).collect()
 }
 
-/// Convenience entry point for the route layer: build a response object
-/// representing an error that occurred mid-generation.
-#[allow(dead_code)]
-pub fn build_failed_response(
-    id: String,
-    model: String,
-    created_at: f64,
-    error_code: &str,
-    error_message: &str,
-    request: &CreateResponseRequest,
-) -> ResponseObject {
-    let ctx = OutboundContext {
-        response_id: id,
-        model_id: model,
-        created_at,
-        completed_at: created_at,
-        status: ResponseStatus::Failed,
-        prompt_tokens: 0,
-        completion_tokens: 0,
-        cached_tokens: 0,
-        reasoning_tokens: 0,
-        text: String::new(),
-        reasoning_text: None,
-        parsed_tool_calls: None,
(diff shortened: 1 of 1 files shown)


DebayanSen96:
Closing — contribution needs local build/test verification before review.

PR #1251, opened 2026-08-19
fix(bench): fail with a context error instead of panicking on --target paths with no file name

### LINKED ISSUE #1242: fix(bench): speculative_bench panics on a user-supplied --target path with no file name
`speculative_bench` calls `.file_name().unwrap()` on a path taken directly from the `--target` CLI argument. `Path::file_name()` returns `None` for paths ending in `..` or a trailing slash component that normalizes away, so a user who passes such a path gets a bare unwrap panic instead of a diagnostic.

## Evidence

src/bin/speculative_bench.rs:829:

```rust
let row = if synthetic.target_subdir == target.file_name().unwrap().to_string_lossy() {
```

`target` comes straight from the command line. `PathBuf::from("models/foo/..").file_name()` and similar inputs return `None`, and the `unwrap()` aborts the whole bench run with a panic backtrace rather than telling the operator what was wrong with the argument.

## Suggested fix

One to three lines. Either tolerate the missing file name:

```rust
let target_name = target.file_name().map(|n| n.to_string_lossy()).unwrap_or_default();
```

or fail early with a context error such as `--target must name a model directory` before the comparison. Either is fine; the second gives the operator a clearer message.

## Acceptance criteria

- [ ] Running `speculative_bench` with a `--target` path ending in `..` or a trailing slash prints a readable error naming the `--target` argument, not a panic
- [ ] The canonical `models/<name>` path comparison at src/bin/speculative_bench.rs:829 behaves exactly as before for normal paths

DebayanSen96:
Fixes #1242.

`speculative_bench` called `.file_name().unwrap()` on the `--target` path. `Path::file_name()` returns `None` for paths ending in `..` or a trailing slash, so those inputs aborted the run with a bare unwrap panic. Now the file name is resolved once, right after the `--target` argument is read, and a missing file name fails with:

```
--target must name a model directory (paths ending in '..' or a trailing slash have no file name)
```

The canonical `models/<name>` comparison behaves exactly as before for normal paths. No Rust toolchain on this machine, so I couldn't compile locally — the change only replaces an unwrap with the same `?`-propagated `Context` pattern used two lines above it.

### DIFF
--- a/src/bin/speculative_bench.rs
+++ b/src/bin/speculative_bench.rs
@@ -805,6 +805,9 @@ fn main() -> Result<()> {
             .as_deref()
             .context("--target is required when --sweep is not set")?
             .to_path_buf();
+        let target_name = target
+            .file_name()
+            .context("--target must name a model directory (paths ending in '..' or a trailing slash have no file name)")?;
         let pairing_name = format!("{} ({})", target.display(), args.kind);
         let synthetic = Pairing {
             name: Box::leak(pairing_name.into_boxed_str()),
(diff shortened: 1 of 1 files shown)


DebayanSen96:
Closing — contribution needs local build/test verification before review.

Issue #1188, opened 2026-08-16
fix(speculative): Gemma 4 MTP breaks byte-identity on generation 15+, and its gate returns unconditionally true


inureyes:
## Status: measured and confirmed on 2026-08-18

This issue was filed as "the mechanism should carry over, nobody has run the comparison". The comparison has now been run on an Apple M5 Max and the mechanism does carry over. Gemma 4 MTP is **not** byte-identical to classic decode on generation 15+ hardware under the default kernel selection, and disabling `qmv_wide` restores it exactly.

The gate change is no longer speculative work. What is new, and what makes it a real decision rather than a formality, is the price.

## The measurement

Host: Apple M5 Max (`Mac17,7`), macOS 26.6.1, mlxcel at `07ab8612`, `[profile.test-fast]`, `--features metal,accelerate`. Target `models/gemma-4-12b-it-4bit`, drafter `models/gemma-4-12b-it-assistant-4bit`, `--draft-kind mtp`, `temperature 0`, 250 tokens, two prompts.

| comparison | prose prompt | code prompt |
|---|---|---|
| MTP vs classic, `qmv_wide` **on** (default) | differs at char 51 | differs at char 55 |
| MTP vs classic, `qmv_wide` **off** | **byte-identical** | **byte-identical** |

Block width does not change the answer: block 2 and block 5 both diverge under the default, at the same character. That is consistent with the mechanism, since `use_qmv_wide` splits at `M >= 2` and a block of 2 is already on the wrong side of it.

Classic decode is deterministic run to run here (three runs, identical), so the divergence is systematic rather than the f16 jitter class.

## One methodological trap, because it cost us a wrong answer first

**Toggling `MLXCEL_QMV_WIDE` changes classic decode too**, so the comparison has to hold the kernel fixed on both sides. Classic's multi-token prefill is itself an `M >= 2` forward, so `classic(wide on)` and `classic(wide off)` differ from each other at char 51 on the prose prompt. Comparing `MTP(wide off)` against `classic(wide on)` conflates the prefill effect with the verify effect and reports a divergence that says nothing about the contract.

Second trap, for whoever reproduces this: the CLI echoes the prompt without a trailing newline and then appends `Loading MTP drafter from "..."` to that same line, so a naive line filter leaves the banner in the compared text and every MTP run "differs from classic" at the character where the banner starts. Extract the generated text by splitting on the prompt echo and dropping the two drafter banner lines.

## What it costs

[...]

Issue #1088, opened 2026-08-08
test: verify-test-cuda is red on clean main because the gate leaves MLX TF32 enabled


inureyes:
## Problem / Background

`make verify-test-cuda` is the documented Linux/NVIDIA merge gate, and it is structurally red on a clean `main`. Four numeric parity tests fail only because MLX defaults TF32 on and the gate does not set `MLX_ENABLE_TF32=0`.

Verified first-hand on `main` at `0f727572` (GB10, CUDA sm_121, Linux aarch64). Full run: 7655 passed, 5 failed. Four of the five failures are TF32 artifacts, not defects. Each one fails with the gate as written and passes with `MLX_ENABLE_TF32=0`, verified individually with `--exact`:

- `models::bailing_moe_linear::tests::chunked_gla_matches_the_sequential_recurrence`
- `models::deepseek_v2::tests::absorbed_mla_attention_matches_the_decompressed_block_step_for_step`
- `models::florence2::florence2_tests::incremental_decode_matches_full_sequence`
- `models::klear::tests::the_prefill_is_causal_without_being_handed_a_mask`

The failure magnitudes are consistent with TF32's 10-bit mantissa rather than with a logic error:

```
klear:     logit 25 at position 0 changed when later tokens were added: -0.1357227 vs -0.13684261 (about 1.1e-3)
florence2: incremental decode diverged from full-sequence decode, max abs diff 0.00014263391
```

`make verify-test-cuda` (`Makefile:572-575`) resolves to `cargo test --workspace --profile test-fast --features cuda --no-fail-fast -- --test-threads=1` and sets no TF32 environment variable, so it inherits MLX's default of TF32 enabled.

The fifth failure, `execution::memory_estimate::tests::resolve_block_budget_explicit_bytes_floors_to_block_count`, is a genuine pre-existing integer underflow and is filed separately. It is not TF32-related and fails identically with TF32 off, so it is out of scope here.

Note that `src/lib/mlxcel-xla/README.md` already records that MLX CUDA's default FAST_TF32 mode is an intentional, lower-precision throughput policy that is not labeled as F32 evidence. The gate as written contradicts that documented project policy.

## Why this matters

[...]

Issue #629, opened 2026-07-02
perf(cuda/moe): fix MoE prefill collapse on CUDA (gather_qmm / grouped GEMM path)


inureyes:
Part of #623. Depends on #625: re-triage first. Upstream #3706 (JIT gather_gemm) and #3632 (gather_qmm NAX name fix) may partially or fully resolve this; the pin-upgrade triage doc states what remains.

## Context

MoE prefill on CUDA is 5-10x slower than on Metal M1 Ultra even though GB10 has more compute. From `benchmarks/cuda_gb10_2026-06-17.csv` vs the 06-15 Metal sweep (prefill tok/s):

| Model | GB10 CUDA | M1 Ultra Metal |
|---|---|---|
| mixtral-8x7b-4bit | 12.5 | 81.0 |
| llama-4-scout-17b-4bit | 28.2 | 119.5 |
| phi-3.5-moe-4bit | 28.9 | 114.0 |
| minimax-m2-3bit | 26.9 | 68.0 |
| gpt-oss-20b-mxfp4 | 126.2 | 282.4 |
| solar-open-100b-4bit | 46.8 | 71.0 |

Dense prefill on the same hardware is fine (CUDA beats Metal by ~1.6x median across 133 common models), so this is specific to the expert-routing matmul path: `gather_qmm` (`src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:2496-2522`, model call sites e.g. `src/models/deepseek_v3.rs:663-702`, `qwen3_moe.rs:134`) and the CUDA grouped-GEMM overlay (`src/lib/mlx-cpp/patches/mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu`, `matmul.cpp`). Note prefill hits the matrix-matrix shape (many tokens per expert), unlike decode's GEMV shape, and the quantized dispatch on CUDA falls through to `qmm_sm80`/`qmm_naive` on GB10 because `qmm_sm90` is Hopper-gated (`patches/mlx/backend/cuda/quantized/quantized.cpp:53-113`).

## Scope

Bring CUDA MoE prefill to at least Metal-M1-Ultra absolute throughput on the table above (GB10 should beat M1 Ultra on any compute-bound path). Decode-side MoE is covered by #626 and #636, not here.

## Implementation plan

1. Reproduce with the long-prompt harness (#624): mixtral-8x7b-4bit at 2048-token prompt is the primary repro; smaller repro phi-3.5-moe-4bit.
[...]

PR #1266, opened 2026-08-20
fix(tests): seed synthetic model fixtures from a sorted key walk

### LINKED ISSUE #1265: test: the klear prefill-causality parity test is nondeterministic run to run on CUDA, and TF32 amplifies it past its tolerance
## Problem / Background

`models::klear::tests::the_prefill_is_causal_without_being_handed_a_mask` (`src/models/klear_tests.rs:643`) produces a different result on every run on GB10, at both precision settings. It is not flaky in the usual "passes alone, fails in a suite" sense; the underlying numeric value moves run to run in an isolated single-test invocation.

Measured on GB10 (CUDA sm_121, Linux aarch64, driver 580.173.02) at `07474633`, MLX pin `9a795735`. The test was temporarily instrumented to print the max absolute logit delta between the single-token forward and row 0 of the 4-token prefill, then run 10 times per mode. The instrumentation was reverted.

| | `MLX_ENABLE_TF32=0` (the pin from #1260) | `MLX_ENABLE_TF32=1` (MLX default) |
| --- | --- | --- |
| max logit delta | 3.6e-7 to 7.2e-7 | 5.9e-4 to 2.0e-3 |
| median | about 4.7e-7 | about 1.2e-3 |
| result vs the test's 1e-3 bound | 10 passed, 0 failed | 2 passed, **8 failed** |

Two things follow. The variance exists at **both** settings, so TF32 is an amplifier and not the cause. And under MLX's default precision the spread straddles the test's own 1e-3 tolerance, which is why the test fails 8 runs in 10 there.

## This is specific to this test, not a general property of the backend

The other three tests #1088 listed were run 6 times each under `MLX_ENABLE_TF32=1` on the same host and commit. All three fail every run with a **bit-identical** magnitude, one distinct value each across 6 runs:

- `models::bailing_moe_linear::tests::chunked_gla_matches_the_sequential_recurrence`: `state[11]: chunked -0.23453808 vs recurrence -0.23468243`
- `models::deepseek_v2::tests::absorbed_mla_attention_matches_the_decompressed_block_step_for_step`: `step 1 drifted by 0.00027007936`
- `models::florence2::florence2_tests::incremental_decode_matches_full_sequence`: `max abs diff 0.00014263391`

The florence2 magnitude is additionally identical to the one recorded in #1088 at `0f727572`, which was a different MLX pin (`2c46b953`), so that path is stable across the pin bump as well.

The structural difference is what these tests exercise. The three stable ones are isolated module-level parity checks over small dense f32 tensors: the bailing test drives the GLA kernel directly on noise arrays (heads 3, len 10, dim 4) with no model, no MoE and no quantization, and the deepseek_v2 test drives the MLA attention module alone. The klear test is the only one that builds and runs a **whole model**, and `small_args()` in `src/models/klear_tests.rs` configures it as **4-bit quantized** (`group_size: 64, bits: 4`) **MoE** with `num_experts = 8` and `num_experts_per_tok = 2`.

## Suspected mechanism

Not confirmed, and the issue should not be closed on the strength of the guess alone. The prime suspect is accumulation order in the quantized MoE path (GatherQMM / grouped GEMM), where nondeterministic fp32 accumulation order is already a documented property of this backend: PR #726 (issue #629) recorded that gpt-oss greedy decode flips near-tie logits from fp32 accumulation order alone. PR #740's notes separately record that the `[1, K]` verify stream on CUDA is load-sensitive run to run.

## Why this matters

- The merge gate is not affected. `verify-test-cuda` runs with the #1260 pin active, where the delta is about 2500x under the tolerance, and the full gate was 8167 passed and 0 failed at `670512c2`.
- Production is affected in principle. Shipped numerics run at MLX defaults, which is the amplified column: a 5.9e-4 to 2.0e-3 spread on logits is enough to flip a near-tie argmax, which would make greedy decode nondeterministic run to run for quantized MoE models on CUDA.
- Anything that measures numerics **once** at default precision on CUDA and turns the measurement into a verdict can get a run-dependent answer. The runtime exactness probes (#1188, #1189) are the in-tree instance worth checking against this.
- It already cost real time. During #1088's CUDA-runner validation, single observations of this test produced two wrong conclusions in a row: first that the #1260 pin was not what fixed the CUDA gate, then that the MLX pin bump had moved klear under its bound. Both were artifacts of sampling a distribution once. That is recorded in #1088's comments and corrected in #1262/#1263.

## Proposed Solution

The choice belongs to the maintainer; these are not mutually exclusive.

1. **Characterize before fixing.** Determine whether the variance is the quantized MoE accumulation order by A/B-ing the fused/grouped path (for example the `MLXCEL_GATHER_QMM_GROUPED=0` and `MLXCEL_FUSED_MOE=0` kill switches) with the same 10-run instrumented protocol. If the spread collapses with the path disabled, that localizes it.
2. **Decide whether it is a defect or a documented property.** If nondeterministic accumulation order is accepted for this path, say so where users will find it, and make the greedy-decode consequence explicit rather than leaving it implied.
3. **Make the test honest either way.** A parity test whose subject varies run to run should either pin the source of variance or state the bound it actually holds. Its real subject, prefill causality, has enormous margin (a deliberate perturbation of `src/models/klear.rs:617` fails it by 0.28, roughly 280x the tolerance), so the tolerance is not protecting causality, it is exposing accumulation noise.

## Acceptance Criteria

- [ ] The source of the run-to-run variance is identified, or explicitly recorded as unlocalized after a stated set of A/Bs.
- [ ] A decision is recorded on whether nondeterministic accumulation on this path is a defect or an accepted property.
- [ ] If accepted, the greedy-decode consequence for quantized MoE models on CUDA is documented where a user would look.
- [ ] The runtime exactness probes (#1188, #1189) are checked against the possibility of a run-dependent verdict at default precision, and either shown to be robust or made so.
- [ ] Re-running the klear test 10 times on GB10 gives a stated, reproducible verdict rather than 8 failures in 10.

## Technical Considerations

- Reproducer: instrument the delta in `src/models/klear_tests.rs:643` and run `cargo test --workspace --profile test-fast --features cuda --no-fail-fast -- --exact models::klear::tests::the_prefill_is_causal_without_being_handed_a_mask --test-threads=1 --nocapture` repeatedly. Do not conclude from one run.
- `--test-threads=1` is required on this backend for whole-suite runs; see the `verify-test-cuda` comment block in the `Makefile`.
- Related but distinct: #997 (flaky VLM parity tests producing non-finite logits under whole-suite runs) is a different mechanism. #1045 (fused decode-MoE not byte-identical on Klear) is a systematic Metal-side offset, not run-to-run variance.
- Related context: #1088, #1259, #1260, #1262, #1263.

inureyes:
## What was actually wrong

The run-to-run variance is a **test fixture defect**, not a backend property. `filled_weights` walks a `WeightMap` and advances one LCG seed per key. `WeightMap` is `HashMap<String, UniquePtr<MlxArray>>`, and Rust's `RandomState` randomizes iteration order per process, so every run handed a different noise block to a different tensor: **each process built a different random model.** The klear causality test then measured a different model's numerics each time, which is why its max logit delta wandered over 5.9e-4 to 2.0e-3 and straddled its own 1e-3 bound.

The fix is to sort the key walk. The same fixture bug was present in three sibling files, so all four are fixed here.

## Two hypotheses in the issue are refuted, with measurements

**"The prime suspect is accumulation order in the quantized MoE path (GatherQMM / grouped GEMM)."** The fixture is not quantized. `synthetic_weights` emits only `.weight` for the expert stacks and no `.scales`, and `SwitchLinear::from_stacked_parts` selects the non-quantized `Regular` path when scales are absent. Probed directly through the public API on GB10: `SwitchGLU::forward_fused_kernel(..).is_some()` is **false**, so neither the fused decode-MoE kernel nor the quantized `gather_qmm` path runs in this test at all.

**"The variance exists at both settings, so TF32 is an amplifier and not the cause."** Both halves needed correcting. Once the fixture is deterministic there is no variance at either setting, and the pin's effect on this test is exactly what it is on the other three in #1088.

Measured on GB10 (CUDA sm_121, MLX pin `9a795735`), 10 processes per mode after the fix, max abs logit delta between the single-token forward and row 0 of the 4-token prefill:

| | `MLX_ENABLE_TF32=0` (the #1260 pin) | `MLX_ENABLE_TF32=1` (MLX default) |
| --- | --- | --- |
| delta | 5.364418e-7 | 1.0485351e-3 |
| distinct values over 10 runs | **1** | **1** |
| verdict over 10 runs | 10 passed, 0 failed | 0 passed, 10 failed |

Before the fix the same test was 8 failures in 10 with a moving value. It is now bit-identical run to run in both modes, so klear stops being the exception among the four tests #1088 listed and behaves like the other three: a deterministic magnitude that clears the bound under the pin and misses it without.

## The backend is not implicated on this path

Repeated identical forwards inside one process are bit-identical at both precision settings, measured separately for each arm (`0e0` for both the single-token and the 4-token prefill arm, at `MLX_ENABLE_TF32=0` and `=1`). Scoped claim on purpose: this says nothing about the quantized MoE accumulation-order nondeterminism recorded in #629 / PR #726, which is a different path that this fixture never reaches.

## What the tolerance is

The 1e-3 bound is a precision bound, not a causality bound, and the test now says so. Measured in the same binary by handing the prefill an all-zero additive mask, a genuinely bidirectional prefill moves row 0 by about **1.6**, roughly 1.5e3x the default-precision delta and 3e6x the pinned one. The assertion keeps its full power over the property it names, so the tolerance is left alone rather than widened.

## Acceptance criteria

- [x] The source of the run-to-run variance is identified: per-process `HashMap` iteration order in `filled_weights`.
- [x] A decision is recorded: it was neither a backend defect nor an accepted backend property, but a fixture defect, and it is fixed.
- [x] Nothing to document for users: the premise (nondeterministic greedy decode for quantized MoE on CUDA, inferred from this test) is refuted, so no such note is added. The corrected reasoning is recorded at the `verify-test-cuda` gate definition in the `Makefile`, replacing the claims #1264 left there.
- [x] The runtime exactness probes (#1188, #1189) are checked. See below.
- [x] Re-running the klear test 10 times on GB10 gives a stated, reproducible verdict.

## On the exactness probes

The probes compare a verify block against a single-token chain byte for byte, within one process, so run-to-run variance would make their verdict noise-determined. The in-process repeat measurement above shows forwards on this path are bit-reproducible, so the probes are robust on that axis.

One asymmetry surfaced while checking and is recorded rather than changed here, since it is not a nondeterminism problem: the Qwen 3.5 gate short-circuits on `!mlxcel_core::metal_is_available()` in both the server path (`models/qwen3_5.rs`) and the offline CLI path (`commands/generate.rs`), while the three Gemma 4 arms have no such guard and therefore do run the probe on CUDA. Not exercised against a real Gemma 4 checkpoint on CUDA here.

## Verification

- `models::klear` 22 passed, `models::afmoe` 23 passed, `models::phixtral` 18 passed, `models::bailing_moe_linear` 35 passed; 98 tests, 0 failed.
- `cargo fmt --all -- --check` clean; `cargo clippy --lib --tests --features cuda -- -D warnings` clean.
- The klear causality test run 10x per precision mode as tabulated above.

Closes #1265


inureyes:
`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at this branch: **8171 passed, 0 failed, 310 ignored**, exit 0.

That is the gate this repository defines as the Linux/NVIDIA merge condition, and it is the configuration whose comment block this PR rewrites, so it is run here rather than inferred. `cargo fmt --all -- --check` and `cargo clippy --lib --tests --features cuda -- -D warnings` are clean on the same tree.

inureyes:

### Intent

Identify and fix the run-to-run variance in `models::klear::tests::the_prefill_is_causal_without_being_handed_a_mask`, record the decision, and withdraw the claims that the earlier nondeterministic measurements left at the `verify-test-cuda` gate definition.

The root cause and the remedy check out. `WeightMap` is `HashMap<String, UniquePtr<MlxArray>>` (`src/lib/mlxcel-core/src/weights.rs:29`), the LCG chain advanced once per key over an unsorted walk, and `RandomState` randomizes that per process, so each process built a different model. `keys.sort()` on a `Vec<String>` is a total, stable, byte-lexicographic order over a fixed key set, and nothing else in `filled_weights` is order-sensitive: `weights.insert` on an already-present key replaces in place and does not change the key set. Downstream the loaders address the map by key (`weights.get` / `contains_key`), never by iteration, so the built model is now fully determined.

### Findings Addressed

- [x] The bidirectional-prefill ratio in the new klear comment was off by 2x (HIGH). `1.6 / 1.0485351e-3 = 1526`, so the figure is roughly `1.5e3x` the default-precision delta, not `3e3x`. The companion `3e6x` is correct (`1.6 / 5.364418e-7 = 2.98e6`). Fixed in `3e71c1f6` and in the PR body, which becomes the squash commit body. This comment block is the fourth pass over the same claim after #1262, #1263 and #1264, so an uncorrected 2x error here would have bought a fifth.

### Remaining Items

- [ ] The hardcoded deltas are now coupled to the fixture key set (MEDIUM). Under a sorted walk the seed chain is positional, so adding any tensor to `synthetic_weights` reseeds every key that sorts after it and silently invalidates the `5.364418e-7` / `1.0485351e-3` figures the new comment states to eight significant digits. Nothing detects that. Two ways out: a one-line caveat in the comment, or derive each tensor's seed from the key so ordering cannot matter at all. The repository already has the latter pattern at `src/models/llada2_moe/tests.rs:83`, where `fill` mixes the element index into the seed instead of carrying state. Not blocking: sorting is the correct minimal fix for this issue, and key-derived seeding is a separate hardening across five call sites in four files.
- [ ] `small_args()` at `src/models/klear_tests.rs:524` still sets `quantization: Some(Quantization { group_size: 64, bits: 4 })` while `synthetic_weights` ships no `.scales` (LOW). That gap is exactly what produced the claim now being withdrawn: a reader checks the config, concludes "4-bit quantized MoE", and never notices that `SwitchLinear::from_stacked_parts` takes the `None => Ok(Self::Regular { weight })` arm and that `forward_fused_kernel` declines at its first `quantized_parts()?`. The Makefile records the correction, but the trap itself is unannotated at the config that springs it.
- [ ] The Makefile paragraph characterizes the withdrawn claim two ways in adjacent sentences (LOW): first as "a property of the backend", then as "the variance was not klear's own numerics". The prior text said the spread was "klear's own, not TF32's", so the second phrasing is the accurate one.

### Notes on the claims, checked against the measurements

Accurate: the `HashMap` mechanism and the per-process consequence; "the same fixture bug was in afmoe, phixtral and bailing_moe_linear"; the two withdrawn claims; "it ships no `.scales`, so the expert stacks load dense and neither the fused nor the `gather_qmm` quantized path runs"; the exactness-probe asymmetry (`metal_is_available()` short-circuits at `src/models/qwen3_5.rs:1515` and `src/commands/generate.rs:1766`/`1820`, with no counterpart in the Gemma 4 arms).

"All four now fail on every run, each at a bit-identical magnitude" holds: the three non-klear tests do not touch the changed fixtures. `chunked_gla_matches_the_sequential_recurrence` seeds from literals at `src/models/bailing_moe_linear_tests.rs:419`, so this change cannot move it, and the deepseek_v2 and florence2 arms drive isolated modules.

One claim sits outside the independently confirmed set: "both arms are bit-identical to themselves across repeats in one process, at both precision settings". It is strongly corroborated by the cross-process result, since a single distinct delta over ten processes per mode would be hard to obtain if either arm varied in-process, but it was not re-measured during this review.

The `8167 passed and 0 failed at 670512c2` line predates the fix and was therefore taken against the randomized fixture. It stands, because under the pin klear's delta stayed in the 3.6e-7 to 7.2e-7 band across all ten runs and never approached the bound, but it is worth knowing that the number was sampled from the old fixture.

### Sweep for the same defect class

Swept independently and wider than the original pass, and found no additional site. `keys().cloned().collect()` appears at 15 further places, all production sanitize and loading code, and every one of them performs a per-element rename or lookup with no state carried across iterations; `src/models/step3p5.rs:1161` parses its index out of the key name rather than from position. Other PRNG constants were checked, not just `1_664_525`: `0x9E3779B9` in `dbrx`, `telechat3`, `openelm`, `llama3` and `helium` tests, `6364136223846793005` in `phi3small_tests.rs`, `2_654_435_761` in `llada2_moe/tests.rs`, `1103515245` in `tensor_compress_tests.rs`. All of them reinitialize the LCG state per call or derive it from the element index, so call order cannot affect the result. `enumerate()` over a map iterator: none. `values()` / `values_mut()` / `into_iter()` / `drain()` feeding a carried accumulator: none. The fifth `synthetic_weights` definition, `src/models/muse_glimmer_tests.rs:90`, assigns explicit per-key scales and never walks the map. Inside the four changed files the only remaining map iteration is `for (k, v) in weights.iter()` at `src/models/klear_tests.rs:735`, which filters by key suffix into a fresh map and is order-independent.

### Verification

- [x] All stated requirements implemented (every acceptance criterion in #1265 is met or explicitly refuted with measurements)
- [x] Integrated into project code flow (not applicable in the usual sense: this is a test-fixture and documentation change with no runtime surface, no new module, and no entry point to wire up; the fixtures are already invoked by the tests that own them)
- [x] Project conventions followed (per-file `synthetic_weights` / `filled_weights` / `noise` helpers match how every other model test module is written, `src/test_support/` carries no weight-fixture helper that was passed over, comment widths match the surrounding blocks, `rustfmt --check` clean on the edited file)
- [x] Tests pass (98 tests across the four modules, 0 failures, plus fmt and clippy, all reported in the PR; the review change is comment text only and was not re-run)

Issue #1261, opened 2026-08-20
perf(speculative): the qmv_wide exactness buy-back pins the whole process narrow, taxing batched decode that never asked for it


inureyes:
## Problem

When the MTP exactness probe fails under `qmv_wide` and passes without it, the gate disables the kernel for the whole process and deliberately never re-enables it (`retry_without_qmv_wide` in `src/models/speculative_exactness.rs`: "it is a per-process kernel selection"). That is correct for the MTP verify forward, which is the block the contract is about, but the switch sits on the dispatch path of **every quantized matmul in the process** (`mlxcel_qmv_wide_flag()` in the `quantized.cpp` overlay), so everything else the same server process runs also drops to the narrow kernel:

- batched decode at `B >= 2`, whose per-step projections are exactly the `M >= 2` shape `qmv_wide` exists for;
- prefill for other requests admitted while MTP is engaged;
- any non-MTP request the scheduler serves alongside the speculative one.

On Apple GPU generation 15+ this fires by default now: the probe fails, the retry passes, and the process is pinned narrow (#1199 for Qwen, #1258 for Gemma 4). The measured verify-side cost is 17 to 20% (Qwen) and ~23% (Gemma 4); what the rest of the process pays has never been measured.

## Step 1: measure the tax before designing the scope

Nothing below is worth building until the collateral cost is a number. On one generation 15+ host, `mlxcel-server` with a batch-capable target:

- batched decode throughput at B = 2, 4, 8 with `MLXCEL_QMV_WIDE=1` pinned vs `=0` pinned (no drafter involved, so the comparison isolates the kernel);
- a mixed workload: one MTP stream plus N classic streams, default env (gate flips the process narrow) vs `MLXCEL_MTP_ALLOW_INEXACT=1` with the switch left wide, reading the classic streams' throughput only.

If the B-sweep delta is small on production shapes, the right fix may be documenting the tax and stopping there.

## Step 2 (if the tax is material): scope the exact kernel to the verify forward

The hard part is recorded in the overlay's own comment: MLX evaluates lazily and dispatch happens at eval time on MLX's scheduler thread, so "set off, run verify, set on" from the caller thread does not bracket the kernels it means to bracket, and a caller-side thread-local never reaches the dispatch site. Candidate shapes, in rough order of invasiveness:

[...]

PR #1268, opened 2026-08-20
feat(mtp-policy): expose the settled verdict at /v1/internal/mtp-policy

### LINKED ISSUE #1257: feat(mtp-policy): expose the settled MTP verdict through a supported interface
## Problem / Background

`MtpPolicy` (issue #333, `src/server/batch/mtp_policy.rs`) settles a per-pairing MTP verdict on the actual machine and persists it as a `PolicyHint` under `${MLXCEL_CACHE_DIR:-$HOME/.cache/mlxcel}/mtp-policy/<key-hash>.json`. That verdict is the only honest, machine-specific answer to "does MTP help for this pairing here", and it is exactly what a host application wants to show a user.

Backend.AI GO is about to surface it (lablup/backend.ai-go#4441, epic lablup/backend.ai-go#4429). There is no API for it, so the only available path is to read mlxcel's cache directory from another process: enumerate `mtp-policy/*.json`, parse each file, and match on the body fields (`target`, `drafter`, `block_size`) rather than recomputing the SHA256 file stem.

That works, and it is coupling to a private on-disk format. `HINT_VERSION` is documented to bump whenever the schema or the verdict semantics change, `HINT_SUBDIR` is a private constant, and `PolicyHint` is `pub(crate)`. Any of those can change in a patch release without notice, and the consumer has no way to learn about it except by silently showing nothing. The consumer already treats an unknown `version`, a missing directory, an unparseable file, and a `block_size` mismatch as "no verdict yet" rather than an error, so a format change degrades to a blank surface rather than a crash, but a blank surface is still a regression the user sees and neither project detects.

## Proposed Solution

Expose the settled verdict through a supported interface. Either of these closes the gap; the first is preferred.

**Option A, server API (preferred).** Add a read-only endpoint that reports the current pairing's policy state, for example `GET /v1/internal/mtp-policy` returning the active `(target, drafter, hardware, block_size)` key plus `verdict`, `acceptance_rate`, `samples`, and whether the policy is still profiling (with `PROFILE_SAMPLE_TARGET` so a consumer can say how many more single-request generations are needed). A consumer then reads the verdict from the process that owns it, over the same transport it already uses, with no filesystem coupling and no cache-root resolution.

**Option B, documented file contract.** Keep the cache files as the interface but make them a contract: document the directory layout, the `PolicyHint` body, and the versioning rule in the repository docs; state which fields are stable; and treat a change to them as a breaking change with a `HINT_VERSION` bump plus a changelog entry.

Either way, the "still profiling" state deserves to be expressible. Today a consumer can only infer it from the absence of a file, which is indistinguishable from "the cache root resolved to the wrong place".

## Acceptance Criteria

- [x] The settled MTP verdict for the running pairing is readable through a supported interface (API endpoint or documented file contract)
- [x] The interface distinguishes "settled: enable", "settled: decline", and "still profiling", and reports how many qualifying samples remain when profiling
- [x] The reported payload carries enough of the key (`target`, `drafter`, `hardware`, `block_size`) for a consumer to confirm the verdict describes the pairing it is showing
- [x] Versioning and the compatibility policy for the interface are documented
- [x] Backend.AI GO's interim cache read can be replaced by the supported interface

## Implementation notes

Option A landed as `GET /v1/internal/mtp-policy` (`src/server/routes/mtp_policy.rs`), documented in `docs/mtp-policy-api.md` including the `schema_version` compatibility policy. The on-disk hints are untouched and keep working, so the shipped Backend.AI GO reader (lablup/backend.ai-go#4441) is not broken by this change; it can migrate to the endpoint at its own pace.

Beyond the three states the body asks for, the endpoint reports a fourth, `forced` (`MLXCEL_ENABLE_MTP_B1` pinned the decision), separately from `settled`, because an operator pin is not a measured verdict. It also reports `unavailable` with a `reason` (`no_mtp_dispatch`, `adaptive_disabled`, `worker_not_ready`), which is what removes the "still profiling is indistinguishable from the cache root resolved to the wrong place" ambiguity named above.

## Technical Considerations

Consumer side: lablup/backend.ai-go#4441 ships the cache read as the interim path, with a module doc comment naming this issue and stating plainly that it couples to a private on-disk format. Replacing it with whichever interface lands here is a small, localized change on that side.

Relevant constants at `4968b72e`: `HINT_VERSION = 3`, `HINT_SUBDIR = "mtp-policy"`, `PROFILE_SAMPLE_TARGET = 4`, `ENABLE_SPEEDUP_FLOOR = 1.5`, `DECLINE_SPEEDUP_CEIL = 1.0`. Cache root resolution is `mlxcel_core::cache_root()` (`MLXCEL_CACHE_DIR`, else `$HOME/.cache/mlxcel`).

Refs #333

inureyes:
## Summary

The adaptive MTP policy's settled verdict was only readable by parsing the private per-pairing hint files under `${MLXCEL_CACHE_DIR:-$HOME/.cache/mlxcel}/mtp-policy/`, and a consumer already ships that read (lablup/backend.ai-go#4441). This adds `GET /v1/internal/mtp-policy`, a supported, versioned read interface served by the process that owns the decision, exposing everything the hint file carries plus the states a file cannot express. The hint files themselves are untouched and keep working, so the shipped consumer is not broken and can migrate at its own pace.

## What changed

- `src/server/routes/mtp_policy.rs` (new): the `GET /v1/internal/mtp-policy` handler and its wire body (`MtpPolicyResponse`, `MTP_POLICY_SCHEMA_VERSION = 1`). The mapping lives in a pure `build_mtp_policy_response` helper so it is testable without an `AppState`, mirroring `cache::build_stats_response`.
- `src/server/app.rs`: registers the route. Always mounted, like `/v1/cache/stats`.
- `src/server/batch/mtp_policy.rs`: adds the published view (`MtpPolicyStatus`, `MtpPolicyUnavailableReason`, `MtpPolicySnapshot`), `MtpPolicy::snapshot()` and `MtpPolicy::is_profiling()`, and makes `Verdict` public. `PolicyState::Settled` becomes a struct variant carrying the acceptance rate and sample count behind the verdict.
- `src/server/batch/observability.rs`: `BatchObservability` gains `set_mtp_policy` / `mtp_policy_snapshot`, and `ObservabilitySnapshot` gains an `mtp_policy` field so `/health` carries the same state.
- `src/server/batch/scheduler.rs`: publishes the resolved policy state from `with_mtp_policy`, and republishes after each profiled B=1 burst while the policy is still profiling.
- `docs/mtp-policy-api.md` (new): the endpoint reference, the state and reason tables, and the versioning and compatibility policy. Linked from `README.md`, `docs/README.md`, `docs/CONTINUOUS_BATCHING.md`, and the `MLXCEL_MTP_ADAPTIVE` row in `docs/environment-variables.md`.
- Tests: `src/server/routes/mtp_policy_tests.rs` (new, 8 tests) for the response shape, and 9 new tests in `src/server/batch/mtp_policy_tests.rs` for the state mapping plus 1 in `observability.rs` for the publish/read seam.

## Design decisions worth reviewing

**Four states, not three.** The issue names `settled: enable`, `settled: decline`, and `still profiling`. `PolicyState` has a fourth, `Forced` (`MLXCEL_ENABLE_MTP_B1` pinned the decision and the policy never profiles), and the endpoint reports it separately rather than flattening it into `settled`. An operator pin is not a measured verdict: a consumer rendering it as "this machine measured MTP as worth it here" would be reporting something nobody measured, and that is exactly the case where an operator is most likely to have overridden the measurement.

**A fifth state, `unavailable`, with a reason.** A server with no MTP dispatch, or with `MLXCEL_MTP_ADAPTIVE=0`, or whose worker has not started, answers with a well-formed body rather than a 404 or a 500. The `reason` field (`no_mtp_dispatch`, `adaptive_disabled`, `worker_not_ready`) is what removes the ambiguity the issue names: with the file read, an empty directory meant "still profiling", "no MTP configured", and "the cache root resolved to the wrong place" all at once. When the adaptive path is disabled the response still reports what the static per-hardware gate decided, via `mtp_enabled`, rather than leaving the consumer to guess.

**Threading.** `MtpPolicy` stays single-threaded and lock-free on the decode path; no lock was added around it and it was not made `Arc<Mutex<..>>`. The scheduler publishes an `MtpPolicySnapshot` into the already `Arc`-shared `BatchObservability` at the points where the state can change (once when `with_mtp_policy` resolves, then after each profiled B=1 burst until the verdict settles), and the handler reads it back from the snapshot. Storage is a `Mutex<Option<MtpPolicySnapshot>>` rather than atomics because the value carries the key's three strings; that matches the existing `PromptCacheRejectCounters` last-reject field in the same module rather than inventing a pattern. It is never written per decode step, and never at all once the verdict is forced or settled.

**Consistency with the hint file.** `PolicyState::Settled` now carries the acceptance rate and sample count, taken from the hint's rounded values, so a verdict restored from disk publishes a view identical to the one the profiling process published, and a consumer comparing the endpoint against the file during migration will not see them differ by a rounding step. A test pins both.

**Always mounted, not flag-gated.** Unlike `/props`, `/slots`, and `/metrics`, this is not behind a CLI flag. Gating it would reproduce the failure it exists to remove: a consumer unable to tell "nothing to report" from "this server does not answer". It is behind the same API-key middleware as every other route, and the payload is as coarse as the already world-readable hint file (a verdict, an acceptance rate, a sample count, the pairing identity; no prompt data, no token ids).

**Path.** `GET /v1/internal/mtp-policy` as the issue proposes, since an external consumer will code against it. There is no other `/v1/internal/*` route in the tree yet.

## Acceptance criteria

- [x] The settled MTP verdict for the running pairing is readable through a supported interface.
- [x] The interface distinguishes settled-enable, settled-decline, and still-profiling, and reports `samples_remaining` while profiling (plus `samples_required` so a consumer can render "2 of 4" without hard-coding the constant).
- [x] The payload carries the full key: `target`, `drafter`, `hardware`, `block_size`.
- [x] Versioning and the compatibility policy are documented in `docs/mtp-policy-api.md`.
- [x] Backend.AI GO's interim cache read can be replaced by the endpoint: it exposes the four key fields, `verdict`, `acceptance_rate`, and `samples`, which is a superset of what the file gives them.

## Out of scope, deliberately

The on-disk hint format, `HINT_VERSION`, and `PolicyStore` are unchanged, and the cache files are neither removed nor deprecated, because a consumer depends on them today. `src/models/speculative_exactness.rs` and the `qmv_wide` machinery are untouched (issue #1261).

## Test plan

Run on a CUDA/Linux aarch64 (GB10) host with the `test-fast` profile. Exit codes captured directly, not from a pipeline.

- [x] `cargo fmt -- --check` (exit 0)
- [x] `cargo test --profile test-fast --features cuda --lib server::batch::mtp_policy` (exit 0, 48 passed, 9 new)
- [x] `cargo test --profile test-fast --features cuda --lib server::routes::mtp_policy` (exit 0, 8 passed, all new)
- [x] `cargo test --profile test-fast --features cuda --lib server::batch::observability` (exit 0, 16 passed, 1 new)
- [x] `cargo test --profile test-fast --features cuda --lib server::routes::health` (exit 0, 8 passed)
- [x] `cargo test --profile test-fast --features cuda --lib server::routes::cache` (exit 0, 19 passed)
- [x] `cargo test --profile test-fast --features cuda --lib server::app` (exit 0, 7 passed)
- [x] `cargo test --profile test-fast --features cuda --lib server::batch::scheduler` (exit 0, 108 passed, 7 ignored)
- [x] `cargo test --profile test-fast --features cuda --lib server::batch::speculative` (exit 0, 77 passed)

Not run here: the full `make verify-test-cuda` workspace gate, and any live-server curl against the endpoint (that needs a loaded MTP pairing). The route's wiring is compile-checked through `create_app`, and the response mapping and the publish/read seam are covered by unit tests.

Refs #333

Closes #1257


inureyes:
`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at `9ff0a3ad`: **8188 passed, 0 failed, 310 ignored**, exit 0.

That is +17 against `main` at `a940c737` (8171 passed), which matches this branch exactly: the diff adds 17 `#[test]` functions (8 in `src/server/batch/mtp_policy_tests.rs`, 8 in `src/server/routes/mtp_policy_tests.rs`, 1 in `src/server/batch/observability.rs`) and removes none.

inureyes:
Re-ran the gate after the accuracy pass, because `0f624651` touches production code (`src/server/model_worker.rs`) and the earlier run was against `9ff0a3ad`.

`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at `0f624651`: **8188 passed, 0 failed, 310 ignored**, exit 0. Same totals as the pre-fix run, which is the expected result: the two new `set_mtp_policy` calls publish into observability at worker startup and add no tests.

`cargo fmt --all -- --check` and `cargo clippy --lib --tests --features cuda -- -D warnings` are clean on the same tree.

PR #1269, opened 2026-08-20
fix(lang-bias): preserve YAML bias: order so priority is deterministic

### LINKED ISSUE #1267: fix(lang-bias): the YAML bias map is a HashMap, so language priority is randomized per process
## Problem / Background

`LangBiasSet` (`src/lib/mlxcel-core/src/lang_analyzer/mod.rs:714-724`) declares `ordered: Vec<(LanguageCode, f32)>` and documents it as "`(language, bias)` pairs in priority order (index 0 = highest priority)". Its consumer `to_token_bias` (`src/lib/mlxcel-core/src/lang_analyzer/mod.rs:864-876`) implements "First-language-wins: only insert if not already claimed", so a language's position inside `ordered` decides which bias value a token shared by several languages receives.

The YAML config path does not preserve any order. `LangBiasYamlConfig::bias` is `Option<HashMap<String, BiasValueStr>>` (`src/lang_bias.rs:180`) and the resolve loop at `src/lang_bias.rs:404-427` iterates that map directly (`for (code_str, BiasValueStr(bias)) in yaml_bias`) and pushes each entry straight into `ordered`. Rust's `HashMap` uses `RandomState`, so the priority order handed to `to_token_bias` is a fresh random permutation of the configured languages, both across processes and across repeated deserializations inside one process.

Measured with a standalone probe against serde_yaml 0.9.34 (the version pinned in `Cargo.lock`), deserializing `ja: -1.0\nzh: -2.0\nko: 5.0` into `HashMap<String, f32>`: 6 separate processes produced 5 distinct key orders, and 10 repeated `from_str` calls inside a single process also produced 5 distinct key orders.

## Current Behavior

Han script is shared by `ja`, `zh` and `ko` (`scripts_for`, `src/lib/mlxcel-core/src/lang_analyzer/mod.rs:730-745`: Japanese includes `Script::Han` under both policies, Chinese is `{Han}`, Korean Conservative includes `Han`). A YAML config naming two or more CJK languages therefore assigns a different bias to every shared Han token on every run, with no error and no warning, so generation behavior changes run to run for the same config file. The documented schema example in the doc comment for `LangBiasYamlConfig` (`src/lang_bias.rs:159-172`) is itself exactly this trigger case: it lists `ja: -inf`, `zh: -10.0` and `ko: +5.0`, so a user copying the shipped example gets `-inf` on the shared Han tokens on some runs and `+5.0` on others.

Nothing surfaces the change. The user-facing line printed by the generate path, "Language bias active: N token entries biased" (`src/commands/generate.rs:2243-2249`), is identical across runs, because the union of biased token ids does not depend on the order, only the assigned values do. The only trace of the ordering is the `languages` field of the DEBUG-level `lang_bias resolved` event (`src/commands/generate.rs:2250-2284`, `src/server/model_worker.rs:366-400`), which is built by joining `bias_set.ordered`; it silently prints a different order per run, at a log level that is off by default.

The CLI sibling `parse_lang_bias_entries` (`src/lang_bias.rs:98-156`) builds the same structure from `s.split(',')` and IS order-preserving, and the server env fallback `env_fallback_lang_bias` (`src/server/cli_input.rs:999-1010`) routes `LLAMA_ARG_LANG_BIAS` through that same parser. Two of the three entry points preserve author order and only the YAML path does not, which confirms that order is intended to be significant and that the entry points currently disagree.

Secondary defect in the same block: the `seen` / `CliError::DuplicateLanguageCode` check at `src/lang_bias.rs:408-411` is unreachable, because a `HashMap` cannot contain duplicate keys and serde's `HashMap` deserializer resolves them before this code runs. Confirmed against serde_yaml 0.9.34: `ja: -1.0\nzh: -2.0\nja: -3.0` deserializes without error into `{"zh": -2.0, "ja": -3.0}` (last occurrence wins, no diagnostic), and serde_yaml's duplicate-entry error path exists only for `Mapping`/`Value` (`mapping.rs`), not for a typed `HashMap`. So a YAML file with a repeated language key is silently accepted while the equivalent CLI string is rejected.

Why this is not caught today: the only multi-language YAML test, `yaml_well_formed_parses` (`src/lang_bias.rs:586-608`), stops at the deserializer and asserts map membership (`bias["ja"]`, `bias["zh"]`, `bias["ko"]`), which is order-independent. Every YAML test that exercises the full `LangBiasCliArgs::resolve` path (`src/lang_bias.rs:640`, `:739`, `:762`, `:813`) uses a single `bias:` entry, so no test ever observes `ordered` holding more than one YAML-sourced language.

## Proposed Solution

Preserve author order from the YAML document so the YAML path matches the CLI path's semantics. The final representation is the maintainer's call; two workable shapes:

- Deserialize `bias` as a sequence of single-entry maps (`Option<Vec<HashMap<String, BiasValueStr>>>`). Order becomes explicit in the document and duplicates survive into the parsed data, at the cost of a schema change (`bias:` becomes a list), which breaks existing config files unless an untagged enum accepts both shapes.
- Keep the current mapping syntax and deserialize into an order-preserving representation: a small `Visitor` collecting `Vec<(String, BiasValueStr)>` through `MapAccess` (no new dependency), or `indexmap::IndexMap` if adding the dependency is acceptable. Existing config files stay valid and the existing duplicate check becomes reachable.

The second shape is preferable if backward compatibility for existing `bias:` mappings matters, and the `MapAccess` visitor needs no new dependency. Under either shape, settle the duplicate question explicitly: either make `CliError::DuplicateLanguageCode` reachable (a `Vec`-backed representation preserves the repeat, so the existing `seen` logic starts working and YAML matches the CLI's rejection), or delete the dead check and document that the last occurrence wins. Do not leave unreachable code claiming a diagnostic that never fires.

## Scope

**In scope:** `src/lang_bias.rs` only: the `LangBiasYamlConfig::bias` field type (`:180`), the YAML resolve loop and its duplicate check (`:404-427`), the schema doc comment example (`:159-172`) if the accepted syntax changes, and new unit tests in the existing `mod tests` (`:472-473`).

**Out of scope:** `LangBiasSet` itself (already `Vec`-backed and correct), `parse_lang_bias_entries` and the `LLAMA_ARG_LANG_BIAS` env fallback (already order-preserving), `to_token_bias`'s first-language-wins rule (the intended semantics, not the bug), and the `scripts_for` script sets.

## Implementation Notes

- **Reuse**: keep `BiasValueStr` and its `parse_bias_f32` handling, the `LanguageCode::from_str` error mapping into `CliError::UnknownLanguageCode`, and the existing `seen` duplicate logic rather than adding a second validation path. `parse_lang_bias_entries` (`src/lang_bias.rs:98-156`) is the reference for intended error behavior; the YAML path should agree with it entry for entry.
- **Constraints**: `#[serde(deny_unknown_fields)]` on `LangBiasYamlConfig` must stay. If the `bias:` syntax changes, either accept both the old mapping and the new sequence, or state the break in the doc comment and in `CHANGELOG.md`; config files in the wild use the mapping form documented at `:159-172`.
- **Edge cases**: an absent or empty `bias:` must still resolve to an empty `LangBiasSet` rather than an error, matching `yaml_missing_bias_field_defaults_to_empty` (`src/lang_bias.rs:612`); a single entry must behave exactly as today; duplicate keys must have a stated, tested outcome; CLI `--lang-bias` still replaces the YAML set entirely (Step 3a, `src/lang_bias.rs:435-438`) and that precedence must not change.
- **Error handling**: unknown language codes keep returning `CliError::UnknownLanguageCode` with the existing `entry: format!("{code_str}: (from YAML)")` context. A rejected duplicate returns `CliError::DuplicateLanguageCode` carrying the offending code, the same variant the CLI path already uses.
- **Test determinism**: a single-run assertion is weak, since a random permutation of two languages passes half the time. Assert the full expected `ordered` vector and repeat the resolve in a loop inside one test. `RandomState` varies per `HashMap` instance, not only per process (10 repeated `from_str` calls in one process produced 5 distinct orders in the probe above), so a looped assertion fails reliably against the current code.

## Acceptance Criteria

- [x] A YAML config whose `bias:` block lists `ja`, `zh`, `ko` in that order resolves to `LangBiasSet.ordered == [(Ja, ..), (Zh, ..), (Ko, ..)]`, identical on every run and on every repeated resolve within one process.
- [x] `to_token_bias` fed that config deterministically assigns the first-listed language's bias to the Han tokens shared by the listed languages.
- [x] The YAML path and the CLI path (`--lang-bias "ja=-inf,zh=-10.0,ko=+5.0"`) produce identical `LangBiasSet` values for equivalent input.
- [x] Duplicate language keys inside a YAML `bias:` block have a defined outcome asserted by a test: either `CliError::DuplicateLanguageCode` (matching the CLI path) or a documented last-wins, with no unreachable code left behind claiming a diagnostic that cannot fire.
- [x] The fix reaches the shipped code flow, not just the parser: the CLI `generate` path and the server `model_worker` path consume the resolved `LangBiasSet` unchanged, and the `languages` field of the DEBUG `lang_bias resolved` event prints the configured order rather than a random one.
- [x] New regression tests live in `src/lang_bias.rs`'s `mod tests`, cover at least two CJK languages, and loop the resolve at least 10 times asserting a fixed order. The test must fail on `a940c737` and pass after the fix.
- [x] The doc comment schema example (`src/lang_bias.rs:159-172`) matches the accepted syntax after the change, and `CHANGELOG.md` records the behavior fix.

## Verification

```bash
cargo fmt --all -- --check
cargo clippy --lib --tests --features cuda -- -D warnings
cargo test --lib --features cuda lang_bias -- --test-threads=1
```

A pass is: fmt and clippy clean, and the `lang_bias` unit tests green including the new multi-CJK ordering test. On Apple Silicon substitute `--features metal,accelerate`. Confirm the new test is load-bearing by running it against the pre-fix `src/lang_bias.rs`: the looped assertion should fail on essentially every run. Manual check on a host with a multilingual model: run the same multi-CJK YAML config 5 times with `RUST_LOG=debug` and confirm the `languages` field of the `lang_bias resolved` event is identical on all 5 runs.

## Technical Considerations

Found while implementing #1265 (PR #1266, merged as `a940c737`), which fixed the same root-cause class, `HashMap` iteration order leaking into ordered state, in four test fixtures. That mechanism is written up at the `verify-test-cuda` gate definition in the `Makefile` (`Makefile:604-614`). This instance is distinct: it sits in production code and changes shipped generation behavior rather than a test fixture.

inureyes:
## Summary

A `--lang-bias-config` YAML file naming two or more CJK languages steered differently on every run. `LangBiasYamlConfig::bias` deserialized into a `HashMap`, and the resolve loop pushed that map's iteration order straight into `LangBiasSet.ordered`, which is the priority order `TokenLanguageIndex::to_token_bias` resolves conflicts with under first-language-wins. `RandomState` randomizes iteration order per map instance, so the priority order was a fresh random permutation on every load. Han is shared by `ja`, `zh` and `ko`, so every token they share got a different bias on every run, with no error and no warning. The schema example in the `LangBiasYamlConfig` doc comment is itself a three-CJK config, so copying the shipped example was enough to trigger it.

The same `HashMap` collapsed repeated keys during deserialization (serde_yaml resolves them last-wins with no diagnostic), which made the `seen` / `DuplicateLanguageCode` check in the resolve loop unreachable and let the YAML path silently accept input the `--lang-bias` parser has always rejected.

## The accepted YAML schema does not change

`bias:` is still a plain mapping. Existing config files keep parsing unchanged, a sequence-shaped `bias:` block is still a parse error (pinned by `yaml_bias_block_rejects_a_sequence`), an absent or empty block still resolves to an empty `LangBiasSet` rather than an error, `#[serde(deny_unknown_fields)]` stays, and CLI `--lang-bias` still replaces the YAML set entirely.

Two behavior changes are visible to users, both recorded in `CHANGELOG.md`. A file that happened to rely on the random ordering now steers consistently toward its first-listed language, which is the documented intent. And a file repeating a language code now errors instead of silently taking the last occurrence.

## Approach

Candidate 1 from the issue's implementation notes: a hand-written `Deserialize` collecting the block through `MapAccess`. `bias` becomes `Option<BiasEntries>`, where `BiasEntries` is a newtype over `Vec<(String, BiasValueStr)>`. Document order survives, and a repeated key arrives as a genuine second entry, so the existing `seen` check starts firing and the YAML path reaches `CliError::DuplicateLanguageCode` (the same variant, not a parallel validation path). No new dependency.

`serde_yaml::Mapping` was the alternative. It preserves order and rejects duplicates, but with serde_yaml's own error rather than the repo's, so the two entry points would still disagree on what a duplicate looks like. `indexmap` as the field type was rejected outright: it preserves order but resolves duplicates last-wins, so the dead check would have stayed dead.

The `seen` / `ordered` loop, `BiasValueStr`, `parse_bias_f32`, and the `UnknownLanguageCode` mapping with its `"{code_str}: (from YAML)"` context are all untouched. `LangBiasSet` and `to_token_bias` are untouched: they were already correct, only the YAML producer was wrong. Nothing outside `src/lang_bias.rs` names `LangBiasYamlConfig`, so the CLI `generate` path and the server `model_worker` path consume the resolved `LangBiasSet` exactly as before, and the `languages` field of the DEBUG `lang_bias resolved` event now prints the configured order.

## Proof the regression tests are load-bearing

The four new tests were run against the pre-fix production path (the `bias` field reverted to `Option<HashMap<String, BiasValueStr>>`, tests unchanged). All four fail:

```
test lang_bias::tests::yaml_and_cli_paths_agree_on_multi_cjk_order ... FAILED
test lang_bias::tests::yaml_multi_cjk_bias_keeps_document_order_across_repeated_resolves ... FAILED
test lang_bias::tests::yaml_multi_cjk_first_language_wins_on_shared_han_tokens ... FAILED
test lang_bias::tests::yaml_duplicate_language_code_is_rejected ... FAILED

---- yaml_multi_cjk_bias_keeps_document_order_across_repeated_resolves stdout ----
assertion `left == right` failed: iteration 0: entry 0 should be Ja but the resolved order was [(Zh, -10.0), (Ko, 5.0), (Ja, -inf)]; YAML bias: order must be the priority order
  left: Zh
 right: Ja

---- yaml_multi_cjk_first_language_wins_on_shared_han_tokens stdout ----
assertion `left == right` failed: iteration 2: the shared Han token must take the first-listed language's bias (ja = -inf), resolved order was [(Zh, -10.0), (Ja, -inf), (Ko, 5.0)]
  left: Some(-10.0)
 right: Some(-inf)

---- yaml_and_cli_paths_agree_on_multi_cjk_order stdout ----
assertion `left == right` failed: iteration 0: the YAML path and the --lang-bias path must resolve equivalent input to the same LangBiasSet
  left: [(Ja, -inf), (Ko, 5.0), (Zh, -10.0)]
 right: [(Ja, -inf), (Zh, -10.0), (Ko, 5.0)]

---- yaml_duplicate_language_code_is_rejected stdout ----
a repeated language code in a YAML bias: block must be rejected: Some(LangBiasConfig { bias_set: LangBiasSet { ordered: [(Zh, -2.0), (Ja, -3.0)] }, ... })

test result: FAILED. 26 passed; 4 failed
```

Three distinct permutations of the same three-key file appeared inside that single process run, which is the per-map-instance randomization the issue measured. Each ordering test performs 32 full `resolve()` calls rather than one, because a single resolve over three languages comes out right often enough to prove nothing; the observed failures landed at iteration 0 and iteration 2.

## What changed

- `src/lang_bias.rs`: new `BiasEntries` newtype over `Vec<(String, BiasValueStr)>` with a `MapAccess`-based `Deserialize`, `IntoIterator`, and `as_slice` / `len` / `is_empty`. The doc comment carries why a `HashMap` was wrong on both counts.
- `src/lang_bias.rs`: `LangBiasYamlConfig::bias` is now `Option<BiasEntries>`. The schema doc comment gains a paragraph stating that `bias:` order is the priority order and that a repeated code is rejected.
- `src/lang_bias.rs`: `CliError::DuplicateLanguageCode`'s message no longer says the duplicate came from `--lang-bias`, since both entry points now raise it. The variant and its `code` field are unchanged.
- `src/lang_bias.rs`: the resolve loop is unchanged apart from a comment; it now receives entries in document order with repeats intact.
- `src/lang_bias.rs` tests: `yaml_well_formed_parses` asserts order instead of map membership, which is the blind spot that hid this. Six new tests: three looping order/parity/`to_token_bias` tests, a duplicate-rejection test, an empty-block test, and a sequence-rejection test pinning the schema.
- `CHANGELOG.md`: an `## [Unreleased]` entry covering both user-visible changes.

## Test plan

- [x] `cargo check --bins --features cuda` (exit 0)
- [x] `cargo test --profile test-fast --features cuda --lib lang_bias` (exit 0, 30 passed / 0 failed)
- [x] The same test run against the pre-fix `bias` field type: 26 passed / 4 failed, output above.

Not run here: the full `make verify-test-cuda` workspace gate.

Closes #1267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Fixed
+
+- **A `--lang-bias-config` YAML file with two or more languages now steers deterministically** (#1267). The `bias:` block deserialized into a `HashMap`, whose iteration order `RandomState` randomizes per map instance, and that iteration order was pushed straight into `LangBiasSet.ordered`, which is the priority order `TokenLanguageIndex::to_token_bias` resolves conflicts with under first-language-wins. Han is shared by `ja`, `zh` and `ko`, so a config naming two or more CJK languages assigned a different bias to every shared Han token on every run, with no error and no warning, and the one place the order surfaced was the `languages` field of a DEBUG-level tracing event that is off by default. The shipped schema example is itself a three-CJK config, so copying it was enough to hit this. The `bias:` block is now collected through `MapAccess` into an ordered `Vec`, so index 0 is the first language written in the file, matching what `--lang-bias ja=-inf,zh=-10.0,ko=+5.0` and the `LLAMA_ARG_LANG_BIAS` env fallback have always done. **The accepted YAML syntax is unchanged**: `bias:` is still a plain mapping and existing config files keep working, though a file relying on the previous random behavior will now steer consistently toward its first-listed language.
+- **A language code repeated inside one YAML `bias:` block is rejected** (#1267). `serde_yaml` resolves a repeated key into a typed `HashMap` last-wins with no diagnostic, which made the duplicate check in the resolve path unreachable and let the YAML path silently accept input the `--lang-bias` parser has always rejected. The ordered representation delivers both occurrences, so both entry points now fail with `duplicate language code '<code>' in language bias entries (ambiguous priority)`. This is a behavior change for any config file that repeats a language code: it previously took the last occurrence and now errors out.
+
 ## [v0.5.2] - 2026-08-18
 
 ### Changed
--- a/TECHNICAL_REPORTS/1269-lang-bias-yaml-order-20260822.en.md
+++ b/TECHNICAL_REPORTS/1269-lang-bias-yaml-order-20260822.en.md
@@ -1,0 +1,70 @@
+# Technical Report: PR #1269 - Deterministic language priority from a YAML bias block
+
+## Executive Summary
+
+`LangBiasSet.ordered` is documented as "pairs in priority order (index 0 = highest priority)" and its consumer `to_token_bias` resolves shared tokens under "first-language-wins". The YAML config path built that order by iterating a `HashMap`, so `RandomState` decided the priority. Han script is shared by `ja`, `zh` and `ko`, which means a `--lang-bias-config` file naming two or more CJK languages assigned a different bias to every shared token on every run, silently.
+
+This is the same root-cause class as #1265 (`HashMap` iteration order leaking into ordered state), but in production code rather than a test fixture. The fix collects the `bias:` block through `MapAccess` into an ordered `Vec`, so the author's document order becomes the priority order, matching what `--lang-bias` and the `LLAMA_ARG_LANG_BIAS` fallback have always done.
+
+## 1. Problem Statement
+
+Three entry points produce a `LangBiasSet` and only one of them was wrong.
+
+`parse_lang_bias_entries` walks `s.split(',')` in document order, uses a `seen` map purely as a membership set, and rejects duplicates with `CliError::DuplicateLanguageCode`. `env_fallback_lang_bias` routes `LLAMA_ARG_LANG_BIAS` through that same parser. The YAML path deserialized `bias:` into `Option<HashMap<String, BiasValueStr>>` and pushed the map's iteration order straight into `ordered`.
+
+Two consequences followed, and the second was invisible. Priority became random per run. And because `serde_yaml` resolves a repeated key into a typed `HashMap` last-wins with no diagnostic, the duplicate check in the resolve loop was unreachable, so the YAML path silently accepted input the `--lang-bias` parser has always rejected.
+
+The trigger is not exotic: the schema example in the `LangBiasYamlConfig` doc comment is itself a three-CJK config, so copying the documented example was enough.
+
+## 2. Technical Decisions
+
+### 2.1 A hand-written `Deserialize` rather than an order-preserving map type
+
+Three candidates were considered. `IndexMap` as the field type preserves order but resolves duplicates last-wins, which would leave the dead check dead and keep the YAML path accepting what the CLI rejects. `serde_yaml::Mapping` is `IndexMap`-backed and does reject duplicates, but with serde_yaml's own error rather than `CliError::DuplicateLanguageCode`, so the two entry points would still disagree on the error surfaced.
+
+The chosen shape is a `BiasEntries` newtype over `Vec<(String, BiasValueStr)>` with a hand-written `Deserialize` that collects through `MapAccess::next_entry`. Document order survives, a repeated key arrives as a second entry that the existing resolve loop rejects with the repository's own error, and no new dependency is added. `indexmap` is present in `Cargo.lock` but not in `Cargo.toml`, so both map-type candidates would have promoted a transitive dependency to a direct one.
+
+### 2.2 The accepted YAML schema does not change
+
+The visitor implements `visit_map` and the entry point is `deserialize_map`, so `bias:` remains a plain mapping. `#[serde(deny_unknown_fields)]` is untouched, an absent or empty block still resolves to an empty set, and a sequence-shaped block is still a parse error. Existing config files keep working. A fix that required users to rewrite `bias:` as a list would have been the wrong shape for a bug report about ordering.
+
+### 2.3 The duplicate error message names both surfaces
+
+Making the check reachable from YAML made the existing text, which named only `--lang-bias`, actively wrong for a user who wrote a YAML file. The variant and its `code` field are unchanged; only the message broadened to name both surfaces.
+
+## 3. Change Summary
+
+| File | Change |
+| --- | --- |
+| `src/lang_bias.rs` | `BiasEntries` newtype plus its `Deserialize`; `LangBiasYamlConfig::bias` retyped; duplicate error message broadened; new tests |
+| `CHANGELOG.md` | `## [Unreleased]` entries for the two user-visible behavior changes |
+
+## 4. Review Findings
+
+The requirement that carried the most weight was not a review finding but a precondition: the regression tests had to be demonstrated failing against the unfixed code. This bug class is unusually good at producing tests that pass before and after, because a single `resolve()` returns the correct order by luck a fair fraction of the time.
+
+The demonstration reverted only the field type (and the one pre-existing test whose membership assertions would not compile against the ordered type), ran the suite, and restored from a copy rather than using `git stash`, so untracked work was never at risk. Four tests failed, and the output showed three distinct permutations of the same three-key file within a single process run:
+
+```
+iteration 0: [(Zh, -10.0), (Ko, 5.0), (Ja, -inf)]
+iteration 2: [(Zh, -10.0), (Ja, -inf), (Ko, 5.0)]
+           : [(Ja, -inf), (Zh, -10.0), (Ko, 5.0)]
+```
+
+That independently reproduces the measurement taken while filing #1267: `RandomState` randomizes per `HashMap` instance, not merely per process (ten maps from the same five keys gave nine distinct orders in one process). It is also why each ordering test runs 32 full `resolve()` calls rather than one.
+
+## 5. Validation
+
+Measured on GB10 (DGX Spark, CUDA sm_121, Linux aarch64).
+
+- `cargo test --profile test-fast --features cuda --lib lang_bias`: 30 passed, exit 0. Against the pre-fix field type: 26 passed, 4 failed, exit 101.
+- `cargo fmt --all -- --check`, `cargo clippy --lib --tests --features cuda -- -D warnings`, `cargo check --lib --tests --features cuda`, `cargo check --bins --features cuda`: all exit 0.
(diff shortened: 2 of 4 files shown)


inureyes:
`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at `6f934905`: **8194 passed, 0 failed, 310 ignored**, 101 suites, run to completion through the doc-test units.

That is +6 against `main` at `fdc19666` (8188 passed), and the diff adds exactly 6 `#[test]` functions and removes none, so the totals reconcile.

`cargo fmt --all -- --check`, `cargo clippy --lib --tests --features cuda -- -D warnings`, `cargo check --lib --tests --features cuda` and `cargo check --bins --features cuda` are clean on the same tree.

PR #1281, opened 2026-08-22
fix(rt-detr-v2): make needs_sanitize independent of HashMap order

### LINKED ISSUE #1276: fix(rt-detr-v2): needs_sanitize returns on the first marker found in HashMap order, so its verdict is order-dependent
## Problem / Background

`needs_sanitize` (`src/vision/detection/rt_detr_v2/sanitize.rs:46-73`) decides whether a freshly loaded weight map is raw HuggingFace layout and therefore needs the rename plus transpose pipeline. It walks `weights.keys()`, and `WeightMap` is a `std::collections::HashMap` (`src/lib/mlxcel-core/src/weights.rs:29`: `pub type WeightMap = HashMap<String, UniquePtr<MlxArray>>`), so the walk order is randomized by `RandomState`, per map instance rather than merely per process. Inside the loop the function sets `has_mlx_marker` for keys starting with `vision.backbone.` or `vision.hybrid_encoder.` (lines 50-52), sets `has_hf_marker` for six HF markers (lines 53-61), and then returns immediately: `if has_mlx_marker { return false }` at lines 63-65, followed by `if has_hf_marker { return true }` at lines 66-68.

The comment at line 62 says "Early out once both decisions are unambiguous", but the code does not wait for both. It returns as soon as EITHER flag is set. For any weight map that contains both marker families, the answer is decided by whichever key the randomized walk happens to reach first, so the same file can produce different verdicts on different runs. The comment should be treated as part of the defect, since it describes a guard the code does not implement.

A wrong answer is expensive in one direction. The module documentation at `src/vision/detection/rt_detr_v2/sanitize.rs:30-32` states that MLX-layout checkpoints must skip the pipeline because "re-running it would double-transpose conv weights". A spurious `true` therefore runs the rename and transpose over already-transposed weights and silently corrupts them: the model still loads through `RtDetrV2Model::load` (`src/vision/detection/rt_detr_v2/model.rs:110-115`) and still emits detections, they are just wrong. A spurious `false` on a genuine HF checkpoint is the milder direction: the rename never runs and the load fails or mismatches loudly.

Reachability, stated precisely so this is not overclaimed. Two of the HF markers are unanchored substring tests, `key.contains(".convolution.")` (line 55) and `key.contains(".normalization.")` (line 56), so they can match anywhere in a key rather than only at a prefix. A correctly converted MLX checkpoint should not trip them, because `rename_stripped` in the same file (lines 102-103, reached through the public `rename_key` at lines 163-165) rewrites `.convolution.` to `.conv.` and `.normalization.` to `.bn.`. So for well formed inputs on both sides this is latent rather than live. It becomes reachable when a weight map carries both families at once: a partially converted or hand merged checkpoint, an MLX checkpoint shipping extra auxiliary tensors under HF field names, or a future MLX layout that keeps any `.normalization.` shaped key. The point of this issue is that the function currently has no defense against that case and fails toward the corrupting direction at random rather than deterministically or loudly.

One more detail worth fixing while there: the trailing `has_hf_marker` return at line 72 can only be reached when no key set either flag, in which case `has_hf_marker` is necessarily `false`. It reads as though it could return `true` and cannot. Returning `false` explicitly, keeping the existing "treat as already-MLX to avoid corrupting an unexpected layout" reasoning from the comment at lines 70-71, would say what actually happens.

## Current Behavior

```rust
pub fn needs_sanitize(weights: &WeightMap) -> bool {
    let mut has_mlx_marker = false;
    let mut has_hf_marker = false;
    for key in weights.keys() {
        if key.starts_with("vision.backbone.") || key.starts_with("vision.hybrid_encoder.") {
            has_mlx_marker = true;
        }
        if key.starts_with("model.backbone.")
            || key.starts_with("backbone.model.")
            || key.contains(".convolution.")
            || key.contains(".normalization.")
            || key.starts_with("model.decoder.")
            || key.starts_with("model.encoder.")
        {
            has_hf_marker = true;
        }
        // Early out once both decisions are unambiguous.
        if has_mlx_marker {
            return false;
        }
        if has_hf_marker {
            return true;
        }
    }
    // No markers at all: treat as already-MLX (no-op) to avoid corrupting an
    // unexpected layout.
    has_hf_marker
}
```

The only caller is `RtDetrV2Model::load` (`src/vision/detection/rt_detr_v2/model.rs:110-115`), which runs `sanitize::sanitize(raw)` when the verdict is `true` and passes `raw` through untouched when it is `false`. The existing coverage, `needs_sanitize_detects_layout` (`src/vision/detection/rt_detr_v2/sanitize.rs:302-321`), builds one single-key MLX map and one single-key HF map, so it never exercises a map holding both families and cannot observe the ordering dependence.

## Proposed Solution

Maintainer's choice on the details, but the shape should be: scan every key before deciding rather than returning from inside the loop, then apply an explicit precedence so the verdict is independent of iteration order. MLX should win, since that is the safe direction: skipping the pipeline on a genuine HF checkpoint fails loudly at weight lookup, while running it on an MLX checkpoint corrupts conv weights silently.

Decide separately whether a map carrying both families should be rejected loudly instead of resolved silently. That input is a checkpoint the loader does not actually understand, and `needs_sanitize` returns a bare `bool` today, so surfacing it would mean changing the signature (for example to `Result<bool, String>`, matching the `Result<Self, String>` error style already used by `RtDetrV2Model::load`) and threading the error out through the caller. If the decision is to keep the silent resolution, say so in the code comment rather than leaving the current misleading "both decisions" wording.

Also replace the trailing `has_hf_marker` at line 72 with an explicit `false`, and correct the line 62 comment so it describes what the code does.

## Scope

**In scope:** `src/vision/detection/rt_detr_v2/sanitize.rs` (the `needs_sanitize` body, its doc comment, the two inline comments at lines 62 and 70-71), the inline `mod tests` in the same file (new regression test), and `src/vision/detection/rt_detr_v2/model.rs:110-115` only if the signature changes to a `Result`.

**Out of scope:** the rename rules in `rename_stripped`, the transpose in `maybe_transpose_conv`, the drop rules in `should_drop`, and any change to the marker set itself. Widening or anchoring the six HF markers is a separate design question and should not ride along here. The sibling `HashMap` ordering instances tracked in #1267 and in the `nodes_with_role` failover ordering issue (#1277 (`src/distributed/registry.rs:146`)) filed alongside this one as #1277 are also out of scope.

## Implementation Notes

- **Reuse**: keep the existing marker string set exactly as it is, and keep the existing "treat as already-MLX to avoid corrupting an unexpected layout" reasoning for the no-marker case. This is a control-flow fix, not a re-derivation of the layout sniffing rules.
- **Constraints**: the verdict must not depend on `weights.keys()` order for any input. Do not switch `WeightMap` to an ordered map to work around this: it is a shared type used across the whole loader (`src/lib/mlxcel-core/src/weights.rs:29`) and changing it here would be a cross-cutting change well outside this issue.
- **Edge cases**: empty map (no keys) keeps the current verdict, `false`; map with only MLX markers is `false`; map with only HF markers is `true`; map with neither is `false`; map with both is the case this issue exists for, and must be either a stable `false` or a loud error, never a coin flip.
- **Error handling**: if the both-families case is made loud, the message must name at least one offending key from each family so the operator can see which tensors caused the ambiguity, and `RtDetrV2Model::load` must propagate it rather than falling back to a default verdict. If it stays silent, no behavior change is needed at the call site.

## Acceptance Criteria

- [x] `needs_sanitize` inspects all keys before returning, so its verdict is a pure function of the key set and not of `HashMap` iteration order.
- [x] A weight map containing both an MLX-prefixed key (for example `vision.backbone.embedder.embedder.0.conv.weight`) and a `.convolution.` shaped key yields one stable verdict, or a deterministic error, across repeated construction of the map.
- [x] The line 62 comment no longer claims a "both decisions" guard that the code does not implement.
- [x] The trailing return at line 72 is an explicit `false` rather than a variable that is provably `false` at that point.
- [x] A regression test in the inline `mod tests` of `sanitize.rs` builds a mixed-marker map and asserts the same verdict over many repetitions (rebuilding the map each iteration, since `RandomState` varies per map instance, so a single call cannot distinguish a fixed answer from a lucky one).
- [x] The existing `needs_sanitize_detects_layout` test still passes unchanged.
- [ ] Integrated into the real load path (signature kept as `bool`, so `RtDetrV2Model::load` is untouched and compiles clean; not exercised against actual checkpoint files, see comment below): `RtDetrV2Model::load` still compiles and behaves correctly against both a pre-converted MLX checkpoint and a raw HF checkpoint (signature change threaded through if the loud-error option is taken).

## Verification

```bash
# Targeted unit tests (macOS)
cargo test --features metal,accelerate rt_detr_v2::sanitize

# Targeted unit tests (Linux / CUDA)
cargo test --features cuda rt_detr_v2::sanitize

# Full local gate
make verify            # macOS: verify-versions, verify-kernel-dtype-keys, verify-fmt, verify-clippy, verify-test
make verify-test-cuda  # Linux/NVIDIA merge gate
```

A pass means the targeted filter runs the existing `needs_sanitize_detects_layout` plus the new mixed-marker regression test with zero failures, and the repetition loop in the new test never observes a differing verdict. Repeating the targeted command several times in a row is a useful sanity check, since the defect this fixes is exactly the kind that a single green run cannot rule out.

## Technical Considerations

Found during the review of PR #1268 while sweeping for the root-cause class fixed by #1265 and PR #1266 (`HashMap` iteration order leaking into an ordered or order-sensitive decision). Sibling instances: #1267 in `src/lang_bias.rs`, and the `nodes_with_role` failover ordering issue in #1277 (`src/distributed/registry.rs:146`) filed alongside this one as #1277.

inureyes:
## Summary

`needs_sanitize` returned from inside its `weights.keys()` walk as soon as either marker family was hit, so for any weight map carrying both families the verdict belonged to whichever key the randomized `HashMap` walk reached first. `WeightMap` is a `std::collections::HashMap` (`src/lib/mlxcel-core/src/weights.rs:29`) and `RandomState` is seeded per map instance, not per process, so the same checkpoint could sanitize on one run and not the next. This scans the full key set before deciding and then applies an explicit precedence, making the verdict a pure function of the key set.

The wrong direction is the expensive one. The module doc in the same file notes that MLX-layout checkpoints must skip the pipeline because re-running it double-transposes conv weights, so a spurious `true` corrupts the weights while the model still loads and still emits detections, just wrong ones.

## Design decision: silent-but-deterministic, plus a warning

The issue asked whether a map carrying both families should be rejected loudly instead. It resolves toward MLX silently, with a `tracing::warn!`, for three reasons.

MLX-wins is the only recoverable direction, so it does not actually need an error to stay safe. If the map really was raw HF, skipping the pipeline fails loudly a moment later: `RtDetrV2Model::from_weights` looks up `vision.backbone.*` names the raw map does not have and returns `Err`. Running the pipeline over already-MLX weights has no such backstop, since a double-transposed conv is a shape-valid tensor that nothing downstream can flag.

Two of the six HF markers, `.convolution.` and `.normalization.`, are unanchored substring tests, and the issue puts anchoring them out of scope. Making the mixed case fatal would build a hard load failure on top of a detector that is known to be over-broad, so any future false positive would refuse a checkpoint that loads correctly today. An imprecise heuristic should fail soft.

`needs_sanitize` is `pub` and returns a bare `bool`. Moving it to `Result<bool, String>` and threading the error through `RtDetrV2Model::load` is real API churn for a case the issue itself describes as latent rather than observed.

The `tracing::warn!` recovers most of what a hard error would have bought: it names one offending key from each family, which is what the issue asked a loud variant to report, and it needs no call-site change. It picks the lexicographically first match of each family rather than the first the walk meets, so the message is reproducible for the same checkpoint instead of reintroducing the ordering nondeterminism in the diagnostic. If a mixed-layout checkpoint ever shows up in practice, that warning is the evidence needed to justify promoting it to an error.

## What changed

`src/vision/detection/rt_detr_v2/sanitize.rs`

- `needs_sanitize` now accumulates `has_mlx_marker` / `has_hf_marker` across every key and decides afterwards. MLX takes precedence over HF; HF wins when only HF markers are present; a map with no markers is unchanged at `false`.
- The marker tests moved into `is_mlx_marker` / `is_hf_marker` so the scan loop is visibly accumulate-only. The six HF marker strings and the two MLX prefixes are byte-for-byte unchanged.
- New `warn_mixed_layout` emits the both-families diagnostic.
- The `// Early out once both decisions are unambiguous.` comment is gone; it described a guard the code never implemented.
- The trailing `has_hf_marker` return is now an explicit `false`. At that point it could only ever have been `false`, but it read as though it could return `true`.
- New test `needs_sanitize_mixed_markers_is_order_independent`: 64 iterations, each building a fresh 7-key map holding both families, asserting a stable `false`. The map is rebuilt per iteration because `RandomState` varies per instance, so a single call cannot distinguish a fixed verdict from a lucky one.
- New test `needs_sanitize_marker_free_maps_are_no_ops`: pins the empty-map and unrecognized-key edge cases from the issue.

No change to `src/vision/detection/rt_detr_v2/model.rs`: the signature stays `bool`, so the call site at lines 110-115 is untouched. The rename rules, the transpose, and the drop rules are untouched.

## Regression test evidence

The new test was written and run against the unfixed function first. It failed as expected:

```
running 14 tests
...
test vision::detection::rt_detr_v2::sanitize::tests::needs_sanitize_detects_layout ... ok
test vision::detection::rt_detr_v2::sanitize::tests::needs_sanitize_mixed_markers_is_order_independent ... FAILED

failures:

---- vision::detection::rt_detr_v2::sanitize::tests::needs_sanitize_mixed_markers_is_order_independent stdout ----

thread '...needs_sanitize_mixed_markers_is_order_independent' panicked at src/vision/detection/rt_detr_v2/sanitize.rs:363:9:
assertion `left == right` failed: needs_sanitize said `true` for 27 of 64 mixed-marker maps freshly built from the same key set; the verdict must be a stable `false` and must not depend on HashMap iteration order. Observed: [false, false, false, false, false, true, false, false, true, true, false, false, false, true, false, false, false, false, false, true, true, false, false, true, true, false, false, true, false, false, true, false, false, true, true, false, true, false, true, true, false, true, true, false, false, true, true, true, false, true, true, false, false, false, true, true, false, false, true, false, true, false, true, false]
  left: 27
 right: 0

test result: FAILED. 13 passed; 1 failed; 0 ignored; 0 measured; 5859 filtered out; finished in 0.37s
```

27 of 64 is close to the 3/7 predicted by the key mix, which is what a first-key-wins walk over that map should produce. Note that `needs_sanitize_detects_layout` passed in the same run: its two single-key maps cannot express the mixed case, so its green status was never evidence about this defect.

After the fix the same filter is green, and repeating it five times in a row (320 independently built maps in total) never observed a differing verdict.

## Test plan

- [x] `cargo test --profile test-fast --features cuda --lib vision::detection::rt_detr_v2::sanitize` (exit 0, 14 passed)
- [x] Same command repeated 5 more times back to back (exit 0 each, 14 passed each)
- [x] `cargo test --profile test-fast --features cuda --lib vision::detection::rt_detr_v2` (exit 0, 32 passed)
- [x] `cargo clippy --lib --tests --features cuda -- -D warnings -A clippy::err_expect` (exit 0, no findings in the touched file)

One note on that last line. Plain `cargo clippy --lib --tests --features cuda -- -D warnings` exits 101 on `main` at `ca114220`, before this branch, on `clippy::err_expect` at `src/multimodal/host_preprocessor_tests.rs:416`. That file is byte-identical to `origin/main` here and this branch does not touch it. The repo's own lint gate is `make verify-clippy`, which runs `--features metal,accelerate`, and the Makefile notes there is deliberately no `verify-clippy-cuda` yet, so the CUDA lint path is currently ungated. Allowing only that one pre-existing lint is what makes the run above meaningful about this change; it is not a suppression added to the code.

The full `make verify-test-cuda` merge gate has not been run here and is left to the maintainer.

Closes #1276


inureyes:
Gate run notes, since the branch was cut before two commits landed on `main`.

The branch base (`ca114220`) is behind `main` (`bf69b83e`), which has since taken #1278 (documentation and benchmark scripts) and #1280 (code plus tests). To gate the tree that will actually merge rather than the older base, the commit was rebased onto `bf69b83e` locally and the gate was run there. The rebase was clean, and after it the branch's diff against `main` is a single file, `src/vision/detection/rt_detr_v2/sanitize.rs` (+134 / -18), which no commit on `main` touches. The rebase was not pushed, so this PR still shows its original commit; a squash merge onto current `main` produces the same tree that was gated.

`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64), rebased onto `bf69b83e`: **8229 passed, 0 failed, 311 ignored**, 101 suites, exit 0. No link or compile errors in the log.

No cross-run delta is quoted on purpose. The previous full gate this session ran at `ca114220` (8194 passed) and `main` has gained tests since, so the two totals are not comparable. What is checkable is this branch's own contribution: the diff adds exactly 2 `#[test]` functions and removes none.

`cargo fmt --all -- --check` and `cargo check --lib --tests --features cuda` are clean on the same tree.

Issue #1283, opened 2026-08-22
fix(lint): verify-clippy is red on main from an err_expect, and no PR-time job runs clippy


inureyes:
## Problem / Background

`main` is red under clippy. At `930fd83e`, `cargo clippy -p mlxcel --lib --tests -- -D warnings` exits 101 with a single `clippy::err_expect`:

```
error: called `.err().expect()` on a `Result` value
   --> src/multimodal/host_preprocessor_tests.rs:416:10
    |
416 |           .err()
    |  __________^
417 | |         .expect("a graph that cannot admit any image must fail at startup");
    | |_______________^ help: try: `expect_err`
    |
    = note: `-D clippy::err-expect` implied by `-D warnings`

error: could not compile `mlxcel` (lib test) due to 1 previous error
```

The lint is not feature-specific, which is the correction this issue exists to record. It was first observed under `--features cuda` on Linux and initially read as a CUDA-only escape, but it reproduces on every feature set that compiles the crate, including the CI feature set on Apple Silicon. Three measurements, all yielding the identical single error at `src/multimodal/host_preprocessor_tests.rs:416:10` and `could not compile mlxcel (lib test) due to 1 previous error`:

| Configuration | Host | Command | Result |
|---|---|---|---|
| `--features cuda` | GB10, Linux aarch64, sm_121 | `cargo clippy --lib --tests --features cuda -- -D warnings` | exit 101 |
| default features (`surgery`, no accelerator) | same host | `cargo clippy -p mlxcel --lib --tests -- -D warnings` | exit 101 |
| `--features metal,accelerate` | self-hosted Apple Silicon, nightly CI | `make verify-clippy` | failed 2026-08-21T18:13:56Z, [run 32511932812](https://github.com/lablup/mlxcel/actions/runs/32511932812) at `dd21ada4` |

The macOS arm of that table is already reported: nightly-verify filed it into #1256 in the [2026-08-21T19:03:51Z comment](https://github.com/lablup/mlxcel/issues/1256#issuecomment-5374086973), whose gate table reads `make verify-clippy` = `failure`, `make verify-test` = `success`. #1256 is a rotating tracker that gets reused for every nightly red, so it names the failing target but not the defect, and its body still describes a different, older red (`verify-test` at `9e2c6675`). This issue names the defect and the escape path so the tracker can close.

## Current Behavior

[...]

PR #1284, opened 2026-08-22
fix(distributed): give registry node accessors a defined order

### LINKED ISSUE #1277: fix(distributed): registry accessors return HashMap order, so node routing and failover re-routing are not reproducible
## Problem / Background

`RegistryInner.nodes` is a `HashMap<String, RegisteredNode>` (`src/distributed/registry.rs:71`). Two of the registry's four list accessors return that map's values without ordering them, and both feed consumers that index the result by position or that tie-break on input order. Rust's `RandomState` randomizes `HashMap` iteration order per map instance, not merely per process, so the order those accessors return is arbitrary and varies between runs and between registries built from the same config.

The consequence is that node selection in the distributed serving path is not reproducible. Request counts stay balanced, so the existing tests pass, but which node a given request lands on is not stable across runs of an identical cluster state.

## Current Behavior

`Registry::nodes_with_role` (`src/distributed/registry.rs:146-156`) collects `inner.nodes.values().filter(|n| n.config.role == role).cloned().collect()` and returns the `Vec` unordered. `Registry::all_nodes` (`src/distributed/registry.rs:135-144`) does the same with no filter.

### Instance 1: failover re-routing (`nodes_with_role`)

`src/distributed/disaggregated/request_router.rs:666-689` builds its failover candidate lists directly from that accessor. `prefill_candidates` is `nodes_with_role(NodeRole::Prefill).into_iter().chain(nodes_with_role(NodeRole::Hybrid))`, filtered to `status == NodeStatus::Online`, `config.id != failed_node_id`, and `!backpressure.is_critical(..)`, then mapped to ids. `decode_candidates` is the same shape for `NodeRole::Decode`.

The re-routing loop indexes those vectors by position with local round-robin counters: `prefill_candidates[prefill_rr % prefill_candidates.len()]` at lines 702 and 724, and `decode_candidates[decode_rr % decode_candidates.len()]` at line 747, across the `Prefilling`, `TransferringCache`, and `Decoding` arms (lines 696-755). So when a node fails, which surviving node each in-flight request is re-routed to is decided by `HashMap` iteration order.

### Instance 2: primary routing (`all_nodes`)

The open question of whether `all_nodes` has an order-sensitive consumer resolves to yes, and its blast radius is larger than instance 1 because it sits on the normal request path rather than the failure path.

`request_router.rs:894-921` (`get_load_infos_for_roles`) starts from `self.registry.all_nodes()` and maps it to `Vec<NodeLoadInfo>` preserving that order. `get_prefill_load_infos` (line 867) and `get_decode_load_infos` (line 872) wrap it, and `route_to_prefill` (line 420) and `route_to_decode` (line 505) pass the result straight into `select_prefill_node` (line 457) and `select_decode_node` (line 534). Every strategy arm in those two functions is order-sensitive:

- `DisaggRoutingStrategy::RoundRobin` indexes positionally: `online[self.prefill_rr_counter.fetch_add(1, Ordering::Relaxed) % online.len()]` (lines 473 and 546).
- `LeastLoaded` uses `online.iter().min_by_key(|n| n.active_requests)` (lines 477 and 550). `min_by_key` returns the first minimum, so ties are broken by input order.
- `MemoryAware` uses `online.iter().max_by_key(|n| n.free_memory())` (lines 481 and 554). `max_by_key` returns the last maximum, so ties are again broken by input order.
- `PromptLengthAware` (lines 484-494 and 557-563) delegates to the same two calls.

A cold or idle cluster is exactly the tie case: every node reports `active_requests == 0` and equal free memory, so `LeastLoaded` and `MemoryAware` pick whichever node `HashMap` iteration happened to place first or last.

The same input reaches the scheduler. `src/distributed/scheduler.rs:413-444` (`build_candidates`) is built from `self.registry.all_nodes()` and is handed to `strategy.select_node` at line 218 and to `find_alternative` at line 226. In `src/distributed/routing.rs`, `RoleBasedRouter` picks `matching.first()` (line 192), `RoundRobinRouter` indexes `% online.len()` (line 334), `PipelineStageRouter` uses `min_by_key` (line 383), and `LoadBalancedRouter` sorts with `sort_by` (line 249), which is stable and therefore preserves input order on ties.

Order-insensitive consumers of `all_nodes` that need no change: `src/distributed/discovery.rs:50`, `src/distributed/heartbeat.rs:131`, and `src/server/router_front.rs:409` all iterate for side effects where sequence does not affect the outcome. `src/server/router_front.rs:561` (`router_stats`) is cosmetic but user-visible: the `nodes` array in the `/stats` JSON comes out in a different order on every process, which makes diffing two stats snapshots noisy.

## The sibling accessors already do the right thing

`nodes_at_stage` (`src/distributed/registry.rs:183-193`) ends with `nodes.sort_by_key(|n| n.config.rank.unwrap_or(u32::MAX))`, and `nodes_at_rank` (`src/distributed/registry.rs:201-211`) ends with `nodes.sort_by_key(|n| n.config.stage.unwrap_or(u32::MAX))`. Both were written to return a deterministic order because their callers depend on position. `nodes_with_role` and `all_nodes` have callers that depend on position just as much and were not given the same treatment.

## Impact

Per-node request counts stay balanced, because round-robin distributes evenly over whatever order it is given. That is why the existing tests do not catch this: `nodes_with_role_filter` (`src/distributed/registry_tests.rs:74-85`) asserts `decode_nodes[0].config.id == "peer-1"` only in a single-element list, which cannot distinguish a fixed order from an arbitrary one.

What is not stable is the assignment itself. Two consequences follow. A failover incident, or a routing decision under a tie, cannot be reproduced from the same cluster state while debugging. And on a heterogeneous cluster, where nodes differ in capacity, resident model, or cache warmth, the quality of the selection varies run to run for reasons no operator can see or control.

## Proposed Solution

Sort inside the accessors by a stable key so every consumer inherits a deterministic order, matching what the two sibling accessors already do. `n.config.id` is the natural key: it is a `String`, it is unique because it is the `HashMap` key, and it is stable across restarts because it comes from the cluster config.

Apply the same `sort_by` to both `nodes_with_role` and `all_nodes`. Fixing only `nodes_with_role` leaves the primary routing path (instance 2) nondeterministic, which is the larger of the two problems.

Sorting at the accessor rather than at each call site is deliberate: there are at least seven consumers across three modules, the set grows, and a per-call-site fix means every future caller has to rediscover the hazard. Rejected alternative: switching `RegistryInner.nodes` to a `BTreeMap`. That would also give a deterministic order and would be a smaller diff, but it changes the type of a field used on the hot lookup path (`get_node`, `local_pp_tp_coords`, `set_node_status`) from O(1) to O(log n) for a determinism property only the list accessors need. Sorting the returned `Vec` confines the cost to the accessors that actually require the guarantee.

## Scope

**In scope:**
- `src/distributed/registry.rs`: sort the `Vec` returned by `nodes_with_role` and by `all_nodes` by `config.id`.
- `src/distributed/registry_tests.rs`: add regression tests for both accessors.

**Out of scope:**
- Changing any selection strategy's semantics. The round-robin, least-loaded, and memory-aware policies keep their current behavior; this change only makes the sequence they operate on deterministic.
- Making tie-breaking explicit at the selection sites (for example, breaking `LeastLoaded` ties on node id inside `select_prefill_node`). Determinism at the accessor makes the existing implicit tie-break reproducible, which is what this issue asks for. A deliberate tie-break policy is a separate design question.
- `src/lang_bias.rs`, already covered by #1267.

## Implementation Notes

- **Reuse**: follow the existing shape in `nodes_at_stage` and `nodes_at_rank`, which bind `let inner = self.inner.read().expect("registry lock poisoned");`, collect into `let mut nodes: Vec<RegisteredNode>`, sort, and return. Do not introduce a different idiom for the two accessors being fixed.
- **Key type**: `config.id` is a `String`, so `sort_by_key(|n| n.config.id.clone())` allocates per comparison. Prefer `nodes.sort_by(|a, b| a.config.id.cmp(&b.config.id));`.
- **Constraints**: no public API change. Both functions keep their signature and return type; only the order of the returned elements becomes defined. The registry read lock is already held for the collect, and the sort happens on the local `Vec`, so the lock is not held any longer than the current code holds it. Node counts are cluster-sized (single or low double digits), so the sort cost is irrelevant next to the existing clone of every `RegisteredNode`.
- **Edge cases**: empty result (no node with that role) must return an empty `Vec` rather than panicking; a single-element result is trivially sorted; duplicate ids are impossible because the id is the `HashMap` key.
- **Error handling**: unchanged. The existing `expect("registry lock poisoned")` on the read lock stays as it is, consistent with every other accessor in the file.

## Acceptance Criteria

- [ ] `nodes_with_role` returns nodes ordered by `config.id` ascending, for every role.
- [ ] `all_nodes` returns nodes ordered by `config.id` ascending.
- [ ] A regression test registers at least four nodes whose ids do not match their insertion order, builds several independent `NodeRegistry` instances from the same config, and asserts every instance returns the identical sequence from `all_nodes` and from `nodes_with_role`. Building fresh registries is required: `RandomState` varies per map instance, so repeated calls on one registry cannot distinguish a fixed order from a lucky one.
- [ ] A test asserts the expected order explicitly (the sorted id list), not merely that two calls agree, so the test fails if the sort key is later changed or dropped.
- [ ] The existing `nodes_with_role_filter` test still passes unchanged.
- [ ] Deterministic order reaches the real consumers: `select_prefill_node` and `select_decode_node` under `RoundRobin`, and their `LeastLoaded` / `MemoryAware` tie cases, select the same node across repeated runs from an identical cluster state. This is the integration requirement; a sorted accessor with no verified effect on the routing path does not close this issue.
- [ ] `make verify-fmt` and `make verify-clippy` pass with no new warnings.

## Verification

```
cargo test --lib distributed::registry -- --test-threads=1
cargo test --lib distributed::disaggregated::request_router -- --test-threads=1
cargo test --lib distributed::scheduler distributed::routing -- --test-threads=1
make verify-fmt
make verify-clippy
```

A pass means the new registry ordering tests succeed and no previously passing distributed test regresses. On this Linux/NVIDIA box the merge gate is `make verify-test-cuda` and it must be run in full, not skipped. **Correction to an earlier draft of this section:** the known abort on this hardware happens when the library suite is run WITHOUT `--features cuda`; `make verify-test-cuda` passes `--features cuda` and is green here. Measured on GB10 (sm_121) on 2026-08-21 across three separate runs while landing #1266 and #1268: 8171 passed / 0 failed / 310 ignored at `a940c737`, and 8188 passed / 0 failed / 310 ignored at both `9ff0a3ad` and `0f624651`. So do not treat a red full gate here as pre-existing without comparing against `main` first.

Manual check, since no automated test covers the multi-node deployment: on a disaggregated cluster with three or more prefill nodes and `prefill_strategy = "round_robin"`, restart the router several times and issue the same sequence of requests from a cold cluster. Before the fix, the node that serves request N differs between restarts. After the fix, it is the same on every restart.

## Technical Considerations

Found while reviewing PR #1268, sweeping for the root-cause class fixed by #1265 and PR #1266, where `HashMap` iteration order leaked into ordered state. #1267 covers the production instance in `src/lang_bias.rs`. This is the third instance of the same class, in a third subsystem, and the first one where the nondeterminism reaches node selection in the serving path.

Worth considering as a follow-up: a lint or a review checklist entry for `HashMap::values()` / `HashMap::iter()` results that are collected into a `Vec` and then indexed, sorted-into, or fed to `min_by_key` / `max_by_key`. Three independent instances in one sweep suggests the pattern is not being caught at review time.

inureyes:
## Summary

`RegistryInner.nodes` is a `HashMap`, and two of its list accessors handed back `.values()` unsorted while their consumers select by position or tie-break on input order. `RandomState` seeds every map instance separately rather than once per process, so identical cluster state produced different node selections between runs. Both accessors now sort by `config.id`, which is what the sibling accessors `nodes_at_stage` and `nodes_at_rank` already did.

## What changed

- `src/distributed/registry.rs`: `all_nodes` and `nodes_with_role` sort the returned `Vec` by `config.id`. Signatures and return types are unchanged; only the order becomes defined.
- `src/distributed/registry.rs`: `peer_addresses` and `topology_summary` get the same treatment. See the scope note below.
- `src/distributed/disaggregated/request_router.rs`: `handle_node_failure` sorts the affected request ids before the round-robin re-routing loop consumes them.
- `src/distributed/registry_tests.rs`: five ordering tests over 64 independently constructed registries each.
- `src/distributed/disaggregated/request_router_tests.rs`: four selection-determinism tests over 32 independently constructed routers each, driving the real routing entry points rather than the accessor.

## Design choices

**Sort in the accessor, not at the call sites.** There are seven consumers across three modules and the set grows; a per-call-site fix means every future caller has to rediscover the hazard. The sort runs on a local `Vec` after the read lock has already been taken for the clone, so it holds no lock longer than before, and node counts are cluster-sized next to the existing clone of every `RegisteredNode`.

**`sort_by`, not `sort_by_key`.** The issue recommends this and the recommendation holds. `sort_by_key` re-evaluates the key on every comparison, so a `String` key allocates per comparison rather than per element. Measured on a 12-element scrambled `Vec<String>`: `sort_by_key(|s| s.clone())` performed 78 key evaluations, each a clone, while `sort_by(|a, b| a.cmp(b))` performed 39 comparisons and allocated nothing. `sort_by_key` also cannot borrow the key out of the element, so a clone is the only way to spell it.

**`BTreeMap` stays rejected.** Per the issue: it would penalize the O(1) `get_node` / `set_node_status` / `local_pp_tp_coords` lookups for a property only the list accessors need. Nothing found during implementation changes that.

## Scope: which accessors actually carry the exposure

The issue describes `nodes` as having four list accessors and puts two of them in scope. Swept the file and the module; the corrected picture:

| Accessor | Before | Consumer sensitivity | Action |
| --- | --- | --- | --- |
| `all_nodes` | unordered | positional and tie-break, primary request path | sorted |
| `nodes_with_role` | unordered | positional, failover path | sorted |
| `nodes_at_stage` / `nodes_at_rank` | already sorted | positional | unchanged |
| `peer_addresses` | unordered, not named in the issue | only consumer (`heartbeat.rs:214`) broadcasts to every peer, so insensitive today | sorted |
| `topology_summary` | unordered, not named in the issue | operator-facing text printed at `discovery.rs:96` and `cluster_init.rs:253` | sorted |
| `find_pp_tp_node` | `.values().find(..)` | deterministic only because `ClusterConfig::validate` rejects duplicate `(stage, rank)` pairs (`config.rs:465`) | unchanged |

`peer_addresses` and `topology_summary` are beyond the issue's stated in-scope list. Including them is deliberate: they are the same defect in the same file, the fix is two lines each, and leaving them means the next caller of `peer_addresses` inherits a hazard the sibling accessors no longer have. `topology_summary` is also the one place where an operator can see the disorder directly, since re-running it on an unchanged cluster reshuffles the node list.

The order-insensitive consumers the issue cleared are confirmed: `discovery.rs:50` (probes run concurrently), `heartbeat.rs:131` (per-node registration), `router_front.rs:409` (per-node health probe). `router_front.rs:561` (`/stats`) is confirmed cosmetic and now stable as a side effect. One consumer the issue does not mention, `router_front.rs:1218`, takes only `.len()` of `nodes_with_role` and is insensitive.

## A gap in the issue's own instance 1

Sorting `nodes_with_role` alone does not make failover re-routing reproducible. `handle_node_failure` collects the affected request ids from `self.requests`, which is also a `HashMap`, and the loop assigns candidates round-robin in exactly that sequence. With sorted candidates but unsorted request ids, the set of targets is stable while the request-to-node pairing still moves between runs, which is the part an operator is trying to reproduce. `affected.sort()` closes it. The new `failover_reroute_assignment_is_stable_across_routers` test asserts the full pairing, so it covers both halves.

## Acceptance criteria

Verified by test:

- [x] `nodes_with_role` returns nodes ordered by `config.id` ascending, for every role. Covered for Prefill, Decode, and Hybrid, plus the empty case for a role with no members.
- [x] `all_nodes` returns nodes ordered by `config.id` ascending.
- [x] A regression test registers at least four nodes whose ids do not match their insertion order and builds several independent registries. Five nodes, 64 registries per test.
- [x] A test asserts the expected order explicitly, so it fails if the sort key is changed or dropped.
- [x] The existing `nodes_with_role_filter` test still passes unchanged. It is untouched, and it structurally cannot observe this bug: it indexes a single-element list.
- [x] Deterministic order reaches `select_prefill_node` and `select_decode_node`, at unit level, under `RoundRobin` and under the `LeastLoaded` / `MemoryAware` tie cases. The four router tests call `route_to_prefill` / `route_to_decode` / `handle_node_failure` on freshly built routers, so they exercise the real path from registry through `get_load_infos_for_roles` into the strategy arm.
- [x] `make verify-fmt` passes. It is exactly `cargo fmt --all -- --check`, run here at exit 0.

Verified by construction, not by observation:

- The claim that a defined accessor order is sufficient for reproducible selection rests on every strategy arm being a pure function of the candidate sequence plus the atomic round-robin counters. That is true by inspection of `select_prefill_node` / `select_decode_node`, and the unit tests confirm it for the tie cases, but a live cluster adds real metrics that break most ties before the order matters.

Not verified by observation:

- The issue's manual check (three or more prefill nodes, restart the router repeatedly, confirm request N lands on the same node each time) cannot be run here. This box is a single node, so there is no multi-node disaggregated deployment to restart. The unit-level equivalent is covered; the end-to-end claim is not.
- `make verify-clippy` as written runs `--features metal,accelerate` and cannot execute on this Linux/CUDA box. The CUDA lint path was run instead; see the test plan.
- The full `make verify-test-cuda` merge gate is not run in this PR's verification and is left to the reviewer.

## Test plan

Commands and their real exit codes, each captured directly rather than read off a pipeline:

- [x] `cargo test --profile test-fast --features cuda --lib --no-run` then the test binary with narrow filters (one filter per invocation), all at exit 0: `distributed::registry` 18 passed, `distributed::disaggregated::request_router` 29 passed, `distributed::scheduler` 16 passed, `distributed::routing` 19 passed, `distributed::heartbeat` 9 passed, `distributed::discovery` 3 passed, `distributed::cluster_init` 14 passed, `server::router_front` 20 passed. 0 failed throughout.
- [x] `cargo fmt --all -- --check`, exit 0.
- [x] The nine new tests were run against the unfixed accessors first and all nine failed, most on iteration 0 or 1. For example `all_nodes` returned `["yankee-prefill", "alpha-decode", "mike-prefill", "bravo-decode", "zulu-hybrid"]` on the first registry, and `round_robin_selection_sequence_is_stable_across_routers` produced `["prefill-0", "prefill-1", "prefill-0", "prefill-1"]` on one router and `["prefill-1", "prefill-0", "prefill-1", "prefill-0"]` on the next.

- [x] `cargo clippy --lib --tests --features cuda -- -D warnings -A clippy::err_expect`, exit 0. Nothing this PR touches produces a lint.
- [ ] `cargo clippy --lib --tests --features cuda -- -D warnings`, exit 101, on one pre-existing error: `clippy::err_expect` at `src/multimodal/host_preprocessor_tests.rs:416`, a file this PR does not touch. The repo's `verify-clippy` lints `--features metal,accelerate`, so the CUDA lint path is ungated and this error predates the branch. It is the only error in the run and none of the four changed files appear in the output.

## Follow-ups found while sweeping, not fixed here

The same defect class appears three more times in `src/distributed/`, all outside this issue's subsystem and all with the identical shape of a `HashMap` iteration collected into a `Vec`, handed to a stable `sort_by_key`, then consumed as a prefix:

- `src/distributed/tensor_parallel/cache_manager.rs:711`: `check_pressure` stops walking the candidate list once the memory target is met, so among allocations tied on `current_offset` or `last_accessed` the hash order decides which sequences `apply_eviction` destroys.
- `src/distributed/pipeline/cache_manager.rs:697`: same stable-sort-over-hash-order defect, published as `PreemptionSignal.sequence_ids`, a field documented as eviction priority order. No in-repo consumer takes a prefix yet.
- `src/distributed/request_tracker.rs:333`: sorts terminal requests by `created_at` and takes a prefix, so same-`Instant` ties are broken by hash order.

These change eviction behavior rather than routing, so they deserve their own issue and their own tests rather than a silent ride along with this one.

Closes #1277

### DIFF
--- a/TECHNICAL_REPORTS/1284-registry-node-order-20260822.en.md
+++ b/TECHNICAL_REPORTS/1284-registry-node-order-20260822.en.md
@@ -1,0 +1,79 @@
+# Technical Report: PR #1284 - A defined order for the distributed registry's node accessors
+
+## Executive Summary
+
+`RegistryInner.nodes` is a `HashMap`, and four of the registry's list accessors returned its `.values()` unordered while their consumers depended on position. The largest consequence was on the primary request path: `all_nodes` feeds `select_prefill_node` and `select_decode_node`, where `RoundRobin` indexes positionally, `LeastLoaded` uses `min_by_key` (first minimum wins) and `MemoryAware` uses `max_by_key` (last maximum wins). A cold or idle cluster is exactly the tie case, and the shipped defaults are `LeastLoaded` for prefill and `MemoryAware` for decode, so on an idle cluster the "least loaded" node was whichever one `HashMap` iteration placed first.
+
+The fix sorts inside the accessors by `config.id`. This is the fourth and last instance of one root-cause class found in a single review sweep, after #1265 (test fixtures), #1267 (`lang_bias.rs`) and #1276 (RT-DETRv2 layout sniffing).
+
+## 1. Problem Statement
+
+Two sibling accessors in the same file, `nodes_at_stage` and `nodes_at_rank`, already ended with a `sort_by_key` because their callers depend on position. `all_nodes` and `nodes_with_role` had callers that depend on position just as much and had never been given the same treatment.
+
+Request counts stayed balanced, because round-robin distributes evenly over whatever order it is handed. That is why the existing tests passed: `nodes_with_role_filter` asserts an index into a single-element list, which cannot distinguish a fixed order from an arbitrary one. What was not stable was the assignment, so a failover incident could not be reproduced from an identical cluster state, and on a heterogeneous cluster the quality of a selection varied for reasons no operator could see.
+
+## 2. Technical Decisions
+
+### 2.1 Sort in the accessors, with `sort_by` rather than `sort_by_key`
+
+Sorting at the source means every consumer inherits determinism rather than each one remembering to sort. `sort_by(|a, b| a.config.id.cmp(&b.config.id))` was chosen over `sort_by_key(|n| n.config.id.clone())` because the key is a `String` that cannot be borrowed out of the element, so `sort_by_key` clones per key evaluation. Measured on a 12-element scrambled vector: 78 key evaluations, each a clone, against 39 comparisons and zero allocations.
+
+Switching `nodes` to a `BTreeMap` was considered and rejected in the issue, because it would penalize the O(1) `get_node` path for a property only the list accessors need. Nothing found during implementation argued against that.
+
+### 2.2 Sorting the candidates was not sufficient on its own
+
+This corrects the issue's own analysis. `handle_node_failure` collects the affected request ids from `self.requests`, which is also a `HashMap`, and the re-routing loop hands candidates out round-robin **in that sequence**. With sorted candidates but an unsorted request list, the request-to-node pairing still moves between runs, which is precisely the property an operator is trying to reproduce. `affected.sort()` was added alongside the accessor fix.
+
+### 2.3 Two more accessors were in scope than the issue named
+
+`peer_addresses` and `topology_summary` also return or render `.values()` unordered and were not named in the issue. `topology_summary` is operator-facing, printed during discovery and cluster init, so re-running it against an unchanged cluster reshuffled the node list. Both are now sorted, giving four ordered accessors in total.
+
+## 3. Change Summary
+
+| File | Change |
+| --- | --- |
+| `src/distributed/registry.rs` | `all_nodes`, `nodes_with_role`, `peer_addresses`, `topology_summary` sort by `config.id` |
+| `src/distributed/disaggregated/request_router.rs` | `handle_node_failure` sorts the affected request ids before the round-robin walk |
+| `src/distributed/registry_tests.rs`, `.../request_router_tests.rs` | 9 new tests |
+
+## 4. Review Findings
+
+Two claims in the issue body were wrong and are corrected here rather than carried forward.
+
+`routing.rs` was listed among the order-sensitive consumers. It is not: `RoundRobinRouter` already performs `online.sort_by(|a, b| a.node_id.cmp(&b.node_id))` before indexing, with a comment saying so. Someone had already made that path deterministic.
+
+`find_pp_tp_node` is safe for a reason worth recording rather than assuming: `ClusterConfig::validate` rejects duplicate `(stage, rank)` pairs, so its `.values().find()` has a unique match regardless of order.
+
+The cleared list in the issue was otherwise correct, and one consumer it omitted, `router_front.rs:1218`, takes only `.len()` and is insensitive.
+
+The regression tests were demonstrated failing against the unfixed accessors before being finalized, since this bug class readily produces tests that pass both before and after. All nine failed, most on iteration 0 or 1, for example:
+
+```
+---- all_nodes_ordered_by_id_across_registries ----
+  left: ["yankee-prefill", "alpha-decode", "mike-prefill", "bravo-decode", "zulu-hybrid"]
+ right: ["alpha-decode", "bravo-decode", "mike-prefill", "yankee-prefill", "zulu-hybrid"]
+---- round_robin_selection_sequence_is_stable_across_routers ----
+  left: ["prefill-0", "prefill-1", "prefill-0", "prefill-1"]
+ right: ["prefill-1", "prefill-0", "prefill-1", "prefill-0"]
+```
+
+The registry tests build 64 fresh registries each and the router tests 32 fresh routers each, rather than reusing one, because `RandomState` randomizes per `HashMap` instance and not merely per process. That property was measured while filing #1267: ten maps from the same five keys gave nine distinct orders inside one process.
+
+## 5. Validation
+
+Measured on GB10 (DGX Spark, CUDA sm_121, Linux aarch64).
+
+- `make verify-test-cuda`: recorded in the PR thread.
+- Narrow filters, all exit 0: `distributed::registry` 18 passed, `distributed::disaggregated::request_router` 29 passed, `distributed::scheduler` 16, `distributed::routing` 19, `distributed::heartbeat` 9, `distributed::discovery` 3, `distributed::cluster_init` 14, `server::router_front` 20.
+- `cargo fmt --all -- --check`: exit 0.
+- `cargo clippy --lib --tests --features cuda -- -D warnings -A clippy::err_expect`: exit 0. Without the allow it exits 101 on a pre-existing `clippy::err_expect` at `src/multimodal/host_preprocessor_tests.rs:416` that predates this branch and is tracked as #1283; none of this PR's files appear in that output.
+
+Not verified by observation: the issue's manual multi-node check, which needs three or more prefill nodes and repeated router restarts. This is a single-node machine. The accessor ordering and the determinism reaching `select_prefill_node` and `select_decode_node` under the tie cases are covered by test; the claim that a defined accessor order suffices for reproducible selection in production rests on each strategy arm being a pure function of the candidate sequence and the atomic counters, which is true by inspection and confirmed for ties by test, but live metrics break most ties before order matters.
+
+## 6. Related Work
+
+- #1277: the issue this closes.
+- #1265 and PR #1266, #1267 and PR #1269, #1276 and PR #1281: the sibling instances of the same class.
+- #1283: the pre-existing clippy failure encountered while validating this branch.
+
+A sweep of the remaining `.values()` and `.iter()` collection sites under `src/distributed/` found three further instances, deliberately left for their own issue because they change eviction behavior rather than routing: `tensor_parallel/cache_manager.rs:711`, where `check_pressure` breaks early once the memory target is met so hash order decides which tied sequences are evicted; `pipeline/cache_manager.rs:697`, which publishes `PreemptionSignal.sequence_ids` documented as eviction priority order; and `request_tracker.rs:333`, where terminal requests sorted by `created_at` take a prefix and same-`Instant` ties fall back to hash order.
--- a/TECHNICAL_REPORTS/1284-registry-node-order-20260822.ko.md
+++ b/TECHNICAL_REPORTS/1284-registry-node-order-20260822.ko.md
@@ -1,0 +1,79 @@
+# 기술 보고서: PR #1284 - 분산 레지스트리 노드 접근자에 정의된 순서 부여
+
+## 요약
+
+`RegistryInner.nodes`는 `HashMap`이고, 레지스트리의 목록 접근자 넷이 `.values()`를 정렬 없이 반환하는데 소비자들은 위치에 의존하고 있었다. 가장 큰 여파는 장애 경로가 아니라 **주 요청 경로**였다. `all_nodes`가 `select_prefill_node` / `select_decode_node`로 흘러가고, 거기서 `RoundRobin`은 위치로 인덱싱하며 `LeastLoaded`는 `min_by_key`(동점이면 첫 번째), `MemoryAware`는 `max_by_key`(동점이면 마지막)를 쓴다. 차갑거나 유휴 상태인 클러스터가 정확히 동점 상황이고, 출하 기본값이 prefill `LeastLoaded` / decode `MemoryAware`라, 유휴 클러스터에서 "least loaded" 노드는 `HashMap` 순회가 앞에 놓은 노드였다.
+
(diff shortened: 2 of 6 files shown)


inureyes:
`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at `a126cdba`: **8238 passed, 0 failed, 311 ignored**, 101 suites, exit 0.

That is +9 against `main` at `930fd83e` (8229 passed), and this branch adds exactly 9 `#[test]` functions and removes none, so the totals reconcile.

`cargo fmt --all -- --check` is clean, and `cargo clippy --lib --tests --features cuda -- -D warnings -A clippy::err_expect` exits 0 with none of this branch's files appearing. Without the allow it exits 101 on the pre-existing `clippy::err_expect` at `src/multimodal/host_preprocessor_tests.rs:416`, which predates this branch and is tracked as #1283.

PR #1288, opened 2026-08-22
fix(distributed): break eviction sort ties on the unique id

### LINKED ISSUE #1286: fix(distributed): eviction candidate sorts are stable, so HashMap order still picks which tied sequence is evicted
## Problem / Background

Three eviction paths under `src/distributed/` build a candidate list from a `HashMap`, sort it, and then consume a prefix. Sorting looks like it makes the result deterministic and does not, because `slice::sort_by_key` is a **stable** sort: entries that compare equal keep their input order, and the input order here is `HashMap` iteration order, which `RandomState` randomizes per map instance. So among candidates that tie on the sort key, which ones land in the consumed prefix varies run to run.

This is the subtle member of the class fixed by #1265, #1267, #1276 and #1277. The other four had no sort at all, so the defect was visible on inspection. These three DO sort, which is exactly why they read as safe.

## Current Behavior

**Instance 1, the most consequential: `src/distributed/tensor_parallel/cache_manager.rs`.** `allocations` is a `HashMap<SequenceId, ShardedCacheAllocation>` (`src/distributed/tensor_parallel/cache_manager.rs:430`). `select_eviction_candidates` (`src/distributed/tensor_parallel/cache_manager.rs:710-721`) does `let mut entries: Vec<_> = self.allocations.values().collect();` at line 711, then `entries.sort_by_key(|a| a.last_accessed)` at line 715 for `EvictionPolicy::LRU`, or `entries.sort_by_key(|a| a.current_offset)` at line 718 for `EvictionPolicy::LeastTokens`. Its doc comment (lines 707-709) says it returns "ALL sequence IDs in eviction priority order (first = evict first). Callers are responsible for selecting a subset sufficient to relieve pressure."

The caller, `check_pressure` (`src/distributed/tensor_parallel/cache_manager.rs:639-673`), does exactly that: it iterates the candidates and breaks out as soon as `projected_used <= target_bytes` (lines 656-657). That break makes it a prefix consumer, so ties in the sort key decide which sequences actually lose their cache. The resulting `EvictionSignal.sequence_ids` is then applied verbatim by `apply_eviction` (`src/distributed/tensor_parallel/cache_manager.rs:679-688`), which removes each listed allocation. Ties are not exotic under `LeastTokens`, where `current_offset` is a token count (`usize`, `src/distributed/tensor_parallel/cache_manager.rs:167`) that many concurrent sequences share.

**Instance 2: `src/distributed/pipeline/cache_manager.rs`.** The same shape at `select_eviction_candidates` (`src/distributed/pipeline/cache_manager.rs:696-709`): `self.allocations.values().collect()` at line 697, then `sort_by_key` on `last_accessed` (line 701, `PreemptionPolicy::LRU`), `current_offset` (line 704, `Shortest`) or `std::cmp::Reverse(current_offset)` (line 707, `Longest`). `allocations` is a `HashMap<SequenceId, StageCacheAllocation>` at `src/distributed/pipeline/cache_manager.rs:459`. The result is published as `PreemptionSignal.sequence_ids` (assigned at `src/distributed/pipeline/cache_manager.rs:686`, field declared at line 355), documented as eviction priority order.

Inside this repository the only consumer reads `.len()` (`src/distributed/pipeline/cache_manager.rs:381`, the `Display` impl), so there is no prefix consumer yet and the live impact today is that a documented priority ordering is not reproducible. It is worth fixing with the other two rather than waiting for a consumer to make it matter.

**Instance 3: `src/distributed/request_tracker.rs`.** `evict_if_needed` (`src/distributed/request_tracker.rs:328-347`) collects terminal requests with `inner.requests.iter().filter(..).map(..).collect()` (lines 334-339) from a `HashMap<String, RequestLifecycle>` (`src/distributed/request_tracker.rs:233`), sorts with `completed.sort_by_key(|(_, t)| *t)` on `created_at` (line 341), then consumes a prefix explicitly: `for (key, _) in completed.into_iter().take(to_remove)` (line 345). Two requests that finish within the same `Instant` tick tie, and hash order picks which is dropped.

## Impact

No corruption and no crash: the eviction still relieves the pressure it set out to relieve, and the counts are right. What is lost is reproducibility of WHICH sequence was sacrificed, which is exactly the thing an operator needs when a request is preempted and they are trying to work out why. Instance 1 is the one with live behavioral consequences today.

## Proposed Solution

Maintainer's choice, but the shape is the same in all three: give the sort a total order so ties cannot fall back to hash order. The tie-break key is the unique id already carried by each entry, `sequence_id` (`SequenceId = u64`, `src/distributed/tensor_parallel/cache_manager.rs:48` and `src/distributed/pipeline/cache_manager.rs:44`) for instances 1 and 2, and the `String` request key for instance 3.

Either of these forms is acceptable, since what matters is that the sort key is a total order, not which sort function is called:

```rust
// Explicit comparator form.
entries.sort_by(|a, b| a.last_accessed.cmp(&b.last_accessed).then_with(|| a.sequence_id.cmp(&b.sequence_id)));

// Tuple-key form (still sort_by_key, but the key is now total).
entries.sort_by_key(|a| (a.last_accessed, a.sequence_id));
entries.sort_by_key(|a| (std::cmp::Reverse(a.current_offset), a.sequence_id));
```

Apply this to every policy arm: both arms in instance 1 (`LRU`, `LeastTokens`), all three arms in instance 2 (`LRU`, `Shortest`, `Longest`), and the `created_at` sort in instance 3, where the tuple becomes `(created_at, key)`.

Note that `sort_unstable_by` would NOT be a fix on its own; it would merely replace one arbitrary tie order with another.

## Scope

**In scope:**

- `src/distributed/tensor_parallel/cache_manager.rs`: total-order sort in `select_eviction_candidates` (lines 713-720), both policy arms.
- `src/distributed/pipeline/cache_manager.rs`: total-order sort in `select_eviction_candidates` (lines 699-708), all three policy arms.
- `src/distributed/request_tracker.rs`: total-order sort in `evict_if_needed` (line 341).
- Regression tests in `src/distributed/tensor_parallel/cache_manager_tests.rs`, `src/distributed/pipeline/cache_manager_tests.rs`, and `src/distributed/request_tracker_tests.rs`.

**Out of scope:**

- `check_pipeline_pressure` (`src/distributed/pipeline/cache_manager.rs:870-879`) selects across a `&[&PipelineCacheManager]` slice with `max_by`, so its input order is caller-controlled and not hash-derived. No change needed.
- `Registry::nodes_at_stage` / `nodes_at_rank` (`src/distributed/registry.rs:193-227`) also stable-sort a `HashMap`-derived list on a possibly-tied key (`rank`, `stage`), but have no non-test consumer in this repository today, and a tie there means two nodes share a `(stage, rank)` slot, which is a configuration error rather than normal operation. Leave them alone unless a consumer appears.
- Changing the eviction policies themselves. The order among tied entries becomes defined, not "better".

## Implementation Notes

- **Reuse**: the tie-break value is already a field on each entry (`ShardedCacheAllocation::sequence_id` at `src/distributed/tensor_parallel/cache_manager.rs:159`, `StageCacheAllocation::sequence_id` at `src/distributed/pipeline/cache_manager.rs:119`). No new field, accessor, or id source is needed. `SequenceId` is `u64` and `String` is `Ord`, so both tie-breaks compile as-is.
- **Test access to private state**: each test file starts with `use super::*;` and is attached with `#[path = "..._tests.rs"] mod tests;` inside the module it tests (`src/distributed/tensor_parallel/cache_manager.rs:791`, `src/distributed/pipeline/cache_manager.rs:1071`, `src/distributed/request_tracker.rs:352`). A child module can reach the parent's private items, so tests can call the private `select_eviction_candidates` and mutate `self.allocations` / `self.inner` directly. Do NOT add a public setter or a `#[cfg(test)]` accessor just to make ties constructible.
- **Forcing ties**: `last_accessed` and `created_at` are `Instant` set from `Instant::now()`, which on Linux has nanosecond resolution and will effectively never tie naturally. Tests must write equal values into the map entries directly rather than hoping for a collision. `current_offset` ties need no such help; allocate several sequences with the same token count.
- **Constraints**: no public API change, no change to `EvictionSignal` / `PreemptionSignal` shape, no behavior change for candidate lists that have no ties. Existing tests that assert on eviction ordering must keep passing unchanged.
- **Edge cases**: empty `allocations` (candidate list empty, `check_pressure` returns `None` at `src/distributed/tensor_parallel/cache_manager.rs:643-645`, unchanged); a single candidate (no tie, unchanged); all candidates tied (the full list must now come out in ascending `sequence_id` order); `to_remove` larger than the number of terminal requests in instance 3 (`take` already saturates, unchanged).
- **Doc discrepancy worth fixing while in there**: the `PreemptionSignal.sequence_ids` doc at `src/distributed/pipeline/cache_manager.rs:354` reads "in priority order (evict first last)", while `select_eviction_candidates` at line 695 says "first = evict first". The parenthetical is garbled and contradicts the producer. Correct it to match the producer.

## Acceptance Criteria

- [x] All three sites produce an identical candidate list across repeated calls and across freshly constructed maps, for every policy arm (`LRU` and `LeastTokens` in instance 1; `LRU`, `Shortest` and `Longest` in instance 2; `created_at` in instance 3).
- [x] Each regression test builds a **fresh** map per iteration and loops at least 32 times, asserting an identical list each time. `RandomState` randomizes per `HashMap` instance and not merely per process: ten maps built from the same five keys gave nine distinct iteration orders inside one process, measured while filing #1267. A test that builds one map and calls the function twice will pass without the fix and is not acceptable.
- [x] Each regression test is shown to fail against the unfixed code, and the failure output is quoted in the PR description. This is how #1266, #1269, #1281 and #1284 were each validated.
- [x] Each test constructs an actual tie (equal `last_accessed`, equal `current_offset`, or equal `created_at`), because a test over distinct keys passes without the fix.
- [x] The tie-break key is documented in a comment at each call site, stating that the id component is what makes the key a total order, so a later reader does not "simplify" it away.
- [x] Instance 1 additionally asserts end to end that `check_pressure` returns the same `EvictionSignal.sequence_ids` prefix across 32 freshly built managers under an identical tied-allocation scenario, not merely that `select_eviction_candidates` is stable.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` is clean. Verified package-scoped as `cargo clippy --features cuda --lib --tests -- -D warnings`, which exits 0 once the one pre-existing `clippy::err-expect` at `src/multimodal/host_preprocessor_tests.rs:416` (issue #1283, fixed in PR #1285, untouched by this branch) is allowed. The workspace-wide form runs in the full gate on PR #1288.
- [ ] `make verify-test-cuda` passes. Running separately against PR #1288.

## Verification

```bash
cargo fmt --all -- --check
cargo test --profile test-fast distributed::tensor_parallel::cache_manager
cargo test --profile test-fast distributed::pipeline::cache_manager
cargo test --profile test-fast distributed::request_tracker
make verify-test-cuda
```

A pass is: every new determinism test green over its full 32-iteration loop, every pre-existing test in those three modules still green, and clippy silent. Before trusting the tests, revert the three sorts to their current `sort_by_key` form and confirm each new test fails; a test that still passes against unfixed code is not testing the tie order.

## Technical Considerations

Found by a sweep of the remaining `.values()` and `.iter()` collection sites under `src/distributed/` while implementing #1277 (PR #1284). Deliberately not folded into that PR, because these change eviction behavior rather than routing and deserve their own tests.

The sweep at `4d4ac957` covered every `sort_by`, `sort_by_key` and `sort()` call under `src/distributed/`. Everything else is already safe: `src/distributed/registry.rs` sorts on the unique `config.id` after #1284; `src/distributed/pipeline/metrics.rs:374,482,535` sorts on keys that are themselves the unique `HashMap` key (`(src_stage, dst_stage)`, `(stage_index, reason)`, `stage_index`); `src/distributed/cluster_init.rs:552` sorts distinct `SocketAddr` strings and dedups; `src/distributed/disaggregated/request_router.rs:668` was already fixed and carries an explanatory comment; the rest sort `Vec`-derived or slice-derived input. `src/distributed/routing.rs:250` is a genuine prefix consumer (`online[0]` after a stable sort that ties on an idle cluster) but its input now arrives already ordered by node id through `Registry::all_nodes` (`src/distributed/registry.rs:143-148`) via `SchedulerCore::build_candidates` (`src/distributed/scheduler.rs:413-438`), so #1284 already cured it.

Related: #1265 and PR #1266, #1267 and PR #1269, #1276 and PR #1281, #1277 and PR #1284, all instances of the same root-cause class.

inureyes:
## Summary

Three eviction paths under `src/distributed/` build a candidate list from a `HashMap`, sort it, and consume a prefix. The sort looked like it made the result deterministic and did not: `slice::sort_by_key` is a STABLE sort, so entries that compare equal keep their input order, and the input order here is `HashMap` iteration order, which `RandomState` randomizes per map instance. Among candidates that tie on the policy key, which ones landed in the consumed prefix varied run to run.

Each sort now carries the entry's unique id as the final key component, which makes the key a total order and removes hash order from the outcome entirely. `sort_unstable_by` would not have been a fix; it substitutes one arbitrary tie order for another.

## What changed

- `src/distributed/tensor_parallel/cache_manager.rs`: `select_eviction_candidates` sorts on `(last_accessed, sequence_id)` for `LRU` and `(current_offset, sequence_id)` for `LeastTokens`. This is the site with live behavioral consequences today, because `check_pressure` breaks out as soon as projected usage falls under the target, so ties decided which sequences actually lost their cache.
- `src/distributed/pipeline/cache_manager.rs`: `select_eviction_candidates` sorts on `(last_accessed, sequence_id)`, `(current_offset, sequence_id)` and `(Reverse(current_offset), sequence_id)` for `LRU`, `Shortest` and `Longest`. The result is published as `PreemptionSignal.sequence_ids`, documented as a priority order that was not reproducible.
- `src/distributed/request_tracker.rs`: `evict_if_needed` sorts on `created_at` then the request key, in comparator form so the `String` key does not have to be cloned into a tuple. Two requests finishing within the same `Instant` tick tied, and hash order picked which one the `take(to_remove)` prefix dropped.
- Each call site carries a comment naming the id component as the thing that makes the key a total order, so a later reader does not simplify it away.
- Corrected the `PreemptionSignal.sequence_ids` doc, which read "in priority order (evict first last)" and contradicted its own producer at `select_eviction_candidates` ("first = evict first").
- Eight regression tests across the three `*_tests.rs` files. Each builds a fresh map per iteration and loops 32 times, because `RandomState` seeds per `HashMap` instance and not merely per process, so a single map probed twice passes without the fix. Each constructs an actual tie: equal `current_offset` comes free from equal token counts, while equal `last_accessed` and `created_at` are written in directly, since `Instant::now()` has nanosecond resolution on Linux and never ties naturally. No public API change, no `#[cfg(test)]` accessor, and no change to `EvictionSignal` / `PreemptionSignal` shape; the tests reach private state through the `use super::*;` child-module relationship that already exists.

## Failing output before the fix

All eight tests were run against the unfixed sources first. Every one failed, and the `left` value in each assertion is the raw hash order that the stable sort passed through.

`src/distributed/tensor_parallel/cache_manager.rs`, including the end-to-end `check_pressure` prefix assertion:

```
---- distributed::tensor_parallel::cache_manager::tests::check_pressure_prefix_is_deterministic_under_ties stdout ----
assertion `left == right` failed
  left: [15, 11, 14]
 right: [11, 12, 13]

---- distributed::tensor_parallel::cache_manager::tests::eviction_candidates_least_tokens_tie_break_is_deterministic stdout ----
assertion `left == right` failed
  left: [8, 2, 4, 3, 7, 5, 1, 6]
 right: [1, 2, 3, 4, 5, 6, 7, 8]

---- distributed::tensor_parallel::cache_manager::tests::eviction_candidates_lru_tie_break_is_deterministic stdout ----
assertion `left == right` failed
  left: [5, 3, 1, 4, 7, 2, 8, 6]
 right: [1, 2, 3, 4, 5, 6, 7, 8]

test result: FAILED. 45 passed; 3 failed; 0 ignored; 0 measured; 5847 filtered out
```

`src/distributed/pipeline/cache_manager.rs`:

```
---- distributed::pipeline::cache_manager::tests::eviction_candidates_longest_tie_break_is_deterministic stdout ----
assertion `left == right` failed
  left: [2, 4, 8, 7, 5, 3, 1, 6]
 right: [1, 2, 3, 4, 5, 6, 7, 8]

---- distributed::pipeline::cache_manager::tests::eviction_candidates_lru_tie_break_is_deterministic stdout ----
assertion `left == right` failed
  left: [6, 4, 7, 8, 3, 5, 2, 1]
 right: [1, 2, 3, 4, 5, 6, 7, 8]

---- distributed::pipeline::cache_manager::tests::eviction_candidates_shortest_tie_break_is_deterministic stdout ----
assertion `left == right` failed
  left: [5, 2, 6, 7, 3, 1, 4, 8]
 right: [1, 2, 3, 4, 5, 6, 7, 8]

---- distributed::pipeline::cache_manager::tests::preemption_signal_order_is_deterministic_under_ties stdout ----
assertion `left == right` failed
  left: [8, 7, 5, 6, 3, 1, 2, 4]
 right: [1, 2, 3, 4, 5, 6, 7, 8]

test result: FAILED. 43 passed; 4 failed; 0 ignored; 0 measured; 5848 filtered out
```

`src/distributed/request_tracker.rs`:

```
---- distributed::request_tracker::tests::eviction_tie_break_is_deterministic stdout ----
assertion `left == right` failed
  left: ["req-0", "req-1", "req-4", "req-5"]
 right: ["req-2", "req-3", "req-4", "req-5"]

test result: FAILED. 20 passed; 1 failed; 0 ignored; 0 measured; 5874 filtered out
```

## Test plan

All commands run on Linux/CUDA against this branch. The lib test binary was built once with `cargo test --profile test-fast --features cuda --lib --no-run` and then invoked directly, since `cargo test` takes only one filter.

- [x] `cargo fmt --all -- --check` exit 0
- [x] `cargo clippy --features cuda --lib --tests -- -D warnings` exit 101, on one pre-existing `clippy::err-expect` at `src/multimodal/host_preprocessor_tests.rs:416`, which is untouched by this branch (`git diff origin/main` reports no change to that file) and is issue #1283, already fixed in PR #1285. Re-running as `cargo clippy --features cuda --lib --tests -- -D warnings -A clippy::err_expect` exits 0, so nothing in this change trips a lint.
- [x] `target/test-fast/deps/mlxcel-<hash> distributed::tensor_parallel::cache_manager` exit 0, 48 passed, 0 failed
- [x] `target/test-fast/deps/mlxcel-<hash> distributed::pipeline::cache_manager` exit 0, 47 passed, 0 failed
- [x] `target/test-fast/deps/mlxcel-<hash> distributed::request_tracker` exit 0, 21 passed, 0 failed
- [x] `MLX_CUDA_GRAPH_CACHE_SIZE=2000 target/test-fast/deps/mlxcel-<hash> --test-threads=1 distributed::` exit 0, 1285 passed, 0 failed, 4 ignored, covering every pre-existing test in the three modules plus the rest of `src/distributed/`
- [x] The three module filters were run in five further separate processes, each of which reseeds `RandomState` again, all exit 0

The default multi-threaded `distributed::` run aborts with `cudaStreamEndCapture ... failed` from an MLX GPU test in `distributed::pipeline::stage_worker`, unrelated to this change and matching the known GB10 CUDA graph-cache thrash; every distributed test that ran had already passed, and the single-threaded run above with `MLX_CUDA_GRAPH_CACHE_SIZE=2000` completes cleanly.

`make verify-test-cuda` is not included here; it is being run separately against this branch.

Closes #1286

### DIFF
--- a/TECHNICAL_REPORTS/1288-eviction-sort-total-order-20260822.en.md
+++ b/TECHNICAL_REPORTS/1288-eviction-sort-total-order-20260822.en.md
@@ -1,0 +1,80 @@
+# Technical Report: PR #1288 - A total order for eviction candidate sorts
+
+## Executive Summary
+
+Three eviction paths built a candidate list from a `HashMap`, sorted it, and consumed a prefix. Sorting looks like it makes the outcome deterministic and does not: `slice::sort_by_key` is a **stable** sort, so entries comparing equal keep their input order, and the input order is `HashMap` iteration order. Among candidates tied on the sort key, which ones landed in the consumed prefix varied per run.
+
+This is the subtle member of the defect class fixed by #1265, #1267, #1276 and #1277. Those four had no sort at all, so the defect was visible on inspection. These three do sort, which is precisely why they read as safe.
+
+## 1. Problem Statement
+
+- `src/distributed/tensor_parallel/cache_manager.rs`: `select_eviction_candidates` collects `allocations.values()` and sorts on `last_accessed` (LRU) or `current_offset` (LeastTokens). Its caller `check_pressure` iterates and breaks as soon as `projected_used <= target_bytes`, making it a prefix consumer. Ties decide which sequences actually lose their cache, and under LeastTokens a shared token count is an ordinary occurrence rather than a corner case.
+- `src/distributed/pipeline/cache_manager.rs`: the same shape across three `PreemptionPolicy` arms, published as `PreemptionSignal.sequence_ids` and documented as eviction priority order. No prefix consumer exists in-tree today, so what was broken was the published contract rather than observed behavior.
+- `src/distributed/request_tracker.rs`: `evict_if_needed` collects terminal requests, sorts on `created_at`, then takes an explicit prefix with `.take(to_remove)`.
+
+Nothing corrupts and nothing crashes: the eviction relieves the pressure it set out to relieve and the counts are right. What was lost is reproducibility of **which** sequence was sacrificed, which is exactly what an operator needs when a request is preempted and they are working out why.
+
+## 2. Technical Decisions
+
+### 2.1 Make the sort key a total order
+
+Each sort now includes the unique id as a tiebreaker, so no two elements compare equal and stability stops mattering. Both the comparator form and the tuple-key form are correct; the request tracker uses the comparator form specifically so its `String` key is not cloned into a tuple on every comparison, while the two cache managers use the tuple form over `Copy` fields.
+
+`sort_unstable_by` was explicitly not the fix. It would replace one arbitrary tie order with another rather than remove the arbitrariness.
+
+### 2.2 Pin the pipeline contract at its boundary anyway
+
+The issue required an end-to-end assertion only for the tensor-parallel site, since it is the one with a live prefix consumer. The pipeline site got one too, asserting on the published `PreemptionSignal.sequence_ids` rather than only on the private helper, so the ordering is already pinned at the boundary a future consumer will read.
+
+## 3. Change Summary
+
+| File | Change |
+| --- | --- |
+| `src/distributed/tensor_parallel/cache_manager.rs` | Both policy arms sort on `(key, sequence_id)` |
+| `src/distributed/pipeline/cache_manager.rs` | All three policy arms sort on `(key, sequence_id)`; the garbled `PreemptionSignal.sequence_ids` doc corrected to `first = evict first` |
+| `src/distributed/request_tracker.rs` | `created_at` sort tie-breaks on the request key via a comparator |
+| the three matching `*_tests.rs` | 8 new tests |
+
+No public setter and no `#[cfg(test)]` accessor was added. All three test files attach via `#[path]` inside the module under test, so the private helpers and the private state were already reachable, and widening the API to make testing easier would have been the wrong trade.
+
+## 4. Review Findings
+
+The tests were written first and run against the still-unfixed sources, rather than the fix being written and then reverted to check. Three ways to write a vacuous test here were identified in advance and all three turned out to matter:
+
+1. A reused map passes without the fix, because `RandomState` randomizes per `HashMap` instance. Each test builds a fresh map per iteration and loops 32 or 64 times.
+2. Test data with distinct sort keys passes without the fix, because distinct keys already give a total order. Every test constructs a real tie.
+3. `Instant::now()` resolves to nanoseconds on this platform and never collides on its own, so the LRU and `created_at` tests write equal `Instant` values in deliberately.
+
+Pre-fix output, where `left` is raw hash order passed through by the stable sort:
+
+```
+---- eviction_candidates_lru_tie_break_is_deterministic ----
+  left: [5, 3, 1, 4, 7, 2, 8, 6]
+ right: [1, 2, 3, 4, 5, 6, 7, 8]
+---- check_pressure_prefix_is_deterministic_under_ties ----
+  left: [15, 11, 14]
+ right: [11, 12, 13]
+---- eviction_tie_break_is_deterministic ----
+  left: ["req-0", "req-1", "req-4", "req-5"]
+ right: ["req-2", "req-3", "req-4", "req-5"]
+```
+
+One trap not named in the issue, recorded here because it will catch the next person: `evict_if_needed` runs inside `submit_with_id` before the insert, so a test that completes each request as it submits triggers an unintended eviction on the sixth submit. The test submits all six, then completes them, then calls `evict_if_needed` directly.
+
+## 5. Validation
+
+Measured on GB10 (DGX Spark, CUDA sm_121, Linux aarch64), rebased onto `main` at `8fcc01f2` before gating so the gate ran on the tree that merges.
+
+- `make verify-test-cuda`: **8246 passed, 0 failed, 311 ignored**, 101 suites, exit 0. That is +8 against `main` (8238), and the diff adds exactly 8 `#[test]` functions and removes none.
+- The three module filters: 48, 47 and 21 passed, exit 0 each, and green again across five further separate processes, which reseeds `RandomState` per run.
+- `cargo fmt --all -- --check`: exit 0. `cargo clippy --lib --tests --features cuda -- -D warnings`: exit 0 (the `err_expect` that previously reddened this was fixed by #1283 / PR #1285).
+
+The gate log was scanned for process-level aborts as well as failing tests. A per-suite tally cannot see a teardown crash: an earlier run of this same gate reported 0 failed and still exited 101, because every test passed and the process then aborted with `Destroy(handle_) failed: driver shutting down` while a second cargo process saturated the GPU. Run alone, both this gate and #1283's are clean.
+
+## 6. Related Work
+
+- #1286: the issue this closes, filed from a sweep of `src/distributed/` during #1277.
+- #1265 and PR #1266, #1267 and PR #1269, #1276 and PR #1281, #1277 and PR #1284: the sibling instances.
+- #1287: the proposal to record this class in `docs/code-guidelines.md` and decide whether a static check can catch it.
+
+Two near-misses were checked and left alone deliberately. `src/distributed/routing.rs` stable-sorts and takes `online[0]`, and an idle cluster ties on every component, but PR #1284 already cured it upstream by giving `Registry::all_nodes` a defined order, so that path silently depends on the accessor fix. And `nodes_at_stage` / `nodes_at_rank` sort on `Option<u32>` keys through `unwrap_or(u32::MAX)`, which looks tie-able, but `ClusterConfig::validate` rejects a PPTP node with either field unset before it can reach the registry, and neither accessor has a non-test consumer.
--- a/TECHNICAL_REPORTS/1288-eviction-sort-total-order-20260822.ko.md
+++ b/TECHNICAL_REPORTS/1288-eviction-sort-total-order-20260822.ko.md
@@ -1,0 +1,80 @@
+# 기술 보고서: PR #1288 - 축출 후보 정렬에 전순서 부여
+
+## 요약
+
+축출 경로 셋이 `HashMap`에서 후보 목록을 만들고, 정렬한 뒤, 접두사를 소비하고 있었다. 정렬은 결과를 결정적으로 만드는 것처럼 보이지만 아니다. `slice::sort_by_key`는 **안정 정렬**이라 같다고 비교되는 원소가 입력 순서를 유지하고, 그 입력 순서가 곧 `HashMap` 순회 순서다. 정렬 키가 동점인 후보들 중 어느 것이 소비되는 접두사에 들어가는지가 실행마다 달라졌다.
+
+#1265, #1267, #1276, #1277이 고친 결함 계열의 미묘한 쪽이다. 그 넷은 정렬이 아예 없어서 눈으로 보였다. 이 셋은 **정렬을 하고 있어서** 안전해 보였다.
+
+## 1. 문제
+
+- `src/distributed/tensor_parallel/cache_manager.rs`: `select_eviction_candidates`가 `allocations.values()`를 모아 `last_accessed`(LRU) 또는 `current_offset`(LeastTokens)로 정렬한다. 호출자 `check_pressure`는 `projected_used <= target_bytes`가 되는 즉시 `break`하므로 접두사 소비자다. 동점이 어느 시퀀스가 실제로 캐시를 잃는지를 정하고, LeastTokens에서는 토큰 수가 겹치는 것이 예외가 아니라 일상이다.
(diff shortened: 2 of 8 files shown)


inureyes:
Rebased onto `main` at `8fcc01f2` locally before gating, so the gate ran on the tree that will actually merge rather than on the `4d4ac957` base this branch was cut from. The rebase was clean and the rebase itself was not pushed, so this PR still shows its original commit; a squash merge onto current `main` produces the tree that was gated.

`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64), rebased: **8246 passed, 0 failed, 311 ignored**, 101 suites, exit 0. That is +8 against `main` (8238 passed), and the diff adds exactly 8 `#[test]` functions and removes none, so the totals reconcile.

The log was scanned for process-level aborts as well as failing tests, because a per-suite tally cannot see a teardown crash. No `SIGABRT`, no `terminate called`, no `error: test failed`. That check is not theoretical here: an earlier run of this same gate reported 0 failed tests and still exited 101, because every test passed and the process then aborted with `Destroy(handle_) failed: driver shutting down` while a second cargo process was saturating the GPU. Gates on this box are run one at a time for that reason.

`cargo fmt --all -- --check` and `cargo clippy --lib --tests --features cuda -- -D warnings` are both clean on the same tree; the `err_expect` that used to red the latter was fixed by #1283 (PR #1285).

Issue #1248, opened 2026-08-19
fix(server): bound the responses and conversation stores by bytes, not just entry count


inureyes:
The `/v1/responses` store and the conversation store cap only the number of entries, not their size. Each entry retains the complete request input, images included, so a client sending well-formed multimodal `store: true` requests can pin gigabytes of server memory inside the TTL window.

## Evidence

src/server/responses_store.rs:43-46 stores the full response object plus the full request input per entry:

```rust
pub struct StoredResponse {
    pub response: ResponseObject,
    pub input_items: Vec<ResponseInputItem>,
}
```

src/server/responses_store.rs:56-67 caps the store by count only:

```rust
pub struct ResponsesStoreConfig {
    pub max_entries: usize,
    pub ttl: Duration,
}

impl Default for ResponsesStoreConfig {
    fn default() -> Self {
        Self {
            max_entries: 1024,
            ttl: Duration::from_secs(3600),
        }
    }
}
```

src/server/conversation_store.rs:53-63 is the same shape (`max_entries: 256`, 1-hour TTL), and its `evict_to_capacity` at conversation_store.rs:147-156 also evicts by count only. For a multimodal request the `ResponseInputItem` values carry the original base64 image data URLs, so 1,024 responses plus 256 conversation transcripts with no byte accounting is an unbounded memory footprint: entry count bounds nothing when a single entry can be tens of megabytes.

The repository already has the right pattern next door: src/server/prompt_cache/policy.rs enforces both a byte budget (`capacity_bytes` at policy.rs:40) and an entry cap, and `prompt_cache_capacity_bytes` is an operator knob.

Secondary: `evict_to_capacity` at src/server/responses_store.rs:237 does an O(n) linear `min_by_key` scan over the whole map to find the LRU victim on every insert (the comment itself flags it as "acceptable for Phase 1"), and conversation_store.rs:147 duplicates the same scan.

## Suggested fix

- Add byte accounting to both stores: track an approximate serialized size per entry at insert time and a running total, and add a `max_bytes` field to `ResponsesStoreConfig` and `ConversationStoreConfig`, mirroring the `capacity_bytes` pattern in src/server/prompt_cache/policy.rs.
- Evict oldest-accessed entries until both the entry count and the byte total are under budget.
- Expose the knob on both server binaries and document it in docs/environment-variables.md, following the naming convention of `prompt_cache_capacity_bytes`.
[...]

PR #1299, opened 2026-08-22
refactor(server): give the four LRU eviction min_by_key calls a total order

### LINKED ISSUE #1291: refactor(server): give the four LRU eviction min_by_key calls a total order
## Problem / Background

Four LRU eviction sites pick their victim with `min_by_key` applied directly to a `HashMap` iterator. `Iterator::min_by_key` returns the FIRST minimum, so when two entries tie on the timestamp key, which entry gets evicted is decided by `HashMap` iteration order, and `RandomState` seeds that order per map instance. The selection is therefore not pinned by the code; it is pinned only by the accident that the key never ties.

**Severity, stated precisely: this is not a live bug, and it should not be described as four bugs.** All four keys are `std::time::Instant`. Every writer computes `let now = Instant::now()` once per call and stamps a single entry with it, so any two entries take their stamps from two separate `Instant::now()` calls, which resolve to nanoseconds on the platforms this project targets and effectively never collide. Nothing misbehaves today, in any of the four.

What this is instead is a fragile pattern with no mechanical guard behind it. It becomes a real defect the moment the key becomes anything coarser than a per-call `Instant`: a seconds-granularity integer, a logical clock, a value hoisted out of a loop and shared across several entries in a batch operation, or a timestamp deserialized from disk at whatever resolution the format carries. The change that would break it is small, and it would not look like it has anything to do with eviction.

## Current Behavior

All four verified at `a0e402ad`:

- `src/server/prompt_cache/store.rs:315`, in `Inner::evict_oldest`, over `entries: HashMap<PromptCacheKeyDigest, EntrySlot>` (declared at `src/server/prompt_cache/store.rs:101`), keyed on `slot.entry.last_used()`. Called from `enforce_caps` at `src/server/prompt_cache/store.rs:355` and `:363`.
- `src/server/prompt_cache/store.rs:336`, in `Inner::evict_oldest_snapshot`, the same shape over `snapshots: HashMap<PromptCacheKeyDigest, SnapshotSlot>` (declared at `src/server/prompt_cache/store.rs:111`). Called from `enforce_snapshot_caps` at `src/server/prompt_cache/store.rs:376` and `:384`.
- `src/server/responses_store.rs:243`, in `ResponsesStore::evict_to_capacity(map: &mut HashMap<String, Entry>, target_size: usize)`, keyed on `e.last_accessed`.
- `src/server/conversation_store.rs:151`, in `ConversationStore::evict_to_capacity(map: &mut HashMap<String, Entry>, target_size: usize)`, keyed on `e.last_accessed`.

Why the tie is unreachable today, per site. For the two stores, the writers are `ResponsesStore::insert` at `src/server/responses_store.rs:161`, `ResponsesStore::get` at `src/server/responses_store.rs:182`, `ConversationStore::get` at `src/server/conversation_store.rs:83`, and `ConversationStore::append` at `src/server/conversation_store.rs:96`. Each takes one `Instant::now()` and stamps one entry. For the prompt cache, `PromptCacheEntry::touch` (`src/server/prompt_cache/entry.rs:274-280`) and `ModelSnapshotEntry::touch` (`src/server/prompt_cache/entry.rs:413-419`) each write `Instant::now()` under their own mutex, one entry per call, and construction stamps the same way (`src/server/prompt_cache/entry.rs:213`, `:329`, `:388`, `:441`).

## Proposed Solution

Give each key a total order by appending the map key, which is unique by construction. `docs/code-guidelines.md` (the "HashMap Iteration Order" section, starting at line 109) records the accepted forms and the rationale. The tuple form suits the `Copy` `PromptCacheKeyDigest` keys, and the comparator form suits the `String` keys, where a tuple key would force a clone:

```rust
// prompt_cache/store.rs, both sites. The digest is Copy, so the tuple is free.
.min_by_key(|(digest, slot)| (slot.entry.last_used(), *digest.as_bytes()))

// responses_store.rs and conversation_store.rs. The key is a String, so compare
// rather than build a tuple key that would clone it.
.min_by(|(key_a, a), (key_b, b)| a.last_accessed.cmp(&b.last_accessed).then_with(|| key_a.cmp(key_b)))
```

Exact call shapes are the implementer's choice; what matters is that no two distinct entries can compare equal, so `HashMap` iteration order cannot reach the result. One detail that will bite otherwise: `PromptCacheKeyDigest` derives only `Clone, Copy, PartialEq, Eq, Hash` (`src/server/prompt_cache/key.rs:53-54`), so it is not `Ord` and cannot go into a `min_by_key` tuple as-is. Either add `PartialOrd, Ord` to that derive, which is well defined for a BLAKE3 digest newtype over `[u8; 32]`, or key on `*digest.as_bytes()` (`src/server/prompt_cache/key.rs:58`) and leave the type alone. Prefer the latter if you would rather not widen the type's public trait surface for one call site.

Add a short comment at each of the four sites naming the key component as the thing that makes the order total, per the same guideline. Without it the component reads as redundant and a later simplification pass removes it.

## Scope

**In scope:** the four call sites listed above, in `src/server/prompt_cache/store.rs`, `src/server/responses_store.rs`, and `src/server/conversation_store.rs`, plus a comment at each.

**Out of scope:** any change to the eviction policy itself, to the cap-enforcement loops, or to what the stores are bounded by. Byte-bounding of the two stores is #1248. No other `min_by_key` or `max_by_key` call in the tree, and no other iteration-order site.

## Implementation Notes

- **Reuse**: follow the forms already recorded in `docs/code-guidelines.md`, and the fixes that landed under #1277 (PR #1284) and #1286 (PR #1288), rather than inventing a new shape.
- **Constraints**: this must not change behavior in any reachable case. Since no reachable input ties, the fix is behavior-preserving by construction; if it changes an existing test's expected victim, that test was depending on a tie and the change needs explaining, not accepting.
- **Do not**: swap the containers to `BTreeMap`. All four are on lookup-hot paths and the guideline's #1277 precedent kept `HashMap` for exactly this reason.
- **Edge cases**: empty map (the existing `else { break; }` and `None => 0` arms already handle it and must keep working); a single entry (the total order is a no-op there); the prompt cache's `last_used` read taking a mutex per entry, which the tuple key must not turn into a double lock, so read it once per element.
- **Error handling**: unchanged. None of the four sites can fail; the poisoned-lock fallbacks in `entry.rs` and the store guards stay as they are.

## Acceptance Criteria

- [ ] All four sites select their victim through a total order, so the choice cannot depend on `HashMap` iteration order.
- [ ] Each of the four sites carries a comment naming the tie-break component as what makes the order total, so a later simplification pass does not strip it.
- [ ] Behavior on every reachable input is unchanged: no existing test in `src/server/prompt_cache/store_tests.rs`, `src/server/responses_store.rs`, or `src/server/conversation_store.rs` changes its expected outcome.
- [ ] A decision is recorded in the PR body on whether a regression test is worth adding, either way, with the reason. A test is **optional** here and deliberately so: the tie is not reachable through the public API, so a test would have to construct entries carrying a deliberately equal `Instant`, which no production path produces. If one is added it must follow the guideline's testing rule (fresh map per iteration, 32 or more iterations, an actual constructed tie, validated by failing against the unfixed code first). If the implementer judges that such a test would only pin an unreachable state and prefers to rely on the comment plus the guideline, that is an acceptable outcome and should be stated.
- [ ] `make verify-test-cuda` passes.

## Verification

```bash
make verify-fmt
make verify-test-cuda
# Targeted, faster than the full gate while iterating:
cargo test --profile test-fast --features cuda prompt_cache -- --test-threads=1
cargo test --profile test-fast --features cuda responses_store conversation_store -- --test-threads=1
```

On Apple Silicon the equivalent gate is `make verify-fmt && make verify-clippy && make verify-test`. `verify-clippy` builds with `--features metal,accelerate` and does not run on a CUDA box; on Linux, `make verify-fmt` plus `make verify-test-cuda` is the gate.

A pass is: fmt clean, no new clippy warnings on the platform where clippy runs, and no test failures attributable to this change.

## Technical Considerations

**Where these four came from.** They are the residue of the sweep recorded in `docs/code-guidelines.md` under "HashMap Iteration Order" (#1287, PR #1290). That guideline documents seven instances of the same class that WERE live and were fixed: #1265 (PR #1266), #1267 (PR #1269), #1276 (PR #1281), #1277 (PR #1284), and #1286 (PR #1288). It also records why no static check was adopted, and these four sites are precisely the reason the `min_by_key` candidate rule was declined: run over all 1,248 `.rs` files under `src/`, it flags exactly these four, catches 0 of the 7 real instances, and gating on it would have meant suppressing four of four findings on the day it landed, with a suppression that does not re-arm on the change that would make the pattern real. Fixing the four by hand removes that tension and leaves the guideline's decision intact.

**Coordination with #1248.** `src/server/responses_store.rs` and `src/server/conversation_store.rs` are already in scope for **#1248** (`status:ready`, open, no PR as of filing), which bounds those two stores by bytes rather than entry count. Two of the four sites are therefore cheaper to fix inside that work than separately, and #1248 may well rewrite the eviction loop around them. Check #1248's state before starting. If it is in flight, folding those two into it and leaving only the two `src/server/prompt_cache/store.rs` sites here is the better split. Note also that a byte-bounded eviction policy is exactly the kind of change that could introduce a coarser key and make this reachable, which raises the value of doing it alongside rather than after.

inureyes:
## Summary

Four LRU eviction sites picked their victim with `min_by_key` applied directly to a `HashMap` iterator: `Inner::evict_oldest` and `Inner::evict_oldest_snapshot` in `src/server/prompt_cache/store.rs:315` and `:336`, and `evict_to_capacity` in both `src/server/responses_store.rs:243` and `src/server/conversation_store.rs:151`. `Iterator::min_by_key` returns the FIRST minimum, so two entries tied on the timestamp key would be separated only by `HashMap` iteration order, which `RandomState` randomizes per map instance.

**This is not a live bug, and it is not described as four bugs here.** All four keys are `std::time::Instant`, and every writer (`touch()`, `insert()`, `get()`, `append()`) stamps a single entry with one `Instant::now()` call. At nanosecond resolution the tie is not reachable through the public API today. What this PR removes is a latent fragility: the pattern would become a real defect the moment one of these keys becomes coarser than a per-call `Instant` (a seconds-granularity timestamp, a logical clock, a value hoisted out of a loop), and nothing in the current code flags that risk.

## What changed

- `src/server/prompt_cache/store.rs`, `evict_oldest` (line ~315) and `evict_oldest_snapshot` (line ~336): both now key on `(slot.entry.last_used(), *digest.as_bytes())`. `PromptCacheKeyDigest` derives only `Clone, Copy, PartialEq, Eq, Hash` and is not `Ord`, so a bare tuple key would not compile. Per the issue's recommendation, I keyed on `*digest.as_bytes()` (`&[u8; 32]`, which is `Ord`) rather than adding `PartialOrd, Ord` to the digest's derive, to avoid widening that type's public trait surface for one call site. `last_used()` is still called exactly once per element (the key-function closure runs once per item), so this does not turn the per-entry mutex lock into a double lock.
- `src/server/responses_store.rs`, `evict_to_capacity` (line ~243): switched to `min_by` with `a.last_accessed.cmp(&b.last_accessed).then_with(|| key_a.cmp(key_b))`. The key is a `String`, so the comparator form avoids the clone a tuple key would force.
- `src/server/conversation_store.rs`, `evict_to_capacity` (line ~151): same comparator form.
- Each site carries a comment naming the tie-break component (digest bytes or map key) as what makes the order total, per `docs/code-guidelines.md`'s "HashMap Iteration Order" section, so a later simplification pass does not read the component as redundant and strip it.

Scope matches the issue exactly: the four call sites plus a comment at each. No change to eviction policy, cap-enforcement loops, or what the stores are bounded by. `#1248` (byte-bounding of the two stores) is open but still `status:ready` with no PR, i.e. not in flight, so this PR keeps all four sites rather than folding two into that work.

## Regression test: decision and reasoning

No regression test was added. Reasoning:

- The tie is not reachable through any production write path on any of the four sites (confirmed per site above), so a test would have to write an identical `Instant` directly into private state, bypassing `touch()`/`insert()`/`get()`/`append()` entirely.
- Such a test would pin the comparator/tuple-key's generic sort behavior, not any reachable production state. It would not catch a real regression that a code reviewer plus the comment at each site wouldn't also catch.
- The general defect class, its testing methodology (fresh map per iteration, 32+ iterations, an actual constructed tie), and precedent test suites are already recorded in `docs/code-guidelines.md`'s "HashMap Iteration Order" section, including its explicit accounting that these exact four sites are "true positives for the lint and zero live bugs."
- A test becomes worth adding the moment a future change makes one of these keys coarser than a per-call `Instant` (the actual risk this PR is guarding against) — and that change would introduce its own new code path, which is the natural place for its own review and its own test, not a synthetic test bolted onto code that cannot exercise the coarser key today.

This is a judgment call per the issue's acceptance criteria, which explicitly allow either outcome with reasoning stated. No existing test's expected outcome changed.

## Test plan

- [x] `cargo fmt --all -- --check` — exit 0
- [x] `cargo check --lib --tests --features cuda` — exit 0
- [x] `cargo clippy --lib --tests --features cuda -- -D warnings` — exit 0
- [x] `cargo test --profile test-fast --features cuda --lib server::prompt_cache` — exit 0, 168 passed, 0 failed
- [x] `cargo test --profile test-fast --features cuda --lib server::responses_store --no-run` then run the built binary filtered to `server::responses_store` — exit 0, 8 passed, 0 failed
- [x] Same binary filtered to `server::conversation_store` — exit 0, 5 passed, 0 failed

`make verify-test-cuda` will be run separately against this branch.

Closes #1291

### DIFF
--- a/TECHNICAL_REPORTS/1299-lru-eviction-total-order-20260822.en.md
+++ b/TECHNICAL_REPORTS/1299-lru-eviction-total-order-20260822.en.md
@@ -1,0 +1,69 @@
+# Technical Report: PR #1299 - A total order for the four LRU eviction victim selections
+
+## Executive Summary
+
+Four LRU eviction sites picked their victim with `min_by_key` directly over a `HashMap`. `Iterator::min_by_key` returns the first minimum, so entries sharing a timestamp were separated by `HashMap` iteration order, which `RandomState` randomizes per map instance. Each now selects through a total order by appending the map key.
+
+**This was not a live defect and the change does not fix an observed bug.** All four keys are `std::time::Instant`, stamped once per call, so the tie was not reachable. What the change removes is a latent fragility that would become real the moment the key got coarser than a per-call `Instant`.
+
+## 1. Problem Statement
+
+The sites, all over an unordered map:
+
+| Site | Function | Map |
+| --- | --- | --- |
+| `src/server/prompt_cache/store.rs:315` | `evict_oldest` | `entries: HashMap<PromptCacheKeyDigest, EntrySlot>` |
+| `src/server/prompt_cache/store.rs:336` | `evict_oldest_snapshot` | `snapshots` |
+| `src/server/responses_store.rs:243` | LRU eviction loop | `HashMap<String, Entry>` |
+| `src/server/conversation_store.rs:151` | `evict_to_capacity` | `HashMap<String, Entry>` |
+
+Why the tie is unreachable today, stated so the severity is not misread later: each writer computes `Instant::now()` once per call and stamps a single entry with it, and `Instant` resolves to nanoseconds on the platforms this project targets, so two entries never take the same value. The reachable-tie counterpart of this pattern is #1293, where the key is `generated_tokens.len()`, a small integer that ties routinely.
+
+These four are also the reason the `min_by_key` static-check candidate was declined in #1287: it flags exactly these, catches none of the seven live instances, and gating on it would have meant suppressing four of four findings on day one. Fixing them by hand removes that tension rather than carrying it.
+
+## 2. Technical Decisions
+
+### 2.1 Two forms, chosen per key type
+
+`PromptCacheKeyDigest` derives only `Clone, Copy, PartialEq, Eq, Hash` (`src/server/prompt_cache/key.rs:54`), so the obvious tuple key does not compile. Rather than widen the type's derive, both prompt-cache sites key on the raw bytes: `min_by_key(|(digest, slot)| (slot.entry.last_used(), *digest.as_bytes()))`. `[u8; 32]` is `Ord`, and the digest type keeps its original surface.
+
+The two `String`-keyed stores use the comparator form, `min_by(|(key_a, a), (key_b, b)| a.last_accessed.cmp(&b.last_accessed).then_with(|| key_a.cmp(key_b)))`, because `min_by_key` would have to clone the `String` into a tuple on every element.
+
+Both forms are recorded as correct in `docs/code-guidelines.md` under "HashMap Iteration Order", and the choice between them there is explicitly stylistic. Each site carries a comment naming the tie-break component as the thing that makes the order total, per the same guideline, since otherwise it reads as redundant and gets simplified away.
+
+### 2.2 No regression test, deliberately
+
+The issue framed this as a judgment call rather than a requirement, and the judgment here is not to add one. Constructing the tie means writing an identical `Instant` directly into private state, bypassing every real write path (`touch`, `insert`, `get`, `append`). A test built that way pins the behavior of a comparator rather than any reachable production state, and it would need private-field access or a test-only setter to exist at all. The reasoning is recorded in the PR body so the absence is a decision rather than an oversight.
+
+This is the opposite call from #1288, where the tie was reachable and eight tests were both possible and necessary. The distinction is reachability, not effort.
+
+## 3. Change Summary
+
+| File | Change |
+| --- | --- |
+| `src/server/prompt_cache/store.rs` | Both sites key on `(last_used(), *digest.as_bytes())`, with comments |
+| `src/server/responses_store.rs` | Comparator form tie-breaking on the map key, with a comment |
+| `src/server/conversation_store.rs` | Same |
+
+33 lines added, 5 removed, across three files. No public API changed, no type derive widened.
+
+## 4. Review Findings
+
+One implementation detail worth recording, checked rather than assumed: `min_by_key` invokes its closure once per element, so `slot.entry.last_used()` still takes its mutex exactly as many times as before. Folding the digest into the key adds no additional locking.
+
+The issue's coordination note about #1248 was re-checked at implementation time: still `status:ready`, no PR, not in flight, so all four sites stayed in this PR rather than two of them moving there.
+
+## 5. Validation
+
+Measured on GB10 (DGX Spark, CUDA sm_121, Linux aarch64). The branch was current with `main` at gate time.
+
+- `make verify-test-cuda`: recorded in the PR thread.
+- `cargo test --profile test-fast --features cuda --lib server::prompt_cache`: 168 passed, exit 0. `server::responses_store`: 8 passed. `server::conversation_store`: 5 passed. Both of the latter through the prebuilt binary, since `cargo test` takes one filter.
+- `cargo fmt --all -- --check`, `cargo check --lib --tests --features cuda`, `cargo clippy --lib --tests --features cuda -- -D warnings`: all exit 0.
+
+## 6. Related Work
+
+- #1291: the issue this closes.
+- #1287 and PR #1290: the guideline that records the class and the accepted remedy forms, and whose declined static check flagged exactly these four sites.
+- #1293: the reachable-tie sibling in `BatchScheduler::select_eviction_victim`, handled separately because its tie occurs in ordinary batch state and therefore does warrant tests.
+- #1265 and PR #1266, #1267 and PR #1269, #1276 and PR #1281, #1277 and PR #1284, #1286 and PR #1288: the seven live instances of the same class.
--- a/TECHNICAL_REPORTS/1299-lru-eviction-total-order-20260822.ko.md
+++ b/TECHNICAL_REPORTS/1299-lru-eviction-total-order-20260822.ko.md
@@ -1,0 +1,69 @@
+# 기술 보고서: PR #1299 - LRU 축출 victim 선택 네 곳에 전순서 부여
+
+## 요약
+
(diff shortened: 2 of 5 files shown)


inureyes:
`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at `bd4bbadf`, run with no other cargo process on the box: **8251 passed, 0 failed, 311 ignored**, 101 suites, exit 0. The log was scanned for process-level aborts as well as failing tests, since a per-suite tally cannot see a teardown crash: no `SIGABRT`, no `terminate called`, no `error: test failed`.

No delta against an earlier baseline is quoted, because `main` moved during this work: #1292, #1295 and #1296 landed after the last full gate run here, and the latter two add tests. What is checkable is this branch's own contribution, which is zero: the diff adds no `#[test]` functions and removes none, so the total matching `main`'s current count is the expected result rather than a coincidence.

`cargo fmt --all -- --check`, `cargo check --lib --tests --features cuda` and `cargo clippy --lib --tests --features cuda -- -D warnings` are all clean on the same tree.

Issue #1287, opened 2026-08-22
docs(guidelines): record the HashMap-iteration-order defect class and decide whether a static check can catch it


inureyes:
## Problem / Background

A single review sweep between 2026-08-20 and 2026-08-22 turned up seven independent instances of one defect class, spread across five modules, and no part of the toolchain flagged any of them. The types are correct, `cargo check` is clean, `clippy` is clean, and the tests pass most of the time. That last clause is the whole problem: the failure is probabilistic, so a green run is not evidence.

The class is: a `HashMap` iteration result becomes ordered state, or feeds a consumer that is sensitive to order. Rust's `RandomState` seeds each map instance separately rather than once per process, so the resulting behavior varies both between runs and between two maps built from identical input inside one run. Measured here with `rustc -O` on this box, building ten `HashMap<&str, u32>` instances from the same five keys inside one process and recording the `.keys()` order of each, repeated over 200 processes: 133 processes produced 10 distinct orders out of 10, 52 produced 9, and 15 produced 8. There is no run in which the ten maps agreed.

The seven instances, all verified, each linked to its fix or its open issue:

- #1265, fixed by PR #1266 (merged). Four test fixtures named `filled_weights` walked a `WeightMap` and advanced one LCG seed per key, so every process built a different random model. `WeightMap` is `HashMap<String, UniquePtr<MlxArray>>`. This is the one that started the sweep, and by the time the cause was found it had already put two wrong conclusions into a tracked file: the note at the `verify-test-cuda` gate definition in the `Makefile` had attributed the wandering delta to klear's own numerics and to a quantized MoE accumulation path that the fixture never reaches. Both claims are now withdrawn in place (`Makefile:604-616`).
- #1267, fixed by PR #1269 (merged). `LangBiasYamlConfig::bias` was `Option<HashMap<String, BiasValueStr>>` feeding `LangBiasSet.ordered`, which is documented as the priority order and consumed first-language-wins, so a multi-CJK YAML config assigned different biases to shared Han tokens on every run. The shipped schema example in the doc comment is itself a three-CJK config, so copying it was enough to trigger this.
- #1276, fixed by PR #1281 (merged). `needs_sanitize` returned from inside a `weights.keys()` walk on the first marker it saw, so a checkpoint carrying both marker families got a coin-flip layout verdict. The wrong direction is the expensive one: re-running the sanitize pipeline over already-MLX weights double-transposes conv weights into a shape-valid tensor that nothing downstream can flag.
[...]

PR #1301, opened 2026-08-22
fix(server): give preemption victim selection a total order

### LINKED ISSUE #1293: fix(server): preemption victim selection falls back to HashMap order when sequences tie
## Problem / Background

`BatchScheduler::select_eviction_victim` (`src/server/batch/scheduler.rs:6803-6826`) decides which in-flight sequence gets preempted when the batch is full and a higher-priority request is waiting. Both policy arms tie-break on `HashMap` iteration order, and unlike the four sites in #1291, **the tie here is the common case rather than an unreachable one**.

- `PreemptionPolicy::LongestFirst` (`src/server/batch/scheduler.rs:6807-6811`) is `self.active_batch.iter_sequences().filter(|seq| seq.structured.is_none()).max_by_key(|seq| seq.generated_tokens.len())`. `Iterator::max_by_key` returns the LAST maximum, so sequences tied on token count are separated by input order.
- `PreemptionPolicy::LowestPriority` (`src/server/batch/scheduler.rs:6815-6823`) is `min_by(|a, b| a.priority.cmp(&b.priority).then_with(|| b.generated_tokens.len().cmp(&a.generated_tokens.len())))`. Two sequences equal on BOTH priority and generated length fall through to input order, and `min_by` returns the FIRST minimum.

The input order is `HashMap` order. `ActiveBatch::iter_sequences` is `self.sequences.values()` (`src/server/batch/active.rs:120-122`) over `sequences: HashMap<SequenceId, SequenceInfo>` (`src/server/batch/active.rs:33`). Rust's `RandomState` seeds each map instance, so that order is arbitrary and varies between processes, per the measurement recorded in `docs/code-guidelines.md:128-130`.

**Why the tie is reachable here and was not in #1291.** #1291's four LRU sites tie on `std::time::Instant`, stamped once per call, so at nanosecond resolution two entries effectively never collide and nothing misbehaves today. The key here is `generated_tokens.len()`, a small integer. Sequences admitted to the same batch decode in lockstep, one token per sequence per decode step, so a batch that admitted several requests together holds several sequences at an identical token count for as long as they stay together. `PreemptionPolicy::LongestFirst` is the default (`src/server/config.rs:224-231`), so the reachable-tie arm is the one that ships on by default. For the `LowestPriority` arm the tie needs equal `RequestPriority` as well, and `Normal` is the default priority (`src/server/batch/sequence.rs:64-73`), so a batch of unlabeled requests ties on that axis too.

**Scope of the harm, stated honestly.** Any tied-longest sequence does satisfy "evict the longest", so the policy's stated intent is not violated, no output is corrupted, and there is no crash. What is lost is reproducibility. An operator investigating why a particular request was preempted (and, per `try_evict_for_preemption` at `src/server/batch/scheduler.rs:6658-6669`, replayed with duplicate streamed tokens) cannot reproduce the choice from the same batch state, and across a fleet two identically configured workers facing identical load pick different victims. That is the same category of harm #1277 was filed for in node selection, and #1286 was filed as `type:bug` for the structurally identical tie-break-falls-to-hash-order shape in the distributed cache managers.

## Current Behavior

Verified at `a0e402ad`:

```rust
// src/server/batch/scheduler.rs:6803-6826
fn select_eviction_victim(&self) -> Option<SequenceId> {
    match self.preemption_policy {
        PreemptionPolicy::LongestFirst => {
            self.active_batch
                .iter_sequences()
                .filter(|seq| seq.structured.is_none())
                .max_by_key(|seq| seq.generated_tokens.len())
                .map(|seq| seq.seq_id)
        }
        PreemptionPolicy::LowestPriority => {
            self.active_batch
                .iter_sequences()
                .filter(|seq| seq.structured.is_none())
                .min_by(|a, b| {
                    a.priority
                        .cmp(&b.priority)
                        .then_with(|| b.generated_tokens.len().cmp(&a.generated_tokens.len()))
                })
                .map(|seq| seq.seq_id)
        }
    }
}
```

Called from `try_evict_for_preemption` at `src/server/batch/scheduler.rs:6659`, whose returned victim is removed from the batch, has its prompt-cache context and KV caches released, is reset for re-prefill, and is reallocated under a fresh `SequenceId` (`src/server/batch/scheduler.rs:6671-6760`).

**The existing tests do not cover this function and cannot catch a regression in it.** `eviction_selects_longest_first_by_default` (`src/server/batch/scheduler_tests.rs:693-719`) and `eviction_selects_lowest_priority_then_longest` (`src/server/batch/scheduler_tests.rs:722-757`) each reimplement the selection expression inline against a locally built `ActiveBatch` instead of calling `select_eviction_victim`. They have already drifted: neither copy carries the `filter(|seq| seq.structured.is_none())` guard that production has. This matters for the fix, because a regression test written in the same shape would pin the copy and pass regardless of what the production code does.

## Proposed Solution

**Give both arms a total order by appending the unique `seq_id`.** `docs/code-guidelines.md:109-199` ("HashMap Iteration Order") records the accepted forms and the reasoning; follow those rather than inventing a new shape. Two details that will bite otherwise:

1. `SequenceId` derives only `Debug, Clone, Copy, PartialEq, Eq, Hash` (`src/lib/mlxcel-core/src/cache.rs:6089-6090`), so it is not `Ord` and cannot go into a comparison as-is. Use `seq.seq_id.as_u64()` (`src/lib/mlxcel-core/src/cache.rs:6102-6104`) rather than widening the type's public trait surface for two call sites.
2. `max_by_key` returns the LAST maximum while `min_by` returns the FIRST minimum, so the two arms need opposite-facing tie components to break the same direction. Verified empirically with `rustc -O` on this box: over `[(1,5),(2,5),(3,5)]` keyed on the second field, `max_by_key` yields `(3,5)` and both `min_by_key` and `min_by` yield `(1,5)`.

Concrete forms, with the direction argued below:

```rust
// LongestFirst. `max_by_key` returns the LAST maximum, so `Reverse` on the id
// makes the smallest seq_id win the tie. The id component is what makes the key
// a TOTAL order, so hash order cannot reach the result; do not simplify it away.
.max_by_key(|seq| (seq.generated_tokens.len(), std::cmp::Reverse(seq.seq_id.as_u64())))

// LowestPriority. `min_by` returns the FIRST minimum; the trailing seq_id
// comparison makes the order total so that "first" is never decided by hash order.
.min_by(|a, b| {
    a.priority
        .cmp(&b.priority)
        .then_with(|| b.generated_tokens.len().cmp(&a.generated_tokens.len()))
        .then_with(|| a.seq_id.as_u64().cmp(&b.seq_id.as_u64()))
})
```

**Tie-break direction: smallest `seq_id`, and it is meaningful, not just total.** `SequenceId` is a `u64` newtype handed out by `CachePool` from `next_id: AtomicU64` (`src/lib/mlxcel-core/src/cache.rs:6315`), initialized to 0 (`:6349`) and advanced by `fetch_add(1, Ordering::Relaxed)` on every allocation (`:6393`), and the type's own doc comment states the monotonic contract (`:6084-6088`). So the id already encodes admission order and "smallest id wins" is "evict the sequence admitted earliest", which is intelligible to an operator without any new field.

There is one nuance worth stating rather than discovering later. Preemption reallocates the victim under a fresh, therefore higher, id (`src/server/batch/scheduler.rs:6734-6737`), so an id records the most recent admission, not the original arrival. That makes smallest-id-wins the safe direction: a just-preempted sequence carries a young id and is the least likely to be picked again, whereas largest-id-wins would make the sequence that was just replayed the preferred next victim, which risks repeatedly starving the same request. If the maintainer wants true arrival order instead, `SequenceInfo::created_at` survives preemption (it is absent from the reset block at `src/server/batch/scheduler.rs:6725-6732`) and is the field to use, but it is an `Instant` and therefore not total on its own, so `seq_id` would still have to be appended behind it. Record whichever is chosen.

**Make the function testable, because right now it is not.** A regression test cannot call `select_eviction_victim`: it is a private method on `BatchScheduler`, which needs a real model to construct, and the two existing tests worked around that by copying the expression. Extract the policy into one pure function that both the scheduler and the tests call, for example a `pub(crate)` free function in `src/server/batch/scheduler.rs`:

```rust
pub(crate) fn select_eviction_victim_from<'a>(
    sequences: impl Iterator<Item = &'a SequenceInfo>,
    policy: PreemptionPolicy,
) -> Option<SequenceId>
```

with `BatchScheduler::select_eviction_victim` reduced to `select_eviction_victim_from(self.active_batch.iter_sequences(), self.preemption_policy)`. A method on `ActiveBatch` taking the policy is the alternative; prefer the free function unless you would rather have `active.rs` depend on `crate::server::config::PreemptionPolicy`, since `ActiveBatch` currently knows nothing about preemption policy. Either way there must be exactly one copy of the selection logic, and the two existing tests at `src/server/batch/scheduler_tests.rs:693` and `:722` must be rewritten to call it, which also restores the `structured.is_none()` filter to their coverage.

**The guideline correction this issue originally asked for is already done.** `docs/code-guidelines.md:137` used to read "`min_by_key` and `max_by_key`. Both return the FIRST extremum", which is wrong for `max_by_key`. It was corrected in PR #1294 (`698f6237`) and now states the two directions separately, so read the current line rather than the text quoted in earlier revisions of this issue. The enforcement block at the end of that section also already records this issue as the finding both static-check candidates missed, so there is nothing left to add there either. Nothing about the fix below changes; the distinction simply no longer has to be rediscovered.

## Scope

**In scope:** `src/server/batch/scheduler.rs:6803-6826` (both arms, plus the extraction described above and a comment at each site naming `seq_id` as the total-order component); `src/server/batch/scheduler_tests.rs:693-757` (rewrite the two existing tests to call the extracted function) plus the new determinism test; `docs/code-guidelines.md` only if you judge it worth adding this instance to the list at `:184-190`, which currently reads "The seven instances" and would become eight. The `max_by_key` correction at `:137` and the enforcement note at `:192-201` are already in place from PR #1294, so do not redo them.

**Out of scope:**

- Any change to the preemption policy itself, to `try_evict_for_preemption`'s reset-and-reallocate sequence, to the `structured.is_none()` exclusion, or to when preemption triggers.
- The four LRU sites in #1291. They are a separate issue and a separate PR.
- Splitting `scheduler.rs` into a directory module, which is #1243. Expect a textual conflict there if both land; this issue keeps its changes local to the function.
- Two adjacent prompt-cache sites were examined and are deliberately excluded. Both are first-wins accumulators over `HashMap` iteration in the #1265 shape rather than `min_by_key`/`max_by_key` calls, and both gate their tie on an `Instant`, so both are unreachable today for the same reason as #1291: the snapshot scan at `src/server/prompt_cache/store.rs:1081-1136`, whose `best` and `diverged` accumulators both replace only on a strict `>` over `last_used`; and `BestCandidate::beats` at `src/server/prompt_cache/lookup.rs:51-63`, fed by the trie DFS that pushes `children.values()` onto a stack (`src/server/prompt_cache/trie.rs:418`, `:694`). The `beats` doc comment already states that ties on both axes resolve first-seen and warns tests off equal `last_used` values, so the arbitrariness is documented at both ends. File them separately if the keys ever coarsen.

## Implementation Notes

- **Reuse**: the forms recorded in `docs/code-guidelines.md:152-166`, and the fixes that landed under #1277 (PR #1284) and #1286 (PR #1288). Do not introduce a new tie-break idiom.
- **Do not** swap `ActiveBatch`'s `HashMap` for a `BTreeMap`. It is documented as backing O(1) lookup by `SequenceId` (`src/server/batch/active.rs:15-35`) and is read on the decode path; the #1277 precedent in the guideline (`docs/code-guidelines.md:180-182`) is explicitly the keep-the-`HashMap`-and-sort direction for exactly this reason.
- **Constraints**: behavior must be unchanged on every input that does not tie. `RequestPriority` is `Low = 0 < Normal = 1 < High = 2` with a derived `Ord` (`src/server/batch/sequence.rs:64-73`), so the existing `min_by` on priority already selects the lowest and the appended component must not perturb that. If an existing test changes its expected victim, it was depending on a tie and that needs explaining, not accepting.
- **Edge cases**: empty batch and an all-filtered batch both return `None` and `try_evict_for_preemption` returns `false` (`src/server/batch/scheduler.rs:6659-6662`); a single candidate makes the tie component a no-op; every candidate tied on every axis must still resolve to exactly one id, which is the case the new test pins.
- **Error handling**: unchanged. `select_eviction_victim` cannot fail, and the `None` path and every downstream error arm in `try_evict_for_preemption` stay as they are.
- **Cost**: `as_u64()` is a `Copy` field read, so the tie component adds nothing measurable to a path that only runs under memory pressure.

## Acceptance Criteria

- [x] Both arms of `select_eviction_victim` select through a total order, so the victim cannot depend on `HashMap` iteration order.
- [x] The selection logic exists in exactly one place, callable from tests, and the two existing tests at `src/server/batch/scheduler_tests.rs:693` and `:722` call it instead of reimplementing it (which also brings the `structured.is_none()` filter under test).
- [x] Each arm carries a comment naming `seq_id` as the component that makes the order total, per `docs/code-guidelines.md:166`, so a later simplification pass does not strip it.
- [x] A decision is recorded in the PR body on the tie-break direction (smallest `seq_id`, largest, or `created_at` with `seq_id` behind it) and why, including whether the anti-starvation argument above was accepted.
- [x] A regression test constructs an actual tie (equal `generated_tokens.len()` for the `LongestFirst` arm; equal priority AND equal length for the `LowestPriority` arm), rebuilds the `ActiveBatch` inside each iteration, loops at least 32 times, and asserts the same victim every time. Per `docs/code-guidelines.md:168-178` the rebuild is mandatory because `RandomState` seeds per map instance.
- [x] That test is demonstrated failing against the unfixed code, with the observed failure count out of the iteration total quoted in the PR body (PR #1281 recorded 27 of 64, PR #1269 recorded first failures at iterations 0, 0 and 2).
- [x] `docs/code-guidelines.md:137` no longer claims `max_by_key` returns the first extremum, and this instance is added to the list at `:184-190`.
- [ ] `make verify-test-cuda` passes.

## Verification

```bash
# The new and rewritten tests, before and after the fix.
cargo test --profile test-fast --features cuda eviction -- --test-threads=1

# Confirm the test actually fails against the unfixed code: revert only the two
# arms (keep the extraction and the test), re-run, and record the failure count.
git stash push -- src/server/batch/scheduler.rs

make verify-fmt
make verify-clippy
make verify-test-cuda
```

A pass is: the determinism test reports the same victim on all 32-plus iterations after the fix, and reports a nonzero failure count before it. Note that the full workspace suite has a pre-existing abort on this CUDA box unrelated to this change; confirm any failure also reproduces on unmodified `main` before attributing it here.

## Technical Considerations

This is the eighth instance of the class recorded in `docs/code-guidelines.md:109-199` (filed as #1287, landed as PR #1290), after #1265 (PR #1266), #1267 (PR #1269), #1276 (PR #1281), #1277 (PR #1284) and the three in #1286 (PR #1288), and separate from the four latent ones in #1291. It is the first in the lineage that is reachable through ordinary batch state rather than through a coarsening of the key.

It is also a live demonstration of the blind spot that guideline documents. The static-check candidates evaluated in #1287 (`docs/code-guidelines.md:194-199`) reported four `min_by_key`/`max_by_key` flags on the current tree and nothing else, and missed this site because the receiver is a cross-module accessor method (`iter_sequences()`) rather than a literal `.values()` on a map declared in the same file. That is exactly the limitation the guideline describes for #1277, and it is the stated reason the checks were declined rather than adopted. The enforcement section should note that the blind spot has now cost a real finding, so the recorded decision stays honest about what it gave up.

inureyes:
## Summary

`BatchScheduler::select_eviction_victim` decides which in-flight sequence gets preempted when the batch is full and a higher-priority request is waiting. Both policy arms selected through a key that was not total, so tied sequences were separated only by `HashMap` iteration order: `LongestFirst` used `max_by_key(|seq| seq.generated_tokens.len())` and `LowestPriority` used `min_by` on priority then length, both over `ActiveBatch::iter_sequences`, which is `HashMap::values`. `RandomState` seeds each map instance, so the victim was not reproducible from the batch state.

Unlike the four latent sites fixed in #1291 / PR #1299, this tie is reachable through ordinary state. The `LongestFirst` key is `generated_tokens.len()`, a small integer that sequences admitted together and decoding in lockstep share routinely, and both `PreemptionPolicy::LongestFirst` and `RequestPriority::Normal` are the shipped defaults, so the reachable arm is the one that ships on. The harm is reproducibility rather than correctness: any tied-longest sequence does satisfy the stated policy, nothing is corrupted and nothing crashes, but two identically configured workers under identical load preempt different requests, and an operator investigating why a given request was preempted (and replayed with duplicate streamed tokens, per `try_evict_for_preemption`) cannot reproduce the choice.

## What changed

- **`src/server/batch/scheduler.rs`**: the policy is extracted into `select_eviction_victim_from(sequences, policy)`, a `pub(crate)` free function, and `BatchScheduler::select_eviction_victim` becomes a one-line wrapper over it. The `structured.is_none()` filter moves inside the function so no caller can lose it. The free-function form was chosen over a method on `ActiveBatch` for the reason the issue gives: `ActiveBatch` currently knows nothing about `PreemptionPolicy`, and a method would make `active.rs` depend on `crate::server::config`.
- **Both arms now append `seq_id`** as the component that makes the order total, using `as_u64()` because `SequenceId` derives only `Debug, Clone, Copy, PartialEq, Eq, Hash` and is not `Ord`. This is the same trap `PromptCacheKeyDigest` posed in #1291, and it is resolved the same way rather than by widening the type's public trait surface for two call sites. `LongestFirst` keys on `(generated_tokens.len(), Reverse(seq_id.as_u64()))`; `LowestPriority` appends `.then_with(|| a.seq_id.as_u64().cmp(&b.seq_id.as_u64()))`. The two directions differ because `max_by_key` returns the LAST maximum while `min_by` returns the FIRST minimum, so opposite-facing tie components are needed to break the same direction. Each arm carries a comment naming `seq_id` as the total-order component so a later simplification pass does not strip it.
- **`src/server/batch/scheduler_tests.rs`**: `eviction_selects_longest_first_by_default` and `eviction_selects_lowest_priority_then_longest` now call `select_eviction_victim_from` instead of reimplementing the selection expression. Neither expected victim changed, because neither test constructs a tie. Two determinism tests and an empty-batch test are added.
- **`docs/code-guidelines.md`**: this instance is added to the list in the "HashMap Iteration Order" section, which becomes eight. The `max_by_key` correction at the "order-sensitive consumers" bullet and the enforcement note were already in place from PR #1294 and are not redone; the only enforcement-section edits are to keep the measured catch rates arithmetically consistent with an eight-item list, scoped to what #1287 actually measured.

## Tie-break direction: smallest `seq_id`, anti-starvation argument accepted

The issue asked for an explicit decision and it is **smallest `seq_id` wins**, with the anti-starvation argument accepted as stated.

`SequenceId` is a `u64` newtype handed out by `CachePool` from an `AtomicU64` advanced with `fetch_add(1)` per allocation, so smallest-id-wins reads as "evict the sequence admitted earliest", which needs no new field to be intelligible.

The decisive part is what preemption does to the id. `try_evict_for_preemption` reallocates its victim under a fresh, therefore higher, id, so a large id marks a sequence that was preempted *recently*, not one that arrived late. Smallest-id-wins therefore rotates preemption onto sequences that have not been hit yet, whereas largest-id-wins would make the request that was just replayed the preferred next victim.

`SequenceInfo::created_at` was considered and rejected. It survives preemption (it is absent from the reset block) and so gives true arrival order, but for exactly that reason it points the wrong way for starvation: smallest-`created_at`-wins would keep selecting the same oldest request, because nothing about being preempted moves the field. It is also an `Instant` and not total on its own, so `seq_id` would have had to sit behind it anyway.

Both arms resolve a tie to the same direction, so the choice does not vary by policy.

## Regression test, and its failure against the unfixed code

Two tests, both built per `docs/code-guidelines.md` "Testing the fix": a fresh `ActiveBatch` inside each iteration (`RandomState` seeds per map instance, so probing one batch repeatedly measures nothing), 64 iterations, and an actual tie constructed on every axis the arm compares. Ids are inserted out of ascending order so a selector that returns the first- or last-inserted sequence cannot pass by coincidence. Each test records a mismatch per iteration rather than panicking on the first, so a failure reports the count out of 64.

The tests were written against the extracted-but-unfixed function and run before the tie-break was added:

```
running 5 tests
test server::batch::scheduler::tests::eviction_returns_none_for_an_empty_batch ... ok
test server::batch::scheduler::tests::eviction_selects_lowest_priority_then_longest ... ok
test server::batch::scheduler::tests::eviction_selects_longest_first_by_default ... ok
test server::batch::scheduler::tests::eviction_longest_first_tie_resolves_to_smallest_seq_id ... FAILED
test server::batch::scheduler::tests::eviction_lowest_priority_tie_resolves_to_smallest_seq_id ... FAILED

---- eviction_longest_first_tie_resolves_to_smallest_seq_id stdout ----
LongestFirst resolved a fully tied batch to something other than seq_id 2 on 58 of 64 freshly built batches (first Some((0, Some(5)))); the tie is falling through to HashMap order

---- eviction_lowest_priority_tie_resolves_to_smallest_seq_id stdout ----
LowestPriority resolved a batch tied on priority and length to something other than seq_id 2 on 39 of 64 freshly built batches (first Some((1, Some(14)))); the tie is falling through to HashMap order

test result: FAILED. 3 passed; 2 failed; 0 ignored; 0 measured; 5895 filtered out
```

So **58 of 64** for `LongestFirst` (first mismatch at iteration 0) and **39 of 64** for `LowestPriority` (first at iteration 1). Both are well under 100 percent, which is the reason the loop exists: a single-shot version of the `LowestPriority` test would have passed about 39 percent of the time. After the fix, all 64 iterations of both tests agree.

No `git stash` was used to produce this. The tests were run against the extraction commit before the tie-break was applied, and `git stash list` is empty.

## Notes on the issue

Everything the issue asserted checked out against the code, with two points worth recording:

- **Line numbers drifted.** The issue was verified at `a0e402ad`; on `c9d53891` (after PR #1299 and PR #1300 landed) `select_eviction_victim` is at `scheduler.rs:6826` rather than `:6803`, and the guideline list is at `docs/code-guidelines.md:184` rather than the quoted span. The code at those sites is otherwise identical to what the issue quotes.
- **One acceptance criterion was already satisfied on `main`.** "`docs/code-guidelines.md:137` no longer claims `max_by_key` returns the first extremum" was fixed by PR #1294; the line already states the two directions separately, so no change was needed there. The issue itself says as much in its "Proposed Solution" section, but the criterion was still listed unchecked.

The `structured.is_none()` filter is now in the tested code path, but no test constructs a `structured: Some(...)` sequence: `StructuredOutputConstraint` has no public test constructor, and building one through `build_json_schema_constraint` needs a real HuggingFace tokenizer with a vocab plus an llguidance grammar compilation, which does not belong in a scheduler policy unit test. Covering the exclusion directly would need a test-only constructor on that type, which is out of scope here.

## Test plan

- [x] `cargo test --profile test-fast --features cuda --lib server::batch::scheduler::tests::eviction` (exit 0, 5 passed, 0 failed)
- [x] `cargo test --profile test-fast --features cuda --lib server::batch::scheduler` (exit 0, 111 passed, 0 failed, 7 ignored)
- [x] `cargo test --profile test-fast --features cuda --lib server::batch` (exit 0, 373 passed, 0 failed, 7 ignored)
- [ ] `make verify-test-cuda` is run separately against this branch.

Closes #1293


inureyes:
`make verify-test-cuda` on GB10 (CUDA sm_121, Linux aarch64) at `478bd6be`, run with no other cargo process on the box: **8256 passed, 0 failed, 311 ignored**, 101 suites, exit 0. The branch was current with `main` at gate time.

That is +3 against `main`, and the diff adds exactly 3 `#[test]` functions and removes none, so the totals reconcile. The two pre-existing eviction tests were rewritten to call the extracted function rather than reimplement it, which changes what they cover but not their count or their expected victim, since neither constructs a tie.

The log was scanned for process-level aborts as well as failing tests, because a per-suite tally cannot see a teardown crash: no `SIGABRT`, no `terminate called`, no `error: test failed`.

`cargo fmt --all -- --check`, `cargo check --lib --tests --features cuda` and `cargo clippy --lib --tests --features cuda -- -D warnings` are all clean on the same tree.

inureyes:
Merging with the `cargo-clippy` check still queued, and recording why rather than leaving it implicit.

That job is queued behind the in-progress `v0.6.0` release build, which holds the GB10 self-hosted runner. It is queued, not failing, and `mergeStateStatus` is `UNSTABLE` rather than `BLOCKED`, so it is not a required check. Every other check on this PR passed.

Rather than wait on an unrelated release, its exact command was run locally on this tree: `cargo clippy -p mlxcel --lib --tests -- -D warnings`, default features, **exit 0** with no diagnostics. That is the same invocation the job runs, so the pending check has no information left to add. `cargo clippy --lib --tests --features cuda -- -D warnings` is also exit 0, as is `make verify-test-cuda` at 8256 passed and 0 failed.

Worth recording as a consequence of the job added in #1285: it runs on the GB10 self-hosted runner to get a warm cache, but GB10 also serves `xla-compile` and the release build, so a release in flight queues clippy behind it for every open PR. The lint this job exists to catch reproduces at default features and needs no accelerator, and `pipeline-parallel-ci.yml` already runs the same command on `ubuntu-latest`, so moving it there would trade a warm cache for independence from the release runner. Worth revisiting.

Issue #1274, opened 2026-08-21
fix: Integration tests cannot link with --features cuda,xla-iree


inureyes:
## Problem / Background

Any integration test built with the default features plus `cuda,xla-iree` fails at link time on a GB10 host (aarch64, sm_121) with `IREE_CUDA_HOME` provisioned by `scripts/iree/setup-cuda.sh`:

```
/usr/bin/ld: .../libiree_runtime_unified.a(call.c.o): undefined reference to symbol '__stack_chk_guard@@GLIBC_2.17'
/usr/bin/ld: /lib/ld-linux-aarch64.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

This is not specific to one test. The failure reproduces identically on `tests/molmo2_xla_vision_parity.rs` (which declares `required-features = ["xla-iree"]`, `Cargo.toml:332-333`) and on the unrelated `tests/cli_help_consistency.rs`, which has no XLA involvement at all. The whole integration-test target class is affected, because the link recipe is emitted once by the root `build.rs` and applies to every target rustc links.

Found while validating PR #916. It is unrelated to that PR's changes: the same failure reproduces on an unrelated test target on the same tree.

## Current Behavior

The root `build.rs` emits the IREE link recipe. The CUDA branch (`build.rs:101-141`, keyed on `IREE_CUDA_HOME`) emits this argument list at `build.rs:117-129`:

```
-Wl,--whole-archive -l:libiree_runtime_unified.a -Wl,--no-whole-archive
-Wl,--start-group -l:libiree_hal_drivers_cuda_registration_registration.a
-l:libflatcc_parsing.a -lgcc -lm -lpthread -ldl -Wl,--end-group
```

rustc links these test binaries with `-nodefaultlibs`, and the group does not include `-lc`, so libc symbols the IREE runtime references (`__stack_chk_guard` in the observed failure) have nothing to resolve against. `libprintf_printf.a` is conditionally inserted at `build.rs:130-136` when the source build produced it, which does not change the libc situation.

Two other link recipes live in the same file and share the same shape of risk:

- The macOS branch at `build.rs:38-86`, keyed on `IREE_MACOS_HOME`. It uses `-Wl,-force_load` rather than `--whole-archive`, relies on Apple ld being multi-pass (so no `--start-group`), links driver archives collected by `collect_driver_archives`, and has no `libgcc`.
[...]

PR #1305, opened 2026-08-22
chore(ci): link an OpenXLA binary in CI so link-only regressions cannot reach main

### LINKED ISSUE #1303: chore(ci): link an OpenXLA binary in CI so link-only regressions cannot reach main
## Problem / Background

The `xla-compile` job added by PR #1282 (`.github/workflows/ci.yml:308-352`) catches compile errors under the OpenXLA feature combinations but cannot catch link errors, because `cargo check` never invokes the linker. The job's own exclusion list at `.github/workflows/ci.yml:296-307` records this gap; this issue is the work of closing it.

That gap is not hypothetical. Issue #1274 was exactly a link failure, and `cargo check --features cuda,xla-iree --all-targets` passed cleanly on the same tree that could not link a single integration test:

```
/usr/bin/ld: libiree_runtime_unified.a(call.c.o): undefined reference to symbol '__stack_chk_guard@@GLIBC_2.17'
/usr/bin/ld: /lib/ld-linux-aarch64.so.1: error adding symbols: DSO missing from command line
```

It reached `main` and was found by hand while validating an unrelated PR. The fix, PR #1275, appends a second `-lc` after the IREE archives in `build.rs` (see the comment block at `build.rs:143-176`, ending with the `"-lc"` entry in the `--start-group` list), because rustc emits its own `-lc` before anything `cargo:rustc-link-arg` can append.

## Current Behavior

`.github/workflows/ci.yml:348-352` runs two commands on the self-hosted `GB10` runner:

```
cargo check --features cuda,xla-iree --all-targets
cargo check --no-default-features --features xla-diagnostics --all-targets
```

Neither links. Nothing else in CI compiles any XLA feature at all: `pipeline-parallel-ci.yml` runs clippy on default features, `nightly-verify.yml` runs `make verify` with `metal,accelerate`, and `release.yml` does not build the `xla-iree` feature. So a regression in the link recipe in `build.rs` reaches `main` unobserved and is caught only when a developer builds locally with these features.

## Proposed Solution

Add a job to `.github/workflows/ci.yml` that links at least one target with `--features cuda,xla-iree` on the `GB10` runner, reusing the `xla-compile` job's provisioning shape: `bash scripts/iree/setup-cuda.sh` followed by `bash scripts/iree/setup-cuda.sh --env | sed 's/^export //' >> "$GITHUB_ENV"`, `MLX_CUDA_ARCHITECTURES: "121"`, and a persistent `CARGO_TARGET_DIR` that is not shared with the release job or with `xla-compile` (a link job emits different artifacts and would repeatedly invalidate a check-only cache).

The design decisions this issue must settle, rather than leaving to the implementer:

**Profile is forced, not chosen.** On this host the debug profile cannot link these targets at all. It fails with hundreds of `relocation truncated to fit: R_AARCH64_CALL26` errors against ordinary `libstd` and `compiler_builtins` symbols, because the unoptimized binary exceeds the AArch64 direct-branch range. A link gate must therefore use `--release`. The cheaper debug link is not an option.

**Cost is the reason this is a separate decision and not a step appended to `xla-compile`.** Linking pulls in the whole IREE runtime and the MLX C++ library, so it is minutes rather than seconds even warm, and a cold release build of `mlxcel-server` with `cuda,xla-iree` measured 17 minutes 6 seconds on the GB10 runner. `xla-compile` is cheap enough to run on every Rust PR; a link job may not be, and that tradeoff is what this issue decides.

**What to link.** Weigh these in the issue's resolution and record the choice in the workflow file:

- `cargo test --release --features cuda,xla-iree --test xla_prepared_prefill --no-run`, the smallest thing that exercises the recipe. `xla_prepared_prefill` and `molmo2_xla_vision_parity` are the two `[[test]]` targets gated on `xla-iree` (`Cargo.toml:327-333`). `--no-run` links without executing, so no GPU contention with development work on the same host.
- `cargo build --release --features cuda,xla-iree --bin mlxcel-server`, closer to what ships but much more expensive, and the target whose 17m06s cold time is measured above.

**When to run it.** Running on every Rust PR is one option. The alternative is a schedule, or a narrower path filter over `build.rs` and `scripts/iree/**`, trading detection latency for runner time. A narrower trigger is already expressible: the `changes` job's `rust` filter (`.github/workflows/ci.yml:77-84`) was extended in #1282 to include `build.rs` and `scripts/iree/**`, so a sibling filter output keyed on just those two paths is a small addition to the existing `dorny/paths-filter` step.

## Scope

**In scope:** a new job in `.github/workflows/ci.yml`, its trigger, its target directory, and the comment recording its cost and coverage next to the existing `xla-compile` exclusion list at `.github/workflows/ci.yml:296-307`. If a narrower path filter is chosen, a new output on the `changes` job.

**Out of scope:** running the XLA test suites (that stays excluded for GPU contention, per `.github/workflows/ci.yml:297-299`); clearing the dead-code backlog and widening `RUSTFLAGS` to `-D warnings` (separate issue); linking the macOS or `IREE_DIST` recipes (see below, no runner exists).

## Implementation Notes

- **Reuse**: copy the provisioning and target-directory steps from `xla-compile` (`.github/workflows/ci.yml:324-347`) rather than writing a new setup path. `scripts/iree/setup-cuda.sh` is idempotent and reuses `~/.cache/mlxcel/iree-cuda-<version>`. Keep the `if: github.repository == 'lablup/mlxcel'` guard so fork PRs never queue on the self-hosted runner.
- **Constraints**: `--release` is mandatory (see above). The GB10 runner is shared with development work, so the job must not execute GPU work, only link. Give it its own `CARGO_TARGET_DIR` under `$HOME/.cargo-target/`, matching the naming of `mlxcel-xla-ci` and `mlxcel-clippy-ci`.
- **Coverage the job will not have**: the `IREE_DIST` recipe (`build.rs:180-220`) and the macOS `IREE_MACOS_HOME` recipe (`build.rs:38-90`) are still unverified by any link, on any machine, because no runner has either distribution. PR #1275 deliberately left both unchanged for that reason. A gate that only covers `IREE_CUDA_HOME` must say so in its comment rather than implying those paths are covered.
- **Edge cases**: a cold runner pays the one-time IREE runtime build; set `timeout-minutes` high enough to absorb it (`xla-compile` uses 120). A purged `CARGO_TARGET_DIR` pays the full cold link, so the timeout must accommodate the 17m06s figure plus the rest of the workspace.

## Acceptance Criteria

- [ ] A job in `.github/workflows/ci.yml` links at least one target under `--features cuda,xla-iree` and fails its check run when the link fails.
- [ ] The job's trigger (every Rust PR, path-filtered, or scheduled) and its measured cost are recorded in a comment in the workflow file, adjacent to the existing `xla-compile` exclusion list.
- [ ] The comment states explicitly that the `IREE_DIST` and macOS `IREE_MACOS_HOME` recipes in `build.rs` remain unverified, so the residual gap stays recorded rather than assumed shut.
- [ ] Demonstrated by reintroducing the defect: temporarily reverting the `-lc` entry added in PR #1275 (`build.rs:143-176`) makes the new job fail, and restoring it makes the job pass. This mirrors how PR #1282 demonstrated its own gate.
- [ ] The job does not execute GPU work, so it does not contend with development on the GB10 host.

## Verification

```
# On the GB10 runner, with the IREE runtime provisioned:
bash scripts/iree/setup-cuda.sh
eval "$(bash scripts/iree/setup-cuda.sh --env)"
export MLX_CUDA_ARCHITECTURES=121

# The chosen link command, e.g.:
cargo test --release --features cuda,xla-iree --test xla_prepared_prefill --no-run

# Negative control: revert the `-lc` entry in build.rs and confirm the same
# command fails with the "DSO missing from command line" error above, then
# restore it.
```

A pass is the link command exiting 0 with the `-lc` entry present, and exiting non-zero with it removed. Record both outcomes in the PR.

## Technical Considerations

Related: #1274 (the link failure), #1275 (the fix), #1282 (the compile gate this extends).

inureyes:
## Summary

`xla-compile` runs `cargo check`, which never invokes the linker, so a regression in the IREE link recipe in `build.rs` can reach `main` unobserved; #1274 was exactly that failure, found by hand after the fact rather than by CI. This PR adds a new `xla-link` job that actually links an OpenXLA target on the self-hosted GB10 runner, closing the gap `xla-compile`'s own comment already recorded as open.

## What changed

- `changes` job: a new `xla_link` output from a sibling `dorny/paths-filter` filter keyed on `build.rs`, `src/lib/mlxcel-xla/build.rs`, `src/lib/mlxcel-xla/csrc/**`, `scripts/iree/**`, `rust-toolchain.toml`, and `.github/workflows/ci.yml`. That is narrower than the existing `rust` filter, so the link job does not run on every Rust PR, but it covers the whole causal surface of a link failure: the root `build.rs` holds the IREE recipe, `scripts/iree/**` pins the distribution whose archive set the recipe names, `mlxcel-xla`'s build script and its `csrc/**` sources produce the shim object whose undefined symbols those archives resolve, and `rust-toolchain.toml` is on the path because #1274 was entirely about where rustc places its own `-lc` relative to appended `rustc-link-arg` entries.
- New `xla-link` job, placed immediately after `xla-compile`, `needs: changes`, gated on `github.repository == 'lablup/mlxcel' && needs.changes.outputs.xla_link == 'true'`, `runs-on: GB10`, `permissions: contents: read`, `timeout-minutes: 120`, a job-scoped `concurrency` group keyed per PR with `cancel-in-progress`, and its own `CARGO_TARGET_DIR` (`$HOME/.cargo-target/mlxcel-xla-link-ci`) separate from `xla-compile`'s and from the release job's.
- The link step runs `cargo test --release --features cuda,xla-iree --test xla_prepared_prefill --no-run`. `--release` is mandatory because the debug profile cannot link these targets on this host at all (hundreds of `relocation truncated to fit: R_AARCH64_CALL26` errors against ordinary libstd symbols, the unoptimized binary exceeding the AArch64 direct-branch range). `--no-run` links without executing, so the job never touches the GPU. Because cargo also builds the package's `[[bin]]` targets whenever an integration test is selected, this command links `mlxcel`, `mlxcel-server`, `speculative_bench`, and `mlxcel-bench-decode` in addition to the test binary, so it is a superset of the `--bin mlxcel-server` alternative rather than a cheaper substitute for it. `RUSTFLAGS` is deliberately left unset, unlike `xla-compile`, so a red run here is unambiguously a link failure rather than a lint failure.
- The `concurrency` group exists because `xla-link` is the first job on the shared GB10 runner whose warm cost is minutes rather than seconds. Without it, three pushes to a `build.rs` PR queue three link jobs of up to 120 minutes each on the one runner that also serves `cargo-clippy` for every Rust PR and the release build.
- Comment blocks recording the trigger rationale, the measured cost, what the gate was demonstrated on, and the residual coverage gaps: the `IREE_DIST` and macOS `IREE_MACOS_HOME` recipes in `build.rs` remain unverified because no runner in this repository holds either distribution, and a link regression arriving through a path outside the filter is still not caught at PR time.
- Corrected the note above the `clippy` job, which asserted that the repository guard means fork PRs never queue on the self-hosted runner. It does not. On a `pull_request` event opened from a fork against this repository, `github.repository` is the base repository, so the guard is true and the job runs here. The guard's real effect is to stop the job queueing forever in a fork of the repository. What gates the fork-PR case is the repository's Actions fork-PR approval policy. That reading applies equally to `xla-compile` and to the new job, and the issue specifying this work repeated the incorrect version, so it was propagating.

## Verification

All of the following ran on the GB10 host with the IREE runtime provisioned (`bash scripts/iree/setup-cuda.sh`, `eval "$(bash scripts/iree/setup-cuda.sh --env)"`, `MLX_CUDA_ARCHITECTURES=121`).

- The job's link command passes from a purged `CARGO_TARGET_DIR`: exit 0 in 15m40s, ending at `Executable tests/xla_prepared_prefill.rs`. Most of that is MLX's CUDA sources compiling from scratch, not the link.
- Warm cost, which is what a `build.rs` edit actually pays because it invalidates the `mlxcel` crate but not `mlxcel-core`'s MLX build: 6m01s and 7m24s across two runs, both exit 0.
- **The gate catches a link-only regression that `xla-compile` cannot.** With `-l:libflatcc_parsing.a` dropped from the `IREE_CUDA_HOME` branch of `build.rs`, `xla-compile`'s command `cargo check --features cuda,xla-iree --all-targets` still exits 0 in 59 seconds, while this job's command exits 101 with `error: linking with cc failed` and undefined references to `flatcc_verify_string_field`, `flatcc_verify_table_vector_field`, and `flatcc_verify_vector_field`. Restoring the archive returns the link to exit 0. That is the same shape as #1274, on the same tree, and is the demonstration the issue asked for.
- The job runs green on this PR itself, since the PR changes `.github/workflows/ci.yml` and so matches the filter.

`build.rs` is unchanged by this PR; every control above was run outside the branch and reverted.

## On the negative control the issue specified

The issue's acceptance criteria asked for the demonstration to be done by reverting the `-lc` entry added in #1275. That control was tried first and did not fail the link: the command exits 0 with `rustc-link-arg=-lc` confirmed absent from the emitted build-script output and the IREE archives confirmed present on the link line.

Why it does not fail is unresolved, and nothing here should be read as evidence that the entry is dead. An earlier revision of this PR claimed the pinned runtime was built without the stack protector; that claim was wrong and has been removed. It came from running `nm` in a shell where `nm` is an alias for an unrelated command, so the search silently matched nothing. Against the pinned distribution, using `/usr/bin/nm`: `__stack_chk_guard` is undefined in 176 objects of `libiree_runtime_unified.a`, including the `call.c.o` named in #1274; it is undefined in `libc.so.6`; and it is defined only in `ld-linux-aarch64.so.1`. Every precondition the `build.rs` comment records still holds today. Two further explanations were tested and ruled out: `-lpthread` and `-ldl` resolve to stub archives on this glibc rather than to scripts that would pull libc in late, and `libm.so` groups only `libm.so.6`.

So `build.rs` is left alone, the flatcc control above is what demonstrated the gate, and the workflow comment records all of this so the next reader does not delete the `-lc` entry on the strength of one control that failed to reproduce.

Closes #1303

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,15 @@ jobs:
   clippy:
     name: cargo-clippy
     needs: changes
-    # Self-hosted, so fork PRs never queue on it. Mirrors `xla-compile` below.
+    # On the repository guard, because it is easy to read more into it than it
+    # does. It keeps this job from queueing forever in a *fork of* this
+    # repository, which has no GB10 runner of its own. It does not stop a pull
+    # request opened *from* a fork against lablup/mlxcel: on that event
+    # `github.repository` is the base repository, so the guard is true and the
+    # job runs here, executing the PR's `build.rs` and `scripts/iree/**` as the
+    # runner's own user. What actually gates that case is the repository's
+    # Actions fork-PR approval policy, currently `first_time_contributors`.
+    # The same reading applies to `xla-compile` and `xla-link` below.
     if: github.repository == 'lablup/mlxcel' && needs.changes.outputs.rust == 'true'
     runs-on: GB10
     permissions:
@@ -472,6 +480,14 @@ jobs:
     runs-on: GB10
     permissions:
       contents: read
+    # The other jobs on this runner are seconds warm; this one is minutes, so
+    # it is the first here that can hold GB10 long enough to matter. Without a
+    # group, three pushes to a `build.rs` PR queue three link jobs of up to
+    # 120 minutes each on the one runner that also serves clippy for every Rust
+    # PR and the release build. Keyed per PR, so PRs never cancel each other.
+    concurrency:
+      group: xla-link-${{ github.event.pull_request.number || github.ref }}
+      cancel-in-progress: true
     timeout-minutes: 120
     env:
       # The value shipped for this runner's architecture. Auto-detection yields
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,8 @@ jobs:
             # every Rust PR. See the `xla-link` job comment for the rationale.
             xla_link:
               - 'build.rs'
+              - 'src/lib/mlxcel-xla/build.rs'
+              - 'src/lib/mlxcel-xla/csrc/**'
               - 'scripts/iree/**'
               - 'rust-toolchain.toml'
               - '.github/workflows/ci.yml'
@@ -406,12 +408,16 @@ jobs:
   # What it links, and why: `cargo test --release --features cuda,xla-iree
   # --test xla_prepared_prefill --no-run`. `build.rs` emits the IREE recipe
   # through `cargo:rustc-link-arg`, which applies to every linked artifact of
-  # the crate, so the smallest linked target exercises the same recipe as
-  # `mlxcel-server` does. `--no-run` links without executing, so this job never
-  # touches the GPU and never contends with development work on the shared
-  # GB10 host. `cargo build --release --features cuda,xla-iree --bin
-  # mlxcel-server` was considered and rejected: it costs 17m06s cold on this
-  # runner and adds no additional recipe coverage over the test binary.
+  # the crate, so any linked target exercises the same recipe `mlxcel-server`
+  # does, and an integration test is the direct reproducer because #1274 was an
+  # integration-test link failure. Cargo also builds the package's `[[bin]]`
+  # targets whenever an integration test is selected, so `mlxcel-server` and
+  # the other three binaries are linked here as well; the costs measured below
+  # already include them, which makes `cargo build --release --features
+  # cuda,xla-iree --bin mlxcel-server` a strict subset of this command rather
+  # than a cheaper alternative to it. `--no-run` links without executing, so
+  # this job never touches the GPU and never contends with development work on
+  # the shared GB10 host.
   #
   # Profile: `--release` is mandatory, not a choice. On this host the debug
   # profile cannot link these targets at all: it fails with hundreds of
@@ -420,12 +426,14 @@ jobs:
   # exceeds the AArch64 direct-branch range.
   #
   # Trigger: path-filtered via the `changes` job's `xla_link` output, not every
-  # Rust PR and not scheduled. `build.rs` holds the recipe and
-  # `scripts/iree/**` pins the IREE distribution whose archive set the recipe
-  # names, so those two are the causal surface. `rust-toolchain.toml` is on the
-  # same causal path because the #1274 failure was entirely about where rustc
-  # places its own `-lc` relative to appended `rustc-link-arg` entries, which a
-  # toolchain bump can move. `.github/workflows/ci.yml` is included so edits to
+  # Rust PR and not scheduled. `build.rs` holds the recipe, `scripts/iree/**`
+  # pins the IREE distribution whose archive set the recipe names, and
+  # `src/lib/mlxcel-xla/build.rs` with its `csrc/**` sources builds the shim
+  # object whose undefined symbols those archives resolve, so those are the
+  # causal surface. `rust-toolchain.toml` is on the same causal path because
+  # the #1274 failure was entirely about where rustc places its own `-lc`
+  # relative to appended `rustc-link-arg` entries, which a toolchain bump can
+  # move. `.github/workflows/ci.yml` is included so edits to
   # this job itself are testable. A schedule was rejected because it decouples
   # the failure from the PR that caused it, which is the exact failure mode of
   # #1274 (found by hand later, not by CI). Running on every Rust PR was
@@ -446,19 +454,22 @@ jobs:
   # unset means a red run here is unambiguously a link failure, not a lint
   # failure wearing a link job's name.
   #
-  # What the gate was demonstrated on, since #1274's own defect no longer
-  # reproduces here: the IREE runtime currently pinned by
-  # `scripts/iree/setup-cuda.sh` is built without the stack protector, so
-  # `libiree_runtime_unified.a` holds no `__stack_chk_guard` reference at all
-  # (`nm` reports zero) and removing the `-lc` entry from `build.rs` links
-  # cleanly today. The `-lc` entry is therefore inert against this particular
-  # distribution, though still correct to keep for one built with the stack
-  # protector. The gate was instead demonstrated by dropping
-  # `-l:libflatcc_parsing.a` from the same recipe: on that tree `cargo check
+  # What the gate was demonstrated on: dropping `-l:libflatcc_parsing.a` from
+  # the `IREE_CUDA_HOME` recipe in `build.rs`. On that tree `cargo check
   # --features cuda,xla-iree --all-targets` still passed in 59 seconds while
   # this job's link command failed with undefined references to
   # `flatcc_verify_*`, which is the same shape as #1274 and the exact gap this
   # job exists to close.
+  #
+  # #1274's own defect, removing the second `-lc`, was tried as the control
+  # first and did not fail the link on the tree this job was validated against.
+  # Why it did not is unresolved, so do not read that run as evidence the entry
+  # is dead: against the pinned runtime, `nm` still reports `__stack_chk_guard`
+  # undefined in 176 objects of `libiree_runtime_unified.a`, including the
+  # `call.c.o` named in #1274; the symbol is still undefined in `libc.so.6`;
+  # and it is still defined only in `ld-linux-aarch64.so.1`. Every precondition
+  # the `build.rs` comment records holds today, so leave that entry alone until
+  # a control that actually reproduces says otherwise.
   #
   # Deliberately NOT covered, so the gap is recorded rather than assumed shut:
   #   - The `IREE_DIST` recipe (`build.rs:180-220`) and the macOS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,7 @@ jobs:
     outputs:
       rust: ${{ steps.filter.outputs.rust }}
       mlx_pin: ${{ steps.filter.outputs.mlx_pin }}
+      xla_link: ${{ steps.filter.outputs.xla_link }}
     steps:
       - uses: actions/checkout@v7
         with:
@@ -88,6 +89,14 @@ jobs:
               - 'src/lib/mlxcel-mlx-pin/**'
               - 'scripts/ci/mlx_pinned_commit.sh'
               - 'scripts/ci/mlx_pinned_commit_test.sh'
+              - '.github/workflows/ci.yml'
+            # Narrower than `rust` above: only the paths that can move the IREE
+            # link recipe itself, so the release link job below does not run on
+            # every Rust PR. See the `xla-link` job comment for the rationale.
+            xla_link:
+              - 'build.rs'
+              - 'scripts/iree/**'
+              - 'rust-toolchain.toml'
               - '.github/workflows/ci.yml'
 
   deny:
@@ -312,6 +321,10 @@ jobs:
   # runner.
   #
   # Deliberately NOT covered, so the gap is recorded rather than assumed shut:
+  #   - Linking. `cargo check` never invokes the linker, so a regression in the
+  #     IREE link recipe in `build.rs` (see #1274/#1275) is invisible here. The
+  #     `xla-link` job immediately below closes that gap for the CUDA/IREE
+  #     recipe; see its comment for what it covers and what it still does not.
   #   - `cargo test`. This is a compile gate; running the XLA suites needs a
   #     GPU that is not contended with development work on the same host.
   #   - Full `-D warnings`. The XLA feature combination carries a pre-existing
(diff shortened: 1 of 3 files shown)


inureyes:

The diff is one workflow file, so this is a CI supply chain and self-hosted-runner review rather than an application-code one. **No CRITICAL or HIGH issue is introduced by this PR.** One HIGH-severity property of the workflow is inherited unchanged from `clippy` and `xla-compile`; it is described below and is explicitly not a finding against this change.

### Verified clean

- **No script injection.** Neither `run:` block in the new job interpolates `${{ github.* }}` or any other untrusted context. Both use only shell variables and literal text, so the discipline the `cross-repo-refs` job documents (base ref passed through `env:`) is not violated here.
- **`permissions: contents: read`** on the job, under a top-level `contents: read`. This repository's default workflow permission is `write`, so the explicit narrowing is doing real work.
- **`persist-credentials: false`** on checkout, so no token is left in `.git/config` for the PR's own `build.rs` or `scripts/iree/setup-cuda.sh` to read.
- **Trigger is `pull_request`, not `pull_request_target`**, so fork code never runs against base-repo secrets or the base ref.
- **`--no-run` really does not execute the linked binary**, so the "never touches the GPU, never contends with development work" claim holds. The contention this job creates is runner occupancy, not GPU occupancy.
- **Provisioning is not an open supply chain hole.** `setup-cuda.sh` verifies the compiler wheel by sha256 and pins the runtime to an exact `iree-org` commit.
- **Action pinning matches the rest of the file** (`actions/checkout@v7`). Nothing here is SHA-pinned and `sha_pinning_required` is off repository-wide, so this is inherited, not introduced.
- **The separate `CARGO_TARGET_DIR` is the right call.** It is distinct from `mlxcel-cuda-gb10` and `mlxcel-cuda-aarch64`, so a CI gate cannot invalidate or poison a release build's cache.

### HIGH, inherited unchanged, not introduced by this PR

Fork pull requests do reach the self-hosted runner. `github.repository == 'lablup/mlxcel'` evaluates **true** on a PR opened from a fork against this repository, because on that event `github.repository` is the base repository. The guard's real effect is to stop the job queueing forever in a *fork of* the repository, which has no GB10 runner. The repository is public with forking enabled, its Actions fork-PR approval policy is `first_time_contributors`, and the runner runs as a personal user account whose home directory holds `.ssh`, `.config/gh`, and the model store. So a contributor with one merged PR can get code execution there through `build.rs`.

**This PR adds no exposure.** `clippy` and `xla-compile` are gated on `rust == 'true'` (`**/*.rs`), which is strictly wider than `xla_link` (`build.rs`, `scripts/iree/**`, `rust-toolchain.toml`, `ci.yml`), and both already run this repository's `setup-cuda.sh` and the PR's `build.rs` on the same runner. Any fork PR that could trigger `xla-link` already triggers both of them today.

Worth a separate hardening issue rather than scope creep here: adding `&& (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)` to all three GB10 jobs, tightening the approval policy to all external contributors, or giving the runner its own unprivileged account.

What I did fix (`ab432ac7`) is the comment above `clippy` that asserted the opposite ("Self-hosted, so fork PRs never queue on it"). #1303 repeated that reading when specifying this job, so it was actively propagating.

### MEDIUM, amplified by this PR, fixed in `ab432ac7`

`ci.yml` has no concurrency control anywhere. That was tolerable while every GB10 job was seconds warm (clippy 27 to 40s, `xla-compile` about a minute); `xla-link` is 6m to 7m30s warm with `timeout-minutes: 120`, a step change. Without a group, three pushes to a `build.rs` PR queue three link jobs on the single runner that also serves clippy for every Rust PR and the release build, and the superseded ones run to completion instead of being cancelled. Added a job-scoped group keyed on the PR number, so superseded runs are cancelled and no other job's behaviour changes. Same pattern `pipeline-parallel-ci.yml` already uses.

### MEDIUM, introduced here, not fixed

This adds a fourth permanent CI-owned target directory on a volume that is 93% full (282G free of 3.7T; `~/.cargo-target` is already 39G, of which `mlxcel-xla-ci` alone is 18G). A release-profile link directory will grow well past its current 2.6G, and nothing prunes any of them. Not a blocker, but the same host runs releases, so a retention step or a follow-up issue is worth having.

### LOW

- **`GITHUB_ENV` injection surface**, inherited verbatim from `xla-compile`: `setup-cuda.sh --env | sed 's/^export //' >> "$GITHUB_ENV"` would let a multi-line value set arbitrary variables for later steps (`RUSTFLAGS`, `LD_PRELOAD`, `PATH`). The emitted values are `$IREE_DIR` and `$IREE_COMPILE`, and the script is in-repo anyway, so this is redundant with simply editing the script. Noted, not changed.
- **A green run does not prove a link happened.** This PR's own `OpenXLA feature link` check passed in 12s with the link step taking 1s, reusing artifacts the manual GB10 validation left in the same directory. That is correct cargo behaviour, and the fingerprint does move for `build.rs`, `rust-toolchain.toml`, and an `IREE_VERSION` bump, which are the triggers that matter. The narrow hole: a `scripts/iree/**` edit that changes how the runtime is built without moving `IREE_VERSION` leaves both `IREE_CUDA_HOME` and `build.rs` unchanged, and the script's idempotence means the archives are not rebuilt either, so the job can green without relinking. Probably worth one line in the "Deliberately NOT covered" list; not worth a design change.
- **The filter omits `Cargo.toml` and `Cargo.lock`**, so a dependency pulling in a conflicting native library is not caught at PR time. Already recorded in the job's own comment, and widening it would fire on every dependency bump, so the tradeoff as written looks right.

### Verdict

No unresolved CRITICAL or HIGH attributable to this PR. The two fixes above are in `ab432ac7`; the remaining items are either inherited posture or follow-up issues.

inureyes:

### Intent

Add a CI job that actually links an OpenXLA target so a regression in the IREE link recipe cannot reach `main` the way #1274 did, path-filtered, release profile, no GPU work.

### Findings addressed in `225586be`

- [x] **(HIGH) The `-lc` paragraph recorded a false fact.** It stated the pinned IREE runtime "is built without the stack protector", that `libiree_runtime_unified.a` "holds no `__stack_chk_guard` reference at all (`nm` reports zero)", and that the `-lc` entry "is therefore inert". Against the pinned distribution on the GB10 host (`~/.cache/mlxcel/iree-cuda-3.12.0rc20260721`), `/usr/bin/nm` reports `U __stack_chk_guard` in 176 objects of that archive, including the `call.c.o` named in #1274's error. `nm -D` on `libc.so.6` shows the symbol undefined there, and `ld-linux-aarch64.so.1` is where it is defined. Those are exactly the preconditions `build.rs` states. The likely source of the zero reading is environmental: in the maintainer's interactive shell `nm` is an alias to a `mosh` command, so `nm <archive> | grep -c ...` returns 0 with no output at all. The comment now records that the `-lc` control did not reproduce, that the reason is unresolved, and that the flatcc control is what demonstrated the gate, without concluding the entry is dead.
- [x] **(MEDIUM) "What it links" was inaccurate about the binaries.** Cargo builds the package's `[[bin]]` targets whenever an integration test is selected, so `cargo test --test xla_prepared_prefill --no-run` links `mlxcel`, `mlxcel-server`, `speculative_bench`, and `mlxcel-bench-decode` as well as the test. The validation run's target directory holds all four (test binary at 20:20:49, the four binaries at 20:21:59 to 20:23:59, same run). The rejected `cargo build --bin mlxcel-server` alternative is therefore a strict subset of the chosen command, not a costlier one. The chosen command is still the right one, since it links the integration test that #1274 failed on, but the rationale now says what actually happens and notes that the measured figures already include the binaries.
- [x] **(HIGH) The path filter missed part of the causal surface.** `src/lib/mlxcel-xla/build.rs` compiles the C shim and emits `rustc-link-lib=static=xla_iree` plus its search path, and `src/lib/mlxcel-xla/csrc/**` produces the object whose undefined symbols the IREE archives resolve. A change there that leaves an unresolved symbol passes `cargo check` and, under the original filter, would not have started this job either: `build.rs` as a picomatch pattern matches only the root build script, and `csrc/**` matches no other filter in the file. Both paths are now in the `xla_link` filter and the trigger comment names them.

### Verified and left alone

- `xla_link` output wiring, the sibling `dorny/paths-filter` entry, and the `if:` condition all parse and evaluate as intended (the inner filter document is valid YAML; the output name matches the filter key).
- Step shape matches `xla-compile` exactly: `actions/checkout@v7` with `persist-credentials: false`, the `CARGO_TARGET_DIR` export into `GITHUB_ENV`, and the `sed 's/^export //'` idiom on the `--env` output.
- Its own `CARGO_TARGET_DIR`, so no cache contention with `xla-compile` or the release job. `--no-run` means no GPU work.
- The four settled design decisions (release profile, which target, path-filtered trigger, `RUSTFLAGS` unset) are sound as decisions; only the recorded reasoning was corrected.

### Remaining item for the author

- [ ] **The PR body still carries the disproved `-lc` narrative** ("`nm` reports zero", "the `-lc` entry is therefore inert", and the suggested follow-up to weaken the `build.rs` comment). Since this repository preserves the PR body as the squash commit body, that text becomes permanent history on merge. Please correct it, and re-run the `-lc` control with an absolute `nm` path before drawing any conclusion about that entry.

PR #1381, opened 2026-08-22
chore: clear the OpenXLA warning backlog and deny all warnings in CI

### LINKED ISSUE #1304: chore: clear the OpenXLA dead-code backlog so the CI gate can deny all warnings
## Problem / Background

The `xla-compile` job added by PR #1282 sets `RUSTFLAGS: "-D unused_imports"` (`.github/workflows/ci.yml:321-323`) rather than `-D warnings`. That was a deliberate choice and its reasoning is recorded in the workflow file at `.github/workflows/ci.yml:300-306`: measured on the tree at that time, `cargo clippy --features cuda,xla-iree --lib --tests -- -D warnings` failed with 4 errors in `mlxcel-xla` alone, before reaching `mlxcel`, and `cargo check --features cuda,xla-iree --all-targets` reported about 10 dead-code warnings across both crates. A job that is red the day it lands gets routed around and then ignored, which is worse than no gate, because a green summary elsewhere then means less than it appears to.

`unused_imports` was chosen because it is the exact lint class of one of the two defects that reached `main` through the OpenXLA coverage gap (a dead `load_weights_from_dir_with_filter` re-export), and compile errors cover the other (a non-exhaustive `match` in the OpenXLA serve worker after a new `ModelRequest` variant landed). Both historical breaks are caught today. The gap is that any new dead code under these features is not.

Clearing the backlog is the precondition for widening the gate.

## Current Behavior

`.github/workflows/ci.yml:348-352` runs `cargo check --features cuda,xla-iree --all-targets` and `cargo check --no-default-features --features xla-diagnostics --all-targets` with `RUSTFLAGS: "-D unused_imports"`. Warnings other than unused imports are printed and ignored.

## The backlog as measured

From `cargo check --features cuda,xla-iree --all-targets`:

In `mlxcel-xla`:
- `src/lib/mlxcel-xla/src/aux.rs`: enum variants `Float16` and `Uint32` never constructed (defined around `src/lib/mlxcel-xla/src/aux.rs:100-101`, matched in the size and code accessors just below).
- `src/lib/mlxcel-xla/src/iree.rs`: methods `decode_ragged_logits` (`src/lib/mlxcel-xla/src/iree.rs:3011`) and `decode_ragged_mrope_logits` (`src/lib/mlxcel-xla/src/iree.rs:3081`) never used. Note their `_with_modes` siblings at `:3022` and `:3096` are the live callees, so these two are the thin wrappers.

In `mlxcel`:
- `src/server/batch/xla_audio_preprocess.rs`: field `0` of `Cancelled(AudioPreprocessCheckpoint)` (`:51`) never read; method `snapshot` (`:223`) never used; field `healthy` (`:412`) never read; struct `AudioPreprocessMetricsSnapshot` (`:342`) never constructed; associated items `spawn` (`:419`), `recv` (`:637`), `metrics` (`:641`) and `is_healthy` (`:645`) never used.
- `src/server/media.rs`: method `validate_xla_raw_counts` (`:200`) never used. Its `_with_audio` sibling at `:210` is the live one.

Clippy adds, in `mlxcel-xla` alone: a collapsible `if`, and a function with too many arguments (8 against the 7 threshold).

## Proposed Solution

Resolve each item above, then widen the gate.

**Deletion is usually the wrong fix here.** Several of these are live under other feature combinations. `decode_ragged_mrope_logits` and much of the audio preprocess surface exist for paths compiled under different features, so the correct change is generally to cfg-gate the item so its `cfg` matches its consumers, or to keep it with `#[allow(dead_code)]` plus a comment stating which feature or caller uses it. Deleting an item that is live under `xla-diagnostics` but dead under `cuda,xla-iree` turns a warning into a build break on the other feature set.

The pattern to follow already exists in-tree: the Molmo2 parity test added in PR #1282 silenced two of its dead-code warnings with `#[allow(dead_code)]` and a comment naming the diagnostics-only caller. See `tests/molmo2_xla_vision_parity.rs:122-128` ("A build without those features still constructs them, so they are allowed rather than cfg-gated") and `tests/molmo2_xla_vision_parity.rs:144-145` ("Only the diagnostics-gated `assert_within` reports this").

Decision criterion for each item: if the item has a real consumer under some feature set, cfg-gate it to that feature set when the gate is expressible without duplicating the item's definition, and otherwise `#[allow(dead_code)]` it with a comment naming the feature or caller. Delete only when a search shows no consumer under any feature combination.

The two clippy findings in `mlxcel-xla` are ordinary fixes: collapse the `if`, and either introduce a parameter struct or add a scoped `#[allow(clippy::too_many_arguments)]` with a one-line reason.

**Then widen the gate.** Change `.github/workflows/ci.yml:321-323` from `RUSTFLAGS: "-D unused_imports"` to `RUSTFLAGS: "-D warnings"`, and update or remove the paragraph at `.github/workflows/ci.yml:300-306` that explains the narrower policy, so the workflow file does not keep describing a state that no longer holds.

## Scope

**In scope:** `src/lib/mlxcel-xla/src/aux.rs`, `src/lib/mlxcel-xla/src/iree.rs`, `src/server/batch/xla_audio_preprocess.rs`, `src/server/media.rs`, and the two clippy findings in `mlxcel-xla`; plus the `RUSTFLAGS` value and the surrounding comment in `.github/workflows/ci.yml`.

**Out of scope:** adding a link gate for the OpenXLA features (#1303); running the XLA test suites in CI (still excluded, `.github/workflows/ci.yml:297-299`); any change to warning policy for feature sets other than `cuda,xla-iree` and `xla-diagnostics`.

## Implementation Notes

- **Reuse**: follow the `#[allow(dead_code)]`-plus-comment pattern already in `tests/molmo2_xla_vision_parity.rs` rather than inventing a new convention.
- **Constraints**: the fix must leave every other feature set building. At minimum re-check default features and `metal,accelerate` (what `nightly-verify.yml` runs) after cfg-gating anything, since a `cfg` that is too narrow breaks the build elsewhere instead of warning.
- **Edge cases**: an item dead under `cuda,xla-iree` but live under `xla-diagnostics` (which implies both `cuda` and `xla-iree`, `Cargo.toml:103`) must stay compiled under the latter. `Cancelled(AudioPreprocessCheckpoint)`'s unread field is a payload that is constructed but never destructured; prefer keeping the payload and documenting why over changing the variant's shape, since `AudioPreprocessOutcome::Cancelled` is constructed in several places in the same file.
- **Error handling**: not applicable; this is a lint and visibility change, and no runtime behavior may change. If a fix would alter behavior, it is the wrong fix.

## Acceptance Criteria

- [ ] `cargo clippy --features cuda,xla-iree --all-targets -- -D warnings` passes on the GB10 runner.
- [ ] `cargo clippy --no-default-features --features xla-diagnostics --all-targets -- -D warnings` passes on the same host.
- [ ] `.github/workflows/ci.yml` sets `RUSTFLAGS: "-D warnings"` for the `xla-compile` job, and the comment explaining the narrower `-D unused_imports` policy is updated or removed so it no longer describes a policy that is gone.
- [ ] Every item resolved by cfg-gating or by an `#[allow(...)]` carries a comment naming the feature or caller that keeps it alive, so the next reader does not delete something that is live elsewhere.
- [ ] No runtime behavior changes: the diff contains no change to any executed code path beyond attribute and `cfg` additions, argument-struct refactors, and the collapsed `if`.
- [ ] Default-feature and `metal,accelerate` builds still compile, confirming no `cfg` was narrowed too far.

## Verification

```
# On the GB10 runner, with the IREE runtime provisioned:
bash scripts/iree/setup-cuda.sh
eval "$(bash scripts/iree/setup-cuda.sh --env)"
export MLX_CUDA_ARCHITECTURES=121

cargo clippy --features cuda,xla-iree --all-targets -- -D warnings
cargo clippy --no-default-features --features xla-diagnostics --all-targets -- -D warnings

# Regression check on the untouched feature sets:
cargo clippy -p mlxcel --lib --tests -- -D warnings
```

A pass is all three commands exiting 0 with no warnings emitted. On a macOS host, additionally run `make verify` to confirm the `metal,accelerate` path is unaffected.

## Technical Considerations

Related: #1282 (the job and its recorded exclusions), #1303 (the link-gate half of the same exclusion list).

inureyes:
## Summary

The `xla-compile` job denied `unused_imports` rather than `warnings` because the `cuda,xla-iree` and `xla-diagnostics` feature sets carried a dead-code and clippy backlog that would have made the job red the day it landed. This clears that backlog and widens the gate to `-D warnings`.

Resolution followed the criterion in the issue: cfg-gate to the feature set that has a real consumer where the gate needs no duplicate definition, otherwise allow with a comment naming the caller, and delete only when a search finds no consumer under any feature combination.

### `mlxcel-xla`

- `src/lib/mlxcel-xla/src/iree.rs`: `IreeRaggedLlama::decode_ragged_logits` is now `#[cfg(feature = "diagnostics")]`. Its only callers are `LlavaReferenceDiagnosticEngine::capture` and the Gemma3n diagnostic runners in `batch.rs`, all of which are already `#[cfg(feature = "diagnostics")]`; serving calls `decode_ragged_logits_with_modes` directly.
- `src/lib/mlxcel-xla/src/iree.rs`: `decode_ragged_mrope_logits` is deleted. This is the one deletion in the PR. `grep -rn "decode_ragged_mrope_logits" . --exclude-dir=.git --exclude-dir=target` filtered of the `_with_modes` sibling and the `xla_llama_` extern returns exactly two hits: the definition itself, and an error-message string inside `decode_ragged_mrope_logits_with_modes` that stays. There is no caller in any crate, test, bench, or example, under any feature combination.
- `src/lib/mlxcel-xla/src/iree.rs`: a doc comment describing `create_ctx` had been stranded above a blank line when #1302 inserted `create_bucket_ctx` and its own doc block between them, which is what `clippy::empty_line_after_doc_comments` was reporting. It is moved back onto `create_ctx`.
- `src/lib/mlxcel-xla/src/aux.rs`: `AuxiliaryWeightDType::Float16` and `Uint32` are allowed under `not(feature = "micro-oracle")`. `numeric_probe.rs`, behind that feature, is their only construction site, while `ffi_code` and `size_bytes` have to keep mapping them for the FFI contract on every build. Gating the variants themselves would force a `cfg` onto each arm of those two matches, including the `Float32 | Uint32` arm.
- `src/lib/mlxcel-xla/src/phi4_audio.rs`: `load_audio_module` takes a scoped `#[allow(clippy::too_many_arguments)]`. A parameter struct would ripple into both call sites for a private single-file helper, so the allow was preferred per the issue's guidance.
- `src/lib/mlxcel-xla/src/aux_manifest.rs`: the nested `if let` in `cache_lock_is_stale` collapses to an edition-2024 let chain.

### `mlxcel`

- `src/server/batch/xla_audio_preprocess.rs`: the test-only surface is allowed under `not(test)`, following the `#[cfg_attr(not(test), allow(dead_code))]` pattern already used in `src/model_metadata.rs` and `src/server/startup.rs`. Each site names `xla_audio_preprocess_tests.rs` as the caller: `AudioPreprocessStage::spawn` (serving uses `spawn_with_loader` to keep MLX handles thread-confined), `recv` (serving must not block, so it drains with `try_recv`), `metrics`, `is_healthy`, the `healthy` field it reads, `AudioPreprocessMetrics::snapshot`, and `AudioPreprocessMetricsSnapshot`. The `Cancelled(AudioPreprocessCheckpoint)` payload keeps its field with the same allow: serving matches it as `Cancelled(_)`, and the checkpoint distinction belongs in the type because `Cancelled` is constructed at several checkpoints in that file.
- `src/server/media.rs`: `validate_xla_raw_counts` gains a `not(test)` arm alongside its existing `not(feature = "xla-iree")` one. `media_tests.rs` is its only caller; the XLA admission path calls `validate_xla_raw_counts_with_audio` directly because it knows whether the loaded family supports audio.
- `src/server/batch/xla_worker_admission.rs`: `drain_preprocessed` takes `#[allow(clippy::while_let_loop)]`. The suggested `while let Some(stage) = self.image_preprocessor.as_ref()` head holds a shared borrow of `self` across a body that calls `&mut self` methods and clears the field, so the `match` shape is load-bearing rather than stylistic.
- `src/loading/vlm.rs` and `src/models/sanitize.rs`: the two `needless_match` sites take `#[cfg_attr(not(feature = "surgery"), allow(clippy::needless_match))]`. The match only collapses to `match transform { Some(t) => Some(t), None => None }` when the `surgery` arm is compiled out, which is why the lint fires under `--no-default-features` and not under the default feature set. Applying clippy's suggestion would delete active-pipeline resolution from the default build, so the lint is silenced only for the builds that see the collapsed shape.

### CI

- `.github/workflows/ci.yml`: `xla-compile` now sets `RUSTFLAGS: "-D warnings"`, and the paragraph explaining the narrower `-D unused_imports` policy is replaced.
- `.github/workflows/ci.yml`: the "Deliberately NOT covered" list records the gap that remains after this change. The job runs `cargo check`, so `-D warnings` denies rustc lints only. `dead_code` and `unused_imports` are gated now, but `collapsible_if`, `needless_match`, `too_many_arguments`, `while_let_loop` and the rest of clippy's own lints are still gated by nothing in CI under these feature sets, because the `clippy` job covers default features only. That half of the backlog can regrow silently; swapping this job's `cargo check` for `cargo clippy` was deliberately left out of scope.

## No runtime behavior change

The diff is attribute additions, `cfg` additions, comments, one collapsed `if`, one doc-comment move, and the deletion of one function that has no caller. No executed code path changes.

## Test plan

Run on the GB10 host with the IREE runtime provisioned (`bash scripts/iree/setup-cuda.sh`, `eval "$(bash scripts/iree/setup-cuda.sh --env)"`, `MLX_CUDA_ARCHITECTURES=121`), and re-run on the final commit after the review follow-ups landed.

Acceptance criteria:

- [x] `cargo clippy --features cuda,xla-iree --all-targets -- -D warnings`: exit 0, no warnings.
- [x] `cargo clippy --no-default-features --features xla-diagnostics --all-targets -- -D warnings`: exit 0, no warnings.
- [x] Default-feature regression, `cargo clippy -p mlxcel --lib --tests -- -D warnings`: exit 0, no warnings.

The two commands the job itself runs, under the widened flag, because `cargo clippy -- -D warnings` applies the flag only to the primary package while `RUSTFLAGS` reaches every path-built unit including `mlxcel-core` and `mlxcel-surgery`. Clippy passing is therefore not sufficient evidence that the job will be green:

- [x] `RUSTFLAGS="-D warnings" cargo check --features cuda,xla-iree --all-targets`: exit 0.
- [x] `RUSTFLAGS="-D warnings" cargo check --no-default-features --features xla-diagnostics --all-targets`: exit 0.

Scoped and mechanical checks:

- [x] `cargo clippy -p mlxcel-xla --lib --features iree --all-targets -- -D warnings`
- [x] `cargo clippy -p mlxcel-xla --lib --features iree,diagnostics --all-targets -- -D warnings`
- [x] `cargo clippy -p mlxcel-xla --lib --features iree,micro-oracle --all-targets -- -D warnings`, confirming the `micro-oracle` cfg arm on the `AuxiliaryWeightDType` variants is not narrowed wrongly
- [x] `rustfmt --edition 2024 --check` on every changed Rust file
- [x] `.github/workflows/ci.yml` parses as YAML

End to end, this PR's own CI exercises the change it makes: `OpenXLA feature compile` ran for 7m44s under the new `RUSTFLAGS: "-D warnings"` and passed, so the widened gate is green on arrival rather than red. `OpenXLA feature link`, added by #1305, also ran because `src/lib/mlxcel-xla/**` is in its path filter.

Not verified:

- [ ] `metal,accelerate`. It cannot be built on this Linux host, so it is unverified by execution rather than confirmed. The static argument that it is unaffected: `make verify-clippy` runs `--features metal,accelerate` without `--no-default-features`, so `surgery` stays on and both `needless_match` attributes expand to nothing; `mlxcel-xla` builds with no features there, so `iree.rs`, `aux.rs`, `aux_manifest.rs` and `phi4_audio.rs` sit behind `#[cfg(feature = "iree")]` and never compile; and on the `mlxcel` side only `media.rs` and `xla_audio_preprocess.rs` compile, where every change is an `allow` attribute that cannot break a build.

## Follow-ups this surfaced, out of scope here

- `mlxcel-xla`'s `#[cfg(test)]` modules are gated by no CI job at all. There is no `default-members`, so `cargo check` in this job resolves to `-p mlxcel` and `--all-targets` expands within that package only, while the `clippy` job builds default features where `mlxcel-xla` is default-off. Recorded in the workflow's exclusion list.
- Clippy-only lints under these feature sets remain ungated, so that half of the backlog can regrow. Closing it means running clippy in this job, which was left out of scope deliberately.
- `media_tests.rs` reaches `validate_xla_raw_counts_with_audio` only through the `supports_audio = false` wrapper, so the `true` branch has no direct unit coverage, and the sign of that flag is the security-relevant bit.

Closes #1304

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -345,6 +345,15 @@ jobs:
   #     `clippy` job above covers default features only. That half of the
   #     backlog cleared by #1304 can therefore regrow silently; closing it means
   #     running clippy here, which was left out of #1304 deliberately.
+  #   - The path dependencies' own test targets. There is no `default-members`
+  #     (see the note at the top of `Cargo.toml`), so `cargo check` run here
+  #     resolves to `-p mlxcel` and `--all-targets` expands within that package
+  #     only. `mlxcel-core`, `mlxcel-surgery` and `mlxcel-xla` are compiled as
+  #     plain library dependencies, so their `#[cfg(test)]` modules are never
+  #     built. That leaves `mlxcel-xla`'s unit tests ungated by every job,
+  #     because the `clippy` job above builds default features and the crate is
+  #     default-off there. `cargo clippy -p mlxcel-xla --all-targets` covers
+  #     them, but only by hand.
   #   - macOS and `IREE_DIST` builds of the same features. Neither has a runner
   #     with the required distribution.
   xla-compile:
@@ -365,6 +374,15 @@ jobs:
       # dead-code backlog that would have made the job red on arrival; #1304
       # cleared it. See the note above for what `-D warnings` on `cargo check`
       # still does not cover.
+      #
+      # Blast radius, deliberately: `RUSTFLAGS` reaches every unit cargo builds
+      # from a path, so the `mlxcel-core`, `mlxcel-surgery` and `mlxcel-xla`
+      # library units and the build scripts are gated here too, and a warning in
+      # any of them reds a job named for OpenXLA. Registry and git dependencies
+      # cannot, because cargo compiles those with `--cap-lints allow`. The
+      # toolchain is pinned exactly in `rust-toolchain.toml`, so a new rustc
+      # release cannot turn this red on its own; bumping that pin is what has to
+      # re-clear these two feature sets.
       RUSTFLAGS: "-D warnings"
     steps:
       - uses: actions/checkout@v7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,13 +337,14 @@ jobs:
   #     recipe; see its comment for what it covers and what it still does not.
   #   - `cargo test`. This is a compile gate; running the XLA suites needs a
   #     GPU that is not contended with development work on the same host.
-  #   - Full `-D warnings`. The XLA feature combination carries a pre-existing
-  #     dead-code backlog in `mlxcel-xla` and `mlxcel`, and clippy fails on it
-  #     today, so denying every lint would make this job red on arrival and
-  #     therefore ignored. It denies `unused_imports` instead, which is the
-  #     exact lint class of the dead re-export above, and compile errors cover
-  #     the other. Broadening to `-D warnings` needs that backlog cleared
-  #     first.
+  #   - Clippy-only lints. This job runs `cargo check`, and `RUSTFLAGS: "-D
+  #     warnings"` denies rustc lints only, so `dead_code` and `unused_imports`
+  #     are now gated but `collapsible_if`, `needless_match`,
+  #     `too_many_arguments`, `while_let_loop` and the rest of clippy's own
+  #     lints are not gated by anything in CI under these feature sets. The
+  #     `clippy` job above covers default features only. That half of the
+  #     backlog cleared by #1304 can therefore regrow silently; closing it means
+  #     running clippy here, which was left out of #1304 deliberately.
   #   - macOS and `IREE_DIST` builds of the same features. Neither has a runner
   #     with the required distribution.
   xla-compile:
@@ -359,9 +360,12 @@ jobs:
       # `121a`, which is numerically identical here but is not what releases
       # build, so the gate compiles what ships.
       MLX_CUDA_ARCHITECTURES: "121"
-      # See the note above: this denies the lint that let a defect through, not
-      # every lint.
-      RUSTFLAGS: "-D unused_imports"
+      # Every rustc lint, not just the `unused_imports` this job started with.
+      # The narrower value existed because the XLA feature combination carried a
+      # dead-code backlog that would have made the job red on arrival; #1304
+      # cleared it. See the note above for what `-D warnings` on `cargo check`
+      # still does not cover.
+      RUSTFLAGS: "-D warnings"
     steps:
       - uses: actions/checkout@v7
         with:
--- a/TECHNICAL_REPORTS/1381-openxla-warning-backlog-20260822.en.md
+++ b/TECHNICAL_REPORTS/1381-openxla-warning-backlog-20260822.en.md
@@ -1,0 +1,146 @@
+# Technical Report: PR #1381 - clear the OpenXLA warning backlog and deny all warnings in CI
+
+**Date**: 2026-08-22
+**Author**: Jeongkyu Shin
+**Status**: Completed
+**Languages**: Rust, YAML (GitHub Actions)
+**Risk Level**: Low (lint and visibility change; no runtime behavior altered)
+
+---
+
+## Executive Summary
+
+The `xla-compile` CI job denied a single lint, `unused_imports`, rather than all warnings. That was deliberate: the OpenXLA feature combinations carried a dead-code and clippy backlog, and a job that is red the day it lands gets routed around and then ignored.
+
+This PR clears the backlog and widens the gate to `RUSTFLAGS: "-D warnings"`. Fourteen findings across two crates were resolved, one by deletion and the rest by cfg gates, scoped `#[allow]`s with comments naming what keeps each item alive, and three mechanical fixes. Four of the fourteen were not in the issue's list at all; they were found by re-measuring the backlog before starting, because the issue's list had drifted since it was filed.
+
+The change is verified by the widened gate running green on the PR that widens it: `OpenXLA feature compile` did a real 7m44s rebuild under the new flag and passed.
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+PR #1282 added `xla-compile` with `RUSTFLAGS: "-D unused_imports"`. That lint was chosen because it is the exact class of one of the two defects that reached `main` through the OpenXLA coverage gap, a dead `load_weights_from_dir_with_filter` re-export; compile errors covered the other, a non-exhaustive `match` after a new `ModelRequest` variant landed. Both historical breaks were caught. New dead code under these features was not.
+
+### 1.2 Existing Issues
+
+Clearing the backlog is the precondition for widening the gate, and the backlog spanned both `mlxcel-xla` and `mlxcel`. Deletion was usually the wrong fix, because several items are live under feature sets other than the one that reports them dead.
+
+### 1.3 Risk Assessment
+
+Low in isolation, compounding over time. Every ungated warning is a place where the next real defect of the same class arrives unremarked, and the two defects that motivated #1282 are the existing proof.
+
+## 2. Change Summary
+
+Ten files: eight Rust, one workflow, and the report.
+
+| File | Resolution |
+|---|---|
+| `mlxcel-xla/src/aux.rs` | `Float16`, `Uint32`: `cfg_attr(not(micro-oracle), allow(dead_code))` |
+| `mlxcel-xla/src/aux_manifest.rs` | collapsible `if` rewritten as an edition-2024 let chain |
+| `mlxcel-xla/src/iree.rs` | `decode_ragged_logits` cfg-gated to `diagnostics`; `decode_ragged_mrope_logits` deleted; stranded doc comment reattached |
+| `mlxcel-xla/src/phi4_audio.rs` | scoped `allow(clippy::too_many_arguments)` |
+| `src/loading/vlm.rs`, `src/models/sanitize.rs` | `cfg_attr(not(surgery), allow(clippy::needless_match))` |
+| `src/server/batch/xla_audio_preprocess.rs` | seven `cfg_attr(not(test), allow(dead_code))`, each with its caller named |
+| `src/server/batch/xla_worker_admission.rs` | `allow(clippy::while_let_loop)` with reason |
+| `src/server/media.rs` | cfg_attr condition widened to `any(not(xla-iree), not(test))` |
+| `.github/workflows/ci.yml` | `RUSTFLAGS` to `-D warnings`; policy comment replaced; two coverage gaps recorded |
+
+## 3. Technical Decisions
+
+### 3.1 Re-measure the backlog before working from it
+
+The issue's list was measured when it was filed. Re-running both clippy commands on the current tree produced fourteen findings against the issue's ten, and the four extras were not incidental: two were plain mechanical fixes, and two were the `needless_match` pair that is the most dangerous item in the whole change (3.3). Working from the stale list would have shipped a PR that fails its own acceptance criteria.
+
+### 3.2 One deletion, and only after proving the item is dead everywhere
+
+`decode_ragged_mrope_logits` is a thin wrapper over `decode_ragged_mrope_logits_with_modes` and is dead under both measured feature sets. It was deleted rather than allowed, because the acceptance criterion requires every `#[allow]` to name a caller keeping the item alive, and there is none. The justification is a repo-wide search across all file types, extended during review to the paths a plain grep misses: macros (`mlxcel-xla` has one `macro_rules!`, for MLIR fixtures, and no `paste!` or `concat_idents!`), doc tests, `pub use` re-exports, trait impls, and feature-gated callers under `micro-oracle` and `xla-diagnostics-cpu`. What remains is the identically-named C function `xla_llama_decode_ragged_mrope_logits`, its extern declaration, and the `_with_modes` sibling, which keeps its live caller.
+
+Its twin `decode_ragged_logits` was **not** deleted. It is dead under `cuda,xla-iree` and live under `xla-diagnostics`, so it was cfg-gated to `feature = "diagnostics"` instead, matching all four of its callers in `batch.rs`, which already carry that cfg. Since `diagnostics = ["iree"]`, the new gate is strictly narrower than the `#[cfg(feature = "iree")] mod iree` containing it and cannot break any combination. Deleting it would have turned a warning into a build break on the other feature set, which is the failure this issue explicitly warned against.
+
+### 3.3 Two clippy suggestions that had to be refused
+
+`src/loading/vlm.rs` and `src/models/sanitize.rs` both contain:
+
+```rust
+let resolved_transform = match transform {
+    Some(t) => Some(t),
+    None => {
+        #[cfg(feature = "surgery")]
+        { active_pipeline.as_deref().map(...) }
+        #[cfg(not(feature = "surgery"))]
+        ...
+    }
+};
+```
+
+Under `--no-default-features` the `surgery` arm compiles out, the `None` arm collapses to `None`, and clippy correctly reports the match as needless with the suggestion "replace it with `transform`". Taking that suggestion would delete active-pipeline resolution from the default build, where `default = ["surgery"]`.
+
+The fix is `#[cfg_attr(not(feature = "surgery"), allow(clippy::needless_match))]` on the `let` statement: silenced exactly for the builds that see the collapsed shape, and still linted in the default build. This is the concrete instance of the issue's rule that a fix which alters behavior is the wrong fix.
+
+### 3.4 `not(test)` rather than a bare allow
+
+Every suppression in `xla_audio_preprocess.rs` and `media.rs` is conditioned on `not(test)`. `cargo check --all-targets` compiles the `mlxcel` lib twice, once under `cfg(test)`, so the suppression is self-enforcing: if the sole test caller is deleted, the lint returns rather than staying silent forever. Each attribute carries a comment naming the caller, which the review checked one by one.
+
+A detail worth recording: `#[allow]` suppresses a lint without marking the symbol live, so the `healthy` field still warned on its own after `is_healthy` was allowed. Each item needed its own attribute.
+
+### 3.5 Widen `RUSTFLAGS`, and record what it still does not cover
+
+`RUSTFLAGS: "-D warnings"` on a `cargo check` job denies rustc lints only. Clippy's own lints, which are most of what this PR fixed, remain gated by nothing under these feature sets, because the `clippy` job builds default features where `mlxcel-xla` is default-off. Swapping the job's `cargo check` for `cargo clippy` would close that, and was deliberately left out of scope. The gap is recorded in the workflow's exclusion list rather than left implied, so the cleared half of the backlog cannot regrow behind a comment that claims full coverage.
+
+## 4. Review Findings
+
+No CRITICAL or HIGH findings survived either review. The follow-up commit fixed four accuracy defects:
+
+- **Error messages naming functions the build does not contain.** Two arity-check messages in `iree.rs` named the thin wrappers rather than their `_with_modes` emitters. Since one wrapper is now deleted and the other is diagnostics-only, a production `cuda,xla-iree` build could return `"decode_ragged_mrope_logits expects ..."` naming a function it does not have.
+- **An incomplete caller list.** `AudioPreprocessStage::spawn`'s comment named one test file; `xla_worker_tests.rs` calls it too. Since the acceptance criterion is that every allow names what keeps the item alive, an incomplete list is precisely what it exists to prevent.
+- **A comment describing half of what it covered.** `drain_preprocessed` has two drain loops, image and audio, and the function-level allow covers both.
+- **Two unrecorded properties of the widened gate**, added to the exclusion list (see 5.2).
+
+## 5. Validation
+
+### 5.1 What passed
+
+All on the GB10 host with the IREE runtime provisioned, and re-run on the final commit.
+
+| Command | Result |
+|---|---|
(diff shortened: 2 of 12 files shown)


inureyes:

### Intent

> Clear the dead-code and clippy backlog under `cuda,xla-iree` and `xla-diagnostics`, then widen the `xla-compile` job from `-D unused_imports` to `-D warnings`, without changing any runtime behavior.

### Findings Addressed

None requiring a code change. Every Set A and Set B item in #1304 is resolved, each `allow` names a real caller, and the "no runtime behavior change" claim holds line by line.

### Line-by-line verification of the no-behavior-change claim

- `iree.rs` `decode_ragged_logits` gate: all four callers (`batch.rs:1190` in `LlavaReferenceDiagnosticEngine::capture`, `batch.rs:1293` in `run_gemma3n_prefix_decode_diagnostic`, `batch.rs:1472` and `:1493` in `run_gemma3n_canonical_diagnostics`) sit under `#[cfg(feature = "diagnostics")]` items, and `diagnostics = ["iree"]` so the gate can never be wider than the module. Serving reaches the same code through `decode_ragged_logits_with_modes`, which keeps its live caller at `batch.rs:1047`.
- `iree.rs` `decode_ragged_mrope_logits` deletion: a repository-wide search (all file types, excluding `.git` and `target`) returns the removed definition, the `_with_modes` sibling, the `xla_llama_decode_ragged_mrope_logits` extern and its C definition, and one error string. No Rust caller in any crate, test, bench, example, or spike, under any feature combination. `micro-oracle` does not reach it: `numeric_probe.rs` and `numeric_oracle.rs` never name either method. Confirmed empirically with `cargo clippy -p mlxcel-xla --features iree,micro-oracle --all-targets -- -D warnings`, which exits 0 on this branch. Removing the wrapper also strands nothing, since `_with_modes` retains its own caller.
- `aux_manifest.rs` let chain: identical control flow. Both shapes are "read succeeds AND pid parses, then probe `/proc`", both have no `else`, and `&&` in a let chain short-circuits, so `cache_lock_owner_pid` is still only called on a successful read. Neither version had a failure path to preserve; every non-matching case falls through to the mtime check, as before.
- `vlm.rs:333` and `sanitize.rs:1668`: the `cfg_attr` sits on the `let resolved_transform` statement whose initializer is the flagged `match`, which is the node clippy reports under. Both match bodies are untouched, and under the default feature set (`default = ["surgery"]`) the attribute expands to nothing, so the default path is byte-identical.
- `xla_audio_preprocess.rs`: every annotated item has a real caller in `xla_audio_preprocess_tests.rs`. `spawn` (8 sites), `recv` (`:122`, `:201`, `:214`, and more), `metrics` (`:132` and more), `is_healthy` (`:140`, `:312`, `:330`), `snapshot` via `stage.metrics().snapshot()`, `AudioPreprocessMetricsSnapshot` through the fields those assertions read, and the `Cancelled` payload at `:263`. The `healthy` field is genuinely written by the production path (`spawn_with_loader`, `:464`-`:516`) and only read back through `is_healthy`, so the comment is accurate. Serving matches `Cancelled(_)` at `xla_worker_admission.rs:395` and `xla_audio_preprocess.rs:812`, so the payload really is unread outside tests. No `allow` here papers over a genuinely dead item.
- `xla_worker_admission.rs`: the borrow argument checks out. A `while let Some(stage) = self.image_preprocessor.as_ref()` head holds a shared borrow across a body that calls `self.handle_preprocessed` and assigns `self.image_preprocessor = None`.
- `ci.yml` comment: every factual claim verified. The job does run `cargo check` (both steps), `collapsible_if` / `needless_match` / `too_many_arguments` / `while_let_loop` are all clippy-only lints, and the `clippy` job above at line 139 does run `cargo clippy -p mlxcel --lib --tests -- -D warnings` at default features only.

### `metal,accelerate`, reasoned statically

Safe. `make verify-clippy` runs `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings`, without `--no-default-features`, so `surgery` stays on and the two `needless_match` attributes expand to nothing. `mlxcel-xla` builds there with no features, so `iree.rs`, `aux.rs`, `aux_manifest.rs` and `phi4_audio.rs` are all behind `#[cfg(feature = "iree")]` and are not compiled at all. On the `mlxcel` side the only two changed files that compile are `media.rs` and `xla_audio_preprocess.rs`, and both changes are `allow` attributes that cannot break a build. `xla_audio_preprocess.rs` additionally still carries the module-level `#[cfg_attr(not(feature = "xla-iree"), allow(dead_code))]` at `batch/mod.rs:70`, which covers the whole module on that feature set.

The `xla-diagnostics-cpu` trap is avoided by construction: no cfg added on the `mlxcel` side keys on a diagnostics feature at all.

### Remaining Items

- [ ] `iree.rs:3103` error string still reads `decode_ragged_mrope_logits expects ...` inside `decode_ragged_mrope_logits_with_modes`, naming a function that no longer exists (LOW) - deliberate and disclosed in the PR body, and rewording it would touch a user-visible string under a no-behavior-change constraint, so leaving it is defensible.
- [ ] `media.rs:204` condition `any(not(feature = "xla-iree"), not(test))` is wider than it needs to be (LOW) - the `not(feature = "xla-iree")` disjunct is inert, because in the one configuration it adds (`xla-iree` off, `test` on) `media_tests.rs` uses the method unconditionally, so no lint could fire there. `not(test)` alone is equivalent and would match the rewritten doc comment, which now explains only the test-caller rationale.
- [ ] `iree.rs:2296`-`:2305` carries the same stranded-doc-comment defect this PR fixed at `:1819` (LOW, pre-existing) - the `IreeRaggedLlama` doc block is attached to `PreparedPrefillDiagnostics`, so `Self::prefill_slot_logits` and `Self::decode_ragged_logits` are broken intra-doc links. It predates this branch and is not a clippy warning (no blank line separates it), so it is outside the backlog this PR was scoped to.

### Note on the widened gate

`RUSTFLAGS: "-D warnings"` applies to every unit cargo compiles, not only the primary package. Registry dependencies are shielded by cargo's `--cap-lints allow`, so the practical blast radius is the workspace path crates and their build scripts, which is the intent. Changing `RUSTFLAGS` does invalidate the persistent target directory, so the first run after merge is a cold rebuild; the 120-minute timeout leaves ample headroom against the measured cold cost on this runner.

### Verification

- [x] All stated requirements implemented (every Set A and Set B item, plus the four not in the issue body)
- [x] Integrated into project code flow (the gate change is in the job that runs it)
- [x] Project conventions followed (`cfg_attr(not(test), allow(dead_code))` already used in `model_metadata.rs`, `server/startup.rs`, `loading/nonstandard.rs`, `loading/config_backed.rs`)
- [x] No unintended structural changes (10 files, all in scope; `build.rs` and `tests/molmo2_xla_vision_parity.rs` untouched)
- [x] `rustfmt --edition 2024 --check` clean on all nine changed Rust files; `ci.yml` parses as YAML
- [x] `cargo clippy -p mlxcel-xla --features iree,micro-oracle --all-targets -- -D warnings` exits 0

No CRITICAL or HIGH findings.

inureyes:
## Security and maintenance-risk review

No CRITICAL or HIGH findings. The ordinary vulnerability surface is genuinely absent here: no new I/O, no new dependencies, no parsing, no trust boundary moved. The review focused on the three risks a suppression PR actually carries.

### 1. Do the suppressions hide a real defect? No.

Every `allow` was checked against its named caller, and every claim holds.

| Item | Claimed live in | Verified |
|---|---|---|
| `AudioPreprocessStage::spawn` | tests | `xla_audio_preprocess_tests.rs:106,166,241` and `xla_worker_tests.rs:247` |
| `recv` / `metrics` / `snapshot` / `is_healthy` | tests | all four called in `xla_audio_preprocess_tests.rs` |
| `AudioPreprocessMetricsSnapshot` | `snapshot()` | yes |
| `healthy` field | worker liveness | written at `spawn_with_loader` 476/483/496 |
| `AuxiliaryWeightDType::Float16` / `Uint32` | `micro-oracle` | `numeric_probe.rs:693,699,705`; `lib.rs:60-61` gates the module on that feature |
| `validate_xla_raw_counts` | `media_tests.rs` | 5 call sites |

The `not(test)` pattern is self-enforcing rather than a permanent silencer: `cargo check --all-targets` compiles the `mlxcel` lib twice, once with `cfg(test)`, so deleting the sole test caller re-reds this job instead of going quiet.

**On the `validate_xla_raw_counts` asymmetry specifically: it is not a latent bug.** The wrapper is `_with_audio(images, audio, videos, false)` verbatim, and admission passes a computed flag (`self.audio_preprocessor.is_some()`, `xla_worker_admission.rs:57-61`), never a hardcoded `false`. Every check in the shared body runs in production; the dead item is a test-facing alias, not a skipped validator. One follow-up worth filing separately: `media_tests.rs` reaches `_with_audio` only through the `false` wrapper, so the `supports_audio = true` branch has no direct unit coverage, and the sign of that flag is the security-relevant bit.

**On `healthy`:** production never reads it, but it is not a missed-failure path. Worker death is detected through `TryRecvError::Disconnected` in `drain_preprocessed`, which clears the stage and errors every pending request. The flag is genuinely redundant in serving, so the allow is the right call.

**On `Cancelled(AudioPreprocessCheckpoint)`:** serving matches `Cancelled(_)` and `observability.audio_preprocess_cancelled` is a single unlabelled counter, so the checkpoint distinction is computed and discarded in production. Keeping the payload is defensible as argued; a checkpoint label on that counter would be a reasonable later improvement.

### 2. The deletion is safe.

`decode_ragged_mrope_logits` has no reachable caller. Repo-wide grep excluding `.git` and `target`, filtered of the `_with_modes` sibling and the `xla_llama_` extern, leaves exactly zero references after this change. Checked for the paths a plain grep misses: `mlxcel-xla` has one `macro_rules!` (`validation.rs:499`, MLIR fixtures) and no `paste!` or `concat_idents!`, so no generated call site; no doc test, no `pub use`, no trait impl, and the C symbol plus the `_with_modes` Rust sibling both remain live via `batch.rs:1052`.

### 3. The cfg gate is correct.

All four `decode_ragged_logits` callers are inside `#[cfg(feature = "diagnostics")]` items: `LlavaReferenceDiagnosticEngine` (`batch.rs:1134`), `run_gemma3n_prefix_decode_diagnostic` (`:1262`), and `run_gemma3n_canonical_diagnostics` (`:1322`). `diagnostics = ["iree"]`, so the new gate is strictly narrower than the module's existing `iree` gate and cannot break a combination. No non-diagnostics caller exists.

### 4. CI maintenance risk from `-D warnings`

Not brittle against compiler churn. `rust-toolchain.toml` pins `channel = "1.97.1"` exactly rather than `stable`, and the job resolves cargo through the rustup shim with no toolchain-install step, so the pin applies. A new rustc release cannot red this job on its own. The cost moves to the pin bump, which must now also re-clear these two feature sets.

Two things worth recording rather than assuming, which the follow-up commit adds to the workflow:

- **Blast radius.** `RUSTFLAGS` reaches every path-built unit, so `mlxcel-core`, `mlxcel-surgery`, `mlxcel-xla` and the build scripts are gated here too, and a cuda-only warning in `mlxcel-core` now reds a job named for OpenXLA. Registry and git dependencies cannot, because cargo compiles those with `--cap-lints allow`.
- **A coverage gap the exclusion list did not have.** There is no `default-members`, so `cargo check` here resolves to `-p mlxcel` and `--all-targets` expands within that package only. The path dependencies are built as plain library units, so `mlxcel-xla`'s `#[cfg(test)]` modules are gated by no job at all: the `clippy` job builds default features, where the crate is default-off. The test plan above ran `-p mlxcel-xla --all-targets` by hand, which CI will not.

One-time cache effect only: `RUSTFLAGS` is part of cargo's fingerprint, so the persistent target dir was invalidated once (the observed 7m44s rebuild) and returns to the ~1 min warm figure the workflow already documents.

### Fixes applied (89c0a5d3)

All LOW, comment and message accuracy only.

1. `iree.rs`: both arity-check messages named the thin wrappers instead of their `_with_modes` emitters, so a production build could return a message naming a function it does not contain.
2. `xla_audio_preprocess.rs`: `spawn`'s comment named one test file; `xla_worker_tests.rs` calls it too. An incomplete caller list is the specific thing #1304's acceptance criterion exists to prevent.
3. `xla_worker_admission.rs`: `drain_preprocessed` has two drain loops and the allow covers both; the comment described only the image one.
4. `ci.yml`: the two items above.

Validated with `cargo clippy -p mlxcel-xla --lib --features iree --all-targets -- -D warnings` under `RUSTFLAGS=-D warnings` (clean), `rustfmt --edition 2024 --check` on all three Rust files, and a YAML parse of the workflow. `metal,accelerate` remains unverified; it cannot be built on this Linux host.

Issue #1436, opened 2026-08-26
feat(inference): complete b10621 sampling and grammar semantics


inureyes:
## Parent

- #1431
- Related: #1377

## Problem

Beyond the low-risk aliases in #1430, b10621 exposes sampling algorithms and controls that mlxcel does not implement or maps differently: `typical_p`, top-n-sigma, dynamic temperature, Mirostat, adaptive-p, XTC edge semantics, sampler ordering, ignore-EOS, grammar/lazy-grammar triggers, preserved tokens, multiple completions, and native cache/progress fields.

## Scope

- Use #1377 for `typical_p` and add the remaining algorithms as independently testable sampler stages.
- Match b10621 value domains, disable sentinels, defaults, and ordering semantics, including XTC probability/threshold behavior.
- Wire `repeat_last_n` / `repetition_context_size` into the shared repetition, frequency, and presence penalty history; the field is currently parsed but not applied to the core sampler, and #1430 intentionally stops before this arithmetic change.
- Correct `dry_penalty_last_n` sentinels: b10621 uses `0` to disable DRY history penalties, while mlxcel currently resolves `0` as full history; preserve an explicit full-history form without conflating it with upstream disablement.
- Add explicit sampler-sequence configuration and validate incompatible/repeated stages.
- Implement ignore-EOS, `n` completions, time/token limits, keep/discard/cache controls, and response-field selection where supported.
- Implement direct GBNF and lazy grammar triggers/preserved-token behavior or reject unsupported grammar forms before generation.
- Keep OpenAI-shaped and native request aliases synchronized without silently dropping recognized fields.

## Acceptance criteria

- [x] A field-by-field schema test covers all b10621 sampling request fields, aliases, defaults, and disable sentinels: the pinned manifest plus `tests/llama_compat_manifest.rs` and `src/server/llama_compat_tests.rs` hold every entry's state, spelling, default, and declaration in both directions; every entry this issue owned left `deferred` (26 `supported`, 3 `not_applicable` with tested rejection, the remainder split to #1485 with the shard owner set extended).
[...]

PR #1454, opened 2026-08-26
fix: align llama-server b10621 compatibility

### LINKED ISSUE #1430: fix(server): align llama-server b10621 aliases and defaults
## Background

The official llama.cpp nightly pointer resolved to `b10621` on 2026-08-26. The macOS arm64 distribution reports build 10621 at commit `c1d0e7a00`; its archive SHA-256 is `429c8270608600188035e5e92f7d78dffb7900904fe7dd7e6a84f48068cd13cf`. Comparing that binary's actual `llama-server --help` and route registration against mlxcel `main` found several low-risk compatibility defects that do not require new inference features.

## Problems

- Canonical environment variables have drifted: upstream uses `LLAMA_ARG_BATCH`, `LLAMA_ARG_UBATCH`, `LLAMA_ARG_SPEC_DRAFT_MODEL`, `LLAMA_ARG_THINK_BUDGET`, `LLAMA_ARG_LOG_FILE`, `LLAMA_ARG_CHAT_TEMPLATE`, `LLAMA_ARG_CHAT_TEMPLATE_FILE`, `LLAMA_ARG_ENDPOINT_METRICS`, `LLAMA_ARG_ENDPOINT_PROPS`, and `LLAMA_ARG_ENDPOINT_SLOTS`; mlxcel either uses stale names or has no env binding.
- `LLAMA_ARG_CACHE_REUSE` is an integer minimum reuse chunk size upstream, but mlxcel currently parses it as a boolean alias for enabling the whole prompt cache. Values such as `0` therefore produce the opposite operational result.
- Upstream b10621 defaults to `top_p=0.95`, `min_p=0.05`, and `dry_penalty_last_n=64`; mlxcel currently defaults to `0.9`, `0.1`, and an all-history sentinel.
- The same request fields are spelled differently: llama-server accepts `n_predict` / `max_completion_tokens`, `repeat_penalty`, `repeat_last_n`, and `reasoning_budget_tokens`; mlxcel's OpenAI-shaped routes currently accept only their native spellings and silently ignore the llama-server names.
- llama-server accepts either one stop string or an array and accepts seed `-1` as random; mlxcel currently requires an array and an unsigned seed.
- Safe route aliases are missing: `GET /v1/health`, `POST /reranking`, and `POST /v1/reranking` can use existing handlers without adding new inference behavior.
- `--temperature` is a llama-server alias for `--temp` but is not accepted by mlxcel-server.

## Scope

- Add canonical b10621 spellings while retaining documented mlxcel legacy spellings as lower-precedence compatibility aliases where doing so is unambiguous.
- Stop interpreting `LLAMA_ARG_CACHE_REUSE` as a boolean prompt-cache enable switch. Accept `0` as the upstream no-extra-reuse default without disabling prompt caching, and fail startup clearly for unsupported positive values rather than silently changing a different setting.
- Align the three server sampling defaults and `/props` reporting with b10621.
- Add request deserialization aliases, one-or-many stop parsing, and negative-seed normalization without changing the inference algorithms.
- Map `repeat_last_n` to the existing `repetition_context_size` request field only. Applying that field as a finite penalty-history window changes arithmetic and is explicitly deferred to #1436 with the numerical validation required by `docs/benchmarks.md`.
- Add the three route aliases and include `/v1/health` in the unauthenticated health allowlist.
- Update CLI consistency tests, request/route tests, environment-variable documentation, and the changelog.

## Out of scope

- New sampler algorithms, TLS/CORS transport features, router mode, LoRA hot-swap, native completion response expansion, slot persistence, or support for low-level GGML device flags.
- Changing `--timeout`, whose mlxcel meaning is currently a decode hang watchdog rather than llama-server's HTTP read/write timeout; that requires a separate migration issue.
- Pretending unsupported hidden flags work by accepting and ignoring values that have observable llama-server semantics.

## Acceptance criteria

- [ ] Both `mlxcel serve` and `mlxcel-server` accept canonical environment variable names with CLI-over-env precedence and preserve explicitly documented legacy aliases at lower precedence.
- [ ] `LLAMA_ARG_CACHE_REUSE=0` leaves prompt caching enabled, positive values fail with an actionable unsupported-setting error, and boolean values are no longer accepted as cache-enable aliases.
- [ ] Server defaults and `/props` report `top_p=0.95`, `min_p=0.05`, and `dry_penalty_last_n=64`.
- [ ] Chat and OpenAI completion requests deserialize the listed llama-server field aliases into their existing mlxcel request fields; a scalar stop string is equivalent to a one-element array, seed `-1` resolves to nondeterministic sampling, and seeds below `-1` are rejected.
- [ ] This issue does not make the parsed repetition window affect sampling; that arithmetic change remains tracked by #1436 and is not claimed as runtime parity here.
- [ ] `GET /v1/health`, `POST /reranking`, and `POST /v1/reranking` reach the existing handlers, with `/v1/health` public under API-key authentication.
- [ ] `--temperature` and `--temp` resolve identically on both server entry points.
- [ ] Focused CLI, request-schema, route, auth, and config-default tests pass; `cargo fmt --check` passes.

## Reference

- Upstream release: https://github.com/ggml-org/llama.cpp/releases/tag/b10621
- Upstream argument definitions: https://github.com/ggml-org/llama.cpp/blob/c1d0e7a004015f23bc0233470b747b596f29b264/common/arg.cpp
- Upstream route registration: https://github.com/ggml-org/llama.cpp/blob/c1d0e7a004015f23bc0233470b747b596f29b264/tools/server/server.cpp

inureyes:
## Summary

- Accept llama-server b10621 canonical environment variables on both server entry points while retaining the documented legacy spellings at lower precedence.
- Treat `LLAMA_ARG_CACHE_REUSE` as an integer reuse-chunk setting: `0` is a no-op, while booleans and unsupported positive values fail startup with actionable errors.
- Align the server defaults for `top_p`, `min_p`, and `dry_penalty_last_n`, including the `/props` configuration source.
- Accept the compatible request spellings, scalar or array stop sequences, and exactly `seed: -1` as nondeterministic sampling; values below `-1` are rejected.
- Mount `/v1/health`, `/reranking`, and `/v1/reranking`, keep every health alias public under API-key authentication, and accept `--temperature` on both binaries.
- Accept the exact `f16` split-cache alias for FP16 while continuing to reject GGML quantizer names such as `q8_0` and `q4_0` rather than translating them to different arithmetic.

## Scope

This PR maps `repeat_last_n` to the pre-existing OpenAI-shaped request field but intentionally does not change sampler arithmetic; runtime penalty-window parity and its numerical validation remain owned by #1436.

## Validation

- `cargo test --release server::types::request::tests::` — 25 passed.
- `cargo test --release --bin mlxcel serve_canonical_llama_envs_and_endpoint_precedence_parse_together` — 1 passed.
- `cargo test --release --bin mlxcel-server canonical_llama_envs_and_endpoint_precedence_parse_together` — 1 passed.
- `cargo test --release --lib cli::turbo_args::tests::split_cache_flags` — 2 passed.
- `cargo test --release --lib prompt_cache_llama_arg_cache_reuse` — 3 passed.
- `cargo test --release --lib prompt_cache_mlxcel_env_does_not_hide_invalid_cache_reuse` — 1 passed.
- `cargo test --release --lib server_default_env_fallback_all_cases` — 1 passed.
- `cargo test --release --lib server_config_default_matches_llama_server_compatibility_defaults` — 1 passed.
- `cargo test --release --lib build_server_config_prompt_cache_default_is_enabled` — 1 passed.
- `cargo test --release --lib all_rerank_aliases_are_mounted` — 1 passed.
- `cargo test --release --lib v1_health_alias_is_public_when_api_key_auth_is_enabled` — 1 passed.
- `cargo test --release --lib all_health_aliases_are_unauthenticated` — 1 passed.
- `cargo test --release --bin mlxcel serve_temp_and_temperature_aliases_resolve_identically` — 1 passed.
- `cargo test --release --bin mlxcel-server temp_and_temperature_aliases_resolve_identically` — 1 passed.
- `cargo test --release --test cli_help_consistency turbo_flags_and_modes` — 4 passed.
- `cargo fmt --check` — passed.

Closes #1430


inureyes:
## Independent real-checkpoint validation

I validated the sampling-default change on Apple GPU (Metal) with `mlx-community/Qwen3-0.6B-4bit`, prompt `Explain why the sky is blue in three concise sentences.`, 128 generated tokens, temperature `0.8`, top-k `40`, and seed `20260826`.

- Previous server defaults (`top_p=0.9`, `min_p=0.1`): finite 128-token reasoning continuation at 491.07 tok/s on the repeated visible run.
- b10621 defaults (`top_p=0.95`, `min_p=0.05`): finite 128-token reasoning continuation at 477.80 tok/s.
- The runs shared the prefix through `I know it's` and then diverged as expected: the previous defaults sampled `due to Rayleigh scattering`, while the b10621 defaults sampled `because of Rayleigh scattering` and continued on a different path.

This confirms that the default change is behaviorally effective and reproducible with a real checkpoint rather than merely changing `/props`. DRY remained disabled in both arms, so the `dry_penalty_last_n=64` default did not affect this sample.

The PR does not modify model-forward, quantized-kernel, or sampler-algorithm code. I nevertheless attempted to build `examples/logit_trace` for a teacher-forced forward check; the local Xcode installation cannot execute its Metal compiler component and returned `cannot execute tool 'metal' due to missing Metal Toolchain`. The already-built release runtime can execute Metal inference, which produced the measurements above, but a new Metal example binary cannot be linked on this host until that Xcode component is installed.

PR #592, opened 2026-07-01
feat(models): port DeepSeek-V4 backbone (MLA + group-limited MoE)

inureyes:
## Summary

Adds DeepSeek-V4 support by porting the **backbone only**. `DeepSeekV4Model` reuses the DeepSeek-V3 stack (MLA attention, group-limited sigmoid-routed MoE, unclamped shared-expert `DenseMLP`, and V3's multi-token-prediction layer handling) as a thin wrapper: `DeepSeekV4Config` is parsed independently and mapped onto `DeepSeekV3Config` via `to_dsv3_config`, mirroring the existing `glm_moe_dsa` wrapper over DeepSeek V3.2. Config defaults track the mlx-vlm `deepseek_v4` reference.

A `deepseek_v4` checkpoint previously errored with "Unsupported model type"; it now detects, loads, and runs on the V3-equivalent backbone.

### Scope

Backbone only, exactly as the issue directs. The two V4-specific features are deferred to separate follow-up issues and are intentionally **not** implemented here:

- HiSA (hierarchical sparse attention)
- the shared-expert SwiGLU `swiglu_limit` clamp

Their config fields (`head_dim`, `swiglu_limit`, `num_nextn_predict_layers`) are parsed for forward-compatibility but not applied by the backbone.

## What changed

- `src/models/deepseek_v4.rs` (new): `DeepSeekV4Config` + `DeepSeekV4Config::to_dsv3_config` + `DeepSeekV4Model` (wraps `DeepSeekV3Model`).
- `src/models/deepseek_v4_tests.rs` (new): config default and V3-mapping unit tests.
- `tests/deepseek_v4_parity.rs` (new): reference-config parity + a synthetic forward-equivalence proof that the V4 backbone is bit-identical to DeepSeek-V3.
- Registration: `ModelType::DeepSeekV4` and the `deepseek_v4` `detection.rs` arm; `LoadedModel::DeepSeekV4` + delegation; `model_metadata.rs` config-backed entry; `tensor_parallel/inference.rs` fallback arch string; `ALL_MODEL_TYPES`, `metadata()`, and the `all_variants!` registry list in `mod.rs`.
- `docs/supported-models.md`: DeepSeek-V4 backbone row.

## Test plan

- [x] `cargo fmt`
- [x] `cargo check --lib --tests`
- [x] `cargo clippy --lib --tests -- -D warnings`
- [x] `cargo test --lib deepseek_v4` (5 passed)
[...]

Issue #550, opened 2026-07-01
feat(deepseek-v4): clamp shared-expert SwiGLU to swiglu_limit


inureyes:
## Summary

Every DeepSeek-V4 feed-forward block is MoE: 256 routed experts plus one shared expert, all SwiGLU MLPs. The model clamps the SwiGLU pre-activations in both expert kinds before the nonlinearity: the gate pre-activation is clamped from above at `swiglu_limit`, the up pre-activation is clamped to `[-swiglu_limit, +swiglu_limit]`, then `silu(gate) * up` proceeds. Omitting the clamp is silent while every pre-activation stays inside the bounds and a real divergence the moment one crosses them.

**This issue was filed on 2026-07-01 against a DeepSeek-V4 module that did not exist yet. It anticipated its own resolution: "If #523 lands with the clamp already in place, this issue reduces to verifying both paths clamp identically and adding the regression tests below." That is what happened.** #523 landed as PR #1455 with the clamp in place in both paths and the regression test written. This body records the verification so the issue can be closed on evidence rather than assumption.

## Current state: implemented by #523 / PR #1455

The clamp is one shared function, `limited_swiglu` in `src/models/deepseek_v4_moe.rs:65`:

```rust
let gate = mlxcel_core::minimum(gate, &limit_arr);      // upper clamp only
let up   = mlxcel_core::clip(up, &neg_limit, &limit_arr); // both sides
// then silu(gate) * up
```

This matches the specified math exactly: upper clamp only on the gate (silu already saturates toward 0 for large negative inputs), both sides on the up branch (it multiplies through linearly), applied to pre-activations before silu, elementwise, with `limit <= 0` disabling the clamp.

**Both expert kinds call the same function**, which is what this issue asked to verify:

- Shared expert: `LimitedMlp`, held at `deepseek_v4_moe.rs:307` as `shared_experts`, constructed at line 361 with `limit: args.swiglu_limit` (line 359).
- Routed experts: `LimitedSwitchGlu`, constructed with `limit: args.swiglu_limit` (line 290).

`swiglu_limit` is read from the checkpoint config with a serde default of 10.0, and DeepSeek-V4-Flash declares 10.0.

### The fused-kernel concern is structurally resolved

[...]

Issue #549, opened 2026-07-01
feat(deepseek-v4): implement HiSA hierarchical sparse attention


inureyes:
## Summary

DeepSeek-V4 attention runs over two KV streams per layer: a sliding-window local stream at full token resolution, and (on layers with a nonzero compress ratio) a pooled stream in which a compressor summarizes every `r` consecutive tokens into one pooled token. On the `r == 4` layers the pooled prefix grows without bound, so a per-layer indexer selects the `index_topk` most relevant pooled positions per query and attention runs over the local window plus only the selected pooled tokens. HiSA (hierarchical indexing for sparse attention) is the selection algorithm: a coarse block-level filter followed by a fine within-block top-k, replacing a flat O(Np) scan.

**This issue was filed on 2026-07-01 against a DeepSeek-V4 module that did not exist yet, and its original plan targeted the layout #523 was expected to introduce. #523 has since landed (PR #1455) and implemented the indexer end to end.** The body below is rewritten against the code as it actually exists. What remains is validation and operability work, not implementation.

## Current state: implemented by #523 / PR #1455

`src/models/deepseek_v4_indexer.rs` implements the full indexer:

- `Indexer` with `wq_b`, `weights_proj`, a private `Compressor` instance at `index_head_dim`, and `n_heads` / `head_dim` / `index_topk` / `index_block` / `index_keep` / `scale`.
- All three selection paths as separate, unit-testable functions: `flat_select` (line 174), `hisa_select_decode` (line 209), `hisa_select_batched` (line 263).
- The dispatch gate (line 153) computes `index_block * index_keep` in `i64` specifically so a large `index_block` cannot wrap an `i32` and let the gate pass while `nb = np / index_block` is 0.
- `index_block` (64) and `index_keep` (16) are serde defaults, absent from the real checkpoint config, as this issue specified.
- The three-way attention branch (local only / dense pooled / sparse) and the split-softmax sparse combine sharing one log-normalizer with the per-head sinks live in `src/models/deepseek_v4_attention.rs`.

Unit tests in `src/models/deepseek_v4_tests.rs`:

- `hisa_selection_agrees_with_flat_fallback` (line 1152): hierarchy equivalence against the flat scan on tie-free inputs.
- `indexer_forward_dispatches_and_respects_visibility` (line 1204): path dispatch and causal visibility.
- `pool_visibility_counts_match_reference_make_mask` (line 888): the `V(t) = min(Np, floor((t + 1) / r))` visibility count.

[...]

PR #1455, opened 2026-08-26
feat(models): port DeepSeek-V4 as a net-new architecture (#523)

### LINKED ISSUE #523: feat(models): port DeepSeek-V4 (MLA + group-limited MoE backbone)
## Summary
Add DeepSeek-V4 support. mlxcel has DeepSeek V2 / V3 / V3.2 but not V4.

## Scope correction (2026-07)
An initial port attempt (PR #592) assumed "V4 = a V3 backbone + HiSA + a shared-expert `swiglu_limit` clamp" and implemented `DeepSeekV4Model` as a thin wrapper over `DeepSeekV3Model`. That premise is incorrect. Verified against the in-tree reference `references/mlx-vlm/mlx_vlm/models/deepseek_v4/` and the real `DeepSeek-V4-Flash` checkpoint, DeepSeek-V4 is a genuinely novel architecture, not a V3 variant. A V3 wrapper does not load or run a real V4 checkpoint, so PR #592 was closed without merging and this issue is re-scoped to a faithful port of the real architecture.

## Real V4 architecture (port target)
Port `references/mlx-vlm/mlx_vlm/models/deepseek_v4/` faithfully. The distinguishing components are:
- HyperConnections in place of plain residual connections.
- A pooled-KV `Compressor` (not V3-style MLA).
- HiSA hierarchical sparse attention via an `Indexer`.
- Hash-routed MoE with `sqrtsoftplus` gating (not V3 group-limited softmax routing).
- A `MultiLinear` output projection.

## Scope
- Add `ModelType::DeepSeekV4` + a `detection.rs` arm for `deepseek_v4` backed by a real V4 backbone (not a V3 alias).
- Implement the components above as net-new modules, reusing V3 pieces only where genuinely shared.
- Register across the arch/metadata tables and `docs/supported-models.md`; add unit + reference-parity tests.
- Validate on the real `DeepSeek-V4-Flash` checkpoint (available locally at `/home/inureyes/models/deepseek-v4-flash-4bit`).

## Effort
High: net-new, multi-component architecture. Consider splitting HyperConnections, the pooled-KV Compressor, the HiSA Indexer, and hash-routed MoE into dedicated sub-issues.

## Notes
The rejected V3-wrapper attempt is preserved in the closed PR #592 for reference only.

inureyes:
## Summary

Port DeepSeek-V4 (DeepSeek-V4-Flash) as a genuinely new architecture from the in-tree reference `references/mlx-vlm/mlx_vlm/models/deepseek_v4/`, replacing the rejected V3-wrapper premise of PR #592. A V3 wrapper cannot load or run a real V4 checkpoint: V4 carries a rank-4 `[B, L, hc_mult, D]` hidden state through Sinkhorn-normalised HyperConnections, a pooled-KV Compressor over a single shared 512-wide KV head in a 128-token rotating window instead of MLA, HiSA hierarchical sparse selection, hash-routed MoE with `sqrtsoftplus` gating, and a grouped MultiLinear output projection with an inverse RoPE on the attention output.

## What changed

- `src/models/deepseek_v4.rs`: config parsing (including the real checkpoint's duplicate `quantization`/`quantization_config` keys and its 44-for-43 `compress_ratios` list), validation of every allocation-sizing scalar and the per-path quantization overrides, the block/model assembly, model-owned heterogeneous sequence state (gemma3/afmoe pattern), and the `LanguageModel` impl.
- `src/models/deepseek_v4_hyper.rs`: HyperConnection / HyperHead / `hc_expand`, the pure-ops Sinkhorn path in float32 (softmax + eps, initial column normalisation, then `hc_sinkhorn_iters - 1` row/column rounds).
- `src/models/deepseek_v4_rope.rs`: full-head proportional RoPE with inf-padded frequency tables, Yarn correction, inverse tables for the output un-rotation, and `freq_scale` division for pooled positions.
- `src/models/deepseek_v4_compress.rs`: `PoolingCache` (prompt and decode branches, remainder buffer) plus the `Compressor` with the NON-interchangeable simple (ratio 128) and overlap (ratio 4, doubled feature width, one-window-back shift) pooling functions, and the additive pooled-visibility mask helpers.
- `src/models/deepseek_v4_indexer.rs`: the HiSA `Indexer` with its own compressor: decode fast path, batched causality-honouring tiled path, and the flat fallback the hierarchy must agree with.
- `src/models/deepseek_v4_attention.rs`: one attention struct across the three `compress_ratios` kinds (`v4_attention_factory` semantics), single shared KV head, weightless Q RMSNorm, per-head sinks, inverse output RoPE, grouped `wo_a` MultiLinear + `wo_b`, and the split-softmax `_sparse_pooled_attention` sharing one log-normalizer with the sinks.
- `src/models/deepseek_v4_moe.rs`: `tid2eid` hash routing for the first `num_hash_layers` layers (indices from token ids, weights from logits, `input_ids` threaded through every block), selection on bias-corrected scores with weights gathered from unbiased ones, `sqrtsoftplus` scoring in float32, and the limited SwiGLU clamp in both routed and shared experts; routed experts load through the shared `switch_layers::SwitchLinear` with the checkpoint's per-path mxfp4/group-32 overrides under an affine/64 top level.
- `src/models/deepseek_v4_sanitize.rs`: both tensor-name planes (mlx-community and the legacy reference plane with `hc_*` renames, `w1/w2/w3`, per-expert stacking, uint8 `view(uint32)` fixups), `mtp.*` and beyond-layer drops, `tid2eid` int32 cast, 2-D `wo_a` reshape, and a strict weight-coverage check that fails the load naming any tensor that maps onto no module path or any required tensor that is missing.
- Registration: `ModelType::DeepSeekV4` in `src/models/mod.rs`, `detection.rs` arm, `LoadedModel` variant via `delegate_language_model!`, `src/model_metadata.rs` config-backed row, tensor-parallel fallback table entry.
- `docs/supported-models.md`: house-style `deepseek_v4` entry; the v1/v2/v3/v3.2 paragraph now states V4 is not a fifth member of that implementation family. `CHANGELOG.md`: an `[Unreleased]` entry for the family addition.
- Tests: 39 unit tests in `src/models/deepseek_v4_tests.rs` with float64 fixture replays of the reference math, a detection test, and `tests/deepseek_v4_real_model.rs` (a non-ignored real-config parse preflight plus three ignore-gated real-checkpoint gates).

## Hardening applied during review

Review closed several boundaries where untrusted checkpoint or config data reached MLX unguarded. Each is now a load-time `Result` rather than an uncatchable abort at first forward, and each is covered by a test.

- `tid2eid` contents were never range checked, only its shape. V4 is the only routing path in the tree whose expert indices come from a checkpoint tensor rather than an `argpartition` over the score row, so the in-range-by-construction argument that covers `bailing_moe` / `afmoe` / `klear` does not reach it. Those indices feed `take_along_axis` and `gather_qmm`, and MLX gathers fold negative indices but never clamp, so an out-of-range expert id in a downloaded checkpoint was an out-of-bounds read.
- `PoolingCache::update_and_fetch` deep-copied the entire pooled buffer on every call, including the no-op branch that dominates decode (3 of every 4 steps at ratio 4, 127 of 128 at ratio 128), roughly 54 MB per decoded token at 8k context. It now returns a borrow, matching the reference.
- The indexer's pooling cache accumulated an unevaluated MLX graph across decode steps, pinning each step's hidden state, because its only consumer is discarded on two of the three sparse branches. `eval_state` after the layer loop mirrors `KVCache::eval_state`.
- `e_score_correction_bias` had no shape check, the 2-D `wo_a` reshape ran unguarded on checkpoint data, architecture scalars were bounded below but not above, and `index_block` / `index_keep` were unvalidated while being multiplied in `i32`.

## Test plan

- [x] `cargo test --lib models::deepseek_v4`: 39/39 passed, including float64 fixture replays of the reference math (Sinkhorn gates, `hc_expand`, yarn tables, inf-prefix/inverse/freq-scale rope composition, both compress functions, limited SwiGLU, `sqrtsoftplus`), HiSA hierarchical-vs-flat agreement on tie-free inputs, the biased-selection/unbiased-weighting gate contract, hash routing via `tid2eid`, hostile config and quantization-override rejection with positive controls, legacy-plane sanitize, strict coverage rejection, load-time shape checks, the tiny three-kind end-to-end model, and window-aligned chunked-prefill parity.
- [x] `cargo test --lib models::deepseek`: 68/68 across the whole DeepSeek family (v1/v2/v3/v3.2/v4), no regression. `models::detection` 43/43 including the new arm; `models::switch_layers` 23/23; `loaded_model` 5/5.
- [x] `cargo clippy --lib --tests -- -D warnings`: exit 0. `cargo fmt --check`: clean.
- [x] Coverage preflight against the real index: the expected-tensor generator cross-checked against all 2481 tensors of `model.safetensors.index.json`, zero missing and zero unknown after ratio truncation. All 641 quantized planes satisfy `packed_in * 32 == bits * num_groups * group_size`.
- [x] **Real-checkpoint load and generation: all three `#[ignore]` gates pass** against `mlx-community/DeepSeek-V4-Flash-4bit` (43 layers, 2481 tensors, ~151 GB on disk) on an M3 Ultra / 512 GB, in 80s. The checkpoint was verified complete before the run two independent ways: every file matching the upstream declared byte count, and every shard self-consistent against its own safetensors header. Results are in https://github.com/lablup/mlxcel/pull/1455#issuecomment-5424590716 and were re-run unchanged after the hardening commits above, producing byte-identical output.

```
[deepseek-v4] long-context prompt: 2234 tokens
[deepseek-v4] long-context answer: " The Pacific Ocean is the largest ocean on Earth. The Pacific"
test deepseek_v4_real_model_long_context_hits_sparse_and_compressed_paths ... ok
[deepseek-v4] prompt: "The capital of France is"
[deepseek-v4] greedy continuation (24 tokens): " Paris. The capital of France is Paris. ..."
test deepseek_v4_real_model_loads_and_generates_coherently ... ok
test deepseek_v4_real_model_decode_crosses_pooling_windows ... ok

test result: ok. 3 passed; 0 failed
```

The long-context gate is the load-bearing one: at 2234 tokens a ratio-4 layer's pooled count passes `index_topk` (512), so `SparseCompressedAttention` takes the HiSA selection and split-softmax branch rather than the dense concat fallback, and it still has to retrieve the right fact from early in the context. The repetition in the short greedy continuation is ordinary temperature-0 decoding on a base checkpoint, not a porting artifact.

## Follow-ups (out of scope here, as scoped in the issue)

- The reference's fused HC Sinkhorn+collapse Metal kernel (`_hc_sinkhorn_collapse_kernel`); this port implements the pure-ops correctness baseline the reference itself uses off-Metal.
- MTP drafting (`num_nextn_predict_layers`; `mtp.*` tensors are dropped at load).
- Tensor-parallel / pipeline sharding (the reference's `shard()`), VLM wrapping, and a fused-MoE decode kernel with a clamp stage (`MLXCEL_FUSED_MOE` stays unfused for this family).
- No `forward_last_logits` override, so a prefill chunk materialises full `[1, 2048, 129280]` logits to sample one row. Deferred deliberately rather than fixed alongside a first-time architecture port.
- `sparse_pooled_attention` promotes to f32 where the reference stays in the activation dtype, roughly 2x on the largest intermediates. Left because changing it moves numerics.
- Issues #549 (HiSA) and #550 (shared-expert `swiglu_limit` clamp) were filed under the old, rejected V3-wrapper scope; this port subsumes both (HiSA in `deepseek_v4_indexer.rs`, the clamp in `deepseek_v4_moe.rs`). Left open for a human to close.

Closes #523


inureyes:

### Intent

> Port DeepSeek-V4 as a genuinely new architecture (rank-4 HyperConnection state, pooled-KV Compressor, HiSA indexer, hash-routed sqrtsoftplus MoE, grouped MultiLinear output), replacing the rejected V3-wrapper premise of PR #592, and register it end to end.

### Findings Addressed

- [x] `cargo clippy --lib --tests -- -D warnings` failed on this branch: `clippy::manual_is_multiple_of` at `tests/deepseek_v4_real_model.rs:164` (`if i % 8 == 0`) failed the whole `deepseek_v4_real_model` test target. The implementer's "clippy zero warnings" result was true when it ran but went stale when c4847ccc added the long-context gate. Fixed in 1c8b07a2. Verified by reintroducing the old expression, confirming clippy rejects it, restoring the fix, and confirming `cargo clippy --lib --tests -- -D warnings` exits 0. (HIGH, CI-breaking)

### Reference-Parity Verification

The reference math was re-derived from `references/mlx-vlm/mlx_vlm/models/deepseek_v4/` in an independent from-scratch transcription rather than read off the port, and the committed float64 fixtures match to 8 decimal places:

- HyperConnection: `post`, the full 4x4 `comb`, and `collapsed` reproduce `_hc_split_sinkhorn_ops` exactly. The softmax then `+eps` then INITIAL column normalisation then `iters - 1` rounds of (row, column) ordering is correct, as are `pre = sigmoid + eps` and `post = 2 * sigmoid`, all in float32 over an f32 copy of the widened state.
- `_simple_compress_kv` and `_overlap_compress_kv` both reproduce exactly, including the feature-axis split, the one-window backshift with the zero / `-inf` prefix, the window-axis concat, and `ape` added before the softmax. The two paths are genuinely distinct and correctly selected on `compress_ratio == 4`.
- `_limited_swiglu` at limit 10 (min on gate, clip on up) and `sqrt(softplus(x))` both reproduce exactly.
- `hc_expand`, `HyperHead`, `PoolingCache` prompt/decode branches, `make_mask` giving `pool_idx < (offset + 1 + j) / ratio` with `None` for `L == 1`, all three HiSA paths including the batched `valid_len` causality, the biased-selection / unbiased-weighting gate contract, `tid2eid` hash routing, the `MultiLinear` `wo_a`/`wo_b` path, the inverse output RoPE, per-head sinks, and the weightless Q RMSNorm were diffed line by line against the reference and match.

### Real-Checkpoint Predictions

Real load and generation have not been run yet, so the load path was exercised statically against `models/deepseek-v4-flash-4bit`:

- Weight coverage: the `sanitize_weights` keep rules and `validate_weight_coverage` required/allowed sets were re-derived from the config and run against all 2481 tensors of `model.safetensors.index.json`. Result: 0 missing, 0 unknown, 0 dropped (the checkpoint ships no `mtp.*` tensors despite `num_nextn_predict_layers: 1`, and the 44-entry `compress_ratios` truncates to 43 so the last layer is ratio 4, matching the reference `__post_init__`).
- Quantization overrides: all 641 quantized planes satisfy `packed_in * 32 == bits * num_groups * group_size` under the per-path override the loader would read, with 0 mode/bias mismatches. The 129 expert planes resolve to mxfp4 group 32 with U8 scales and no `.biases`; the other 512 overrides are all `(64, 4, affine)`, identical to the top level, so reading the top-level pair for non-expert paths is correct on this checkpoint.
- Shapes and dtypes spot-checked against what each loader expects: `attn_hc.fn` F32 `[24, 16384]`, `hc_head.fn` F32 `[4, 16384]`, `attn_sink` F32 `[64]`, `tid2eid` I64 `[129280, 6]`, `wo_a.weight` 2-D `[8192, 512]` reshaping cleanly to `[8, 1024, 512]`, ratio-4 `compressor.ape` `[4, 1024]` against ratio-128 `[128, 512]`, and indexer `compressor.ape` `[4, 256]`. Nothing in `deepseek_v4_sanitize.rs` fires a legacy-plane arm on this checkpoint: `.scales` (plural) correctly does not match the `.scale` legacy quant fixup.
- Prefill masking: `create_sliding_window_prefill_mask` is the UNCAPPED builder, giving `[L, min(offset, window - 1) + L]`, which is exactly the shape `RotatingKVCache::update_concat` returns and exactly what the reference's `RotatingKVCache.make_mask` produces, so `align_local_mask` is a no-op on the real path and there is no all-masked-row degeneration at `L > 128`.
- Chunked prefill: the default `MLXCEL_PREFILL_CHUNK` of 2048 is a multiple of both 4 and 128, so chunk boundaries land on window boundaries for every compressed layer.

### Remaining Items

- [ ] `LimitedSwitchGlu` skips the `gather_sort` / `scatter_unsort` fast path that the shared `SwitchGLU` takes once `tokens * top_k >= 64`, so V4 prefill routes every expert dispatch through unsorted `gather_qmm`. Correctness is unaffected (sorting is an optimisation), but a 2048-token chunk at top-k 6 is 12288 unsorted gathers per layer. Follow-up, not a blocker. (LOW)
- [ ] No `forward_last_logits` override. With `vocab_size` 129280 the default computes the full `[1, chunk, 129280]` logits per prefill chunk (about 530 MB in bf16 at chunk 2048) to sample one row. Several large-vocab families in this tree override it. Deliberately not fixed here: adding an untested projection shortcut to a port whose real-checkpoint run has not happened yet trades a transient allocation on a 512 GB host for new risk. (MEDIUM, follow-up)
- [ ] `src/models/deepseek_v4.rs` imports `crate::models::bailing_moe::TokenIdField`. Every other family that needs this helper declares its own copy (11 of them do), and this is the only text family importing an unrelated text family's config type. Left as is rather than churning the PR: reuse is defensible and duplicating an 11th copy is not obviously better. (LOW)
- [ ] `TECHNICAL_REPORTS/.keep-reports` is present and no report exists for this PR yet. Not generated in this pass.

### Notes On Faithfully Ported Reference Quirks

These are correct ports, documented in the module docs, and are called out here only so they are not mistaken for defects later: the overlap compressor's backshift applies WITHIN each ready batch, so every decode-completed ratio-4 window and the first window a chunked continuation completes takes the zero / `-inf` prefix instead of its real predecessor's feature half. During decode that means the ratio-4 compressor degenerates to a softmax over the second feature half alone. The reference `_overlap_compress_kv` behaves identically, and `tiny_model_chunked_prefill_matches_single_pass` scopes its parity claim honestly around it.

### Verification

- [x] Integrated into project code flow (`ModelType`, detection arm plus test, `LoadedModel` via `delegate_language_model!`, `model_metadata.rs` row, TP fallback table, `docs/supported-models.md`; nothing orphaned)
- [x] Project conventions followed (load-time rejection of every allocation-sizing scalar, `compress_ratios` values and length, `scoring_func`, `num_experts_per_tok`, and every per-path quantization override, so no MLX C++ throw can cross the cxx bridge at first forward)
- [x] Existing modules reused where applicable (`MultiLinear`, `SwitchLinear`, `RMSNorm`, `UnifiedLinear`, `UnifiedEmbedding`, `RotatingKVCache`, `ModelOwnedSequenceState`, `moe_weighted_sum`, `validate_expert_quantization_*`)
- [x] No unintended structural changes (all shared-file edits are single-line registrations)
- [x] Tests pass (`cargo clippy --lib --tests -- -D warnings` exits 0 after the fix; `cargo fmt --check` clean on every V4 file; `cargo test --lib models::deepseek_v4` 26/26, `models::detection` 43/43, `loaded_model` 5/5 confirmed separately)
- [ ] Real-checkpoint load and generation: still outstanding, as the PR body states

inureyes:
## Real-checkpoint validation

Ran against the real `DeepSeek-V4-Flash` 4-bit checkpoint (43 layers, 33 shards, 2481 tensors, ~151 GB) on an M3 Ultra / 512 GB. The checkpoint was verified complete two independent ways before the run: every file matches the upstream declared byte count, and every shard is self-consistent against its own safetensors header (`filesize == 8 + header_len + max(data_offsets)`).

```
cargo test --test deepseek_v4_real_model --release --features metal,accelerate -- --ignored --nocapture

running 3 tests
[deepseek-v4] long-context prompt: 2234 tokens
[deepseek-v4] long-context answer: " The Pacific Ocean is the largest ocean on Earth. The Pacific"
test deepseek_v4_real_model_long_context_hits_sparse_and_compressed_paths ... ok
[deepseek-v4] prompt: "The capital of France is"
[deepseek-v4] greedy continuation (24 tokens): " Paris. The capital of France is Paris. The capital of France is Paris. ..."
test deepseek_v4_real_model_loads_and_generates_coherently ... ok
[deepseek-v4] 48-token decode: "\",\"answer\":\"The ocean is vast and mysterious.\",\"model\":\"gpt-4o-mini\", ..."
test deepseek_v4_real_model_decode_crosses_pooling_windows ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 71.93s
```

What each gate covers:

- **loads_and_generates_coherently** exercises the full weight-load path (both tensor-name planes, the 641 per-module-path quantization overrides mixing `mxfp4` group-32 experts with `affine` group-64 everything else) and a greedy prefill plus decode. The repetition in the continuation is ordinary greedy decoding on a base checkpoint, not a porting artifact.
- **decode_crosses_pooling_windows** drives 48 decode steps so the `PoolingCache` remainder buffer fills and flushes across window boundaries in both the ratio-4 and ratio-128 layers, which is where the prompt-mode and decode-mode branches diverge.
- **long_context_hits_sparse_and_compressed_paths** is the one that matters most: at 2234 tokens the pooled prefix exceeds `index_topk`, so `SparseCompressedAttention` takes the `_sparse_pooled_attention` branch with HiSA top-k selection rather than the dense concat fallback. Retrieving the right fact from deep in the prompt is the signal that the indexer selects real positions and the split softmax shares its log-normalizer correctly.

Note on layer composition: the checkpoint ships 44 `compress_ratios` entries for 43 layers. Following the reference `__post_init__`, the list truncates to the first 43, which makes the final layer ratio 4 (sparse) rather than the trailing 0 (local). A port that required `len == num_hidden_layers` would reject this checkpoint outright, and one that kept the last 43 entries would build the wrong attention class in every layer.

inureyes:

Static analysis only. The real checkpoint under `models/` was not loaded (four shards still downloading), so nothing here is backed by a real-model run.

Two commits land the fixes: `db05a80c` (four load-boundary guards plus the pooled-cache copy) and `233d4279` (the pooling-cache eval barrier). `cargo fmt --check` clean, `cargo clippy --lib --tests -- -D warnings` exit 0, `cargo test --lib models::deepseek_v4` 31 passed / 0 failed (26 before).

### Fixed

**HIGH, security. `tid2eid` values were never range checked.** `MoEGate::from_weights` validated the table's shape but not its contents. At forward time those entries feed `take_along_axis` over a `[B, L, n_routed_experts]` score row and `gather_qmm` over the stacked `[n_routed_experts, out, packed_in]` expert planes. MLX gathers only fold negative indices (`offset_neg_idx` in `mlx/backend/cpu/indexing.cpp`) and never clamp, so an out-of-range expert id in a downloaded checkpoint was a straight out-of-bounds read. This is the only routing path in the tree whose expert indices come from a checkpoint tensor rather than an `argpartition` over the score row, so the in-range-by-construction argument that covers `bailing_moe` / `afmoe` / `klear` does not reach it. Now bounded to `0..n_routed_experts` at load with one `min_all` / `max_all` pair per hash layer.

**HIGH, performance. `PoolingCache::update_and_fetch` deep-copied the whole pooled buffer on every call.** `mlxcel_core::copy` is `mlx::core::copy` (`mlx/ops.cpp:340`), a real `Copy` primitive, not a handle alias, so both branches materialised `Np * D` elements per call, including the `n_new == 0` branch that dominates decode (3 of every 4 steps at compress ratio 4, 127 of every 128 at ratio 128). `Np` grows with context, so the copy traffic grew without bound: roughly 54 MB per decoded token at 8k context and 860 MB per decoded token at 128k across the real config's 40 compressed and sparse layers. The reference `PoolingCache.update_and_fetch` returns `self.pooled` directly. `update_and_fetch` and `Compressor::forward` now return a borrow of the cache's buffer; the sparse arm relies on `cache.pool` and `cache.idx_pool` being disjoint field projections through the same `&mut V4LayerCache`.

**MEDIUM, performance. The indexer pooling cache accumulated an unevaluated MLX graph across decode steps.** The only consumer of the indexer's pooled buffer is the top-k selection `Indexer::forward` returns, and the `AttnKind::Sparse` arm discards it on both the `np == 0` and the `np <= index_topk` branches. Nothing else reads `idx_pool`, so while the sparse selection path is not being taken its `pooled` concat chain grew by one unevaluated node per completed window and its `buf_kv` / `buf_gate` `slice_update` chains grew by one per decode step, each node pinning that step's normed hidden state plus its `wkv` / `wgate` projections. The sparse branch only fires past `index_topk * compress_ratio` tokens (2048 on the real config), so every sequence paid this over its first ~2048 tokens and a shorter sequence paid it for its whole life: about 190 KB per decode step across 20 sparse layers, then forced in one burst down a chain thousands of nodes deep. `PoolingCache::eval_state` and `V4LayerCache::eval_state` now run after the layer loop and before `hc_head`, mirroring `KVCache::eval_state`, so the layer stack is forced but the `[B, L, vocab]` logits tensor is not.

**MEDIUM, security. `e_score_correction_bias` had no shape check** while both sibling tensors in the same constructor did. `MoEGate::forward` adds it straight onto a `[B, L, n_routed_experts]` score row, so any other length made MLX's `add` throw at the first forward pass instead of failing the load, and that throw is an uncatchable `std::terminate`, not a catchable Rust error. Now checked against `[n_routed_experts]`.

**MEDIUM, security. The 2-D `wo_a` reshape in `sanitize_weights` Pass 5 ran unguarded on checkpoint data.** The element count comes from the checkpoint and the divisor from `config.json`. MLX's `reshape` asserts the `-1` axis divides evenly, and that pass runs before `validate_weight_coverage`, so a mismatch aborted the process mid-load rather than returning an error naming the tensor. Now rejected when the size is zero or not a multiple of `o_groups * o_lora_rank`, computed in `i64`.

**MEDIUM, security. `ModelArgs::validate` bounded every architecture scalar below but never above.** Each one is cast `as i32` on its way into MLX, and a `usize` past `i32::MAX` does not fail that cast: it truncates silently to zero or to a negative dimension that then reaches `reshape` / `broadcast_to` / `zeros`. `index_block` and `index_keep` were not validated at all despite gating the HiSA hierarchical path and dividing `Np` inside it. All scalars now capped at `i32::MAX`, `index_block` / `index_keep` added to the checked set, and `hidden_size * hc_mult` rejected when the product does not fit `i32`. `Indexer::forward`'s hierarchical gate now computes `index_block * index_keep` in `i64`: a wrapped product could pass the gate while `nb = np / index_block` was 0, driving `kb` to 0 and reaching `argpartition(kth = -1)`.

### Reported, not fixed

**MEDIUM, performance. `sparse_pooled_attention` promotes to float32 where the reference stays in the activation dtype.** `gathered` is `[B, L, index_topk, head_dim]`, which on the real config at a 2048-token prefill chunk is 537M elements: 1.07 GB in bf16, 2.15 GB in f32, with both live across the cast. `local_scores` goes from 0.57 GB to 1.14 GB and `local_weights` again the same. That is roughly a 2x multiplier on the largest intermediates in the model. The reference `_sparse_pooled_attention` runs the whole split softmax in the activation dtype. This looks deliberate (the doc comment says so, and `2b875ca4` tuned the sink dtype here), and changing it moves numerics, so it should be decided against the real-model run rather than in review.

**LOW, performance. The local KV is duplicated to fill the rotating cache's value side.** `V4Attention::forward` does `let kv_values = mlxcel_core::copy(&kv);` because K and V are the same tensor for a single shared KV head. The reference passes a zero-width `mx.zeros((B, 1, L, 0))` placeholder instead, so it stores nothing on the V side. The port copies `L * head_dim` per layer per call and keeps a values buffer it never reads: about 96 MB of both during a 2048-token prefill chunk across 43 layers. Left alone because a zero-width value tensor interacts with this tree's `RotatingKVCache` quantization modes and that is not safe to change without the real-model run.

**LOW, performance. `Indexer::flat_select` materialises `[B, H, L, Np]` f32 scores with no tiling**, about 268 MB for a 2048-token prefill chunk at `Np = 512`, and it is the path taken for the entire first prefill chunk because `hierarchical` needs `Np >= index_block * index_keep = 1024`. The tiled `hisa_select_batched` bound does hold: `cand` is `[B, min(512, L), 1024, 128]` f32 per tile with an `eval` barrier between tiles. The flat fallback that runs first has no equivalent bound, but it matches the reference, so it is a shared property rather than a port defect.

**LOW, security. RMSNorm weight shapes are never validated** (`attn_norm.weight`, `ffn_norm.weight`, `attn.q_norm.weight`, `attn.kv_norm.weight`, `model.norm.weight`, `compressor.norm.weight`). `validate_weight_coverage` checks presence only, so a wrong-width norm weight reaches `mx.fast.rms_norm` and throws. This is a repo-wide pattern rather than something this PR introduced, so it belongs in a follow-up that covers every family at once.

**Informational.** Token ids are not bounded before `take(tid2eid, input_ids, 0)`, the same exposure the embedding lookup already carries in every family. `PoolingCache::pooled` grows monotonically with context by design (the pooled KV is the long-term memory), reaching roughly 7 GB across all layers at the declared 1048576-token `max_position_embeddings`; that is architectural and matches the reference.

### Checked and clear

Registration edits in `mod.rs`, `detection.rs`, `loaded_model.rs`, `model_metadata.rs` and the tensor-parallel fallback table are mechanical and complete. Quantization bounds meet `docs/adding-models.md`: the top-level pair, the mode string, and every one of the 641 per-path overrides are validated in `ModelArgs::validate`, and `LimitedSwitchGlu` runs both `validate_expert_quantization_params` and `validate_expert_quantization_shapes` rather than bypassing the shared loader's bound. `PoolingCache::accumulate_windows` slice arithmetic holds on every branch, including the `usable == 0` and remainder-boundary cases. The `topk_indices` `kth` argument stays in range on all three selection paths. `pool_visible_counts`, `align_local_mask` and `extend_mask` are additive-f32 throughout and agree with the mask `create_sliding_window_prefill_mask` produces. RoPE frequency tables are precomputed per `(head_dim, inverse)` pair at load, and the prefill mask is built once outside the layer loop. The `transpose` of the HC `fn` plane and the MoE gate weight on every forward is not a real cost: MLX's `Transpose` shares the buffer and `matmul` consumes the transposed strides directly.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: added 8 tests closing gaps the implementation and security/performance reviews did not target (they covered reference-math parity, gate contracts, sanitize, coverage, HiSA agreement, and the new bounds checks). New coverage: `MoEGate::from_weights`'s router-weight and `tid2eid` shape checks, `V4Attention::from_weights`'s `attn_sink` shape check, `Compressor::from_weights`'s `ape` shape check, `sanitize_weights`'s rejection of a legacy per-expert plane missing one index, `v4_rope_base_freqs`'s yarn-scaling field checks (missing `factor`/`original_max_position_embeddings`, non-positive factor, zero orig), `AttnKind::for_ratio`'s reject-unknown-ratio arm, and `ModelArgs::eos_token_ids`'s default-fallback branch. DeepSeek-V4 lib suite: 31 -> 39, all passing.
- **Documentation**: `docs/supported-models.md`'s DeepSeek-V4 entry checked field-by-field against the implementation (HyperConnections/Sinkhorn params, attention shape, compress-ratio dispatch, HiSA defaults, MoE hash routing, quantization overrides) and confirmed accurate; the checkpoint size was corrected from the pre-fetch "~145 GB" estimate to the observed "~151 GB" (matched in `tests/deepseek_v4_real_model.rs`'s doc comment), and the validation sentence now names what each of the three real-checkpoint gates actually checks. The "DeepSeek v1 / v2 / v3 / v3.2" summary line already closed with an explicit pointer to the V4 entry, so it needed no change. Added a `CHANGELOG.md [Unreleased]` entry, which this port had none of. `docs/adding-models.md`'s shared-helper files are untouched by this PR, so no comment update was needed there; this repo carries no `docs/ko` tree, so no Korean translation exists to update.
- **Lint/Format**: `cargo fmt` and `cargo clippy --lib --tests -- -D warnings` clean on all touched files.

### Left alone (per scope)
- Reference-math parity, gate contracts, sanitize/coverage correctness, HiSA agreement, and the review's bounds-check fixes: already covered, not re-reviewed.
- The three `#[ignore]`d real-checkpoint tests: not re-run.
- Everything under `models/`.

Issue #1456, opened 2026-08-27
feat(server): add Vertex AI (GCP) predict route adapter for llama-server b10621 compatibility


inureyes:
## Parent

- #1435 (classification sub-issue of the llama-server b10621 epic #1431)

## Problem

llama-server b10621 registers Google Cloud Vertex AI custom-container compatibility purely from environment variables (`tools/server/server-http.cpp`, `gcp_params` and `register_gcp_compat`, reference commit c1d0e7a004015f23bc0233470b747b596f29b264). mlxcel-server has none of it, although every route the adapter fans out to already exists in mlxcel. #1435 classified this surface as Supported because it is a pure HTTP adapter over existing handlers and matters for the Linux/CUDA deployment target where Vertex custom containers run.

## Upstream semantics to match

- Activation: only when `AIP_MODE=PREDICTION`. Otherwise nothing is registered and the variables are ignored.
- `AIP_HTTP_PORT` (default `8080`) overrides `--port`; upstream logs a warning when the two differ.
- `AIP_HEALTH_ROUTE` (default empty, meaning the existing `/health` only): when set, a leading slash is ensured and the value is registered as a GET alias of the health handler.
- `AIP_PREDICT_ROUTE` (default `/predict`): a leading slash is ensured; startup fails if the path collides with an already registered route.
- Request: JSON object with an `instances` array (max 128 entries). Each instance is an object that must carry a string `@requestFormat`; the field is stripped and the remainder is forwarded as the request body. A client-provided `stream` field is ignored with a warning; streaming responses are rejected because the predict route is non-streaming.
- Alias table: every registered route path is converted with `path_to_gcp_format`: strip a leading `/v1`, strip the leading `/`, stop at the first `:` path parameter, and convert `/`, `-`, `_` boundaries to camelCase (`/v1/chat/completions` becomes `chatCompletions`, `/v1/embeddings` becomes `embeddings`, `/v1/messages` becomes `messages`, `/tokenize` becomes `tokenize`). First registration wins on collision.
- Instances are executed concurrently and results are returned in order as `{"predictions":[...]}`; per-instance failures become an error object in that slot rather than failing the whole batch. Malformed envelopes return 400 with the llama-server error envelope.

## Scope

- Add an env-gated adapter in the axum app builder that registers the health alias and the predict route and applies the `AIP_HTTP_PORT` override with the same precedence as upstream, on both `mlxcel serve` and `mlxcel-server`.
[...]

Issue #1453, opened 2026-08-26
feat(server): implement b10621 prompt-cache and slot-scheduling semantics


inureyes:
## Parent

- #1431
- Immediate collision fix: #1430

## Problem

#1430 stops treating integer `LLAMA_ARG_CACHE_REUSE` as a boolean prompt-cache switch, but it does not implement b10621 KV-shift chunk reuse. Server-wide prompt caching, per-request `cache_prompt`, slot prompt similarity, explicit slot selection, continuous batching, and parallel-slot scheduling also have distinct externally visible behavior.

## Scope

- Implement `--cache-prompt` and `--cache-reuse N` with the b10621 dependency, default, and minimum-chunk semantics.
- Align per-request `cache_prompt`, `id_slot`, prompt similarity, slot reuse, and cache invalidation behavior.
- Classify static/dynamic lookup caches, context/SWA checkpoints, checkpoint spacing, cache RAM limits, unified KV topology, idle-slot caching, and context shifting as implemented, not applicable, or deferred.
- Align `--parallel`, continuous-batching toggles, queue selection, cancellation, and fairness semantics where MLX execution permits it.
- Define cache identity across model, tokenizer, template, reasoning kwargs, adapters, modalities, and speculative configuration.
- Expose cache hit/reuse counters without logging prompt content or creating unbounded labels.

## Acceptance criteria

- [x] `cache_reuse=0` means no KV-shift chunk reuse while prompt caching remains independently enabled. (`--cache-reuse` is now a real option with the `LLAMA_ARG_CACHE_REUSE` binding, so a command-line value beats the environment; `0` is accepted and inert and says nothing about whether the cache is on.)
- [x] Positive reuse thresholds produce measurable compatible prefix reuse or fail explicitly until implemented. (Fails explicitly: no KV operation in this tree re-bases the rotary positions of a cached span, so chunk reuse cannot be expressed. The diagnostic names the missing operation.)
- [x] Per-request cache disablement forces reprocessing and does not mutate reusable state. (`cache_prompt: false` on the chat-shaped routes withholds the request's cache context, which is the single handle the scheduler reaches the store through for both lookup and donate-back. Proven end to end: the opt-out reports `cached_tokens == 0` and a following identical request still hits.)
[...]

Issue #1452, opened 2026-08-26
fix(server): align embedding and reranking mode, pooling, and normalization


inureyes:
## Parent

- #1431

## Problem

b10621 uses `--embedding` to restrict the server to embedding workloads, `--rerank` to enable reranking, `--pooling` to select pooling behavior, and `--embd-normalize` to control vector normalization. mlxcel has dedicated embedding and reranking workers, but its route enablement, pooling/normalization vocabulary, native envelopes, and model-capability failures are not b10621 aliases by construction.

## Scope

- Align `--embedding` / `--embeddings`, `--rerank` / `--reranking`, `--pooling`, and `--embd-normalize` with canonical environment bindings.
- Define whether compatible mode flags restrict unrelated generation routes or select a dedicated worker, and expose the resolved capability in `/props` and `/v1/models`.
- Implement b10621 pooling choices `none`, `mean`, `cls`, `last`, and `rank` where the model architecture supports them.
- Implement normalization values `-1`, `0`, `1`, `2`, and `p > 2` with compatible numeric behavior.
- Align native embedding and reranking request/response schemas, errors, document limits, and route gates without changing the OpenAI `/v1/embeddings` contract unintentionally.

## Acceptance criteria

- [x] CLI/environment parsing and mode restriction tests cover every pooling and normalization value.
- [x] Unsupported model/pooling combinations fail before serving requests with an actionable capability diagnostic.
- [x] Numerical tests validate normalization invariants and stable ranking order, including zero vectors and ties.
- [x] Native reranking aliases return compatible schemas and are unavailable when reranking is disabled as in b10621.
- [x] `/v1/embeddings` remains regression-tested separately from native `/embedding` and `/embeddings`.

## Validation

- Run one real embedding checkpoint and one real reranker checkpoint through all supported pooling/normalization paths.

## Remaining work (kept open after the implementation PR)

PR #1483 implements the whole surface and flips the six route entries this issue owns in `compat/llama-server/b10621/routes.json` to `supported`. The four option entries in `compat/llama-server/b10621/embeddings-and-rerank.json` stay `deferred`, each with one or two observable differences recorded on it, so this issue stays open and owns them:

[...]

Issue #1451, opened 2026-08-26
fix(server): classify projector flags and align multimodal media handling


inureyes:
## Parent

- #1431

## Problem

b10621 loads a separate multimodal projector and exposes projector URL/auto/offload/device controls plus `--media-path` for local `file://` inputs. mlxcel has direct MLX VLM checkpoints and multimodal request handling, so exact feature equivalence requires translation at the model capability and request boundary rather than accepting projector flags as no-ops.

## Scope

- Classify `--mmproj`, `--mmproj-url`, `--mmproj-auto`, `--mmproj-offload`, `--mmproj-device`, their aliases, and environment variables.
- Translate only inert or capability-equivalent forms for integrated MLX VLM checkpoints; explicitly reject requests to load a separate GGUF projector.
- Implement compatible image/audio content-part decoding, media limits, model modality checks, and error envelopes on shared chat/responses/transcription routes.
- Align `--image-min-tokens`, `--image-max-tokens`, and `--mtmd-batch-max-tokens` with integrated MLX vision encoders where exact token-budget translation is possible.
- Implement `--media-path` only with a confined root and b10621-compatible relative `file://` handling, or reject it with a supported upload alternative.
- Keep remote URL policy explicit and protect local files, loopback services, metadata services, redirects, and decompression limits.

## Acceptance criteria

- [ ] Every projector and media option is classified with a parser or diagnostic test.
- [ ] Text-only models reject media before generation, while supported VLM/audio models accept compatible content-part shapes.
- [ ] Local media paths cannot traverse or escape the configured root through symlinks, encoding, or URL normalization.
- [ ] Remote media handling has request-size, decode-size, timeout, redirect, and network-address protections.
- [ ] Image-token minima, maxima, and batch limits either constrain the real encoder path or fail before serving with an architecture-specific diagnostic.
- [ ] Help does not imply that a GGUF `mmproj` file can be attached to an MLX checkpoint.

## Validation

- Smoke-test one real VLM with an uploaded image and one audio-capable model with a short clip; include malformed and oversized media cases.

PR #1458, opened 2026-08-27
test(server): add pinned b10621 compatibility manifest and gates

### LINKED ISSUE #1443: test(server): add a pinned b10621 compatibility manifest and conformance harness
## Parent

- #1431

## Problem

The b10621 audit currently exists as an analysis snapshot. Without a checked-in, machine-readable inventory, new llama.cpp nightlies can add or change flags, environment variables, routes, and native request fields without producing a reviewable mlxcel diff.

## Scope

- Add a versioned compatibility manifest pinned to llama-server `b10621` / commit `c1d0e7a004015f23bc0233470b747b596f29b264`.
- Inventory every public and hidden help entry, long-option spelling, `LLAMA_ARG_*` / `LLAMA_API_KEY` environment variable, registered HTTP method/path, and native request field.
- Give every entry one state: `supported`, `aliased`, `not_applicable`, or `deferred`.
- Record the mlxcel spelling or route, semantic notes, test identifier, and linked issue for every non-supported entry.
- Add an extraction/update script that consumes an official nightly binary and a pinned llama.cpp source checkout.
- Add CI checks that compare the manifest against both `mlxcel serve --help` and `mlxcel-server --help`, including hidden compatibility options.

## Acceptance criteria

- [x] The manifest accounts for all 249 b10621 help entries, 323 distinct long-option spellings, 134 environment-variable names, registered routes, and native schema fields.
- [x] Duplicate aliases are represented without inflating the feature count.
- [x] CI fails when an mlxcel spelling/default/env binding drifts from a claimed supported or aliased entry.
- [x] CI fails when a deferred entry has no open issue or a not-applicable entry lacks a diagnostic/documentation test.
- [x] Regeneration is deterministic and documents the official archive SHA-256 `429c8270608600188035e5e92f7d78dffb7900904fe7dd7e6a84f48068cd13cf`.
- [x] Operator documentation links to the generated compatibility boundary.

## Validation

- Run the extractor twice and verify a clean worktree after the second run.
- Run the conformance test against the official macOS arm64 b10621 archive and the current release `mlxcel-server` binary.

inureyes:
## Summary

Turns the llama-server b10621 compatibility audit into a checked-in, machine-readable manifest plus executable conformance gates, so a future llama.cpp nightly that adds or changes a flag, environment variable, route, or native request field produces a merge-blocking, reviewable mlxcel diff instead of silent drift. This is the foundation issue of epic #1431; the twenty-two later sub-issues flip manifest entries from `deferred` to `supported` / `aliased` / `not_applicable`.

## The manifest

`compat/llama-server/b10621/` pins release `b10621` (commit `c1d0e7a004015f23bc0233470b747b596f29b264`, archive SHA-256 `429c8270608600188035e5e92f7d78dffb7900904fe7dd7e6a84f48068cd13cf`) and inventories the full surface reproduced from the official macOS arm64 binary and the pinned sources: 249 help entries carrying 323 distinct long-option spellings (all of them declared in entry heads once the full comma list of each head line is parsed, including mixed-order heads such as `--spec-draft-threads-batch, -tbd, --threads-batch-draft` and negation pairs such as `--cont-batching, --no-cont-batching`; each entry records `discovery: "help-entry-head"`), 136 environment variables (134 `LLAMA_*` plus `HF_TOKEN` and `MTMD_BACKEND_DEVICE`), 53 registered HTTP method/path pairs from `server.cpp` / `server-http.cpp` (router-mode, 403-fallback, GCP, and Web UI static registrations annotated via `condition`), and 74 native `/completion` request fields from `server-schema.cpp` with their aliases. Aliases live inside their entry, so the 323 spellings never inflate the 249-entry feature count, and the validator enforces that no spelling appears in two entries.

Every entry carries exactly one policy state per epic #1431: currently 22 `supported`, 354 `deferred` (each linked to its owning open issue among #1432-#1442, #1444-#1453, #1456, and #1377 for `typical_p`), 0 `aliased`, 0 `not_applicable`; #1430's settled surface (canonical env spellings, aligned defaults, rerank/health route aliases, `--temp`/`--temperature`, seed and stop semantics) is what makes the supported set non-empty, and everything ambiguous stays deferred rather than optimistically supported. Accepted-but-ignored hidden flags are `deferred` with their acceptance pinned in an `mlxcel` claim block, which is restricted to a nine-key allowlist (`accepted_spellings`, `accepted_on_one_binary_only`, `env`, `env_binding`, `env_test`, `defaults`, `hidden`, `route`, `field`) so a typo'd key fails the gate instead of silently recording a claim nothing checks. Semantic collisions from the epic audit are marked explicitly in `notes` (`--timeout`, `--models-dir`, `--cache-type-k/v`, `--props` gating GET instead of POST, `POST /completions` and `POST /embeddings` selecting OpenAI-shaped handlers, `dry_sequence_breakers` taking token ids instead of strings, `--parallel` defaulting to 4 instead of -1). Claim blocks are CI-verified regardless of state, so partial support cannot silently regress either.

Regeneration binds the verified archive to the binary it actually runs: the extractor's `--archive` verifies the download against the pinned SHA-256 and then requires the `--llama-server` binary to be byte-identical to a member of that same archive (compared through `tarfile` in read mode, nothing ever extracted to disk), because the extractor executes that binary with `DYLD_LIBRARY_PATH` / `LD_LIBRARY_PATH` pointed at its own directory. Hashing the tarball while running an unrelated path would only prove something about a file nobody executed.

## `divergence`, and why `supported` started out overclaimed

Epic #1431 defines `supported` as "the spelling, value domain, default, precedence, and externally observable behavior match", and its Definition of Done says no option with observable b10621 semantics is accepted and silently ignored. Twenty-two entries claimed `supported` anyway, nineteen of them documenting the divergence in their own `notes` (two opening with "BEHAVIOR DRIFT"), which is worse than an undocumented gap because the CI gate then locks the false claim in. All twenty-two are now `deferred` with the divergence moved into a structured field naming its owning issue: the health and models routes (#1440, #1438), the four rerank route spellings (#1452), `--metrics` (#1440), `--model` (#1434), `--help` and `--verbose` (#1448), and `--dry-base` / `field:dry_base` / `--seed` / `field:seed` / `--temp` / `--top-k` / `--top-p` (#1436).

Three of the twenty-two were found by probing both binaries rather than by reading notes. mlxcel's clap parser rejects the space-separated form of a negative value, so `llama-server --predict -1`, `--reasoning-budget -1` and `--seed -1` all fail to start under mlxcel while b10621 parses them and reaches model load; only the `--predict=-1` form works. `--predict` (#1450) and `--reasoning-budget` (#1447) recorded nothing about this, and `--seed`'s note attributed the below-minus-one rejection to the flag when it actually lives on the HTTP field. That class is likely wider than the three entries verified here, and the owning sub-issues should sweep every negative-valued option.

Every entry now carries a `divergence` list: short strings, one per externally observable way mlxcel differs from b10621 for that entry. A non-empty `divergence` makes `supported` a hard error whose message names the three honest alternatives (`aliased`, `not_applicable`, `deferred`) and asks for the owning issue. `notes` keeps the context and the reasoning; `divergence` carries the difference a machine can check, because prose scanning was never going to be that gate and every remaining sub-issue of the epic will flip more entries into `supported`. Entry keys are an exact allowlist in canonical order too (`state`, `issue`, `test`, `notes`, `divergence`, `mlxcel` after the kind's fact keys), so `divergance` and a `divergence` misfiled inside the `mlxcel` claim block are both gate failures rather than silent no-ops.

## Sharding scheme (for the downstream chains)

One directory `compat/llama-server/b10621/` with `pin.json` (frozen reference, counts, shard ownership map, hand-maintained `mlxcel_baseline` snapshot) plus 17 area shard files. Ownership is machine-readable, not prose: `pin.json`'s `shards` map records, per shard, the set of implementation issue numbers allowed to own entries in it (`shards["authentication"].owners == [1437]`, for example), and `scripts/ci/check_llama_compat_manifest.py` fails an entry whose `issue` is not a member of its own shard's owner set. That is what stops two concurrent chains from editing the same file: the file itself rejects the second chain's entry, not just a merge conflict or a reviewer. Routes and native request fields live in the shard of the issue that owns them, so no two chains touch one file:

| Shard file | Chain | Owning issues |
|---|---|---|
| `transport-tls-cors.json` | A | #1432 (also `field:sse_ping_interval`) |
| `authentication.json` | A | #1437 |
| `routes.json` | A | #1441, #1442, #1452 (all general/native/OpenAI/tokenize routes plus the four rerank route entries, envelope-level native fields) |
| `embeddings-and-rerank.json` | A | #1452 (option entries) |
| `observability-and-slots.json` | A | #1440 (`/metrics`, `/props`, `/slots` routes included) |
| `router-models.json` | A | #1438 (router `/models*` routes included) |
| `lora-adapters.json` | A | #1439 (`/lora-adapters` routes and `field:lora` included) |
| `model-source.json` | B | #1434 |
| `ggml-runtime.json` | B | #1445 |
| `chat-templates.json` | B | #1447 (chat-parsing and reasoning-budget native fields included) |
| `multimodal-and-audio.json` | B | #1451, #1446 (audio-transcription routes included) |
| `ui-tools-mcp-gcp.json` | B | #1435, #1456 (tools/cors-proxy/GCP/Web-UI-static routes included) |
| `streams-and-realtime.json` | B | #1444 (`/v1/stream*`, `POST /v1/chat/completions/control`, `field:reasoning_control`) |
| `runtime-and-context.json` | C | #1450, #1453, #1449 (`field:n_keep`, `field:n_discard`, `field:cache_prompt`, `field:n_cache_reuse` included) |
| `sampling-and-grammar.json` | C | #1436, #1377 (all sampling/grammar native fields included) |
| `speculative.json` | C | #1433 (`speculative.*` native fields included) |
| `logging-and-presets.json` | C | #1448 |

A chain flips an entry by editing `state`, `issue`, `test`, `notes`, `divergence`, and `mlxcel` in its shard, always with an `issue` from its own shard's owner set; the extractor never touches those fields or the owner sets, which it carries forward verbatim by shard name across regeneration, the same way it carries `mlxcel_baseline` forward. New upstream entries (on a future pin bump) land in `_unclassified.json` with `state: "unclassified"`, which the validator rejects until a human classifies them. `pin.json` and every shard carry `schema_version: 3`, bumped together with the validator, the extractor, and the two Rust conformance suites: 2 when the owner-set map replaced the bare shard-name list, 3 when every entry gained the structured `divergence` list.

## Hidden clap arguments

The acceptance criterion "including hidden compatibility options" is met with a runtime mechanism, not a source scan: both binaries answer a hidden `--dump-flag-surface` machine interface (`src/cli/flag_surface.rs`) that prints the complete built clap surface as deterministic JSON (primary longs, all aliases, shorts, env bindings, defaults, hidden-ness, headings) and exits. It is intercepted positionally before `Cli::parse` (`mlxcel serve --dump-flag-surface`, `mlxcel-server --dump-flag-surface`) and is not a clap argument, so `--help` and the operator-facing surface are unchanged, and ordinary argument values can still contain the token. Argv is read as `args_os` and compared as `OsStr`, so an unrelated non-UTF-8 argument (a legal Unix path, accepted today by every `PathBuf` argument including `--model`) still reaches clap instead of aborting the process. A source scan was rejected because it drifts from what the built binary actually parses; the dump is produced by the same `clap::Command` the binary parses with, so the twenty `hide = true` compatibility arguments (10 per binary) are first-class in the CI comparison.

## Enforcement

- `scripts/ci/check_llama_compat_manifest.py` (new `llama-compat manifest` CI job, `make verify-llama-compat`, wired into `make verify`): offline structural validation of pinned counts, one-state-per-entry, alias non-inflation, issue links on deferred entries, test ids on not_applicable/aliased entries, claim self-consistency, shard ownership (every entry's `issue` must belong to its own shard's `pin.json` owner set), the entry key allowlist, the `mlxcel` claim-key allowlist, the rule that a non-empty `divergence` forbids `supported`, repo-relative test-path existence (a repo-relative POSIX path with no `..` component, resolving under the repository root; an absolute or upward-walking value is rejected), and canonical serialization of both `pin.json` and every shard. The CI job adds `--check-issues-open`, so a deferred entry pointing at a closed issue fails CI.
- `scripts/ci/check_llama_compat_manifest_test.sh` (same CI job, same make target): the validator's own negative coverage. It copies the manifest to a temporary directory, mutates one entry, points the validator at the copy through the new `--manifest-dir` flag, and asserts the expected rejection, so a rule cannot degrade into one that only ever passes on a manifest that already satisfies it. Eight cases: the untouched control, a `supported` entry carrying a `divergence` (rejected, with the guidance text asserted), a `divergence` on a `deferred` entry (accepted), a misspelled `divergance` key, a `divergence` buried in the `mlxcel` claim block, a bare-string `divergence`, an empty-string element, and a dropped `divergence` key.
- `tests/llama_compat_manifest.rs`: every option claim is verified against the dumped surfaces of BOTH `mlxcel serve` and `mlxcel-server` (spellings, clap env bindings, defaults, hidden-ness), an option entry carrying no claim must be accepted by neither binary so adding a b10621 flag without flipping its entry fails, the sentinel is asserted absent from both `--help` renderings, and an archive-gated test re-derives the full option inventory from the real `llama-server --help` and compares it exactly when `MLXCEL_LLAMA_B10621_DIR` is set. CI never needs the archive or network; the test skips with a clear message otherwise. Complements `tests/cli_help_consistency.rs` (which pins the two mlxcel surfaces against each other) rather than duplicating it.
- `src/server/llama_compat_tests.rs`: it restates the `divergence` rule so a `cargo test` run cannot pass a manifest `make verify` would reject, and route claims are probed against the real axum router (claimed routes must resolve, unclaimed b10621 routes must NOT resolve) and native-field claims against `NativeCompletionRequest` in both directions, so implementing part of the surface without flipping its manifest entry fails. A claim naming a different mlxcel identity is an `aliased` mapping and is checked both ways too: the alias must resolve or be declared and the b10621 identity must not, which is what keeps `aliased` mechanically distinct from `supported` for routes and request fields.

## Known divergences

`--repeat-last-n` / `field:repeat_last_n` are parsed and echoed by `GET /props` but never reach the sampler, and because `SamplingConfig` carries no penalty-window field, `--repeat-penalty` / `--frequency-penalty` / `--presence-penalty` and their request fields scan the whole history where b10621 bounds them to the window. `--dry-penalty-last-n` inverts the disable sentinel (0 disables DRY upstream, means "scan everything" here), `--dry-allowed-length` compares with `>` where b10621 uses `>=`, `--dry-sequence-breaker` starts from an empty breaker set instead of b10621's four defaults and rejects the `"none"` sentinel and multi-token breakers, `field:stop` is consumed only by the default-off OpenXLA worker, `field:n_predict` is `Option<usize>` so b10621's documented `-1` is rejected, and `--alias` takes one opaque id where b10621 takes a comma-separated list. All of these are recorded as `deferred` with the divergence spelled out in both `divergence` and `notes` rather than optimistically marked `supported`. `--slots` does bind `LLAMA_ARG_ENDPOINT_SLOTS`, through a runtime bridge that is claimed (`env_binding: "runtime"`, `env_test` pointing at its covering test) and covered.

## Docs

`docs/llama-server-compat.md` documents the boundary, the states, the `divergence` field and the rule that a divergence forbids `supported`, the sharding and its machine-enforced ownership, the enforcement gates, and the deterministic regeneration procedure including the pinned SHA-256 and the owner-set preservation across regeneration; `README.md`'s llama-server bullet and `docs/README.md` link to it. The stale pre-#1430 audit numbers are recorded as such in `pin.json`'s `mlxcel_baseline` block together with the current measured numbers and the counting method for each.

## Test plan

- [x] `python3 scripts/compat/extract_b10621_manifest.py --llama-server ... --source-dir ... --archive ...` verifies the archive SHA-256, requires the binary to match an archive member, and reproduces 249 / 323 / 136 / 53 / 74 with the `shards[name].owners` map and every `divergence` preserved; run in place twice back to back the manifest is byte-identical both times, and identical to a hand-migration from the previous commit.
- [x] `python3 scripts/ci/check_llama_compat_manifest.py` passes offline and with `--check-issues-open` (all 23 linked issues verified open).
- [x] `bash scripts/ci/check_llama_compat_manifest_test.sh` passes (8 cases), including the negative case proving a `supported` entry carrying a `divergence` is rejected.
- [x] `cargo test --profile test-fast --features metal,accelerate --test llama_compat_manifest` passes, including the archive conformance test against the extracted official b10621 archive (`MLXCEL_LLAMA_B10621_DIR` set).
- [x] `cargo test --profile test-fast --features metal,accelerate --lib server::llama_compat_tests` and `--lib cli::flag_surface` pass.
- [x] `cargo test --profile test-fast --features metal,accelerate --test cli_help_consistency` passes (25 tests).
- [x] `cargo fmt --all -- --check` and scoped `cargo clippy -- -D warnings` pass.
- [x] `make verify-llama-compat` passes.

Closes #1443


inureyes:

### Intent

Turn the llama-server b10621 compatibility audit into a checked-in, sharded, machine-readable manifest plus executable conformance gates, so a future llama.cpp nightly produces a reviewable diff instead of silent drift, and so the twenty-two later sub-issues of epic #1431 have a stable contract to flip entries in.

### Findings Addressed

- [x] Both shipped binaries panicked (exit 101) on any non-UTF-8 argument (CRITICAL). The sentinel interception read `std::env::args()`, which panics on invalid Unicode, before clap ran. A non-UTF-8 path is legal on Unix and accepted today by every `PathBuf` argument including `--model`; on `main` the same invocation gives a clean resolver error and exit 1. Now compares `std::env::args_os()` as `OsStr`, with a `#[cfg(unix)]` regression test. Reproduced against the pre-change binaries and re-checked against the rebuilt ones.
- [x] Sixteen entries were marked `supported` for behavior mlxcel does not provide (CRITICAL), which is the one failure mode the issue calls out as worse than `deferred`. All flipped to `deferred` with the divergence recorded in `notes`: `--repeat-last-n` and `field:repeat_last_n` are parsed and echoed by `GET /props` but never reach the sampler (`SamplingConfig` has no penalty-window field), which also makes `--repeat-penalty` / `--frequency-penalty` / `--presence-penalty` and their fields scan the whole history where b10621 bounds them to the window and disables them at 0; `--dry-penalty-last-n` and `field:dry_penalty_last_n` invert the disable sentinel (0 disables DRY upstream, means "scan everything" here); `--dry-allowed-length` and `field:dry_allowed_length` compare with `>` where b10621 uses `>=`, so the `base^0` tier is never penalized; `--dry-sequence-breaker` starts from an empty breaker set rather than b10621's `'\n' ':' '"' '*'`, does not recognize `"none"`, and rejects multi-token breakers; `field:stop` is deserialized and carried to `ServerGenerateOptions` but only the default-off OpenXLA worker consumes it, so the shipped MLX decode path never truncates; `field:n_predict` is `Option<usize>` so b10621's documented `-1` is rejected during deserialization; `--alias` takes one opaque id where b10621 takes a comma-separated list, matching the `--api-key` precedent already deferred for that reason. Seeded states move from 59 supported / 317 deferred to 43 / 333.
- [x] `aliased` was structurally unrepresentable for routes and native request fields (HIGH), even though the docs promise it for both. The conformance test asserted `mlxcel.route == entry id` and `mlxcel.field == b10621 field`, so a claim naming a different mlxcel identity could not exist, while the validator required a non-empty mapping. Both halves now accept a differing identity and assert it in both directions (the alias must resolve or be declared, the b10621 identity must not), and the validator rejects an `aliased` entry whose mapping is the b10621 identity itself.
- [x] The option half of the gate had no negative direction (MEDIUM). Routes and native fields already failed when mlxcel implemented them without flipping the entry; options did not, which is the epic's central risk with twenty-two chains adding flags. An option entry carrying no claim must now be accepted by neither binary. It passes as seeded (zero violations across all 195 unclaimed entries), so it locks in the current truth rather than papering over a gap.
- [x] The `--slots` note claimed mlxcel does not bind `LLAMA_ARG_ENDPOINT_SLOTS` (MEDIUM, misleading #1440). It does, through `env_fallback_endpoint_slots`, called from both entry points. Note corrected and the runtime env binding claimed with its covering test.
- [x] Nothing asserted the operator-facing surface is unchanged (LOW). Added a test that neither binary's `--help` renders the sentinel.
- [x] Narrower divergences recorded in `notes` without changing state: the checkpoint's `generation_config.json` overriding the effective `--temp` / `--top-k` / `--top-p` server default, `dry_base` below 1.0 used verbatim rather than clamped, `--metrics` leaving `/metrics` unmounted (404) instead of answering with a diagnostic, `--verbose` losing to `RUST_LOG`, `--help` not accepting the `--usage` twin, `GET /health` reporting queue saturation as 503 where b10621 reports it on `/slots`, the `GET /models` response-field subset, and the rerank envelope missing `"object": "list"`.

### Remaining Items

- [ ] `pin.json` records no machine-readable shard ownership (MEDIUM). The shard-to-chain-and-issue assignment exists only as a prose table in `docs/llama-server-compat.md`. It is collision-free as seeded (every shard's issue set maps to exactly one chain, verified entry by entry), but nothing enforces it. Recommend a `shard_owners` block in `pin.json` plus a validator rule that every entry's `issue` belongs to its shard's owner set, and preserving that block in the extractor alongside `mlxcel_baseline`. Not done here because it is an additive design decision better made by the epic owner.
- [ ] No comparison between b10621's recorded `default` and the mlxcel claim's `defaults` (MEDIUM). CI catches mlxcel drifting away from its own recorded value, but not a `supported` claim whose recorded default never matched b10621's. The upstream field is free-form help text, so this needs a normalizing comparison for the numeric cases. All current pairs agree.
- [ ] The `mlxcel` claim-block key names are undocumented and unvalidated (MEDIUM). Neither gate rejects unknown keys, so a downstream chain writing `accepted_spelling` (singular) gets a claim that checks nothing, silently. Worth an allowlist in the validator plus the key names spelled out in the doc.
- [ ] `accepted_on_one_binary_only` resolves its binary key by substring (LOW). A key of `"mlxcel"` silently matches `mlxcel serve`. Exact-match with a panic on an unknown key would be safer.
- [ ] `test` ids are only checked for file existence, not that the named function exists (LOW).
- [ ] Two arguable shard placements (LOW). `--backend-sampling` sits in `ggml-runtime.json` under #1445 (chain B) while sampling semantics are #1436 (chain C), and `--spec-draft-type-k/v` sit in `speculative.json` under #1433 (chain C) while KV-format rejection is #1445 (chain B). Each stays inside one chain's file so the collision-free property holds, but #1445's "reject unsupported KV formats" would naturally cover the draft twins and would then have to edit another chain's shard.
- [ ] `--metrics` remains `supported` while its only observable effect is the `deferred` `GET /metrics` route (LOW). The split is documented in the entry's note, but `--props` and `--slots` were deferred for milder versions of the same class. Left for #1440 to adjudicate.
- [ ] `nightly-verify.yml` runs `verify-fmt` / `verify-clippy` / `verify-test` / `verify-test-video` individually, so `verify-llama-compat` is not in the nightly (LOW). Consistent with `verify-versions` and `verify-kernel-dtype-keys`, which are also absent and covered by their own CI jobs, and this PR adds the equivalent job. The workflow comment saying the steps are `make verify` split apart is now stale for three targets.
- [ ] `docs/llama-server-compat.md` is not in CLAUDE.md's reference-document table (LOW), though llama-server compatibility is one of the project's stated core principles.

### Verification

- [x] Integrated into project code flow (`make verify`, a CI job matching the `check_crate_versions.py` / `check_kernel_dtype_keys.py` precedent exactly, and a runtime interface on both shipped binaries)
- [x] Project conventions followed (upstream references are full GitHub URLs at the pinned commit, no local checkout paths, no unqualified cross-repo refs, no em dashes, no AI attribution, no hard-wrapped body text)
- [x] Existing modules reused (`resolve_repo_binary`, `ModelProvider::recording_for_route_tests`, the `verify-*` target and CI-job pattern); complements `tests/cli_help_consistency.rs` rather than duplicating it, since that suite pins the two rendered `--help` surfaces against each other while this one pins manifest claims against the dumped surfaces including hidden arguments
- [x] Tests pass: `make verify-llama-compat`, `cargo fmt --all --check`, scoped clippy with `-D warnings`, `--test llama_compat_manifest` (4 tests, including the archive conformance test against the official b10621 archive), `--lib server::llama_compat_tests`, `--lib cli::flag_surface`, `--test cli_help_consistency` (25 tests)

inureyes:

Five findings, all fixed in 8c3749ce2. The `--dump-flag-surface` interface, the archive handling in the extractor, and the Rust conformance tests were each examined and found clean on the points they were flagged for; details below.

### MEDIUM: the "referenced test exists" gate could be satisfied from outside the repository

`scripts/ci/check_llama_compat_manifest.py` checked `(REPO / value).is_file()`, which does not answer "is this file in the repository". Under pathlib join semantics an absolute value discards `REPO` entirely, so a manifest entry carrying `"test": "/etc/passwd"` passed the gate (confirmed empirically), and a `..`-laden relative value walks out of the tree the same way. `mlxcel.env_test` had the same hole. The acceptance criterion "CI fails when a not-applicable entry lacks a diagnostic/documentation test" was therefore satisfiable by a path that is not in the tree, and the check doubled as a filesystem-existence oracle on the CI runner. Both now go through `repo_relative_file`, which requires a repo-relative POSIX path with no upward component and confirms the resolved path is still under `REPO`.

### MEDIUM: the extractor verified a file it never used

`--archive` hashed the tarball, but `run_help` then executed the separately supplied `--llama-server` path with `DYLD_LIBRARY_PATH` and `LD_LIBRARY_PATH` pointed at its own directory, so every dylib beside the executed binary is loaded too. Nothing bound the verified archive to the executed artifact, and `--archive` was optional, so the default invocation executed an unverified binary with no signal at all. `docs/llama-server-compat.md` claimed "`--archive` verifies the download against the pinned SHA-256 before anything else", which overstated the guarantee. After verifying the archive digest, the binary must now be byte-identical to a member of that archive, compared through `tarfile` in read mode so nothing is ever extracted or written to disk; omitting `--archive` warns on stderr; the doc sentence was corrected.

Everything else asked about the extractor was already correct and is unchanged: it never extracts the archive (so there is no path-traversal or symlink handling to get wrong), it invokes `llama-server` through `subprocess.run` in list form with no shell, verification runs before any use of the binary, and a mismatch does `raise SystemExit` and aborts.

### MEDIUM: route and native-field inventories had no frozen invariant

Help entries, spellings, and `LLAMA_*` variables are pinned to constants in the validator (249 / 323 / 134), but routes (53) and native request fields (74) were only compared against `pin.json` counts. `pin.json` is written by the same extractor run that writes the shards, so a `parse_routes` or `parse_schema` regression that drops entries also lowers the recorded count and the two agree on the wrong number. `EXPECTED_ROUTES` and `EXPECTED_NATIVE_FIELDS` now sit alongside the other three; both were confirmed to fire when perturbed.

### LOW: `--check-issues-open` cost about 14s of serial CI time per run

One `gh issue view` round trip per deferred issue, issued serially at roughly 0.63s each. It now runs through a bounded thread pool with results collected in sorted issue order so the error output stays deterministic: 14.5s to 2.8s at the current 23 issues. A missing `gh` binary is now a validation error rather than an `OSError` traceback.

### LOW: validator robustness on hand-edited entries

`isinstance(True, int)` is True in Python, so a JSON `true` passed the deferred-issue check and would have reached `gh issue view` as the literal string `"True"`; issue numbers must now be positive and non-bool. Unchecked subscripts on `long_spellings`, `field` and `id` raised a `KeyError` traceback out of the gate instead of reporting a violation. The duplicate-spelling report used an O(n squared) `list.count` scan (failure path only, n = 323) and is now a `Counter`; each shard is read once instead of twice. In the extractor, an unknown `kind` made every key look like policy so the merge silently overwrote freshly extracted facts with stale ones, and is now a hard error. Hashing is chunked rather than whole-file, all reads and writes pin UTF-8 so the "run it twice, the worktree stays clean" contract does not depend on the ambient locale, and a missing `--llama-server` exits cleanly.

### No finding: `--dump-flag-surface` discloses nothing sensitive

Checked empirically against the current build rather than by reading the serializer. The document carries `id`, `long`, `long_aliases`, `short`, `short_aliases`, `env`, `defaults`, `hidden`, `takes_value`, `heading`, and nothing else. `env` is the variable NAME from `Arg::get_env`, never its value: with `LLAMA_API_KEY` and `HF_TOKEN` set in the environment, neither value appears in either binary's dump. Every default is a static literal (numbers, booleans, `127.0.0.1`, `8080`, `auto`, `entropy-bound`) and none embeds a path, home directory, token, or model-store location; `--api-key`, `--api-key-file`, `--model`, `--models-dir`, `--log-file` and `--lora` all carry an empty `defaults` array. No help text is serialized. Reaching the interface requires the ability to execute the binary locally, and positional matching keeps the sentinel out of the way of ordinary argument values.

### No finding: the conformance tests are linear

Measured, not inferred. `tests/llama_compat_manifest.rs` runs in 3.7s, dominated by the seven subprocess spawns of the two large binaries rather than by manifest traversal; every lookup goes through a `BTreeMap`, so the passes over 376 entries and 323 spellings are linear. `src/server/llama_compat_tests.rs` runs in 0.01s despite rebuilding the axum router once per probe across 53 route entries, so the obvious quadratic-looking shape there costs nothing in practice. The offline validator is 0.25s.

### Validation

`make verify-llama-compat` passes. `check_llama_compat_manifest.py --check-issues-open` passes in 2.8s with all 23 linked issues open. `repo_relative_file` was unit-checked against absolute, upward-walking, backslash, empty, non-string and valid inputs. The archive binding was exercised against a synthetic archive covering the matching, tampered-binary and bad-digest paths, confirming nothing is written outside it. `cargo test --profile test-fast --features metal,accelerate` passes for `--test llama_compat_manifest` (4), `--lib server::llama_compat_tests` (2) and `--lib cli::flag_surface` (5).

PR #1460, opened 2026-08-27
test(deepseek-v4): run the HiSA hierarchy on real weights, add its kill switches (#549, #550)

### LINKED ISSUE #549: feat(deepseek-v4): implement HiSA hierarchical sparse attention

### LINKED ISSUE #550: feat(deepseek-v4): clamp shared-expert SwiGLU to swiglu_limit

inureyes:
## Summary

Closes out the two DeepSeek-V4 follow-ups that #523 left open. Both issues were filed on 2026-07-01 against a module that did not exist yet; #523 (PR #1455) has since implemented the substance of both, so their bodies were rewritten against the merged code before this work started. What was actually missing was validation and operability, not implementation.

## The gap this closes

The existing long-context real-model gate runs at 2234 tokens. At ratio 4 that is 558 pooled rows: past `index_topk` (512), so the sparse split-softmax combine engages, but short of `index_block * index_keep` (1024), so selection takes the **flat scan**. Both hierarchical selection functions, `hisa_select_decode` and `hisa_select_batched`, had therefore only ever run on synthetic unit fixtures, never on real weights. The engagement bands at ratio 4:

| pooled rows | tokens | path |
|---|---|---|
| <= 512 | <= ~2048 | dense over pooled, selection unused |
| 512 to 1024 | ~2048 to ~4096 | flat scan |
| >= 1024 | >= ~4096 | HiSA hierarchy |

## What changed

- `MLXCEL_V4_FLAT_INDEX` forces the flat scan, and `MLXCEL_V4_DENSE` keeps sparse layers on the dense local+pooled concat past `index_topk`, following the `MLXCEL_DSA_DENSE` precedent in `deepseek_v32`. The flat-scan switch is the bisection tool a long-context regression needs: both paths score the same pooled rows the same way, so running one prompt each way separates a hierarchy that retained the wrong blocks from a sparse combine that is wrong.
- Both reads are cached in a `OnceLock`, which is right for a value consulted per sparse layer per forward and makes the gate untestable in process. The decision therefore moved into a pure `use_hierarchy()`, which also gives the `i64` widening its own tested home: an `i32` product that wraps lets the gate pass while `nb` is 0, which drives `kb` to 0 and reaches `argpartition(kth = -1)`, an MLX throw and so an uncatchable abort.
- `deepseek_v4_real_model_long_context_engages_hisa_hierarchy` clears `ratio * index_block * index_keep` tokens (4410 in practice, ~1102 pooled rows) so prefill takes the batched hierarchical path and decode takes the fast path. Thresholds are read from the real `config.json` rather than hardcoded, so a config change fails the test instead of silently downgrading it to the flat scan again.
- `hisa_selection_never_returns_a_tail_row` makes the single tail row the most attractive candidate in the pool, then asserts the flat scan takes it and neither hierarchical path does. The positive control is the point: the existing agreement test plants its score mass inside the retained blocks, so it cannot distinguish a tail excluded by design from one that was unreachable anyway.
- `routed_and_shared_experts_clamp_swiglu_identically` (#550) reduces the MoE block to `2.5 * shared(x)` with one routed expert and `norm_topk_prob`, an identity that holds only if both expert kinds clamp the SwiGLU pre-activations the same way.

## A real-model test-harness bug this surfaced

Adding a fourth gate to `tests/deepseek_v4_real_model.rs` made the suite OOM. Each gate loads its own ~96 GB copy of the checkpoint; three fit on a 512 GB host and the fourth pushes the long prefill's activations past the limit, at which point Metal aborts the process with `Insufficient Memory (kIOGPUCommandBufferCallbackErrorOutOfMemory)`. That surfaces as a SIGABRT from C++ rather than a test failure, so it reads like a port bug rather than a host-capacity one. The documented invocation now carries `--test-threads=1` and says why.

## Test plan

- [x] `cargo test --lib models::deepseek_v4`: 42/42 (was 39).
- [x] **Negative control on the clamp identity test**: disabling the clamp on the routed path alone (`limited_swiglu(&gate, &up, 0.0)` in `LimitedSwitchGlu::forward`) makes `routed_and_shared_experts_clamp_swiglu_identically` fail, so it is not passing for the wrong reason. Reverted.
- [x] `cargo clippy --lib --tests -- -D warnings`: exit 0. `cargo fmt --check`: clean.
- [x] Real checkpoint, all four gates, serially:

```
cargo test --test deepseek_v4_real_model --release --features metal,accelerate -- --ignored --nocapture --test-threads=1

[deepseek-v4] hisa prompt: 4410 tokens -> ~1102 pooled rows at ratio 4 (hierarchy needs 1024)
[deepseek-v4] hisa answer: " The Pacific Ocean is the largest ocean on Earth. The Nile"
test deepseek_v4_real_model_long_context_engages_hisa_hierarchy ... ok

test result: ok. 4 passed; 0 failed; finished in 262.08s
```

## The fourth acceptance criterion, measured and not met

#549 also asked that decode step time stay roughly flat as context grows past the hierarchy threshold, on the reasoning that the hierarchy scores `nb` representatives plus `index_keep * index_block` candidates, both constant in `Np`. That claim does not survive measurement, so this PR reports it rather than quietly dropping it.

| prompt tokens | pooled rows | path | ms/token | with `MLXCEL_V4_FLAT_INDEX=1` |
|---:|---:|---|---:|---:|
| 2257 | 564 | flat scan | 659.3 | 665.6 |
| 4513 | 1128 | hierarchy | 858.1 | 854.9 |
| 8900 | 2225 | hierarchy | 1634.8 | 1640.3 |

Pooled rows x1.97 gives ms/token x1.91, and forcing the flat scan changes nothing within 0.4% at every length. The switch was confirmed active by its stderr notice, so this is a real no-difference rather than a no-op flag.

The cause is the coarse stage: `hisa_select_decode` and `hisa_select_batched` both rebuild their block-mean representatives from the entire pooled prefix on every call. Scoring is `O(nb + Kb*b)` exactly as specified; the input to that scoring is not. So the hierarchy trades "score `Np` rows" for "mean-reduce `Np` rows, then score about `Kb*b`", and both are `O(Np)` traffic over the same buffer. The upstream reference has the identical structure, so this is the algorithm as specified rather than a defect introduced here, and fixing it is an optimization beyond the reference. Filed as **#1531**, where the fix is exact: pooled rows are append-only, so a completed block's mean never changes and only the trailing partial block needs recomputing.

Two corrections this forced, both worth naming because each would have shipped a false pass:

- The test's original assertion was `growth < pooled_growth`, which passed at 1.91 against 1.97 and would have passed on anything sublinear by a hair. It was a flatness gate that could not detect the absence of flatness. The test is renamed `deepseek_v4_real_model_hisa_decode_cost_scaling` and now reports the numbers while asserting only against a pathological regression (cost outrunning the prefix itself).
- The kill switches originally did their work silently. An A/B whose two arms are secretly the same arm produces a confident wrong conclusion, so both now announce themselves once on stderr, and the table above is only trustworthy because of it.

## What this does not prove

The retrieval assertion uses a fact that is also world knowledge, so a correct answer does not prove the answer came through the pooled path. A purely synthetic key would test retrieval harder and be markedly less stable on a base checkpoint at this length. The gate's load-bearing claim is the threshold arithmetic, which is asserted directly against the real config: at 1102 pooled rows the hierarchy is the path taken.

Three of #549's four acceptance criteria are met here; the fourth is measured above and carried to #1531.

Closes #549
Closes #550


inureyes:
Correcting the record on this PR's "fourth acceptance criterion, measured and not met" section. The conclusion there is wrong, and so are the numbers behind it.

The measurement timed a whole `generator.generate()` call and divided by the tokens generated. That call includes **prefill**, and at 8900 tokens prefill is roughly 37 seconds against about 4 seconds of decode. So the column labelled ms/token was prefill cost divided by a decode-step count. It tracked the pooled prefix because prefill tracks prompt length, not because selection scales with `Np`, and it made the `MLXCEL_V4_FLAT_INDEX` arm look identical for the same reason: prefill dominated both and masked whatever difference existed.

Timing two decode lengths per context and subtracting so prefill cancels, same checkpoint and host:

| prompt tokens | pooled rows | prefill | decode, hierarchy | decode, `MLXCEL_V4_FLAT_INDEX=1` |
|---:|---:|---:|---:|---:|
| 4513 | 1128 | ~18.3 s | 98.3 ms/step | 98.1 ms/step |
| 8900 | 2225 | ~36.9 s | 99.6 ms/step | 97.3 ms/step |

Decode is flat in `Np`: x1.01 against a x1.97 growth in the prefix. **#549's fourth criterion is met**, and #549's body has been corrected to say so.

Consequences:

- **#1531 has been closed as invalid.** It was filed against the bad numbers and would have optimized a non-problem. Its caching argument stays sound in principle (pooled rows are append-only, so a completed block's mean is immutable) and is worth refiling if a workload ever makes selection cost visible.
- The claim in this PR body that the coarse stage's `O(Np)` representative rebuild causes decode to scale is unsupported. The rebuild is still `O(Np)` in principle; it is simply not measurable beside the per-step expert weight traffic.
- The hierarchy and the flat scan being indistinguishable is a statement about where decode time goes at these lengths, not evidence that the hierarchy fails to help.

The test has been corrected in a follow-up: it subtracts prefill, warms up before timing (the first `generate()` of a process pays one-time init, which produced a negative decode cost at the shortest context), and asserts flatness as a real gate. The previous assertion, `growth < pooled_growth`, passed at 1.91 against 1.97 and would have passed on anything sublinear by a hair, so it could not have caught either the original error or a genuine regression.

inureyes:
Retracting the numbers in my previous comment. They do not reproduce, and the method behind them is unsound.

Re-running the same measurement on the same checkpoint and host gives, at the 4513-token point:

| run | decode | prefill (same prompt) |
|---|---:|---:|
| A | 98.3 ms/step | ~18.3 s |
| B | 369.5 ms/step | ~16.1 s |
| C | 186.4 ms/step | ~38.4 s |

A 2x to 4x spread on the quantity being measured, and prefill for an identical prompt ranging from 16 to 54 seconds across runs.

The method is why. `t_short` and `t_long` are two separate `generate()` calls, each paying its own prefill. Subtracting them cancels prefill only if the two prefills are equal; when they differ by δ, the whole δ lands in the decode estimate divided by the step delta. With prefill varying by tens of seconds, the subtraction does not cancel prefill variance, it amplifies it. So this approach is not a fix for the original prefill-dominated measurement, it is a second flawed measurement.

**What stands and what does not:**

- The original measurement was definitely wrong. Timing one whole `generate()` and dividing by generated tokens reports prefill cost, and that part of the diagnosis holds.
- "Decode is flat in `Np` (x1.01)" does **not** stand. I asserted it as established; it was one noisy sample.
- Therefore #549's fourth criterion is **unverified**, neither met nor unmet, and I have corrected #549 again to say exactly that rather than flipping it a second time on thin evidence.
- #1531 stays closed, but on the narrower and defensible ground that its premise was never established, not on a claim that decode is provably flat.

This repository already has the right tool for the question: `src/bin/bench_decode.rs` (`mlxcel-bench-decode`) loads the model once, warms up, and measures in one process, and its header documents precisely the cold-prefill artifact I walked into. Answering this properly means driving that harness at several context lengths with and without `MLXCEL_V4_FLAT_INDEX`, not an ad-hoc subtraction inside an integration test.

The test has accordingly been demoted from a gate to a reporting harness with the caveat recorded in its doc comment, since a gate whose measurement swings 4x would fail at random on CI and teach people to ignore it.

Issue #1448, opened 2026-08-26
fix(server): align logging, introspection, completion, and preset flags


inureyes:
## Parent

- #1431

## Problem

b10621 provides cache listing, Bash completion generation, log file/color/verbosity/prefix/timestamp controls, prompt-directory logging, and several built-in model presets. mlxcel overlaps only part of this surface, uses a noncanonical log-file environment variable, and must avoid accepting observability options whose output format or security behavior differs.

## Scope

- Align `--log-disable`, `--log-file`, colors, verbose/verbosity, prefix, timestamp, and canonical environment bindings.
- Define log destination precedence and rotation/error behavior without silently losing logs.
- Implement secure `--log-prompts-dir` handling or reject it; include permissions, redaction policy, atomic files, and retention limits.
- Implement or explicitly reject `--cache-list` and `--completion-bash` with stable, testable output.
- Classify built-in EmbeddingGemma, FIM Qwen, GPT-OSS, vision Gemma, and speculative preset flags against mlxcel's direct checkpoint support.
- Keep human logs separate from machine-readable HTTP/SSE responses.

## Acceptance criteria

- [x] Each logging and introspection option has exact CLI/environment/default tests.
- [x] Log files and prompt logs use restrictive permissions and never contain API keys or repository tokens.
- [x] Invalid/unwritable destinations fail at startup instead of falling back silently.
- [x] Shell completion output is syntax-checked and includes visible compatibility options without exposing hidden unsafe aliases.
- [x] Preset flags either resolve to documented immutable checkpoint identifiers or fail with an explicit download example.

Issue #1438, opened 2026-08-26
feat(server): implement b10621 router mode and resolve models-dir semantics


inureyes:
## Parent

- #1431

## Problem

`--models-dir` is currently a semantic collision. In b10621 it enables router-server model discovery and management, while mlxcel treats it as the root of its local model store. b10621 also exposes model presets, aliases, autoload limits, load/unload/delete routes, and router event streams that mlxcel does not provide.

## Scope

- [x] Reserve `--models-dir` and its canonical environment binding for b10621 router semantics.
- [x] Move the current mlxcel model-store-root behavior to a clearly native option with a documented migration path.
- [x] Implement directory discovery, model metadata, stable aliases, preset/config loading, maximum-loaded-model policy, and autoload behavior. (`--models-preset` INI presets are translated since the cache-source PR: sections define models or overlay configuration, `[*]` cascades globally, explicitly-given CLI flags win, and untranslated keys fail startup with a diagnostic.)
- [x] Implement compatible router load, unload, delete, status, and server-sent-event routes. (The cache source exists since the cache-source PR: the mlxcel model store lists as removable `source: "cache"` entries, `POST /models` downloads into it with SSE progress, and `DELETE /models` removes from disk with containment checks.)
- [x] Define concurrency and eviction behavior when active requests reference a model being unloaded.
- [x] Prevent path traversal and symlink escapes outside the configured models directory.

## Acceptance criteria

- [x] Starting with `--models-dir` exposes the compatible router model inventory rather than changing an unrelated downloader store root.
- [x] Load/unload/delete operations have compatible status codes and event ordering.
- [x] Alias collision, invalid preset, missing checkpoint, capacity, active-request, and restart recovery cases are tested. (Preset parsing, unknown-key refusal, alias routing, dedup hiding, and download/cancel/remove flows are covered by unit and route tests plus a real-checkpoint E2E.)
- [x] Directory traversal, symlink escape, and unauthorized management-route tests pass.
- [x] Existing mlxcel store-root users receive a migration diagnostic and replacement option.

## Validation

- Exercise two small real checkpoints through discovery, alternating requests, unload/reload, and process restart.

## Resolution of the first PR's deferred remainder

[...]

Issue #1435, opened 2026-08-26
feat(server): classify and implement optional UI, tools, MCP, and GCP surfaces


inureyes:
## Parent

- #1431

## Problem

b10621 includes a Web UI/static path, built-in tool and MCP endpoints, CORS proxy support, agent mode, and generated GCP-compatible routes. mlxcel currently lacks these surfaces or accepts related hidden flags without providing the requested behavior. Treating every surface as mandatory would also overstate what is useful or safe for an inference runtime.

## Classification

Each surface below was checked against the pinned b10621 sources (`common/arg.cpp`, `tools/server/server.cpp`, `tools/server/server-http.cpp`, `tools/server/server-tools.h`, `tools/server/server-mcp.h`, `tools/server/server-cors-proxy.h`, `tools/server/README.md`, `tools/server/README-dev.md`). The compatibility-policy state follows #1431.

| Surface | b10621 flags / routes | What it does upstream | mlxcel state | Decision |
|---|---|---|---|---|
| Web UI | `--ui` / `--no-ui` (`--webui` / `--no-webui` aliases, `LLAMA_ARG_UI`), `--ui-config`, `--ui-config-file` | Serves an embedded SvelteKit bundle at `/` that is coupled to `/props`, `/slots`, and `/tools` | `--no-webui` is accepted as a hidden no-op; the other spellings are rejected by the parser with no guidance | Not applicable. Accept `--no-ui` and `--no-webui` as inert forms; reject `--ui`, `--webui`, `--ui-config`, `--ui-config-file` with a diagnostic that states mlxcel ships no web UI and points at the `/v1` API |
| Static path | `--path` (`LLAMA_ARG_STATIC_PATH`) | Only takes effect when the UI is enabled; replaces the embedded bundle with a directory mounted at `/`. With `--no-ui` upstream ignores it too | Absent | Not applicable. Reject with the same UI diagnostic; there is no independent static-serving feature to alias to |
[...]

Issue #1431, opened 2026-08-26
epic(server): close llama-server b10621 compatibility gaps


inureyes:
## Objective

Make `mlxcel-server` a truthful, testable compatibility target for the official llama.cpp `llama-server` nightly `b10621`, without treating successful argument parsing as equivalent to behavior parity and without silently ignoring options whose values have observable semantics.

## Frozen reference

- Release: https://github.com/ggml-org/llama.cpp/releases/tag/b10621
- Published: 2026-08-25
- Build: 10621
- Commit: `c1d0e7a004015f23bc0233470b747b596f29b264`
- macOS arm64 archive: `llama-b10621-bin-macos-arm64.tar.gz`
- SHA-256: `429c8270608600188035e5e92f7d78dffb7900904fe7dd7e6a84f48068cd13cf`

## Audit summary

The distributed binary exposes 249 help entries containing 323 distinct long-option spellings and 134 `LLAMA_ARG_*` / `LLAMA_API_KEY` environment variables. The current mlxcel help exposes 149 distinct long-option spellings, with 47 exact long-option spelling intersections and 15 exact environment-variable intersections. These raw counts are not a parity score because many upstream aliases map to one feature and many GGML hardware flags are not meaningful on MLX, but they establish that the current “drop-in replacement” surface is incomplete.

The higher-risk findings are semantic collisions rather than missing names: `LLAMA_ARG_CACHE_REUSE` is parsed as a boolean although upstream defines an integer chunk size; `--timeout` controls a decode watchdog rather than HTTP read/write timeout; `--models-dir` names mlxcel's model-store root rather than llama-server router mode; `/completions` and `/embeddings` currently select OpenAI-shaped handlers while b10621 uses native llama-server response schemas; `--cache-type-k/v` use overlapping names with incompatible accepted values; and several hidden flags are accepted and ignored even when the requested value would change upstream behavior.

## Compatibility policy

Every b10621 surface must end in exactly one documented state:

1. **Supported:** the spelling, value domain, default, precedence, and externally observable behavior match.
2. **Correctly aliased:** a different mlxcel implementation provides equivalent behavior, with a tested translation and no loss of requested semantics.
3. **Not applicable:** the concept has no MLX/CUDA equivalent; mlxcel accepts only semantically inert forms or rejects the option with an actionable diagnostic, and documents why.
4. **Deferred:** a linked implementation issue exists with explicit acceptance criteria; the current binary must not claim support by silently ignoring it.

[...]

PR #1496, opened 2026-08-28
fix(server): repair three regressions on merged main after wave 3

inureyes:
Three regressions the orchestrator's full `make verify` found on merged `main` at `df3accb61`, all confirmed present on a clean checkout of that commit before any fix, so none is an artifact of another.

## The duplicate `--no-webui` declaration

Both binaries declared the long name `no-webui` twice, and clap resolves a duplicate silently in this profile, so the second definition shadowed the first with no error. A bare `_no_webui: bool` compatibility field predates this epic on `ServeArgs` and the server's `Cli`; #1435 then implemented the surface properly, with `UiCompatArgs::no_ui` declaring `--no-ui` and `no-webui` as an alias alongside the twelve enabling forms that are now refused with a diagnostic. The stale field is removed rather than the new alias renamed, because the new declaration carries the classification, the diagnostic and the tests while the old one carried nothing but its own acceptance. Verified on the built binary: `--no-webui` and `--no-ui` are still accepted and inert, and `--webui` is still refused with the `mlxcel ships no web UI` diagnostic.

## The store-root resolver test

#1438 resolved the `--models-dir` semantic collision deliberately: on the two llama-server surfaces that spelling now selects b10621 router-mode discovery, and the mlxcel store-root meaning moved to `--model-store-root`. `legacy_server_mode_resolves_repo_id_from_models_dir_override` predates the swap and still asserted the old meaning, so it drove the resolver into a network fetch for a repo id that exists nowhere. It moves to the new spelling and is renamed to match.

A second test pins the other half of the swap, because renaming alone would have left the reserved spelling unguarded: nothing would notice if `--models-dir` silently went back to resolving a repo id against itself as a store, since the renamed test no longer uses that spelling. `models_dir_no_longer_acts_as_the_model_store_root` asserts it does not, accepting either an error or a different resolved path, since the router branch may legitimately refuse the combination rather than resolve it.

## The `--cache-list` format test

`cache_list_reports_the_store_in_b10621s_format` reported ten cached models where it seeded and expected two. The test was already careful, seeding a temporary store and naming it on the command line rather than through the environment precisely so the list it asserts is the one that invocation would load from, but it named that store with `--models-dir`. After #1438 the flag no longer pointed anywhere, so `--cache-list` fell back to the developer's real cache.

This one is worth calling out as a class. It is a cross-chain collision at the level of flag meaning rather than source text: #1448 and #1438 ran concurrently in different worktrees of the same wave, `--models-dir` was still the store root when the test was written and was not once both had merged. Manifest shard ownership keeps two chains out of each other's files but cannot see a shared flag changing meaning underneath them. The failure is also order-dependent, which is why it did not fail at merge time: it passed while the real cache happened to hold few enough models, and only broke once sibling chains downloaded checkpoints for their own real-model validation.

## Verification

`make verify` on this branch: exit 0, 113 binaries, 9754 passed, 0 failed, 320 ignored. The same gate on merged `main` reported 2 failed, then 1 failed after the first fix.

Part of #1431.

### DIFF
--- a/src/bin/mlx_server.rs
+++ b/src/bin/mlx_server.rs
@@ -2713,7 +2713,32 @@ mod tests {
     }
 
     #[test]
-    fn legacy_server_mode_resolves_repo_id_from_models_dir_override() {
+    fn legacy_server_mode_resolves_repo_id_from_model_store_root_override() {
+        let tmp = tempfile::tempdir().unwrap();
+        let models_root = tmp.path().join("custom-model-store");
+        let repo_id = "zz-mlxcel-test-owner/zz-mlxcel-test-model";
+        let expected = make_complete_snapshot(&models_root, repo_id);
+        let models_root_arg = models_root.to_string_lossy().to_string();
+
+        let args = parse_server_args(&[
+            "mlxcel-server",
+            "-m",
+            repo_id,
+            "--model-store-root",
+            &models_root_arg,
+        ]);
+        let input = build_startup_input(args).expect("repo-id should resolve from override store");
+
+        assert_eq!(input.model_path, expected);
+    }
+
+    /// The other half of #1438's `--models-dir` swap: the spelling that used to
+    /// mean "store root" now selects b10621 router-mode discovery, so it must
+    /// NOT resolve a repo id against itself as a store. Without this the swap
+    /// could silently regress back and only the renamed test above would notice,
+    /// which it would not, because it no longer uses the spelling.
+    #[test]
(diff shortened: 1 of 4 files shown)


inureyes:
Superseded. Chain A's PR #1497 independently fixed the duplicate `--no-webui` declaration and the `--models-dir` store-root bin tests while this branch was in review, so two of the three commits here are now redundant against `main` at `884b6ebe1`. The third, the `--cache-list` format test still naming the reserved `--models-dir` spelling, is still failing on `main` and moves to a fresh single-commit PR.

Issue #1536, opened 2026-08-31
epic: NVIDIA Volta (sm_70) inference acceleration program


inureyes:
## Summary

Program of work to make the CUDA backend usable on NVIDIA **Volta (sm_70)**. The headline finding from a first-principles audit plus an nsys/measured baseline on a Tesla V100-PCIE-32GB: **mlxcel never issues a single tensor-core instruction on Volta.** Every quantized GEMM falls into a scalar-FMA CuTe path, and the activation dtype (bf16) is one that Volta has no hardware for at all — no ALU, no tensor core, no cuBLAS GEMM.

This is the pre-Ampere counterpart to #623 (which targeted GB10/sm_121 and treated sm_80 as the floor). #623 explicitly scoped Blackwell; nothing in that program looked below Ampere, and the arch matrix shipped in the release workflow (`80;86;89;90a;100;120` for x86_64) starts at sm_80. Volta is reachable today only because `build.rs` auto-detects the host compute capability.

## Baseline (measured)

Host: Tesla V100-PCIE-32GB (sm_70, 900 GB/s HBM2, 14 TFLOPS FP32, 112 TFLOPS FP16 tensor), driver 575.51.03, CUDA 12.9.41, mlxcel 0.6.0, MLX pin `9a79573`. Build: `make release-cuda` with `MLX_CUDA_ARCHITECTURES` auto-detected to `70` (verified: `cuobjdump --list-elf libmlx.a` reports 96 cubins, all `sm_70`).

Model: `qwen3.8-27B-4bit` (15 GB, affine 4-bit / group 64; all non-quantized tensors BF16; hybrid GDN + full attention every 4 layers, 64 layers, hidden 5120, head_dim 256).

| Phase | Measured | Roofline | Attained |
|---|---|---|---|
| Decode | **4.2 tok/s** (239 ms/tok) | ~60 tok/s (900 GB/s / 15 GB) | **7%** |
| Prefill | **~7.7 tok/s** (~600-token prompt in 78.2 s) | ~0.41 TFLOPS achieved vs 14 TFLOPS FP32 peak | **~3%** |
| TTFT (3-token prompt) | **~13 s** | — | — |

Decode rate is the slope of two runs (`-n 8` → 14.89 s, `-n 40`/31 generated → 20.39 s); the ~13 s intercept is the fixed first-token cost. Prefill is the delta between a 3-token and a ~600-token prompt at equal `-n`.

[...]

PR #1551, opened 2026-08-31
chore(core): expose CUDA compute capability to the mlxcel runtime

### LINKED ISSUE #1537: chore(core): expose CUDA compute capability to the mlxcel runtime, build, and diagnostics
Part of #1536. Phase 0. Gates #1539, #1541, #1542, #1543, #1544 — all of which need a runtime and/or build-time answer to "what compute capability is this?".

## Context

`grep -rn "compute_capability" src/ --include=*.rs` returns nothing. mlxcel has no notion of CUDA compute capability anywhere in its own code. Every architecture decision is delegated to MLX's C++ gates (`compute_capability_major()` inside `mlx/backend/cuda/**`), which means mlxcel cannot:

- log or report which arch-dependent kernel path a run took;
- gate its own overlay patches or defaults on the host arch;
- warn when a prebuilt binary's cubin set does not cover the host (the release matrix starts at `80`, so a released x86_64 archive will not load on a V100 at all);
- express "this default is correct on Ampere+ and wrong below it", which is the shape of every finding in #1536.

The build side is half-present: `build.rs:392-422` (`detect_cuda_arch`) already shells out to `nvidia-smi --query-gpu=compute_cap` and derives `MLX_CUDA_ARCHITECTURES`, but the result is consumed only by CMake and never surfaced to the Rust crate or to the user.

## Scope

Make compute capability a first-class, testable value in mlxcel. This is plumbing plus diagnostics only — no kernel or dispatch behavior changes in this issue.

Out of scope: acting on the value. Each consumer is its own sub-issue.

## Implementation plan

1. **Runtime probe.** Add a `cuda_compute_capability() -> Option<(u32, u32)>` to `src/lib/mlxcel-core/src/hardware.rs`, backed by a small FFI shim over `cudaDeviceGetAttribute(cudaDevAttrComputeCapabilityMajor/Minor)` (MLX already caches these in `cu::Device`; prefer exposing the existing value through the bridge over a second CUDA call). `None` on non-CUDA builds and when no device is present. Cache in a `OnceLock` — it cannot change within a process.
2. **Build-time record.** Have `build.rs` emit the arch list it resolved as `cargo:rustc-env=MLXCEL_CUDA_ARCHITECTURES=<list>` so the binary knows what it was compiled for, not just what it is running on.
3. **Mismatch check.** At device init, compare the running device's SM against the compiled arch list. If the device SM has neither a matching cubin nor a PTX-forward-compatible entry, fail with a message naming both, instead of the current opaque CUDA load error. This is the check that would have made the "release binaries do not run on V100" case self-diagnosing.
4. **Diagnostics.** Include compute capability and the compiled arch list in the startup device line (next to `Detected N GPU(s).`) and in whatever structured hardware report `hardware.rs` already feeds. Add `MLXCEL_TRACE_ARCH` (presence-enables, matching the conventions on this page in `docs/environment-variables.md`) to print, once per process, the resolved capability, the compiled arch list, and — where MLX exposes it — the quantized-matmul path selected on the first call.
5. **Docs.** `docs/environment-variables.md` entry for the new trace variable; `docs/installation.md` cross-reference from the CUDA architecture selection section to the new mismatch error.

## Acceptance criteria

- [ ] `cuda_compute_capability()` returns `Some((7, 0))` on the V100 host and `Some((12, 1))` on GB10; `None` on a Metal build. (Partially verified in PR #1551: `Some((7, 0))` confirmed on the V100. GB10 and Metal need hardware this host does not have; the Linux CPU-only build exercises the same `None` branch a Metal build takes, but that is not a Metal run.)
- [x] Unit tests for the arch-list parser and the mismatch predicate cover: exact match, PTX-forward-compatible match, `90a`-suffixed entries, and the no-match case. No GPU required for these.
- [x] Running an x86_64 release-matrix build (`80;86;89;90a;100;120`) on the V100 produces the new named error rather than a CUDA load failure.
- [x] `MLXCEL_TRACE_ARCH=1` prints the capability and compiled arch list exactly once.
- [ ] No behavior change on any existing platform with the variable unset; GB10 baseline unmoved. (Verified on sm_70 only in PR #1551. No GB10 or Apple Silicon baseline was measured.)
- [ ] `cargo test --features cuda` green; `cargo test` (no CUDA) green. (PR #1551 ran the scoped `cuda_arch` and `execution::runtime` suites green under both feature sets, plus `cargo clippy --lib --tests -- -D warnings`; the full suites were not run on this host because each cold MLX/CUTLASS/cuDNN configuration takes 30 to 47 minutes. CI covers them.)

## Validation

```
MLX_CUDA_ARCHITECTURES=70 make release-cuda
MLXCEL_TRACE_ARCH=1 ./target/release/mlxcel generate -m ./models/qwen3.8-27B-4bit -p "Hi." -n 4
MLX_CUDA_ARCHITECTURES="80;86;89;90a;100;120" cargo build --release --features cuda   # then run on the V100: expect the named error
```

## References

- Build-side detection that already exists: `src/lib/mlxcel-core/build.rs:358-435`.
- MLX's cached values: `mlx/backend/cuda/device.cpp:42-44`, `device.h:181-185`.
- Release arch matrix: `.github/workflows/release.yml:541` (aarch64 `90a;100;121`), `:764` (x86_64 `80;86;89;90a;100;120`).

inureyes:
## Summary

Makes CUDA compute capability a first-class, testable value in mlxcel. Before this change `grep -rn "compute_capability" src/ --include=*.rs` returned nothing: every architecture decision was delegated to MLX's C++ gates, so mlxcel could not report which architecture it was running on, did not record which architectures it had been compiled for, and failed with an opaque CUDA load error at the first kernel launch when the two did not agree. This is plumbing and diagnostics only. No kernel or dispatch behavior changes; acting on the value is left to the consumers in epic #1536.

## What changed

**Runtime probe.** `mlxcel_core::hardware::cuda_compute_capability() -> Option<(u32, u32)>` (implemented in the new `src/lib/mlxcel-core/src/cuda_arch.rs`, re-exported from `hardware.rs`) returns the running GPU's compute capability, cached in a `OnceLock` because it cannot change within a process. It is backed by a new `gpu_compute_capability(index)` bridge function that reads MLX's own `device_info` map, whose CUDA backend already fills `compute_capability_major` / `compute_capability_minor` from the `cudaDeviceGetAttribute` values it caches per device, so no second CUDA call is issued and no CUDA header is required. The same symbol links on Metal and CPU-only builds, where those keys are absent and the probe reports `None`. Reading device properties never loads a cubin, which is what keeps the probe callable on exactly the binaries the mismatch check has to report on.

**Build-time record.** `build.rs` now resolves the CUDA architecture list once in `main` via a new `resolve_cuda_architectures()`, passes that same string to CMake, and records it in the crate as `cargo:rustc-env=MLXCEL_CUDA_ARCHITECTURES=<list>`. The list CMake compiles for and the list the binary reports are therefore the same string by construction. Empty off CUDA.

**Mismatch check.** `cuda_arch_mismatch()` compares the running device against the compiled list using CUDA's own compatibility rules: a cubin carries forward across minor revisions within one major version, PTX JITs forward including across majors, `a`-suffixed (architecture-specific) entries cover their exact target only, and `f`-suffixed (family-specific) entries carry forward within their major. `enforce_cuda_arch_compatibility()` turns a no-coverage verdict into a named error, and the new `initialize_runtime_checked()` calls it at device init, so `generate`, `chat`, `embed`, `rerank`, `detect` and the server refuse to start with a message naming both the compiled list and the device found, instead of continuing to a CUDA load failure that names neither. Two safety valves are deliberate: an architecture list that parses to nothing (`native`, `all-major`, or a spelling this parser has not learned) is treated as unknown rather than as "covers nothing", so it can never manufacture a startup failure; and `MLXCEL_DEVICE=cpu` bypasses the refusal, since running on the CPU is the one workaround left to someone holding an archive built for the wrong architecture.

**Diagnostics.** The startup device report now carries the architecture picture next to `Detected N GPU(s)`, in both the CLI (`src/commands/generate.rs`) and the server (`src/server/startup.rs`, which also emits it as structured fields alongside the existing `Hardware capabilities detected` record). `MLXCEL_TRACE_ARCH` (presence-enables, matching the other diagnostic switches) prints the resolved capability, the compiled list, and the coverage verdict once per process to stderr. The CUDA `QuantizedMatmul` overlay (`src/lib/mlx-cpp/patches/mlx/backend/cuda/quantized/quantized.cpp`) additionally reports which quantized-matmul path the dispatcher picked on its first call, which is the most architecture-dependent choice in decode and which MLX exposes no hook for; its environment read is cached because it sits on the decode path.

**Docs.** `docs/environment-variables.md` gains the `MLXCEL_TRACE_ARCH` row; `docs/installation.md` cross-references the new refusal from the CUDA architecture selection section, naming the two cases that produce it (a published x86_64 archive on a pre-Ampere card, and a source build made where `nvidia-smi` was unavailable and `build.rs` fell back to `90a`).

## Test plan

Host: Tesla V100-PCIE-32GB (compute capability 7.0, sm_70), CUDA 12.9.41, driver 575.51.03, x86_64. This is the only GPU on the machine, which bounds what could be verified; see the section below.

- [x] `cargo test -p mlxcel-core --release --features cuda --lib cuda_arch` (26 passed). The parser and coverage tests cover exact match, PTX-forward-compatible match, `90a`-suffixed entries, `f`-suffixed entries, `-real` / `-virtual` qualifiers, cubin-preferred-over-PTX, the no-match case, and the error message. None of them need a GPU.
- [x] The same 26 tests on a build without the `cuda` feature (`cargo test -p mlxcel-core --release --lib cuda_arch`). There the recorded architecture list is empty and the capability probe reports `None`, which is the same `device_info`-without-capability-keys path a Metal build takes.
- [x] `cargo test --release --features mlxcel/cuda --lib execution::runtime` (13 passed), including that the refusal tracks `cuda_arch_mismatch()` exactly and that it applies only when the GPU was actually requested.
- [x] `cargo clippy --release --features mlxcel/cuda -p mlxcel-core -p mlxcel --lib --tests -- -D warnings`, `cargo fmt --check`, `git diff --check`.
- [x] `MLX_CUDA_ARCHITECTURES=70 cargo build --release --features cuda`, then `MLXCEL_TRACE_ARCH=1 mlxcel generate -m ./models/mlx-community/qwen3.8-27B-4bit -p "Hi." -n 8`. The probe reports the real device: `[mlxcel arch] compute capability 7.0 (sm_70); compiled for [70]; coverage: cubin`, printed exactly once across 8 decode steps, followed once by `[mlxcel arch] quantized matmul path on first call: qmm_naive (device sm_70)`. The startup device report shows `Detected 1 GPU(s).` then `CUDA compute capability 7.0 (sm_70); compiled for [70] (cubin)`.
- [x] The same command with `MLXCEL_TRACE_ARCH` unset prints neither trace line and generates identically.
- [x] `MLX_CUDA_ARCHITECTURES="80;86;89;90a;100;120" cargo build --release --features cuda --bin mlxcel` (the x86_64 release matrix; its `libmlx.a` is 249 MB against 155 MB for the single-architecture sm_70 build, so the six architectures really are compiled in), then running that binary on this V100. It exits non-zero with the new named error before touching a kernel: `this mlxcel binary was compiled for CUDA architectures [80;86;89;90a;100;120] and cannot run on the GPU in this host, which is compute capability 7.0 (sm_70). No compiled architecture provides a cubin for sm_70, and none is old enough for the driver to JIT its PTX forward, so every kernel launch would fail. Rebuild for this device with MLX_CUDA_ARCHITECTURES=70 ... To run this binary as it is, set MLXCEL_DEVICE=cpu.`
- [x] The same mismatched binary under `MLXCEL_TRACE_ARCH=1` reports `coverage: not covered` before refusing, and under `MLXCEL_DEVICE=cpu` starts normally on the CPU with the startup line reading `compiled for [80;86;89;90a;100;120] (not covered)`.

Not run: the full `cargo test --features cuda` and `cargo test` suites. This unit builds MLX, CUTLASS and cuDNN from source per feature set, and each cold configuration here took 30 to 47 minutes, so validation was scoped to the modules this change touches plus the end-to-end runs above. CI covers the rest.

## Deferred to GB10

This machine has exactly one GPU, a Tesla V100 at compute capability 7.0, and no Ampere-or-later device of any kind. The following acceptance criteria from #1537 therefore could not be exercised here and are left unticked rather than assumed. They belong to the `## GB10 (sm_121) continuation` section of epic #1536.

- `cuda_compute_capability()` returns `Some((12, 1))` on GB10. Only the `Some((7, 0))` half was verified. The unpacking of the bridge's `major * 1000 + minor` encoding and the `121` list entry are covered by unit tests, but no sm_121 device ran this code.
- `cuda_compute_capability()` returns `None` on a Metal build. No macOS host was available. The nearest reachable evidence is the Linux CPU-only build above, which exercises the same branch: `device_info` returns a map without the two capability keys and the probe reports `None`. That is the identical code path, but it is not a Metal run.
- No behavior change on any existing platform with the variable unset, and the GB10 baseline unmoved. Verified only on sm_70, where generation with `MLXCEL_TRACE_ARCH` unset is unchanged. No GB10 or Apple Silicon baseline was measured, and no throughput claim is made for either.

Closes #1537


inureyes:
Follow-up datum on what the check replaces, since the acceptance criterion is phrased as "the new named error rather than a CUDA load failure".

`mlxcel-bench-decode` is the one entry point this PR deliberately leaves on the unchecked `initialize_runtime()`, so building it from the same x86_64 release-matrix tree gives a like-for-like before/after on one binary. Running it on this V100 with the same checkpoint and prompt, it printed `[Load] wall: 0.795 s MLX peak: 0.00 GB` and then produced nothing at all for 15 minutes, holding no GPU memory, until the harness timeout killed it. The checked `mlxcel generate` path on the same tree refuses in well under a second with the named error, and the matching sm_70 build finishes the same 2-token run in about 15 seconds.

I am not attributing the stall to a specific CUDA call, since I did not attach a debugger and MLX also JIT-compiles some kernels with NVRTC at first use, which could account for part of it. The point that stands is the one the issue makes: without the check the failure mode carries no diagnostic naming either the architectures the binary holds or the device it found.

PR #1570, opened 2026-09-02
chore(build): guard webpage-build against missing MkDocs sources

### LINKED ISSUE #1138: chore(build): guard `webpage-build` against the missing MkDocs manual sources
## Problem / Background

`make webpage-build` (`Makefile:809-817`) has the same unmet MkDocs dependency that #1111 documented for the thirteen `docs-*` targets, but it is not a `docs-*` target, so #1111's enumerated scope did not cover it and PR #1122 left it unguarded. It was recorded as a follow-up rather than changed silently.

The target's body is:

```
rm -rf webpage/site/public/en/manual webpage/site/public/ko/manual
uv run zensical build -f mkdocs.yml -d webpage/site/public/en/manual
uv run zensical build -f mkdocs.ko.yml -d webpage/site/public/ko/manual
cd webpage/site && pnpm install && pnpm build
```

`mkdocs.yml` sets `docs_dir: docs/en` and `mkdocs.ko.yml` sets `docs_dir: docs/ko`, and neither directory exists in this checkout. `docs/shared`, `docs/requirements.txt` and `docs/scripts` are also absent, and `docs/overrides` exists but is empty and untracked, so it is absent from a fresh clone. Those manual sources are maintained in a separate documentation tree, exactly as `docs-guard` already explains.

The consequence today is that `make webpage-build` fails partway: the `rm -rf` has already deleted `webpage/site/public/en/manual` and `webpage/site/public/ko/manual`, and only then does zensical error out against a nonexistent `docs_dir`. The failure is opaque in the same way the `docs-*` targets were before the guard.

## Proposed Solution

Add the existing `docs-guard` prerequisite to `webpage-build` so it fails immediately with the same clear message (which already points readers at the published manual at https://mlxcel.lablup.ai/en/manual/) instead of failing partway with a zensical error against a missing `docs_dir`.

Keep the guard a presence check on `docs/en`, not an unconditional refusal, so `webpage-build` still runs exactly as before in a tree that does hold the manual sources. `git show 2c6b3465 -- Makefile` has the exact shape used for the thirteen guarded targets.

Also update the target's `##` help string to carry the same `(manual sources not in this checkout)` suffix the guarded `docs-*` targets use, so `make help` states the dependency rather than making the user discover it by running the target.

Separately, decide whether `webpage-deploy` (`Makefile:819-822`, which runs `scripts/deploy_webpage.sh`) needs the same treatment. The script does not invoke zensical or reference the manual output: it runs `pnpm install && pnpm run build` in `webpage/site` and publishes `out/`. So it does not fail on the missing sources, but it will publish a site whose `/en/manual` and `/ko/manual` paths are missing or stale unless `webpage-build` ran successfully first. Confirm that reading of the script and choose deliberately between guarding it, warning inside it, or leaving it alone with a comment explaining why.

## Acceptance Criteria

- [x] `webpage-build` declares `docs-guard` as a prerequisite.
- [x] On a checkout without `docs/en`, `make webpage-build` fails at the guard before deleting `webpage/site/public/en/manual` or `webpage/site/public/ko/manual` and before invoking `uv`.
- [x] The guard stays a presence check, so `webpage-build` still runs end to end in a tree that does hold the manual sources.
- [x] The `##` help string for `webpage-build` states that it needs the manual sources, matching the `(manual sources not in this checkout)` suffix carried by the thirteen guarded targets.
- [x] `webpage-deploy` is evaluated against `scripts/deploy_webpage.sh`, and the outcome (guarded, warned, or intentionally left alone) is reflected in the Makefile.

## Technical Considerations

- `docs-guard` and `DOCS_MANUAL_DIR` are defined below the webpage section in the Makefile. Make resolves prerequisites and variables after the whole file is parsed, so no reordering is required, but consider whether moving the guard definition above both sections reads better.
- `webpage-dev` (`Makefile:805-807`) only runs `pnpm dev` and does not touch the manual, so it needs no guard.

inureyes:
## Summary

`make webpage-build` deleted `webpage/site/public/{en,ko}/manual` and then failed opaquely inside zensical against a nonexistent `docs_dir`, because it ran the same unmet MkDocs dependency the thirteen `docs-*` targets already guard, but it was not itself a `docs-*` target and stayed out of that earlier pass. This adds `docs-guard` to `webpage-build` so it fails fast with the same explanation instead, and makes a deliberate, documented choice for `webpage-deploy` rather than leaving that question open.

## What changed

- `Makefile`: `webpage-build` now declares `docs-guard` as a prerequisite and its `##` help string carries the same `(manual sources not in this checkout)` suffix as the other guarded targets, so `make help` states the dependency up front.
- `Makefile`: `webpage-deploy` is intentionally left off `docs-guard`, since `scripts/deploy_webpage.sh` never touches zensical or `docs_dir` and cannot fail the way `webpage-build` does, and `docs-guard` checks `docs/en` (a build input) rather than the manual output the script actually reads. It now warns, without blocking, when `webpage/site/public/en/manual` or `.../ko/manual` is missing, and a comment above the target records the reasoning.

## Test plan

- [x] On a checkout without `docs/en`: `make webpage-build` fails at `docs-guard` before the `rm -rf` runs, and `webpage/site/public/{en,ko}/manual` stay absent before and after (`ls` checked both times).
- [x] `make help` shows the new `(manual sources not in this checkout)` suffix on `webpage-build`.
- [x] Overriding `DOCS_MANUAL_DIR` to an existing directory lets the guard pass through unchanged into the `rm -rf` and `uv run zensical` steps (fails only on the unrelated absence of `zensical` in this environment, confirming the guard itself is a presence check, not a hard refusal).
- [x] `make -n webpage-deploy` confirms the new warning block is syntactically valid Make/shell, and the warning condition was exercised in isolation against the current manual-less checkout.

Closes #1138

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -939,7 +939,7 @@ webpage-dev: ## Run download webpage dev server
 	cd webpage/site && pnpm install && pnpm dev
 
 .PHONY: webpage-build
-webpage-build: ## Build download webpage (static export)
+webpage-build: docs-guard ## Build download webpage (static export) (manual sources not in this checkout)
 	@echo "$(CYAN)Building documentation for webpage...$(RESET)"
 	rm -rf webpage/site/public/en/manual webpage/site/public/ko/manual
 	uv run zensical build -f mkdocs.yml -d webpage/site/public/en/manual
@@ -948,9 +948,22 @@ webpage-build: ## Build download webpage (static export)
 	cd webpage/site && pnpm install && pnpm build
 	@echo "$(GREEN)Build complete. Output in webpage/site/out/$(RESET)"
 
+# webpage-deploy is intentionally not gated behind docs-guard. deploy_webpage.sh
+# never invokes zensical or reads docs_dir, so it cannot fail the way
+# webpage-build does, and docs-guard checks for docs/en (a build input) rather
+# than for the manual output deploy actually reads. What deploy needs is the
+# manual already built into webpage/site/public/{en,ko}/manual by a prior
+# 'make webpage-build' (or copied in from elsewhere); a missing docs/en does
+# not imply that output is missing. So this checks the actual precondition and
+# warns without blocking, since a maintainer may be re-deploying an unchanged
+# site from a previous build.
 .PHONY: webpage-deploy
 webpage-deploy: ## Deploy download webpage to GitHub Pages
 	@echo "$(CYAN)Deploying webpage...$(RESET)"
+	@if [ ! -d webpage/site/public/en/manual ] || [ ! -d webpage/site/public/ko/manual ]; then \
+		echo "$(YELLOW)Warning: webpage/site/public/en/manual or .../ko/manual is missing.$(RESET)"; \
+		echo "  Run 'make webpage-build' first, or the deployed site will be missing (or serving a stale copy of) the manual pages."; \
+	fi
 	./scripts/deploy_webpage.sh
 
 # ============================================================================
--- a/TECHNICAL_REPORTS/1570-guard-webpage-build-mkdocs-sources-20260902.en.md
+++ b/TECHNICAL_REPORTS/1570-guard-webpage-build-mkdocs-sources-20260902.en.md
@@ -1,0 +1,173 @@
+# Technical Report: PR #1570 - chore(build): guard webpage-build against missing MkDocs sources
+
+**Date**: 2026-09-02
+**Author**: mlxcel maintainers
+**Reviewer**: automated pr-reviewer and pr-security-checker passes
+**Status**: Completed
+**Languages**: Makefile (GNU Make, POSIX shell)
+**Risk Level**: Low
+
+---
+
+## Executive Summary
+
+`make webpage-build` ran the same unmet MkDocs manual dependency that thirteen `docs-*` targets already guard, but predated the `docs-guard` prerequisite and stayed unguarded through an earlier pass (#1111, PR #1122). On a checkout without the manual sources it deleted `webpage/site/public/en/manual` and `webpage/site/public/ko/manual`, then failed with an opaque zensical error. This PR adds `docs-guard` as a prerequisite of `webpage-build`, matching the existing pattern exactly, and makes a deliberate, documented choice for the neighboring `webpage-deploy` target rather than leaving that question open.
+
+---
+
+## 1. Problem Statement
+
+### 1.1 Background
+
+`mkdocs.yml` sets `docs_dir: docs/en` and `mkdocs.ko.yml` sets `docs_dir: docs/ko`. Those directories, along with `docs/shared`, `docs/requirements.txt` and `docs/scripts`, are maintained in a separate documentation tree and are absent from this repository's checkout. Thirteen `docs-*` targets already declare `docs-guard` as a prerequisite for exactly this reason, so each of them fails immediately with an explanation instead of an opaque `uv`, `ln`, or zensical error. `webpage-build` runs the same zensical build against the same missing `docs_dir`, but because it is not itself a `docs-*` target, it was out of scope for the pass that added the guard and was recorded as a follow-up.
+
+### 1.2 Existing Issues
+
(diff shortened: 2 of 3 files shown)


inureyes:

### Intent

Add the existing `docs-guard` prerequisite to `webpage-build` so it fails fast on a checkout without the MkDocs manual sources instead of deleting `webpage/site/public/{en,ko}/manual` and then failing opaquely inside zensical, and settle the open question about `webpage-deploy` deliberately rather than leaving it unaddressed.

### Findings Addressed

None. No CRITICAL or HIGH findings were found, and no MEDIUM findings warranted a change. Nothing was committed.

### Remaining Items

- [ ] Warnings print to stdout rather than stderr (LOW). This is consistent with `docs-guard` and every other `echo` in this Makefile, so changing it here alone would be the inconsistency. No action recommended.
- [ ] `webpage/site/public/{en,ko}/manual` is spelled literally in the new check rather than factored into a variable (LOW). The pre-existing `rm -rf` line in `webpage-build` already spells both paths literally, so the new code matches the local convention. Extracting a variable would widen the diff past the issue's scope.

### Verification

- [x] All stated requirements implemented. All five acceptance criteria in #1138 verified independently against the working tree, not just read from the diff.
- [x] AC1: `webpage-build: docs-guard ## ...` at `Makefile:942`, prerequisite placed before the `##` marker exactly as `docs-install:1003` and the other twelve guarded targets do.
- [x] AC2: `make webpage-build` in this checkout (no `docs/en`) stops at `docs-guard` with the published-manual message. Confirmed live: the target's own `Building documentation for webpage...` echo never fires, `rm -rf` never runs, `uv` is never invoked, and `webpage/site/public` still holds only `brands/` afterwards.
- [x] AC3: the guard stays a presence check. `make -n webpage-build DOCS_MANUAL_DIR=webpage` passes straight through into the unchanged `rm -rf`, two `uv run zensical build` and `pnpm build` steps.
- [x] AC4: `make help` parses the new line and renders `webpage-build  Build download webpage (static export) (manual sources not in this checkout)` under Webpage Targets. The suffix is byte-identical to the thirteen pre-existing ones (twelve declared `target: docs-guard ##`, plus `docs-pdf` which carries extra prerequisites).
- [x] AC5: `scripts/deploy_webpage.sh` was read end to end and the comment's claims hold. It runs `pnpm install && pnpm run build` in `webpage/site`, then commits and force-pushes `out/` to the `gh-pages` branch of `mlxcel-releases`. It never mentions zensical, `mkdocs.yml` or `docs_dir`, so gating it on `docs-guard` (a `docs/en` presence check) would test the wrong precondition. Warning on the actual precondition, the built manual output, without blocking is the right call and the reasoning is recorded above the target.
- [x] No placeholder/mock code remaining. No TODO, stub, or debug output in the diff.
- [x] Integrated into project code flow. `docs-guard` is a real `.PHONY` target at `Makefile:984`; Make resolves it fine despite being defined below the webpage section, confirmed by the live run.
- [x] Project conventions followed. Recipe lines are tab-indented, `\` continuations are correct, and `$(YELLOW)`/`$(RESET)` are the variables defined at `Makefile:54` and `:56`.
- [x] Existing modules reused where applicable. The change reuses `docs-guard` rather than writing a second presence check.
- [x] No unintended structural changes. `git diff origin/main...HEAD` is one file, 14 insertions and 1 deletion, entirely within `Makefile:941-967`. The `verify-*` targets and every other section are untouched.
- [x] Tests pass. No Rust code changed, so no test run applies. `git diff --check` is clean, `scripts/ci/check_cross_repo_refs.py` reports OK, and CI is green with the Rust jobs correctly skipped by path filtering.

### Behavior checks run

- `make webpage-build` (fails at the guard, which is the intended behavior in this checkout).
- `make -n webpage-build DOCS_MANUAL_DIR=webpage`.
- `make -n webpage-deploy` (the warn block expands to valid `/bin/sh`).
- The warn block executed standalone against this manual-less checkout: it prints both lines and exits 0, so `./scripts/deploy_webpage.sh` still runs unconditionally on the next recipe line.
- `make help`.

Issue #1316, opened 2026-08-22
perf(server): apply structured-output token masks as a packed u32 bitmask instead of a full f32 bias


inureyes:
## Summary

Every constrained-decoding step unpacks the grammar engine's packed token bitmask into a `Vec<bool>` the size of the vocabulary, turns that into a `Vec<f32>` bias of the same size, uploads the bias to the device and adds it to the logits. For a 248k-row head that is two 248k-element host loops plus a 1 MB host-to-device copy per token per constrained sequence, all on the scheduler thread between the forward pass and sampling. The change keeps the mask in its packed form (32 tokens per `u32`, 31 KB for the same vocabulary), uploads that, and expands it on the device with four graph ops feeding `where_cond`, removing both host loops and 97% of the copy.

## Current behavior

`src/server/structured.rs:345` (`compute_mask`) unpacks the packed mask the grammar engine returns:

```rust
self.mask_buf.resize(self.vocab_size, false);
...
vob.iter_set_entries(|idx| { if idx < vocab_size { mask[idx] = true; } });
```

`src/server/structured.rs:736` (`apply_structured_mask_to_logits`) then walks the whole vocabulary a second time to build the bias and ships it:

```rust
constraint.bias_buf.resize(vocab_size, 0.0);
for (i, slot) in constraint.bias_buf.iter_mut().enumerate() {
    let allowed = i < mask_len && constraint.mask_buf[i];
    if !allowed { *slot = f32::NEG_INFINITY; }
}
let bias_arr = mlxcel_core::from_slice_f32(&constraint.bias_buf, &[1, vocab_size as i32]);
Ok(mlxcel_core::add(logits, &bias_arr))
```

`vocab_size_hint` is the logits width (padded head rows, 248320 for Qwen3.8 against a 248077-token tokenizer), so the bias is `4 * 248320 = 993 KB` per step. The scheduler calls this inside `apply_structured_mask` (`src/server/batch/scheduler.rs:3131`) for each constrained sequence on each tick, before sampling, so with B concurrent constrained sequences the cost is B times per tick. The grammar engine's own mask computation is roughly 50 microseconds (`src/server/structured.rs:44`); the unpack, the bias fill and the copy are the part this change removes. The existing `mask_buf` / `bias_buf` reuse (`src/server/structured.rs:288` onward) avoids allocation but not the loops.

[...]

Issue #1314, opened 2026-08-22
feat(server): expose per-request speculative decoding acceptance stats in response timings


inureyes:
## Summary

When a request is served by a drafter (DFlash, MTP, internal MTP), the scheduler computes rounds, proposed and accepted draft-token counts for that request and then only logs them through `tracing`. A client has no way to learn whether speculation helped its own request, which is what operators tuning `--draft-max` / `--draft-block-size` and clients doing A/B comparisons need. The change carries those counters on `GenerationResult` and surfaces them on the `/completion` `timings` object and as an optional `timings` object on chat completions, using the llama-server field names (`draft_n`, `draft_n_accepted`) plus `draft_rounds` and `draft_kind`.

## Current behavior

- `src/server/types/response.rs:552` (`TimingInfo`) has only prompt / predicted counts and rates.
- `src/server/model_provider.rs:152` (`GenerationResult`) carries no speculative field.
- The numbers exist but die in logs:
  - `src/server/batch/speculative_burst.rs:1776`: `diagnostics.rounds`, `proposed_tokens`, `accepted_tokens` logged at `info` from `DFlashDiagnostics` (`src/lib/mlxcel-core/src/drafter/dflash/round_loop.rs:482`).
  - `src/server/batch/scheduler.rs:4808`: `mtp_profile.rounds` / `accepted_draft_tokens` (`MtpBurstProfile`, `src/server/batch/mtp_policy.rs:266`, which also has `proposed_tokens`) logged when a B=1 burst finalizes.
- `GenerateEvent::Done(result)` is built by `seq.decode_state.finish_with_cache(...)` at `src/server/batch/speculative_burst.rs:2091`, `src/server/batch/scheduler.rs:6680` and `src/server/batch/scheduler.rs:8131`, none of which know the diagnostics.
- `/metrics` exposes only aggregate MTP policy state (`src/server/batch/observability.rs`, `acceptance_rate` on the policy snapshot), not per request.

## Expected behavior

`GenerationResult` gains:

```rust
pub speculative: Option<SpeculativeStats>,
pub struct SpeculativeStats {
    pub draft_kind: String,      // "dflash" | "mtp" | "internal-mtp" (DrafterKind::as_str)
    pub draft_rounds: usize,     // verify rounds executed for this request
    pub draft_n: usize,          // draft tokens proposed across all rounds
    pub draft_n_accepted: usize, // draft tokens accepted across all rounds
}
```

[...]

PR #1581, opened 2026-09-02
feat(server): enforce required and named tool_choice on chat (#1319)

### LINKED ISSUE #1319: feat(server): enforce tool_choice required and named-function semantics on chat completions
## Summary

`/v1/chat/completions` accepts `tool_choice` in all four OpenAI shapes but only `none` changes behaviour. `required` is validated and then ignored: the model is free to answer in prose and the response carries no tool call. A named function (`{"type":"function","function":{"name":"f"}}`) still renders every declared tool into the prompt and only filters the parsed calls afterwards, so a model that picks another tool returns an empty `tool_calls` array with `finish_reason: "stop"`. The change makes `required` and the named form actually steer generation: the tool list handed to the template is narrowed, an instruction is injected into the prompt, the request is rejected with a 400 when it cannot be satisfied, and on tool-call formats whose wire shape is JSON the call is grammar-forced through the existing structured-output constraint.

## Current behavior

- `src/server/routes/chat.rs:1149` (`validate_chat_tool_inputs`) accepts `auto`, `none`, `required` and the object form, and that is the last time `required` is looked at.
- `src/server/chat_request.rs:748` (`effective_tools`) strips tools only for `none`:

  ```rust
  if let Some(ref tc) = request.tool_choice && tc.is_none() { return None; }
  request.tools.as_deref()
  ```

  so a named choice still renders every tool.
- `src/server/tool_calls/mod.rs:101` (`build_tool_call_responses`) and `src/server/routes/chat.rs:1026` (the streaming finish path) drop parsed calls whose name differs from the named function, after generation. Nothing prevents the model from choosing another function or from not calling at all.
- `src/server/types/request.rs:66` (`ToolChoice`) has no notion of `required` beyond the string.
- The Anthropic translator (`src/server/anthropic_translator.rs:596`, `convert_tool_choice`) and the Responses translator (`src/server/responses_translator.rs:198`) both forward into the same field, so they inherit the same gap (`{"type":"any"}` and `{"type":"tool","name":...}` are accepted and not enforced).
- The grammar engine already exists: `src/server/structured.rs:535` (`build_json_schema_constraint`) compiles a JSON schema into an llguidance matcher and `ServerGenerateOptions.structured` (`src/server/config.rs:187`) carries it to the scheduler, which applies it per step at `src/server/batch/scheduler.rs:3141`.

## Expected behavior

Semantics, resolved once per request before template rendering:

| `tool_choice` | tools handed to the template | prompt instruction | constraint | error cases |
|---|---|---|---|---|
| absent / `"auto"` | all | none | none | none |
| `"none"` | none (current behaviour) | none | none | none |
| `"required"` | all | "You must call one or more of the available functions to answer the user's request. Do not answer directly without calling a function." | forced tool call when the resolved format is grammar-capable | 400 `invalid_request_error` when `tools` is absent or empty |
| `{"type":"function","function":{"name":"f"}}` | only `f` | "You must call the 'f' function to answer the user's request. Do not call any other function and do not answer directly." | forced call to `f` when grammar-capable | 400 when `tools` is empty or no tool named `f` exists; 400 when `type` is not `function` or `name` is empty |

Instruction placement: appended to the first `system` message as `"\n\n" + instruction` when one exists; otherwise appended to the last `user` message whose content is plain text; otherwise inserted as a new leading `system` message. The original request object is not mutated for response echoing; the injection happens on the message list handed to the renderer.

Grammar-capable formats are the ones whose emitted call is a JSON object with a fixed wrapper and whose wrapper can be read off the loaded chat template before generation: `ToolCallFormat::Hermes` when the template text contains `<tool_call>` (Qwen2.5 / Qwen3 / Hermes templates), `ToolCallFormat::MistralNemo` when it contains `[TOOL_CALLS]`, `ToolCallFormat::Llama3` when it contains `<|python_tag|>` or renders `"parameters":`. There is no operator parser flag today (`src/server/chat_template.rs:297`, `resolve_tool_call_format`, only infers ATEM from template identity), so this template sniffing is the selection mechanism. For those, when the request carries no `response_format`, the server builds a constraint from the tool schemas:

```text
schema(f) = {"type":"object",
             "properties": {"name": {"const": f.name},
                            "arguments": f.parameters or {"type":"object"}},
             "required": ["name","arguments"],
             "additionalProperties": false}
required: anyOf over schema(f) for every declared f
named:    schema(f) for the named f only
```

wrapped in the format's delimiters with an llguidance Lark grammar:

```text
start: "<tool_call>" %json { <schema> } "</tool_call>"          # Hermes
start: %json { <schema> }                                         # Llama3
start: "[TOOL_CALLS] [" %json { <schema> } "]"                    # MistralNemo
```

(Llama3 uses the key `parameters` instead of `arguments` in its JSON; the schema uses the format's key.) Formats without a JSON wire shape (ATEM, Gemma 4, FunctionGemma, Command R, Functionary, Qwen3-Coder XML, MiniMax, Kimi K2, Pythonic) get the instruction and tool narrowing only. When the constrained generation finishes, `finish_reason` is `tool_calls`.

A request that combines `tool_choice: "required"` or a named function with `response_format: json_schema` is rejected with a 400 (two constraints cannot be composed).

## Implementation plan

1. `src/server/types/request.rs`: add to `impl ToolChoice`:
   - `pub fn is_required(&self) -> bool` (`Mode("required")`).
   - `pub fn validate(&self, tools: Option<&[Tool]>) -> Result<(), String>` implementing the 400 cases of the table above (unknown mode, `required` without tools, named function not declared, `type != "function"`, empty name). Move the string-mode check out of `validate_chat_tool_inputs` into this method so the disaggregated router front (`src/server/router_front.rs`) and `routes/chat.rs` share it.
2. `src/server/chat_request.rs`:
   - Change `effective_tools(request)` to return only the named tool for the named form (keep `None` for `none`, all tools otherwise). Because `chat_carries_loop_amplifier` (`src/server/request_options.rs:133`) reads the same helper, update its doc comment: a single narrowed tool still counts as "tools rendered".
   - Add `pub(crate) fn tool_choice_instruction(choice: &ToolChoice) -> Option<String>` returning the two instruction strings.
   - Add `pub(crate) fn inject_tool_choice_instruction(messages: &mut Vec<Message>, instruction: &str)` with the placement rule above (`Message.content` is `MessageContent::Text` for the plain-text cases; a `Parts` user message gets a trailing `ContentPart::Text`).
   - Call both from `prepare_chat_request_with_cache` on a cloned message list before `render_*`, so the prompt-cache key (`template_sig` / `history_prompt`) is derived from the prompt that is actually rendered.
3. `src/server/chat_template.rs`: add `pub fn forced_tool_call_format(&self) -> Option<ToolCallFormat>` next to `default_tool_call_format` (`src/server/chat_template.rs:293`), returning `Hermes` / `MistralNemo` / `Llama3` from the template-text markers above and `None` otherwise (ATEM templates return `None` here even though `default_tool_call_format` reports `Atem`). Cache the answer in the processor at construction, like `supports_tools`.
   `src/server/tool_calls/mod.rs`: add `pub fn tool_choice_grammar(format: ToolCallFormat, tools: &[Tool], choice: &ToolChoice) -> Option<String>` that returns the Lark grammar text for the three grammar-capable formats and `None` otherwise.
4. `src/server/structured.rs`: add `pub fn build_lark_constraint(tokenizer, lark: &str) -> Result<Arc<Mutex<StructuredOutputConstraint>>, StructuredOutputError>` next to `build_json_schema_constraint`, using `llguidance::api::TopLevelGrammar::from_lark(lark)` with the same `ParserLimits`, the same `TOK_ENV_CACHE`, and the same size cap (`MAX_SCHEMA_BYTES`) applied to the serialized tool schemas.
5. `src/server/routes/chat.rs`: in both handlers (non-streaming at ~line 311 where `structured` is built, streaming at ~line 741), after `validate`: if `tool_choice` is `required` or named and the resolved format is grammar-capable and `request.response_format` is absent, build the constraint through `build_lark_constraint` on the blocking pool exactly like the `response_format` path and store it in `options.structured`. Reject the `response_format` + forced-tool combination with 400 before any work.
6. `src/server/routes/chat.rs` finish paths (non-streaming `parse_tools` block and streaming `finish_events` block around line 1020): when `tool_choice` is `required` or named and the parse produced no call, log at `warn` with the format name and leave `finish_reason` as the model produced (the instruction-only formats cannot guarantee a call). Keep the existing name filter for the named form as a last line of defence.
7. `src/server/anthropic_translator.rs::convert_tool_choice`: map `{"type":"any"}` to `ToolChoice::Mode("required")` and `{"type":"tool","name":"f"}` to `ToolChoice::Specific`, so the Anthropic route gets the same enforcement. `src/server/responses_translator.rs` already forwards `tool_choice` unchanged.
8. `docs/responses-api.md` and the server README section on tool calling: document the four modes, the grammar-capable format list, and the 400 cases.

## Validation

(a) Unit tests:

- `src/server/chat_request_tests.rs`: `effective_tools_narrows_to_named_function`, `tool_choice_instruction_appends_to_system_message`, `tool_choice_instruction_appends_to_last_text_user_message`, `tool_choice_instruction_inserts_system_when_no_system_and_no_text_user`, `named_tool_choice_changes_rendered_prompt_and_template_sig`.
- `src/server/types/request.rs` tests: `tool_choice_required_without_tools_is_rejected`, `tool_choice_unknown_function_is_rejected`, `tool_choice_object_without_function_type_is_rejected`.
- `src/server/tool_calls/mod.rs` tests: `tool_choice_grammar_hermes_required_is_anyof_over_all_tools`, `tool_choice_grammar_named_pins_name_const`, `tool_choice_grammar_is_none_for_atem_and_gemma4`.
- `src/server/chat_template.rs` tests: `forced_tool_call_format_detects_hermes_marker`, `forced_tool_call_format_is_none_for_atem_template`.
- `src/server/structured_tests.rs`: `lark_constraint_forces_hermes_wrapper` (feed the matcher `<tool_call>{"name":"get_weather","arguments":{}}</tool_call>` token by token via `consume_token`, assert every step is allowed by `compute_mask` and the matcher reports stopped at the end; assert a plain-text first token is masked out).
- `src/server/anthropic_translator.rs` tests: `anthropic_tool_choice_any_maps_to_required`, `anthropic_tool_choice_tool_maps_to_named`.

(b) Real checkpoint: `mlx-community/Qwen3-4B-4bit` (Hermes tool format, grammar-capable).

```bash
./target/release/mlxcel-server -m models/qwen3-4b-4bit --port 8080
# required
curl -s localhost:8080/v1/chat/completions -H 'content-type: application/json' -d '{
  "model":"q","messages":[{"role":"user","content":"Say hello."}],
  "tool_choice":"required",
  "tools":[{"type":"function","function":{"name":"get_time","parameters":{"type":"object","properties":{"tz":{"type":"string"}},"required":["tz"]}}}]
}' | jq '.choices[0] | {finish_reason, tool_calls: .message.tool_calls}'
# named, with two tools declared
curl -s localhost:8080/v1/chat/completions -H 'content-type: application/json' -d '{
  "model":"q","messages":[{"role":"user","content":"What time is it in Seoul?"}],
  "tool_choice":{"type":"function","function":{"name":"get_weather"}},
  "tools":[{"type":"function","function":{"name":"get_time","parameters":{"type":"object","properties":{"tz":{"type":"string"}}}}},
           {"type":"function","function":{"name":"get_weather","parameters":{"type":"object","properties":{"city":{"type":"string"}}}}}]
}' | jq '.choices[0] | {finish_reason, name: .message.tool_calls[0].function.name}'
```

Acceptance: the first call returns `finish_reason: "tool_calls"` with a `get_time` call on 10 of 10 runs at `temperature: 0.7` (the grammar makes this deterministic, not probabilistic); the second returns a `get_weather` call and never `get_time` on 10 of 10 runs; `tool_choice: "required"` with `"tools": []` returns HTTP 400. Repeat the `required` case with `stream: true` and confirm the tool-call deltas arrive and the final chunk carries `finish_reason: "tool_calls"`. Run the same two requests against `mlx-community/Muse-Glimmer-30B-4bit` (ATEM, instruction-only path) and confirm the named form yields a `get_weather` call on at least 9 of 10 runs.

## Acceptance criteria

- [ ] `tool_choice: "required"` with no tools returns 400; a named function not in `tools` returns 400.
- [ ] The named form renders only the named tool into the prompt and changes `template_sig`.
- [ ] Both forced forms inject the instruction with the placement rule above.
- [ ] Templates carrying `<tool_call>`, `[TOOL_CALLS]` or `<|python_tag|>` grammar-force the call; other templates get instruction and narrowing only; `response_format` plus a forced tool is a 400.
- [ ] Anthropic `tool_choice: {"type":"any"}` and `{"type":"tool"}` map to the same enforcement.
- [ ] Unit tests listed above added and passing; real-checkpoint runs above pass.

## Out of scope

- `parallel_tool_calls: false` enforcement.
- Grammar forcing for XML and ATEM tool formats (would need per-format Lark grammars with string escaping rules).
- `/v1/completions` (no tool concept).

adrysn:
## Summary

`/v1/chat/completions` accepted `tool_choice` in all four OpenAI shapes but only `none` changed behaviour: `required` was validated and then ignored, and a named function still rendered every declared tool and only filtered the parsed calls afterwards, so a model that picked another tool returned an empty `tool_calls` array with `finish_reason: "stop"`. This PR makes `required` and the named form steer generation: the tool list handed to the template is narrowed, an instruction is injected into the rendered prompt, unsatisfiable requests are rejected with 400, and on tool-call formats whose wire shape is a JSON object in a fixed wrapper the call is grammar-forced through the existing structured-output constraint.

Closes #1319

## Semantics

| `tool_choice` | tools rendered | prompt instruction | constraint | 400 when |
|---|---|---|---|---|
| absent / `"auto"` | all | none | none | never |
| `"none"` | none | none | none | never |
| `"required"` | all | "You must call one or more of the available functions ..." | forced call on grammar-capable formats | `tools` absent or empty |
| `{"type":"function","function":{"name":"f"}}` | only `f` | "You must call the 'f' function ..." | forced call to `f` on grammar-capable formats | `f` undeclared (matched exactly as sent), `type` not `function`, empty name |

The instruction is appended to the first system message as `"\n\n" + instruction`, otherwise to the last user message (a content-parts message gets a trailing text part), otherwise inserted as a new leading system message. The injection happens on the message list handed to the renderer; the request object is never mutated. A forced choice combined with a constraining `response_format` is rejected with 400.

Grammar forcing is best effort on top of the prompt enforcement: when the loaded template's tool format has no fixed JSON wrapper, when the declared tool schemas exceed the grammar size or `$ref` budgets, when the grammar does not compile, or on the disaggregated router (whose prefill frame cannot carry a constraint to the decode node), the request proceeds with the narrowed tool list and the injected instruction only and the server logs the reason at `warn`. A forced choice that finishes without a parsed call is logged at `warn` as well, and `finish_reason` stays what the model produced.

## Changes

- `src/server/types/request.rs`: `ToolChoice::is_required`, `ToolChoice::is_forced`, and `ToolChoice::validate(tools)` carrying the 400 rules, shared by `routes/chat.rs`, `routes/responses.rs`, `routes/anthropic.rs`, and the disaggregated `router_front.rs` through `validate_chat_tool_inputs`. The named function is matched exactly as sent, because the renderer, the grammar builder and the response filter compare the raw string; a padded name is a 400 rather than a 200 with an empty `tool_calls` array.
- `src/server/chat_request.rs`: `effective_tools` narrows to the named tool; `tool_choice_instruction`, `inject_tool_choice_instruction`, and `with_tool_choice_instruction` (a `Cow` clone only when a choice is forced). `prepare_chat_request_with_cache` and `render_next_turn_history` render from that list, so `history_prompt` stays a prefix of the prompt and the next-turn warm-up matches what a follow-up turn with the same `tool_choice` renders. `request_options.rs` documents that one narrowed tool still counts as tools rendered for the loop-detection gate.
- `src/server/chat_template.rs`: `forced_tool_call_format()` cached at construction from template markers: Hermes on `<tool_call>` (excluding the XML-body dialects Qwen3-Coder `<function=`, GLM `<arg_key>`, and `<function_name>`), Mistral Nemo on `[TOOL_CALLS]` (excluding bracketed `[ARGS]`), Llama 3 on `<|python_tag|>` or `"parameters":`; ATEM templates return `None`.
- `src/server/tool_calls/mod.rs`: `tool_choice_grammar` and `tool_choice_grammar_with` build the Lark grammar (`start: <wrapper> %json { schema } <wrapper>`) from per-tool schemas (`name` pinned by `const`, `arguments` or `parameters` typed by the tool's own schema, `additionalProperties: false`; `anyOf` over all tools for `required`). `relocate_root_refs` re-points a tool schema's root-relative `$ref`s (`#/$defs/...`, the shape Pydantic, the OpenAI SDK and LangChain emit for nested models) at the place the schema comes to rest inside the wrapper, so those schemas compile and two tools' identically named definitions stay apart. `special_token_id` resolves wrapper markers that the llguidance trie treats as special tokens (an added token that is flagged special or wrapped in `<...>`, which is how Qwen ships `<tool_call>`) so they are referenced by id (`<[151657]>`); a byte literal there masks out the model's own token, which is exactly what the first real-model run showed. Before anything is cloned, `tool_schemas_fit` measures the selected tools' names and schemas against `MAX_SCHEMA_BYTES` with a counting sink and `count_refs` applies `MAX_SCHEMA_REFS`, so a request cannot buy runtime-worker time or transient allocation proportional to a body sized for base64 images; either budget exceeded is the instruction-only fallback.
- `src/server/structured.rs`: `build_lark_constraint(tokenizer, lark)` next to `build_json_schema_constraint`, sharing `TOK_ENV_CACHE`, the tightened `ParserLimits`, and the `MAX_SCHEMA_BYTES` cap applied to the grammar text. Nothing else in the file changes (issue #1316 owns the rest).
- `src/server/routes/chat.rs`: `build_chat_constraint` builds either the `response_format` constraint or the forced tool-call grammar on the blocking pool and is shared by the chat, Responses, and Anthropic routes; `forced_constraint_or_instruction_only` turns a grammar compile failure into the instruction-only fallback with a `warn` (a `response_format` the client asked for explicitly still yields a 400); `validate_chat_tool_inputs` delegates to `ToolChoice::validate` and rejects `response_format` plus a forced choice; both finish paths log at `warn` when a forced choice ends without a call (format name included), and the streaming path sets `finish_reason: "tool_calls"` only when at least one tool-call delta reached the client (previously it claimed `tool_calls` even when the named-function filter had dropped every call). The name filter for the named form stays as the last line of defence.
- `src/server/routes/responses.rs` and `src/server/routes/anthropic.rs`: run the shared validation and constraint build, so `text.format` plus a forced choice is a 400 on Responses and the Anthropic `{"type": "any"}` / `{"type": "tool"}` mappings (already present in `anthropic_translator.rs`) get the same enforcement; the constraint is threaded into both Anthropic generation paths.
- `src/server/router_front.rs`: a forced choice is accepted as before and enforced through the prompt only, with a `warn` saying the grammar cannot travel to the decode node; `router_front_tests.rs` pins that the router forwards such a request instead of rejecting it the way it rejects `response_format`.
- `src/server/routes/chat.rs` streaming finish gate: `tool_call_matches_choice` and `count_tool_calls_reaching_client` are extracted from the `spawn_blocking` closure so the `finish_reason: "tool_calls"` gate has unit coverage for the `required`, named-match, and named-miss cases; the SSE emission flow is unchanged.
- Docs: `docs/responses-api.md` (`tool_choice` row plus a "Tool choice enforcement" section), `docs/server-features.md` ("Tool calling and `tool_choice`"), and `docs/CONTINUOUS_BATCHING.md` (the disaggregated router's instruction-only behaviour next to its existing `response_format` limitation); the repository has no `src/server/README`, so the server feature doc is the closest equivalent to the README section the issue names.

## Validation

All commands ran under `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer` with `--profile test-fast --features metal,accelerate` on Apple Silicon (M1, Metal + Accelerate).

- `cargo check --lib --tests`: clean, no warnings.
- `cargo clippy --lib --tests -- -D warnings`: clean.
- `cargo test --lib -- server::chat_request server::types server::tool_calls server::chat_template server::structured server::anthropic_translator server::routes::chat server::request_options server::responses_translator server::routes::anthropic server::routes::responses server::router_front`: all passing, including the new tests `effective_tools_narrows_to_named_function`, `tool_choice_instruction_appends_to_system_message`, `tool_choice_instruction_appends_to_last_text_user_message`, `tool_choice_instruction_inserts_system_when_no_system_and_no_text_user`, `named_tool_choice_changes_rendered_prompt_and_template_sig`, `tool_choice_required_without_tools_is_rejected`, `tool_choice_unknown_function_is_rejected`, `tool_choice_object_without_function_type_is_rejected`, `tool_choice_grammar_hermes_required_is_anyof_over_all_tools`, `tool_choice_grammar_named_pins_name_const`, `tool_choice_grammar_is_none_for_atem_and_gemma4`, `tool_choice_grammar_references_special_wrapper_tokens_by_id`, `tool_choice_grammar_is_none_when_the_declared_schemas_exceed_the_size_cap`, `tool_choice_grammar_is_none_when_the_declared_schemas_exceed_the_ref_cap`, `special_token_id_follows_the_trie_classification`, `forced_tool_call_format_detects_hermes_marker`, `forced_tool_call_format_is_none_for_atem_template`, `lark_constraint_forces_hermes_wrapper` (token-by-token `consume_token` with `compute_mask` on every step, stopped at the end, prose masked out as the first token), `lark_constraint_named_choice_refuses_other_tool_names`, `a_grammar_that_does_not_compile_degrades_to_instruction_only`, `anthropic_tool_choice_any_maps_to_required`, `anthropic_tool_choice_tool_maps_to_named`, the streaming finish-gate cases, and `router_chat_forwards_forced_tool_choice_instead_of_rejecting_it` (972 tests in the filter at the final commit).
- Real checkpoint `mlx-community/Qwen3-4B-4bit` (Hermes format, grammar-capable) served by `target/test-fast/mlxcel-server`, 5 runs per request at `temperature: 0.7`: `tool_choice: "required"` returned a `get_time` call with `finish_reason: "tool_calls"` on 5 of 5 runs; the named `get_weather` form with two tools declared returned `get_weather` on 5 of 5 runs and `get_time` on 0 of 5; `required` with `"tools": []` returned HTTP 400, an undeclared named function returned HTTP 400, and `required` plus `response_format: json_schema` returned HTTP 400; `required` with `stream: true` streamed the tool-call start and arguments deltas and the final chunk carried `finish_reason: "tool_calls"`; `/v1/messages` with `tool_choice: {"type": "any"}` ended with `stop_reason: "tool_use"` and a `get_time` block. The server log carried no forced-choice warning across these runs. The review cycle re-ran the same checkpoint after the `$ref` relocation: a tool whose `parameters` carry `$defs` returned `book_trip({"to":{"city":"Seoul"}})` with `finish_reason: "tool_calls"` on 3 of 3 runs and over SSE, an unresolvable external `$ref` degraded to instruction-only with one fallback warning and still returned 200, and the original cases were unchanged.

## Changes during review

- The first real-model run failed: llguidance's trie marks every added token of the form `<...>` as special, so a byte-literal `"<tool_call>"` masked out Qwen's own token. The grammar now references such markers by token id.
- Nested tool schemas (`$defs` + `#/$defs/...`) failed to compile inside the wrapper and turned a request that worked on `main` into a 400; the schema's root-relative pointers are now relocated, and a grammar that still does not compile falls back to instruction-only enforcement instead of failing the request.
- The grammar builder ran on a Tokio runtime worker with work proportional to the declared tool schemas (a 36 MB `tools` array cost about 350 ms and 72 MB of transient allocation); the size and `$ref` budgets are now applied before anything is cloned (133 us for the same array).
- A padded function name passed validation and then rendered no tool; the name is now matched exactly as sent.

## Not verified on this host

- `mlx-community/Muse-Glimmer-30B-4bit` (ATEM, instruction-only path) is not available locally, so the "named form yields `get_weather` on at least 9 of 10 runs" acceptance for the instruction-only path was not run. The ATEM detection (`forced_tool_call_format() == None` with `default_tool_call_format() == Some(Atem)`) is unit-tested.
- Mistral Nemo and Llama 3 grammar forcing is covered by the grammar-text and matcher unit tests only; no checkpoint of either family is present on this host.
- `cargo test --workspace` and `cargo clippy --workspace --all-targets` (the issue's full-workspace gates) were not run here; the narrow module runs above and the path-filtered CI cover the touched surface. No CUDA code is touched.

## Follow-ups (not in this PR)

- The Responses API spells a named choice as `{"type": "function", "name": "f"}`; `ToolChoice` only deserializes the Chat Completions nesting, so that shape is rejected at deserialization as it was before this PR. `required` and the string modes work on `/v1/responses`.
- The `required` grammar admits exactly one call; parallel calls on grammar-capable templates and `parallel_tool_calls: false` enforcement are out of scope per the issue.

## Notes

PR for #1314 touches the same file; whichever merges second needs a rebase. The `routes/chat.rs` edits here are scoped to validation, constraint setup, and the finish-path warning.


adrysn:

### Intent

Make `tool_choice: "required"` and the named-function form actually steer generation on `/v1/chat/completions` and the Responses / Anthropic translations of it: narrow the rendered tool list, inject a prompt instruction, reject unsatisfiable requests with 400, and grammar-force the call on JSON-wire tool formats.

The implementation matches the issue's specification closely. All eight deliverables are present and wired into the request path, all sixteen mandated unit tests exist under the names the issue gives, `effective_tools` / `forced_tool_call_format` / `tool_choice_grammar` / `build_lark_constraint` are each reached from a real handler, and the change to `structured.rs` stays inside the `build_lark_constraint` addition that issue #1316 leaves free. Extending the enforcement to `routes/responses.rs` and `routes/anthropic.rs` goes past the literal deliverable list but is what the "Anthropic `any` and `tool` map to the same enforcement" acceptance criterion needs, so it is in scope rather than drift.

### Findings addressed

- [x] **HIGH — a forced `tool_choice` returned 400 for any tool whose `parameters` schema carries its own `$defs`.** `tool_call_schema` embedded the tool schema verbatim under `properties/arguments` of a synthesized wrapper, so `{"$defs": {"Addr": ...}, "properties": {"to": {"$ref": "#/$defs/Addr"}}}` resolved `#/$defs/Addr` against the wrapper's root, found nothing, and failed to compile. That is the exact shape Pydantic, the OpenAI SDK and LangChain emit for a nested model, and because the compile failure was returned to the client, `tool_choice: "required"` with such a tool answered HTTP 400 on a request that returned 200 before the choice was enforced at all. Confirmed by feeding the schema through `build_lark_constraint` (`failed to compile grammar`), not inferred. Fixed in fa7332f6 by `relocate_root_refs`, which re-points a tool schema's root-relative pointers at where the schema comes to rest (`#/properties/arguments/...` for a named choice, `#/anyOf/<i>/properties/arguments/...` for `required`, which also keeps two tools' identically-named definitions apart), plus a fallback: a grammar that still does not compile now degrades to the instruction-only path with a `warn` instead of failing the request, while a `response_format` schema the client did ask for explicitly still turns a compile failure into a 400. Covered by `lark_constraint_compiles_tool_schemas_that_use_defs_refs` and `a_grammar_that_does_not_compile_degrades_to_instruction_only`, and documented in both docs pages.

### Remaining items (reported, not changed)

- [ ] **MEDIUM — `ToolChoice::validate` matches a trimmed name against untrimmed tool names** (`src/server/types/request.rs`). `name` is `choice.function.name.trim()` for the empty check and then reused for the declared-tool lookup, so `{"function": {"name": " get_weather "}}` validates against a declared `get_weather`. Every consumer afterwards compares the untrimmed string: `effective_tools` renders no tool, `tool_choice_grammar` builds no constraint, and `build_tool_call_responses` filters out every parsed call, so the request answers 200 with an empty `tool_calls` array instead of a 400. Using the raw name for the declared lookup makes a padded name a clean rejection.
- [ ] **MEDIUM — the Llama 3 marker `"parameters":` is the loosest of the three detectors** (`src/server/chat_template.rs`). Any template that is not Hermes, Mistral Nemo or ATEM but happens to render `"parameters":` anywhere, which a generic tools preamble that dumps the OpenAI function JSON does, is classified as Llama 3 and gets a wrapper-less bare-JSON grammar its model was never trained to emit. The issue names this marker, so the code is spec-conformant; pairing it with a Llama-family identity check, or requiring `<|python_tag|>`, would narrow the blast radius.
- [ ] **MEDIUM — grammar forcing suppresses the reasoning block.** The Lark grammar applies from the first generated token, so a thinking-enabled template cannot emit `<think> ... </think>` before the call. This matches the existing `response_format` path (both pass `lazy: None` to `build_constraint`), so it is at least consistent, but there the client asked for constrained decoding and here it did not. `build_constraint` already accepts a `LazyGate`; a trigger on the thinking close marker would let both coexist. Worth a documented limitation either way.
- [ ] **MEDIUM — `required` promises "one or more" calls and the grammar admits exactly one.** On a grammar-capable template a `required` request can never produce parallel calls, which diverges from the instruction text the same request injects. `parallel_tool_calls` enforcement is out of scope for this issue, but the divergence deserves a line in the docs.
- [ ] **LOW — `with_tool_choice_instruction` deep-clones the whole request** when only `messages` needs a copy. One extra allocation per forced request, proportional to conversation length.
- [ ] **LOW — the streaming `finish_reason` correction is untested.** Gating `"tool_calls"` on `emitted > 0` fixes a real prior bug (the stream claimed `tool_calls` even when the named-function filter had dropped every call), but it sits inside the finish closure with no unit test; only the real-model streaming run exercises it.

### Verification

- [x] All stated requirements implemented (eight deliverables, four `tool_choice` shapes, three grammar-capable formats, the 400 cases)
- [x] No placeholder, mock, or stub code remaining
- [x] Integrated into the project code flow (chat, Responses, Anthropic Messages, the disaggregated router front, and `/apply-template` all reach `ToolChoice::validate`; `build_chat_constraint` feeds `options.structured` on every generating path)
- [x] Project conventions followed (`thiserror`-style error plumbing, inline format arguments, `// Used by:` markers, doc comments carrying the reasoning)
- [x] Existing modules reused (`build_constraint`, `TOK_ENV_CACHE`, `ParserLimits`, `MAX_SCHEMA_BYTES`, `structured_error_to_response`, `validate_chat_tool_inputs`, `template_sig`)
- [x] No unintended structural changes (`structured.rs` limited to the new `build_lark_constraint`; `routes/chat.rs` limited to validation, constraint setup and the finish-path warning)
- [x] Tests pass: `cargo fmt --all -- --check`, `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings`, and the module filter covering `chat_request`, `types`, `tool_calls`, `chat_template`, `structured`, `anthropic_translator`, `request_options`, `responses_translator`, `router_front`, `prompt_cache` and the three routes: 1139 passing
- [x] Real checkpoint re-run on `mlx-community/Qwen3-4B-4bit` after the fix: nested `$defs` tool with `required` returns `finish_reason: "tool_calls"` with `book_trip({"to":{"city":"Seoul"}})` 3 of 3 and over SSE; an unresolvable external `$ref` degrades to instruction-only with the fallback warning and still returns 200; `required` to `get_time` 3 of 3; named `get_weather` with two tools 3 of 3 and never `get_time`; the empty-`tools`, undeclared-function and `response_format` combinations each return 400

Not verified on this host: the ATEM instruction-only checkpoint (`Muse-Glimmer-30B-4bit` is not present), Mistral Nemo and Llama 3 grammar forcing against a real checkpoint of either family, and any CUDA path.

adrysn:

Scope: the changed surface of this PR, with attention on the request-controlled JSON schemas that now flow from `tools[].function.parameters` into a Lark grammar, the constraint build on the chat / Messages / Responses routes, and the shared `tool_choice` validation.

### Fixed in 4933f6ff

**HIGH -- unbounded reactor-thread work and transient allocation in the forced-grammar build.** `build_chat_constraint` computes `forced_tool_choice_grammar` *before* entering `spawn_blocking`, so `tool_choice_grammar_with` ran on a Tokio runtime worker: it deep-clones every selected tool's `parameters`, walks the clone recursively in `relocate_root_refs`, and re-serializes the result. Nothing bounded that work -- `tools` is capped by count (`MAX_TOOLS`) and not by size, and `build_lark_constraint`'s 64 KiB `MAX_SCHEMA_BYTES` cap only sees the finished text, on the blocking pool, after the work is done and about to be discarded. The JSON body limit is sized for base64 images (roughly 1.3 GB at the default image knobs), so one `tool_choice: "required"` request could stall a runtime worker and allocate a multiple of the body size. Measured on an M-series host: a 36 MB `tools` array cost 352 ms and about 72 MB of transient allocation inside the builder, scaling linearly with the body. Tokio's worker count is the core count, so a handful of concurrent requests stalls the HTTP surface including `/health` and in-flight SSE. The fix measures the selected tools' names and `parameters` against a `MAX_SCHEMA_BYTES` budget with a counting sink that fails once the budget is exhausted, before anything is cloned; the same 36 MB array now leaves the builder in 133 us. Over-budget returns `None`, the same instruction-only fallback a grammar that fails to compile already takes, so no request that worked before is newly refused.

### Reported, not changed

**MEDIUM -- `build_lark_constraint` applies only one of the three pre-compilation guards the `response_format` path applies.** `build_json_schema_constraint` runs `validate_schema_bounds`, which enforces `MAX_SCHEMA_BYTES`, `MAX_SCHEMA_DEPTH` (32) and `MAX_SCHEMA_REFS` (64). `build_lark_constraint` enforces the size cap only, so `tools[].function.parameters` reaches llguidance without the `$ref`-count guard whose stated rationale is exactly this attack ("a tiny schema that references itself 10k times to explode compilation cost"). Measured: 64 `$ref`s compile in 4.2 ms, 1000 `$ref`s (49 KB of Lark, inside the cap) in 101 ms. The 64 KiB cap keeps this bounded and the compile is on the blocking pool, so it is an amplification factor rather than an outage, but the asymmetry is worth closing -- `tools` is a far more commonly attacker-reachable field than `response_format`. Depth is not a practical concern: serde_json's own 128-level recursion limit rejects a `properties` chain past roughly 40 levels before the schema is ever built.

**MEDIUM -- the disaggregated router front silently drops the forced grammar.** `router_front.rs` rejects `response_format` with a 400 on the stated principle that `PrefillRequestFrame` cannot carry a compiled constraint and the router must not "return unconstrained output for a request that explicitly asked for structured output". It now runs the shared `validate_chat_tool_inputs`, so `tool_choice: "required"` is accepted -- and then always degrades to instruction-only, with no constraint and no warning, on every grammar-capable model. On the single-node path that degradation is the compile-failure exception; on the router it is the rule. At minimum this deserves the same `warn` the compile-failure path emits, and a line in the docs.

**MEDIUM -- the named `tool_choice` form is unreachable on `/v1/responses`.** The Responses API spells it `{"type": "function", "name": "f"}`, but `ToolChoice` is `#[serde(untagged)]` over `Mode(String)` and `Specific(ToolChoiceFunction)`, and `ToolChoiceFunction` requires the Chat Completions nesting (`function.name`). A Responses-shaped named choice matches neither variant, so the request fails at deserialization with a generic untagged-enum 400 before any of the new validation runs. The route now wires validation and constraint building for a shape it cannot receive.

**LOW -- `lark_string` is not a general-purpose escaper.** It escapes `"`, `\`, `\n`, `\t` and `\r` but passes other C0/C1 control bytes and DEL through, which the Lark string-literal lexeme rejects. Today it only ever sees the `&'static str` wrapper markers, so this is safe, and grammar injection is not possible either way (the JSON payload is `serde_json`-serialized, so it carries no raw newline or unescaped quote, and llguidance's `%json` is parsed by `parse_json_prefix` rather than by the comment-stripping lexer). The risk is a future caller passing a tool name or description and getting a silent compile failure. A debug assertion or a doc note that the input must be a fixed marker would pin the invariant.

**LOW -- `MAX_TOOLS` is checked after `ToolChoice::validate`.** The named-choice branch does a linear scan over `tools` before the count guard runs. Pre-existing ordering and negligible next to parsing the body, but the cheap guard should come first.

**LOW -- `with_tool_choice_instruction` clones the whole request twice per turn.** Once in `prepare_chat_request_with_cache` and once in `render_next_turn_history`'s warm-up probe. Already noted in the implementation review; the second copy is the part worth remembering.

### Checked and clear

- No grammar injection through `tools[].function.name` or `parameters`: the schema is embedded as `serde_json` output, which cannot contain a raw newline or an unescaped quote, and llguidance hands `%json` to `parse_json_prefix` without applying its `#` / `//` comment skip inside the value.
- No prompt-cache cross-contamination: `template_sig` already hashes `tool_choice` and `tools`, so a forced render and an unforced render of the same conversation cannot share a bucket.
- No stack-overflow path through `relocate_root_refs`: serde_json's 128-level deserializer recursion limit bounds the input depth.
- Error messages stay sanitized: a forced-grammar compile failure is swallowed by `forced_constraint_or_instruction_only` and never reaches the client, and the `StructuredOutputError` variants that do surface carry fixed strings.
- No log forging through the tool name in `warn_unmet_tool_choice` or the `debug` in `forced_tool_choice_grammar`: the default `tracing_subscriber::fmt` visitor records `&str` fields through `Debug`, which escapes newlines.
- `special_token_id` borrows the added-token map rather than cloning it (`AddedVocabulary::get_added_tokens_decoder` returns a reference in tokenizers 0.22), so the per-request marker lookup is cheap.
- The Llama 3 grammar emits a bare JSON object with no `<|python_tag|>` prefix, and `try_llama3` accepts exactly that shape, so the forced output still parses.

### Validation

`cargo fmt --all -- --check`, `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings`, and the chat / tool-call / structured-output / translator / route test filters: 967 passed, 0 failed. The measurements above come from throwaway probes run against this branch and removed before the commit.

PR #1587, opened 2026-09-02
refactor(loading): remove dead is_special_weight_model_type wrapper

reiden21:
## Summary

Deletes the dead `is_special_weight_model_type` pass-through wrapper in `src/loading/special.rs`, together with the `#[allow(dead_code)]` that kept it compiling. The only consumer of the predicate, `src/model_metadata_tests.rs`, already calls the `model_metadata` original.

## Related issues

Closes #1225

## Type of change

- [x] `refactor` — internal restructuring without behavior change

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo check --workspace --all-targets --features metal,accelerate` (initially on a CPU-only MLX build via `MLXCEL_BUILD_METAL=OFF`; superseded by the Metal gate below)
- [x] `cargo test --features metal,accelerate -p mlxcel --lib model_metadata_tests` (7 passed)
- [x] `grep -rn is_special_weight_model_type src/` now matches only `src/model_metadata.rs` and `src/model_metadata_tests.rs`
- [x] Full local gate from CONTRIBUTING.md on a Metal MLX build, run on a local branch that merges all eight re-implementation PRs (#1583, #1584, #1585, #1586, #1587, #1589, #1590, #1591) onto `main`: `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings` (clean) and `cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1` (114 test binaries, 10385 passed, 0 failed, 326 ignored)

## Notes for reviewers

This re-implements the closed PR #1237 against current `main`. Not an inference change, so no real-checkpoint validation applies.

PR #1585, opened 2026-09-02
docs: correct the MLXCEL_DEVICE row in installation.md

reiden21:
## Summary

Corrects the `MLXCEL_DEVICE` row in `docs/installation.md` (default was documented as `auto`, and `metal` was missing from the accepted values) and adds `metal` to the `mlxcel --help` environment-variable line, so installation.md, environment-variables.md and the CLI help all match `parse_runtime_device` in `src/execution/runtime.rs`.

## Related issues

Closes #1223

## Type of change

- [x] `docs` — documentation only

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo check --workspace --all-targets --features metal,accelerate` (initially on a CPU-only MLX build via `MLXCEL_BUILD_METAL=OFF`; superseded by the Metal gate below)
- [x] Cross-checked against `src/execution/runtime.rs`: unset resolves to `RuntimeDevice::Gpu`, and `"gpu" | "metal"` both map to `RuntimeDevice::Gpu`
- [x] Full local gate from CONTRIBUTING.md on a Metal MLX build, run on a local branch that merges all eight re-implementation PRs (#1583, #1584, #1585, #1586, #1587, #1589, #1590, #1591) onto `main`: `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings` (clean) and `cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1` (114 test binaries, 10385 passed, 0 failed, 326 ignored)

## Notes for reviewers

This re-implements the closed PR #1239 against current `main`. The hint text in `src/cli/ggml_compat_args.rs` (`MLXCEL_DEVICE=gpu|cpu to choose the execution device`) was left as is since it describes the two execution choices rather than enumerating every accepted spelling; say the word if you would like `metal` added there too.

PR #1584, opened 2026-09-02
docs(bench): clean scrub artifacts in benchmark_results

reiden21:
## Summary

Cleans the reference-scrubbing leftovers in `docs/benchmark_results/model_tests_m5max.md` and `model_tests_m1ultra.md`: the one broken relative link in the repo, a sentence missing its subject, four bold spans that rendered as literal asterisks, and the dangling "epic-" / "(post-)" fragments where issue numbers were stripped.

## Related issues

Closes #1233

## Type of change

- [x] `docs` — documentation only

## Test plan

- [x] `docs/turbo-kv-cache.md` exists at the new `../turbo-kv-cache.md` link target (matches the sibling `model_tests.md`)
- [x] `grep -nE '^\*\* |epic-|\(post-\)| as\.$'` returns nothing in either file
- [x] Rendered the four `**...**` spans mentally against CommonMark: no leading space after the opening delimiter, so they now render as bold
- [x] Full local gate from CONTRIBUTING.md on a Metal MLX build, run on a local branch that merges all eight re-implementation PRs (#1583, #1584, #1585, #1586, #1587, #1589, #1590, #1591) onto `main`: `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings` (clean) and `cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1` (114 test binaries, 10385 passed, 0 failed, 326 ignored)

## Notes for reviewers

This re-implements the closed PR #1244 against current `main`. Where the scrub removed a referent I named it from context rather than guessing the issue number: the "epic-" matrix is the TurboQuant KV speed gate matrix that the surrounding headings describe, the "(post-)" table rows are the post-Phase 3 steel-envelope readings, and the "exactly as." sentence refers to the Phase 3 kernel described a few paragraphs earlier. The "This PR" phrasing in the follow-up paragraph was also rewritten, since the document is a benchmark log rather than a PR body.

PR #1589, opened 2026-09-02
refactor(server): remove unused build_failed_response helper

### LINKED ISSUE #1226: refactor(server): remove or wire up the unused build_failed_response helper in responses_translator.rs

reiden21:
## Summary

Removes the never-called `build_failed_response` helper from `src/server/responses_translator.rs`, taking the "delete" branch of #1226. The `/v1/responses` route layer already reports mid-generation failures as HTTP `ErrorResponse`s via `generation_error_to_response`, so there is no place a failed `ResponseObject` builder would be used without changing the endpoint's error contract.

## Related issues

Closes #1226

## Type of change

- [x] `refactor` — internal restructuring without behavior change

## Test plan

- [x] `cargo check --workspace --all-targets --features metal,accelerate` (initially on a CPU-only MLX build via `MLXCEL_BUILD_METAL=OFF`; superseded by the Metal gate below). No new warnings, so no import went stale.
- [x] `cargo test --features metal,accelerate -p mlxcel --lib server::responses_translator` (24 passed)
- [x] `grep -rn build_failed_response src/ docs/` returns nothing
- [x] Full local gate from CONTRIBUTING.md on a Metal MLX build, run on a local branch that merges all eight re-implementation PRs (#1583, #1584, #1585, #1586, #1587, #1589, #1590, #1591) onto `main`: `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings` (clean) and `cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1` (114 test binaries, 10385 passed, 0 failed, 326 ignored)

## Notes for reviewers

This re-implements the closed PR #1246 against current `main`. If a failed `ResponseObject` shape is wanted on `/v1/responses` in the future, it is a five-line `OutboundContext` construction with `status: ResponseStatus::Failed` and `error: Some(..)`, and should be added together with the route change and a test rather than kept as an uncalled helper.

### DIFF
--- a/src/server/responses_translator.rs
+++ b/src/server/responses_translator.rs
@@ -762,42 +762,6 @@ pub(crate) fn short_uuid() -> String {
     raw.chars().take(16).collect()
 }
 
-/// Convenience entry point for the route layer: build a response object
-/// representing an error that occurred mid-generation.
-#[allow(dead_code)]
-pub fn build_failed_response(
-    id: String,
-    model: String,
-    created_at: f64,
-    error_code: &str,
-    error_message: &str,
-    request: &CreateResponseRequest,
-) -> ResponseObject {
-    let ctx = OutboundContext {
-        response_id: id,
-        model_id: model,
-        created_at,
-        completed_at: created_at,
-        status: ResponseStatus::Failed,
-        prompt_tokens: 0,
-        completion_tokens: 0,
-        cached_tokens: 0,
-        reasoning_tokens: 0,
-        text: String::new(),
-        reasoning_text: None,
-        parsed_tool_calls: None,
-        max_tool_calls: None,
-        request,
-        error: Some(ResponseErrorBody {
-            code: error_code.to_string(),
-            message: error_message.to_string(),
-        }),
-        incomplete_reason: None,
-        finish_reason: "error".to_string(),
-    };
-    build_response_object(ctx)
-}
-
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;


inureyes (approved):
LGTM

PR #1593, opened 2026-09-02
feat(models): route text-only Youtu-LLM to its MLA decoder

### LINKED ISSUE #1371: feat(models): detect and load text-only Youtu-LLM (youtu, youtu_llm) through the existing Youtu MLA decoder
## Summary

Tencent's text-only Youtu-LLM-2B is the decoder that mlxcel already runs inside Youtu-VL (`src/models/youtu_vl_lm.rs`), but no text-only route exists: the original checkpoint declares `model_type: "youtu"`, one community conversion declares `"youtu_llm"`, and a third declares `"deepseek_v2"`. The first two are rejected at detection. This issue adds the two detection strings, a `ModelType::YoutuLLM` text route onto `YoutuLanguageModel::load`, and confirms the `deepseek_v2`-labelled conversion already loads through the DeepSeek-V2 path.

## Current behavior

`src/models/detection.rs:329` has only the multimodal arm:

```rust
"youtu_vl" => Ok(ModelType::YoutuVLM),
```

`"youtu"` and `"youtu_llm"` fall through to the `_ => Err("Unsupported model type")` arm at `src/models/detection.rs:379`. A standalone text loader already exists and is only reached from tests:

```rust
// src/models/youtu_vl_lm.rs:309-326
// Standalone text-only loader (chiefly used by tests and for ad-hoc CLI runs
// that target a Youtu-VL `model.safetensors` directory directly).
impl YoutuLanguageModel {
    pub fn load<P: AsRef<Path>>(model_dir: P) -> Result<(Self, YoutuTextConfig), String> {
```

`YoutuTextConfig` (`src/models/youtu_vl_lm_config.rs:32-80`) already carries every key of the text-only config (`kv_lora_rank`, `q_lora_rank`, `qk_rope_head_dim`, `qk_nope_head_dim`, `v_head_dim`, `rope_theta`, `rope_traditional`, `rope_interleave`, `tie_word_embeddings`, `rope_scaling`), and `sanitize_text_weights` (`src/models/youtu_vl_lm_sanitize.rs:41`) performs the `kv_b_proj` per-head split and drops `lm_head.*` when tied. No `LoadedModel` variant wraps `YoutuLanguageModel` on its own (`grep YoutuLanguageModel src/` hits only `src/loading/vlm_youtu_vl.rs` and `src/vision/youtu_vl.rs`).

## Expected behavior

All three published labels load and generate the same text:

| checkpoint | `model_type` | route |
|---|---|---|
| `tencent/Youtu-LLM-2B` (bf16, about 4.5 GB) | `youtu` | new `ModelType::YoutuLLM` |
| `mlx-community/Youtu-LLM-2B-mlx-4bit` | `youtu_llm` | new `ModelType::YoutuLLM` |
| `mlx-community/Youtu-LLM-2B-4bit` (about 1.3 GB) | `deepseek_v2` | existing `ModelType::DeepSeekV2` (verify, no code change expected) |

### Architecture (for the record; it is what `youtu_vl_lm.rs` already implements)

Config: `vocab_size` 128256, `hidden_size` 2048, `intermediate_size` 6144, `num_hidden_layers` 32, `num_attention_heads` 16 (= `num_key_value_heads`), `kv_lora_rank` 512, `q_lora_rank` 1536, `qk_rope_head_dim` 64, `qk_nope_head_dim` 128, `v_head_dim` 128, `rms_norm_eps` 1e-6, `rope_theta` 1600000, `rope_interleave` true (traditional RoPE), `rope_scaling` null in the original (one conversion writes `{"type": "yarn", "factor": 1.0, "mscale_all_dim": 0}`, which is the identity and must parse), `max_position_embeddings` 131072, `tie_word_embeddings` true, `attention_bias` false, `mlp_bias` false, `bos_token_id` 128000, `eos_token_id` 128001.

Attention is the DeepSeek-V2 low-rank layout without absorption: `q = q_b_proj(RMSNorm(q_a_proj(x)))` split into 128 nope + 64 rope dims; `kv_a_proj_with_mqa(x)` splits into a 512-dim latent and a 64-dim shared rope key; `kv_b_proj(RMSNorm(latent))` gives per-head 128 nope key + 128 value; rope applied (traditional) to the 64-dim parts; the rope key is broadcast over all 16 heads; scale `(128 + 64) ** -0.5`; cache stores the full 192-dim keys and 128-dim values (`KVCache`). MLP is SwiGLU. Weight keys: `model.embed_tokens.weight`, `model.layers.{i}.self_attn.{q_a_proj,q_a_layernorm,q_b_proj,kv_a_proj_with_mqa,kv_a_layernorm,kv_b_proj,o_proj}`, `model.layers.{i}.mlp.{gate,up,down}_proj`, `model.layers.{i}.{input_layernorm,post_attention_layernorm}`, `model.norm`; no `lm_head` (tied).

## Implementation plan

1. `src/models/detection.rs`: add `"youtu" | "youtu_llm" => Ok(ModelType::YoutuLLM)` immediately before the `"youtu_vl"` arm, plus `youtu_llm_model_type_is_detected` (both spellings) in `src/models/detection_tests.rs`.
2. `src/models/mod.rs`: `ModelType::YoutuLLM` (enum, text list, description `("Youtu-LLM (DeepSeek-V2-style MLA decoder, text-only)", "Specialized")`).
3. `src/loaded_model.rs`: `LoadedModel::YoutuLLM(models::youtu_vl_lm::YoutuLanguageModel)` and the `LanguageModel` forwarding arm.
4. `src/model_metadata.rs`: registration row with `kind: Text, directory: ConfigBacked` (or `Nonstandard` through `src/loading/nonstandard.rs` with `load_pair_from_dir(path, YoutuLanguageModel::load)` if the loader's `(model, config)` return shape fits that path better, as `KimiLinear` does at `src/loading/nonstandard.rs:98`).
5. `YoutuTextConfig::model_type` defaults to `"youtu_vl"`; no change needed, but make `q_lora_rank` tolerate `null` (`Option<usize>`, falling back to a plain `q_proj`) so a future no-LoRA variant does not fail to parse; the published 2B always sets 1536.
6. `src/distributed/tensor_parallel/inference.rs`: arch-string arm to keep the dispatch total (not TP-enabled).
7. `docs/supported-models.md`: add the text-only entry next to Youtu-VL with the three labels and checkpoints.

## Validation

(a) Tests: `youtu_llm_model_type_is_detected` (both strings); in `src/models/youtu_vl_lm_tests.rs` add `text_only_config_with_identity_yarn_parses` (the `{"type": "yarn", "factor": 1.0, "mscale_all_dim": 0}` block parses and the rope table equals the unscaled one) and `load_routes_text_only_dir` (a temp dir with a 2-layer synthetic config and weights loads through `ModelType::YoutuLLM`).

(b) Real checkpoints:

```
./target/release/mlxcel download mlx-community/Youtu-LLM-2B-mlx-4bit
./target/release/mlxcel generate -m models/Youtu-LLM-2B-mlx-4bit -p "Give three reasons why the sky is blue." -n 64 --temp 0
./target/release/mlxcel download mlx-community/Youtu-LLM-2B-4bit
./target/release/mlxcel generate -m models/Youtu-LLM-2B-4bit -p "Give three reasons why the sky is blue." -n 64 --temp 0
```

Acceptance: both produce finite, fluent output; greedy ids of `Youtu-LLM-2B-mlx-4bit` match an oracle run of the same checkpoint through at least the first 32 tokens on three prompts; `tencent/Youtu-LLM-2B` (bf16, `model_type: youtu`) loads and produces the same first 16 greedy tokens as the 4-bit conversion on a 64+ token prompt (quantization noise allowed after that).

## Acceptance criteria

- [x] `"youtu"` and `"youtu_llm"` resolve to `ModelType::YoutuLLM` and load through `YoutuLanguageModel::load`
- [x] Greedy parity against an mlx-lm oracle established on the local `mlx-community/Youtu-LLM-2B-4bit` weights: prefill argmax identical at every position, teacher-forced argmax agreement 118 of 126, and token-exact for 32 tokens on two chat-templated prompts. Raw completion prompts are not usable as a parity gate for this checkpoint: mlx-lm's own bf16 and float32 runs agree on only 91.3 percent of positions and produce different text from the sixth token of the Fibonacci prompt. `mlx-community/Youtu-LLM-2B-mlx-4bit` and `tencent/Youtu-LLM-2B` are not on the validation host; both take plain label arms covered by unit tests.
- [x] `mlx-community/Youtu-LLM-2B-4bit` (`deepseek_v2` label) verified to load through the existing DeepSeek-V2 route. Measured against an mlx-lm `deepseek_v2` oracle on the same weights: token-exact for all 32 greedy tokens on a chat-templated prompt, and identical to the oracle's first 18 tokens on a raw prompt. An earlier revision of the PR claimed otherwise and added an `architectures[0]` detection split; that claim did not survive measurement and the split was removed, so this criterion stands as originally written.
- [ ] `mlxcel arch` reports the family (the supported-architecture listing; `mlxcel list` enumerates local checkpoints and has no architecture column). Needs a release build; commands in the PR body.
- [x] detection table in src/models/detection.rs updated with a test
- [x] cargo clippy (scoped to `--lib --tests`) and cargo fmt --all -- --check pass

inureyes:
## Summary

Text-only Youtu-LLM is the decoder mlxcel already runs as Youtu-VL's text tower, but no text-only route existed: `model_type: "youtu"` and `"youtu_llm"` were both rejected at detection. This adds `ModelType::YoutuLLM` on top of the existing `YoutuLanguageModel`, with no new architecture code, plus three config properties the VLM sibling checkpoint never exercises. The change is purely additive: two new `model_type` arms, no existing arm touched.

## Detection

| checkpoint | `model_type` | before | after |
|---|---|---|---|
| `tencent/Youtu-LLM-2B` | `youtu` | rejected, "Unsupported model type" | `ModelType::YoutuLLM` |
| `mlx-community/Youtu-LLM-2B-mlx-4bit` | `youtu_llm` | rejected | `ModelType::YoutuLLM` |
| `mlx-community/Youtu-LLM-2B-4bit` | `deepseek_v2` | `ModelType::DeepSeekV2` | unchanged |

The third row was a split in the first revision of this PR and is not one now. See "What review changed" below; `deepseek_v2_label_keeps_the_deepseek_v2_route` guards it against being re-added.

## Config properties now honored

`rope_interleave` reaches the rope call. The vendor decoder branches on it (`YoutuMLAttention.forward` in https://huggingface.co/tencent/Youtu-LLM-2B/blob/main/modeling_youtu.py) to choose between the interleaved and half-split rotary pair layouts, and mlxcel's shared `DeepSeekV3Attention` hardcoded the interleaved form. It now carries a `rope_traditional` field that `from_weights` still sets to `true` for every DeepSeek backbone, so nothing about DeepSeek-V3 / V3.2 / V4 / Kimi-VL changes, and the Youtu decoder layer overrides it through `with_rope_traditional`. A test asserts that flipping the flag changes the logits, which is the only way to catch a flag being parsed and then dropped.

`rope_scaling` is validated at load on the text route. The published block is `{"type": "yarn", "factor": 1.0, "mscale_all_dim": 0}`, which is the identity: at factor 1 YaRN's extrapolation and interpolation frequencies coincide and the attention mscale is 1. That block is accepted and a test pins the resulting attention scale to the unscaled `(qk_nope + qk_rope) ** -0.5`. A factor above 1 asks for a frequency interpolation this decoder does not implement, so it is refused at load rather than decoded on the plain table, which would give fluent but positionally wrong long-context output. The VLM loader is deliberately untouched; no published Youtu-VL checkpoint carries a `rope_scaling` block.

`q_lora_rank` becomes `Option<usize>`, so a variant that projects the query directly through `q_proj` parses instead of failing the whole load. `tie_word_embeddings` was already handled; this adds the whole-model assertion that a tied build has `lm_head == None` and still produces finite logits.

## What review changed

The first revision also split the `deepseek_v2` label on `architectures[0]`, on the belief that mlxcel's DeepSeek-V2 decoder mishandled the relabelled Youtu conversion. Measuring both decoders against an mlx-lm `deepseek_v2` oracle on the same weights refuted that: on a chat-templated prompt the DeepSeek-V2 route is token-exact with the oracle for all 32 greedy tokens, and on the raw five-token prompt that started the investigation it matches for 18. The split was removed.

The apparent failure was two artifacts stacked. The CLI applies the checkpoint's chat template while the oracle run did not, so the two sides were answering different prompts, and the "Okay, so the question is..." output is this reasoning model's `<think>` opener, which is the right answer to the templated prompt. And on raw completion prompts this checkpoint is close to undecided: over a 126-token passage the median top-1 to top-2 logit gap is 0.625 and one position in eight is an exact bf16 tie, so mlx-lm's own bf16 and float32 runs of the same weights agree on only 91.3 percent of positions, and on that Fibonacci prompt they produce entirely different text from the sixth token on.

## Numerical verification (measured, not assumed)

Against mlx-lm 0.31.3 on the same `mlx-community/Youtu-LLM-2B-4bit` weights, identical prompt ids, greedy at temperature 0.

- **Prefill stage diff, 5-token prompt.** Bit-identical through `q_a_proj`, `q_a_layernorm`, `q_b_proj`, the nope/rope split, the compressed KV, the latent, and both rope calls. First difference is the per-head key and value materialization (0.57 percent and 0.74 percent mean relative), where the absorbed path contracts a dequantized weight and mlx-lm a still-quantized one. Layer 0 attention output 0.33 percent, per-layer hidden state 1.0 percent at layer 0 rising to 5.8 percent at layer 31, logits 5.4 percent. Argmax identical at all five positions.
- **Prefill versus decode, inside mlxcel.** Zero argmax mismatches over 33 positions, so the absorbed `l == 1` arm agrees with the multi-token arm. No decode-path defect.
- **Teacher-forced, 126-token passage.** mlxcel agrees with the bf16 oracle on 118 of 126 argmax positions (93.7 percent). Every disagreement sits at a top-2 gap between 0.0 and 0.75 logits; three of the eight are exact ties. The bf16 oracle agrees with its own float32 self on only 115 of 126 (91.3 percent) at the same positions.
- **Greedy, chat-templated prompt.** Median top-2 gap rises to 3.875 and the bf16 and float32 oracles agree on all 32 tokens. mlxcel is token-exact against both for 32 tokens on two prompts.

## What changed

- `src/models/detection.rs`: `"youtu" | "youtu_llm"` arm; the `deepseek_v2` arm gains only a comment recording the measurement behind leaving it alone.
- `src/models/deepseek_v3.rs`: `DeepSeekV3Attention::rope_traditional` (still `true` from `from_weights`) plus a `with_rope_traditional` override, used by the two `fast_rope` calls.
- `src/models/youtu_vl_lm_config.rs`: `q_lora_rank: Option<usize>`, `rope_is_interleaved()` folding the `rope_interleave` / `rope_traditional` spellings, and `validate_rope_scaling()`.
- `src/models/youtu_vl_lm.rs`: the decoder layer applies the rope flag; `YoutuLanguageModel::load` validates `rope_scaling` and parses `eos_token_id` out of `config.json`.
- `src/models/mod.rs`, `src/loaded_model.rs`, `src/model_metadata.rs` (Text / Nonstandard), `src/loading/nonstandard.rs` (`load_pair_from_dir`), `src/distributed/tensor_parallel/inference.rs` (dispatch-table arm, not TP-enabled): registration.
- `docs/supported-models.md`: the text family entry, including the parity guidance below.
- `TECHNICAL_REPORTS/1593-youtu-llm-text-route-20260902.{en,ko}.md`: the report pair, carrying every number above, the alternatives weighed, and the diagnosis method.

## Test plan

- [x] `cargo test --profile test-fast --features metal,accelerate --lib -- models::youtu_vl_lm models::detection_tests loading::nonstandard` (60 passed, 8 new)
- [x] `cargo test --profile test-fast --features metal,accelerate --lib -- models::deepseek models::metadata_tests model_metadata_tests loading::tests` (197 passed; the shared MLA change moves no DeepSeek family)
- [x] `cargo test --profile test-fast --features metal,accelerate --bin mlxcel -- family_order all_model_types supported_models arch` (10 passed)
- [x] `cargo clippy --profile test-fast --lib --tests --features metal,accelerate -- -D warnings`
- [x] Real-checkpoint numerical verification against the mlx-lm oracle, above.

## Reviewer gate

Parity for this family must use the chat template. A raw-prompt greedy diff against any reference produces spurious failures, because the reference does not agree with its own higher-precision self on those prompts.

```bash
./target/release/mlxcel arch | grep -i youtu

DIR=models/mlx/youtu-llm-2b-4bit
COPY=/tmp/youtu-llm-2b-4bit-youtu
mkdir -p "$COPY" && for f in "$DIR"/*; do ln -sf "$(cd "$(dirname "$f")" && pwd)/$(basename "$f")" "$COPY/"; done
rm -f "$COPY/config.json"
python3 -c "import json;c=json.load(open('$DIR/config.json'));c['model_type']='youtu';json.dump(c,open('$COPY/config.json','w'))"

./target/release/mlxcel generate -m "$COPY" -p "The Fibonacci sequence begins with" -n 32 --temp 0
```

The local checkpoint declares `deepseek_v2` and therefore keeps the DeepSeek-V2 route by design, so the copied directory with `model_type` rewritten to `youtu` is what exercises the new arm. `mlxcel arch` must list `Youtu-LLM (DeepSeek-V2-style MLA decoder, text-only)` under `Specialized:` alongside the existing Youtu-VL entry under `Other VLM:`. The generation must reproduce `<think>\nOkay, the user is asking about the Fibonacci sequence beginning. Let me start by recalling what the Fibonacci sequence is. It's a series of numbers where each`, because the CLI applies the checkpoint's chat template by default; the oracle produces the identical 32 ids in both bf16 and float32.

Two published checkpoints are not on this host and are not required: `tencent/Youtu-LLM-2B` (bf16, `model_type: youtu`) and `mlx-community/Youtu-LLM-2B-mlx-4bit` (`model_type: youtu_llm`). Both take the plain label arms that unit tests cover.

Closes #1371


inureyes:
Stage-by-stage diff against the mlx-lm oracle is done. The forward is correct; the failing gate was measuring something else. Numbers below, all against mlx-lm 0.31.3 on the same `mlx-community/Youtu-LLM-2B-4bit` weights, identical prompt ids, greedy at temperature 0.

## Prefill stage diff, 5-token raw prompt `[837, 91949, 12082, 13328, 458]`

| stage | mean abs | mean abs diff | max abs diff | relative |
|---|---|---|---|---|
| embedding | 0.08657 | 0.000000 | 0.00000 | 0.00% |
| `q_a_proj`, `q_a_layernorm`, `q_b_proj`, q_nope, q_pe | | 0.000000 | 0.00000 | 0.00% |
| `kv_a_proj_with_mqa`, latent, k_pe | | 0.000000 | 0.00000 | 0.00% |
| q_pe after rope, k_pe after rope | | 0.000000 | 0.00000 | 0.00% |
| k_nope (per-head key) | 0.051333 | 0.0002947 | 0.001953 | 0.574% |
| values | 0.015301 | 0.0001136 | 0.000977 | 0.743% |
| layer 0 attention output | 0.074063 | 0.0002411 | 0.015625 | 0.326% |
| layer 0 block output | 0.07108 | 0.000718 | 0.06250 | 1.01% |
| layer 31 hidden | 4.34448 | 0.249703 | 14.0 | 5.75% |
| final norm | 0.75743 | 0.054879 | 0.50000 | 7.25% |
| logits | 5.34840 | 0.286125 | 2.09375 | 5.35% |

Everything up to and including both rope calls is bit-identical, so `rope_interleave`, `rope_theta` 1600000, the q LoRA chain, the `q_a_layernorm` eps, the nope/rope split order, `kv_a_layernorm` placement, and the 1/sqrt(192) scale with mscale exactly 1.0 are all confirmed correct. The first divergence is the per-head key and value materialization, where mlxcel's absorbed MLA contracts a dequantized `kv_b_proj` and mlx-lm contracts a still-quantized one. Argmax is identical at all five prefill positions.

## The decode path is clean

Teacher-forced inside mlxcel, prefilling 15 tokens then feeding 32 forced tokens one at a time: zero argmax mismatches over 33 positions between the `l == 1` absorbed decode arm and a single multi-token prefill over the same sequence. There is no decode-path defect, which also explains why the server output matched a direct library run exactly.

## Why token-exactness against the bf16 oracle was the wrong gate

The reference does not agree with itself. Running the identical mlx-lm model with `set_dtype(mx.float32)`:

- On `The Fibonacci sequence begins with`, bf16 and float32 diverge at step 5 (top-2 gap 0.250) and produce entirely different text: `" 1,1, and each subsequent number is the sum"` against `" 1,1,1,4,34,"`. **mlxcel reproduces the float32 sequence exactly for 12 tokens**, so on this prompt mlxcel tracks the higher-precision answer and the bf16 oracle is the outlier.
- Teacher-forced over a 126-token passage: bf16 versus float32 agree on 91.3 percent of argmax positions, mlxcel versus bf16 on 93.7 percent, mlxcel versus float32 on 89.7 percent. All three sit in the same band. Median top-1 to top-2 gap is 0.625 and 12 percent of positions are exact bf16 ties. Every mlxcel disagreement is at a gap of 0.75 logits or less.

This checkpoint is a reasoning model being asked for raw completion, which is out of distribution and leaves it close to undecided.

## The CLI versus server discrepancy

The CLI applies the checkpoint's chat template. Rendering the same request through it and running the oracle gives `<think>\nOkay, the user is asking about the Fibonacci sequence beginning...`, which is the `"Okay, so the question is..."` output you saw. The CLI and the server were not disagreeing about the same input; the server had the raw prompt and matched a direct library run token for token.

## Chat-templated parity: token-exact

With the chat template the median top-2 gap rises from 0.625 to 3.875, bf16 and float32 oracles agree on all 32 tokens, and mlxcel is token-exact against both:

```
prompt ids  [128000, 128236, 837, 91949, 12082, 13328, 458, 128237]
mlxcel      [128227, 198, 37317, 11, 290, 1483, 371, 11935, 913, 290, 91949, 12082, 7963, 13, 6846, 611, 1311, 603, 125377, 1165, 290, 91949, 12082, 371, 13, 5902, 261, 4326, 324, 6578, 1551, 1981]
oracle      identical, in both bf16 and float32
```

`In distributed systems, consensus protocols such as Raft` through the chat template is also 32 of 32.

## The PR premise was wrong and the detection split is gone

Running mlxcel's existing DeepSeek-V2 decoder on the same checkpoint: chat-templated, 32 of 32 identical to the oracle; raw prompt, identical to the oracle's first 18 tokens, which is the full `" 1,1, and each subsequent number is the sum of the two preceding numbers."` The DeepSeek-V2 route handles this export correctly, so the `architectures[0]` split I added had nothing to fix and would only have moved working checkpoints onto a different decoder. It is removed, `deepseek_v2` keeps its route unconditionally, and `deepseek_v2_label_keeps_the_deepseek_v2_route` pins the Youtu relabel, a genuine `DeepseekV2ForCausalLM`, and a config with no `architectures` array, with the reasoning recorded so it is not re-added. That also restores the issue's original acceptance criterion, which said this checkpoint already loads through the existing DeepSeek-V2 route.

I should have measured that before encoding it in a routing rule rather than after; the report's section 5.5 records it.

## Tests added

`prefill_and_decode_paths_agree_position_by_position` pins the invariant the diagnosis turned on: the absorbed `l == 1` decode arm and the multi-token prefill arm must agree at every position on a synthetic model. Nothing else in the suite exercised the decode arm. `deepseek_v2_label_keeps_the_deepseek_v2_route` is the detection guard above.

The gate commands in the PR body now use the chat template and a config-rewritten copy of the local checkpoint, since the local directory declares `deepseek_v2` and therefore keeps the DeepSeek-V2 route by design.

PR #1595, opened 2026-09-02
fix(core): separate GPU backend availability from the default device

### LINKED ISSUE #1421: fix(core): is_gpu_available reports the current default device, not hardware availability
## Problem / Background
`mlxcel_core::is_gpu_available()` does not answer whether a GPU backend exists. It answers whether the process-wide MLX default device is currently the GPU, so it flips to `false` the moment any caller runs `set_default_device(false)`, on a machine that has a GPU, in a `--features cuda` build. Two consequences were hit during the epic #1348 gate:

- `src/multimodal/host_preprocessor_tests.rs:31-34` (`ensure_cpu_device`, a `Once` calling `mlxcel_core::set_default_device(false)`) moves the default device to the CPU and never restores it. Under `make verify-test-cuda` (`--test-threads=1`, modules in name order) every real-checkpoint gate that sorts after `multimodal::` then measured the CPU backend: `rerank::real_checkpoint_tests` scored the 4-bit Qwen3 reranker at 0.3486 instead of 0.9883 and the bf16 Qwen3-VL reranker returned `[NaN, NaN]` for image documents, while both passed in isolation.
- PR #1420 fixed the symptom by pinning `set_default_device(true)` unconditionally inside `mlx_test_guard` (`src/models/embedding_test_support.rs:70-79`). A first attempt guarded the pin with `is_gpu_available()` and never fired, for exactly the reason above. The pin is now taken by 101 call sites and papers over a leak instead of removing it.
- `src/execution/runtime.rs:84-96` (`initialize_runtime`) calls `set_default_device(false)` when `MLXCEL_DEVICE` requests the CPU and then reads `is_gpu_available()` to fill `RuntimeSetup.device`, so the field reports the requested device rather than availability, and `should_warn_cpu_only_on_nvidia_host` (`src/execution/runtime.rs:256-262`) is fed a "resolved" device that is really the request echoed back.

## Current Behavior
- `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:4676-4678`: `bool is_gpu_available() { return mlx::core::default_device() == mlx::core::Device::gpu; }`.
- `src/lib/mlxcel-core/src/lib.rs:2060-2061` declares it in the cxx bridge with the honest doc comment "Check whether the current default device is GPU"; the name is what misleads.
- `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:228-234` (`gpu_device_count`) already calls `mlx::core::device_count(mlx::core::Device::gpu)` but clamps the result to `>= 1` so that `Device::gpu` is always selectable, which makes it unusable as an availability answer on its own.
- Callers that actually want the "is the default device the GPU" semantics: `src/lib/mlxcel-core/src/streams.rs:59` (`new_thread_local_generation_stream`) and `src/lib/mlxcel-core/src/streams.rs:163` (deprecated `new_generation_stream`) create a GPU stream only when the GPU is the default; that behaviour must not change.
- Other test code that moves the default device and never restores it: `src/lib/mlxcel-core/src/ffi_tests.rs:2535` and `src/lib/mlxcel-core/src/streams.rs:343` (`set_default_device(true)`), and the six `ensure_cpu_device()` calls in `src/multimodal/host_preprocessor_tests.rs`.

## Proposed Solution
1. In `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp` and the cxx bridge block in `src/lib/mlxcel-core/src/lib.rs`, rename the existing function to `default_device_is_gpu()` (same body, doc comment unchanged) and add `gpu_backend_available() -> bool` that returns the unclamped `mlx::core::device_count(mlx::core::Device::gpu) > 0`. On a CPU-only build this is `false` (the count is 0, as the comment at `mlx_cxx_bridge.cpp:230-232` records); on Metal and CUDA builds with a usable device it is `true`. Do not derive it from `cfg!(feature = "cuda")` on the Rust side: a CUDA build on a host without a driver must answer `false`, which only the runtime query does. Keep `is_gpu_available` as a `#[deprecated(note = "use default_device_is_gpu or gpu_backend_available")]` Rust shim in `src/lib/mlxcel-core/src/lib.rs` for one release so out-of-tree callers get a compile warning instead of a break, and switch every in-tree caller.
2. `src/execution/runtime.rs::initialize_runtime`: set `device` from `gpu_backend_available()` and apply the CPU override separately, so `RuntimeSetup.device` is `Cpu` either because no backend exists or because `MLXCEL_DEVICE=cpu` asked for it, and the two cases are distinguishable. Add a `cpu_override: bool` field to `RuntimeSetup` (or reuse `invalid_device_override` semantics if a reviewer prefers) and document both in the struct doc. `should_warn_cpu_only_on_nvidia_host` keeps its signature; it now receives a real resolved device.
3. Add `pub struct DefaultDeviceGuard` to `src/lib/mlxcel-core/src/streams.rs` next to the existing `DefaultStreamGuard` (`streams.rs:128`): `DefaultDeviceGuard::cpu()` and `::gpu()` record `default_device_is_gpu()` and call `set_default_device(..)`, and `Drop` restores the recorded value. Replace `ensure_cpu_device()` in `src/multimodal/host_preprocessor_tests.rs` with a guard held for the duration of each test (a `Once` cannot restore anything). Apply the same guard at `src/lib/mlxcel-core/src/ffi_tests.rs:2535` and `src/lib/mlxcel-core/src/streams.rs:343`.
4. Leave the unconditional pin in `mlx_test_guard` in place until the guards land, then reduce it to a debug assertion (`debug_assert!(default_device_is_gpu() || !gpu_backend_available())`) so a future leak fails loudly instead of being silently repaired.

Rejected alternative: making `is_gpu_available` return the hardware answer under the same name. Every existing caller in `streams.rs` depends on the default-device semantics, and a silent semantic change under an unchanged name is how this bug was created.

## Scope
**In scope:** `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp`, `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.h`, `src/lib/mlxcel-core/src/lib.rs`, `src/lib/mlxcel-core/src/streams.rs`, `src/lib/mlxcel-core/src/ffi_tests.rs`, `src/execution/runtime.rs`, `src/multimodal/host_preprocessor_tests.rs`, `src/models/embedding_test_support.rs`.

**Out of scope:** the MLX CUDA teardown abort (separate issue), the OpenXLA device selection (`MLXCEL_XLA_DEVICE`), multi-GPU index selection (`set_default_gpu_device`).

## Implementation Notes
- **Reuse**: `DefaultStreamGuard` (`streams.rs:128`) is the RAII pattern to copy; `gpu_device_count` shows the portable `device_count` call; `metal_is_available` / `cuda_is_available` (`lib.rs:2063-2072`) show how backend-specific answers are already exposed with `no_cuda` stubs.
- **Constraints**: the bridge must keep compiling on the Metal, CUDA and CPU-only builds; `gpu_backend_available` must not touch the default device or create a stream (it is called before `initialize_runtime` finishes).
- **Edge cases**: CUDA build with no driver (count 0, `false`); `MLXCEL_DEVICE=cpu` on a GPU host (`device == Cpu`, `cpu_override == true`, no warning); nested guards (inner restores to the outer's value, which the outer then restores again); a guard dropped during a panic (Drop still runs, so a failing test does not poison later modules).
- **Error handling**: none of these calls can fail; `device_count` on a CPU-only build returns 0, not an exception.

## Acceptance Criteria
- [ ] `mlxcel_core::gpu_backend_available()` returns `true` on the GB10 CUDA build and on Apple Silicon, `false` on a CPU-only build, and its value does not change after `set_default_device(false)`.
- [ ] `mlxcel_core::default_device_is_gpu()` replaces every in-tree `is_gpu_available()` call and the deprecated shim compiles with a warning.
- [ ] `initialize_runtime()` under `MLXCEL_DEVICE=cpu` on a CUDA host reports `device == RuntimeDevice::Cpu` and the new override flag, and `cargo run -- generate` still runs on the CPU.
- [ ] A test in `src/multimodal/host_preprocessor_tests.rs` (or a sibling module that sorts after it) asserts `mlxcel_core::default_device_is_gpu()` equals the value recorded before the module ran; this test passes under `--test-threads=1` with the unconditional pin in `mlx_test_guard` reduced to the debug assertion.
- [ ] `cargo test -p mlxcel --lib -- --test-threads=1 multimodal:: rerank::real_checkpoint_tests` (the PR #1420 reproduction) passes with the pin reduced.
- [ ] Integrated into the real code flow: `RuntimeSetup` consumers and the startup warning use the new field, not a standalone helper.

## Verification
```bash
cargo fmt --all -- --check
cargo clippy --lib --tests --profile test-fast --features cuda -- -D warnings
cargo test --profile test-fast --features cuda -p mlxcel-core --lib streams::
cargo test --profile test-fast --features cuda -p mlxcel --lib -- --test-threads=1 multimodal:: rerank::real_checkpoint_tests
MLXCEL_DEVICE=cpu cargo test --profile test-fast --features cuda -p mlxcel --lib execution::runtime
```
Pass: every command exits 0, the reranker gates report 0.9883 for the Beijing document and finite image scores, and the device-restore test passes when run both alone and after `multimodal::host_preprocessor_tests`.

## Technical Considerations
Related: PR #1420 (the unconditional pin and the `is_gpu_available` finding in its body), issue #487 (`gpu_device_count` and the multi-GPU device-index surface), issue #626 (`metal_is_available`).

Originating PR: #1420.

Follow-up from epic #1348.

adrysn:
## Summary

`mlxcel_core::is_gpu_available()` answered whether the process-wide MLX default device was currently the GPU, not whether a GPU backend exists, so it turned `false` the moment any caller ran `set_default_device(false)` on a machine that has a GPU. Two things went wrong because of that: `initialize_runtime()` read it after applying the `MLXCEL_DEVICE=cpu` override and echoed the request back as the resolved device, and test modules that moved the default device to the CPU inside a `Once` and never moved it back made every real-checkpoint gate sorting after them under `--test-threads=1` measure the CPU backend (the `rerank::real_checkpoint_tests` scores in PR #1420). PR #1420 papered over the leak with an unconditional GPU pin in `mlx_test_guard`; this PR removes the leak and reduces the pin to an assertion.

## What changed

- `src/lib/mlxcel-core/cpp/mlx_cxx_bridge.{cpp,h}`, `src/lib/mlxcel-core/src/lib.rs`: `is_gpu_available` is renamed `default_device_is_gpu` (same body, same doc), and `gpu_backend_available()` is added as the unclamped `mlx::core::device_count(Device::gpu) > 0`. In the pinned MLX, `no_gpu::device_count()` returns 0, Metal returns 1, and the CUDA backend returns whatever `cudaGetDeviceCount` reports (0 when it fails, no throw), so the answer is `false` on a CPU-only build and on a CUDA build without a driver and `true` on Metal and on a CUDA build with a device. `device_count` is portable, so the function needs no backend `#ifdef`, and it never moves the default device or creates a stream. `is_gpu_available()` stays as a `#[deprecated(note = "use default_device_is_gpu or gpu_backend_available")]` shim at the `mlxcel-core` root for one release.
- Every in-tree caller is switched: `streams.rs` keeps the default-device meaning for the generation-stream helpers, while `execution/runtime.rs`, `tests/sampling_*_kill_switch.rs`, and the two sampling microbench examples want backend availability. `make verify-clippy` lints `--all-targets`, so the examples had to move too.
- `src/execution/runtime.rs`: `initialize_runtime()` resolves `RuntimeSetup.device` from `gpu_backend_available()` through a pure `resolve_device(requested, available)` and applies the resolution separately: every `Cpu` resolution is installed with `set_default_device(false)`, not only the operator's override, because MLX seeds its default device from `gpu::is_available()`, which the CUDA backend answers `true` unconditionally, so a CUDA build on a driverless host would otherwise report `Cpu` while MLX kept dispatching to an unusable GPU. The new `RuntimeSetup.cpu_override: bool` says when `MLXCEL_DEVICE=cpu` is why the device is `Cpu`, both fields are documented on the struct, and `should_warn_cpu_only_on_nvidia_host` keeps its signature but now receives a real resolved device. The server startup log (`src/server/startup.rs`) and the CLI runtime printouts (`src/commands/generate.rs`, `src/commands/chat.rs`) report the override next to whether a GPU backend was available, so a CPU line on a GPU host is not mistaken for a missing backend.
- `src/lib/mlxcel-core/src/streams.rs`: `pub struct DefaultDeviceGuard` next to `DefaultStreamGuard`, with `cpu()`, `gpu()`, and `capture()` (records without moving, for code that changes the device through `set_default_gpu_device`). `Drop` restores the recorded value, guards nest, and restoration runs during unwinding. `gpu()` is a no-op without a GPU backend because the pinned MLX throws from `set_default_device(Device::gpu)` there; the pin this PR removes from `mlx_test_guard` would have terminated a CPU-only test binary for the same reason. `lock_default_device()` (returning a `DefaultDeviceLock`) is the one process-wide lock for the one device: every test that moves the default device takes it, and `mlx_test_guard` holds it for the duration of the tests it serializes, so a measurement and a move can no longer overlap (a lock private to each module only excluded that module's own tests, and `multimodal::host_preprocessor` and `vision::merge` still interleaved). The guard deliberately does not take the lock, so guards nest inside a held span. `default_device_guards_held()` reports how many guards are alive process-wide; the leak assertion below uses it as a backstop for a guard created without the lock.
- `src/multimodal/host_preprocessor_tests.rs` and `src/vision/merge_tests.rs`: the `Once` is replaced by a per-test guard held for the test's duration, serialized by the shared `lock_default_device()` (the helpers return the device guard before the lock so the device is restored before the lock is released; tuple fields drop in declaration order). `host_preprocessor_tests` also records the default device before its first move, and a new test, `the_default_device_is_restored_after_the_export_tests`, sorts after every test that moves the device and asserts `default_device_is_gpu()` equals the recorded value, plus the same invariant `mlx_test_guard` checks. `vision/merge_tests.rs` was not in the issue's file list but carried the identical `Once` leak, so it is converted the same way. The guard also covers `ffi_tests.rs` (the shapeless-compile audit harness) and the `streams.rs` index-validation test.
- `src/models/embedding_test_support.rs` (issue step 4, done in this PR): with the leaks removed, the unconditional `set_default_device(true)` pin in `mlx_test_guard` is reduced to an assertion that `default_device_is_gpu() || !gpu_backend_available()`, exempting an explicit `MLXCEL_DEVICE=cpu` (`initialize_runtime()` moves the default device on purpose under it) and any live `DefaultDeviceGuard` (a backstop for a guard created without the shared lock; under the `--test-threads=1` gates no guard can be alive while an unrelated test runs, so a real leak still fails loudly). The guard now returns an `MlxTestGuard` that holds both its serialization mutex and `lock_default_device()`, taken in that order and never reversed, so a real-checkpoint gate cannot be measuring while another module holds the device on the CPU; `models::bert_tests` and `models::modernbert_tests` delegate to it and only changed their return types. It is a plain `assert!` rather than the `debug_assert!` the issue text suggests: every gate the repository runs (`make verify-test`, `make verify-test-cuda`, `make test-fast`) uses `--profile test-fast`, which inherits `release` and compiles debug assertions out, so a `debug_assert!` would never fire where the leak was found. The helper is `#[cfg(test)]` code, so the check costs nothing in production.
- `CHANGELOG.md` and `docs/environment-variables.md` describe the rename, the new field, and the startup line.

## Changes during review

The implementation review found three defects in the first commit and fixed them in `d1f2a38c`: `initialize_runtime()` applied only the operator's CPU override rather than every CPU resolution (the CUDA-without-driver case above); the `host_preprocessor_tests` helper returned `(lock, guard)`, which released the lock before the device was restored; and `vision::merge_tests` had no lock at all. Together those made a parallel `cargo test --lib` leak the CPU default and trip the new assertion in 31 of 40 runs of a representative filter, and 0 of 40 after the fix. The live-guard count in `DefaultDeviceGuard` removed the remaining false positives.

The security and performance review found that per-module locks do not serialize a process-global device and fixed it in `3741c8fd`: `multimodal::host_preprocessor`'s restore check read the CPU default a `vision::merge` guard was legitimately holding in 4 of 300 parallel runs, and a gate under `mlx_test_guard` could have scored a real checkpoint on the CPU backend for the same reason without saying so. `mlxcel_core::streams::lock_default_device()` is now the one lock for the one device (0 of 300 after; 200 parallel runs of the core `streams::` filter and 60 runs of a 2432-test filter spanning `models::`, `rerank::`, `embeddings::`, `vision::` and `multimodal::` with no failure and no hang). It also checked that `gpu_backend_available()` cannot throw across the cxx boundary on any backend, that the guard's restore cannot hit MLX's throwing `set_default_device(gpu)`, that the live-guard counter's atomic ordering is sound, that the extra availability calls cost nothing beyond what the first array allocation already pays on Metal, and that the new log lines carry only a boolean. Reported and left as is: MLX keeps its default device in a non-atomic global, so `initialize_runtime()`'s startup-only `set_default_device(false)` stays the one unlocked mover (commented); `RuntimeSetup` is `pub` without `#[non_exhaustive]`, so the new field breaks out-of-tree struct literals (pre-existing style).

Finalization added the review-suggested comments and the bilingual technical report `TECHNICAL_REPORTS/1595-gpu-backend-availability-device-guard-20260902.{en,ko}.md`, which the repository's `.keep-reports` marker asks to be committed inside the PR.

## Test plan

All on Apple Silicon with `--profile test-fast --features metal,accelerate` (the issue's commands use `--features cuda`; this host has Metal). Re-run after each review commit with the same results; the review runs added `models::bert` (28 passed) and `models::modernbert` (19 passed), which delegate to `mlx_test_guard`.

- [x] `cargo clippy --profile test-fast --features metal,accelerate --lib --tests -- -D warnings`
- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel-core --lib streams::`: 17 passed (new: `gpu_backend_available_does_not_track_the_default_device`, which asserts `true` on Metal and unchanged inside a CPU guard; nested, capture, unwinding, live-count, and deprecated-shim tests)
- [x] `cargo test --profile test-fast --features metal,accelerate --lib execution::runtime`: 19 passed
- [x] `MLXCEL_DEVICE=cpu cargo test --profile test-fast --features metal,accelerate --lib execution::runtime`: 19 passed (`device == Cpu`, `cpu_override == true`, backend answer unchanged, default device restored by the guard)
- [x] `cargo test --profile test-fast --features metal,accelerate --lib multimodal::host_preprocessor::tests::the_default_device_is_restored_after_the_export_tests` alone: 1 passed
- [x] `cargo test --profile test-fast --features metal,accelerate --lib multimodal::host_preprocessor`: 24 passed, 1 ignored
- [x] `cargo test --profile test-fast --features metal,accelerate --lib vision::merge`: 4 passed
- [x] `cargo test --profile test-fast --features metal,accelerate --lib models::gemma3_embedding`: 11 passed (exercises the reduced `mlx_test_guard`)
- [x] `cargo test --profile test-fast --features metal,accelerate --lib -- --test-threads=1 multimodal:: rerank::real_checkpoint_tests` (the PR #1420 reproduction): 262 passed, 26 ignored; the reranker gates soft-skip because their checkpoints are not on this host
- [x] `cargo test --profile test-fast --features metal,accelerate --test sampling_gumbel_kill_switch` and `--test sampling_rejection_kill_switch`: 1 passed each
- [x] `target/test-fast/mlxcel generate -m mlx-community/Phi-3.5-mini-instruct-4bit -p "Say hello." -n 3`: prints `Runtime device: Apple GPU (Metal)` and decodes at 13.27 tok/s; with `MLXCEL_DEVICE=cpu` it prints `Runtime device: CPU` and `Running on the CPU because MLXCEL_DEVICE=cpu asked for it (GPU backend available: true).`, then decodes on the CPU at 0.44 tok/s

## Not verified on this host

- CUDA build of the bridge (`--features cuda`): the new function uses only `mlx::core::device_count`, which `gpu_device_count` already calls on every backend without an `#ifdef`, and `no_cuda` stubs are untouched, but no CUDA toolchain is available here.
- `gpu_backend_available()` returning `true` on the GB10 CUDA build, and `false` on a CUDA build without a driver (reasoned from the pinned MLX `cudaGetDeviceCount` path), including the `initialize_runtime()` behavior in that driverless case.
- The reranker gate numbers themselves (0.9883 for the Beijing document, finite image scores): the checkpoints are not on this host, so those gates soft-skip.

Closes #1421


inureyes:
## Final implementation hardening

Audited the full PR diff and history against [issue #1421](https://github.com/lablup/mlxcel/issues/1421), including the pinned MLX device implementation, all in-tree default-device movers, runtime consumers, test guards, and existing review context.

### Findings fixed

- **MEDIUM — redundant process-global writes:** `initialize_runtime()` wrote the CPU default on every CPU resolution even when MLX was already on the CPU. The write is now limited to an actual GPU→CPU transition, preserving the CUDA-without-driver behavior while avoiding repeated mutation of MLX's plain non-atomic global during CPU-only and repeated test initialization.
- **MEDIUM — panic-unsafe multi-GPU test restoration:** `trivial_op_runs_on_non_default_gpu_index` manually reset GPU 0 only after evaluation. It now captures the prior default with `DefaultDeviceGuard`, so a panic restores the original device before the shared lock is released.

### Review outcome

- **Correctness:** The split between backend availability and current default-device state matches the original issue intent, runtime resolution applies the required CPU transition, and every in-tree test mover is serialized and restored.
- **Security:** No authentication, untrusted-input, secret-handling, or privilege boundary is introduced. The bridge exposes only device state/count booleans; no security defect was found.
- **Reliability/performance:** The two findings above are fixed in commit [`6d6ef1b7`](https://github.com/lablup/mlxcel/commit/6d6ef1b782c3dbcfe740d2073350dbea646aa358). The bilingual technical reports now record the final findings and NVIDIA validation.

### Validation

- `cargo fmt --all -- --check` and `git diff --check` — clean.
- `cargo clippy --workspace --all-targets --profile test-fast --features cuda -- -D warnings` — clean; pre-existing C++ compiler warnings remain outside this diff.
- `cargo test --profile test-fast --features cuda -p mlxcel-core --lib streams:: -- --test-threads=1` — 18 passed both before and after the hardening patch.
- `cargo test --profile test-fast --features cuda --lib execution::runtime -- --test-threads=1` — 19 passed.
- `MLXCEL_DEVICE=cpu cargo test --profile test-fast --features cuda --lib execution::runtime -- --test-threads=1` — 19 passed.
- `cargo test --profile test-fast --features cuda --lib -- --test-threads=1 multimodal:: rerank::real_checkpoint_tests` — 262 passed, 26 ignored; available real reranker checkpoints executed successfully.
- CPU-only backend-availability regression and root runtime suites — 1 passed and 19 passed, respectively.
- Host: NVIDIA GB10, compute capability 12.1. A CUDA-without-driver environment was not available; that branch remains supported by pinned-source analysis and pure runtime resolution tests.

The new GitHub checks for `6d6ef1b7` are running and will be required before merge.

Issue #1600, opened 2026-09-02
fix(youtu_vl): reverse_window_indices returns the permutation, not its inverse


inureyes:
## Problem / Background

`reverse_window_indices` in `src/vision/encoders/youtu_vl_window.rs:147-159` builds the wrong permutation. It sorts `(value, index)` pairs and writes `reverse[orig_idx] = rank`. For a permutation of `0..N`, which is exactly what `get_window_index` in the same file returns, the element with value `v` always lands at rank `v`, so the loop degenerates to `reverse[i] = window_index[i]`: it reproduces `window_index` itself instead of `argsort(window_index)`. The call site at `src/vision/encoders/youtu_vl.rs:420-427` even documents the intent as "Equivalent to `argsort(window_index)` (Python upstream)", so the comment and the code disagree.

Applying `take(h, window_index)` twice restores the original order only when the permutation is an involution. That happens whenever every merged grid dimension is a multiple of the window width, and fails otherwise, leaving the merged vision tokens scrambled before they reach the language model.

This is the same construction that `src/vision/encoders/qwen2_5_vl.rs` carried until #1596 (PR to follow). It was deliberately left out of that PR because Youtu-VL is a separate family with its own encoder, checkpoint and validation path, and the fix should be validated on a Youtu-VL checkpoint rather than assumed from the Qwen2.5-VL result.

## Evidence

Measured for Qwen2.5-VL, whose `get_window_index` has the same shape (4x4 merged-token windows from `window_size / spatial_merge_size / patch_size`):

- 448x448 image, 16x16 merged grid: the permutation is an involution, 0 of 256 merged tokens misplaced, and the defect is invisible.
- 336x336 image, 12x12 merged grid: 120 of 144 merged tokens misplaced. The mean absolute difference of the merger output against a transformers float32 oracle was 0.85 relative with the incorrect inverse and 3.2e-3 with `argsort`, and the model described a three-shape image as "A red circle and a blue triangle" instead of naming all three shapes with their own colors.
- 560x560: 360 of 400 misplaced. 448x672: 352 of 384 misplaced.

Whether the same grid sizes reach Youtu-VL depends on its own processor and `merger_window_size`, which is why this needs its own measurement.

## Proposed Solution

[...]

Issue #1592, opened 2026-09-02
fix(server): speculative decoding mis-attributes decode time to prompt_ms in native timings


inureyes:
## Problem / Background

On the speculative burst path (`--model-draft`, MTP or DFlash), the native `/completion` response's `timings` block reports almost the entire request wall time under `prompt_ms` and 0 to 1 ms under `predicted_ms`. `predicted_per_second` is therefore meaningless for any drafted request (95000 tok/s on the measurement below), and the per-token `predicted_per_token_ms` is equally wrong. The llama-server `timings` contract that `docs/llama-server-compat.md` pins to b10621 has `prompt_ms` cover prompt evaluation only and `predicted_ms` cover token generation measured from the first token (`docs/llama-server-compat.md:111`, which is also why `predicted_per_second` divides by `predicted_n - 1`).

This predates #1588. The same request measured before that PR showed `prompt_ms = 4075.0` and `predicted_ms = 0.0`; #1588 only added the `draft_*` acceptance counters (documented in `docs/speculative-acceptance.md`, "Per-request acceptance on the response itself"), which is what made the 48 verify rounds visible on the response and exposed where the time is going.

## Evidence

Measured 2026-09-02 on an Apple M1 Ultra, release build at main `2fddec101`, `mlxcel-server -m models/mlx/qwen3.5-4b-4bit`, `POST /completion` with `{"prompt":"Write a short paragraph about the history of the bicycle.","n_predict":96,"temperature":0}`.

| Configuration | prompt_n | prompt_ms | predicted_n | predicted_ms | predicted_per_second | draft counters |
|---|---|---|---|---|---|---|
| No drafter | 11 | 73.0 | 96 | 855.0 | not captured | absent |
| `--model-draft models/mlx/qwen3.5-4b-dflash` (auto-detected `--draft-kind dflash`) | 11 | 3930.0 (357 ms per prompt token) | 96 | 1.0 | 95000.0 | `draft_n = 655`, `draft_n_accepted = 47`, `draft_rounds = 48`, `draft_kind = "dflash"` |
| Same drafter, build before #1588 | 11 | 4075.0 | 96 | 0.0 | not captured | not reported yet |

Interpretation: on the drafted request, drafter warmup plus every verify round lands in `prompt_ms`, and the decode phase reports the few microseconds it takes to replay the finished token vector.

## Current Behavior

[...]

PR #1602, opened 2026-09-03
fix(server): let a malformed `MLXCEL_PAGED_SLAB_BLOCKS` fall back to the derived slab size

### LINKED ISSUE #1137: fix: Unparseable MLXCEL_PAGED_SLAB_BLOCKS warns about a derived size it never uses
## Problem / Background

When `MLXCEL_PAGED_SLAB_BLOCKS` is set to something that does not parse as a `usize`, the server logs that it is falling back to the derived slab size, but it actually keeps the fixed 32-block pool default. The log line describes a behavior the code does not perform.

`src/execution/memory_estimate.rs`, `resolve_paged_slab_blocks` (around line 994):

```rust
if let Ok(raw) = std::env::var(PAGED_SLAB_BLOCKS_ENV) {
    return match raw.trim().parse::<usize>() {
        Ok(0) => None,
        Ok(n) => Some(n),
        Err(_) => {
            tracing::warn!(
                "{PAGED_SLAB_BLOCKS_ENV}={raw:?} is not a non-negative integer; \
                 using the derived slab size"
            );
            None
        }
    };
}
// derivation logic follows here (byte clamp, geometry probe, etc.)
```

The chain of facts:

1. The `Err(_)` arm returns `None` from an early return, before the derivation logic below it ever runs. The derived slab size is never computed, let alone used.
2. `Err(_)` is therefore byte-for-byte identical in effect to the `Ok(0)` arm above it, which is documented as pinning the historical 32-block pool default.
3. The single production call site, `src/server/model_worker.rs:550-559`, consumes the result as `if let Some(blocks) = paged_slab_blocks { ... }`, so `None` skips the pool override entirely.
4. The pool then falls back to `POOL_SLAB_BLOCKS`, which is `pub const POOL_SLAB_BLOCKS: usize = 32;` at `src/lib/mlxcel-core/src/cache/paged.rs:509`.

So the operator is told the derived size is in use while the run actually gets the fixed 32-block default. Since a layer whose rows outgrow its first slab silently returns to the gather path, this can quietly cost the fused v2 decode path on a long-context deployment, and the log line points the operator away from the cause.

## Proposed Solution

Two options, and they are not equivalent. The maintainer should pick one.

**(a) Fix the message to describe what happens.** Change the warning text to say that the 32-block pool default is being kept, for example `"... is not a non-negative integer; keeping the 32-block pool default"`. One-line change, no runtime behavior change, and it leaves the existing documented contract intact.

**(b) Fix the behavior to match the message.** Fall through to the derivation on a parse error instead of returning early, so a malformed value really does land on the derived slab size. This is what the text promises today and is arguably the more useful behavior: a typo in the env var would then degrade to the tuned default rather than silently disabling the #899 sizing. This does change runtime behavior for a malformed env value.

If (b) is chosen, the documentation row from #1119 must also be updated: `docs/environment-variables.md:244` currently documents the *current* behavior with "A value that is not a non-negative integer warns and leaves that same 32-block default in place". Under (b) that sentence becomes wrong. Option (a) leaves that row correct as written.

## Acceptance Criteria

- [ ] A decision is recorded on this issue for whether the fix is (a) message-only or (b) behavior change.
- [ ] The warning text and the code path agree: whatever the log line claims is what `resolve_paged_slab_blocks` actually returns for an unparseable value.
- [ ] A unit test pins the chosen contract by asserting the return value of `resolve_paged_slab_blocks` for an unparseable `MLXCEL_PAGED_SLAB_BLOCKS` value (under (a): `None`, distinct from the derived value; under (b): equal to the value returned when the variable is unset). No existing test covers this today, which is why the log-versus-behavior drift went unnoticed.
- [ ] The test also distinguishes the parse-error case from the `Ok(0)` case, so the two arms cannot silently converge again.
- [ ] If (b) is chosen, the `MLXCEL_PAGED_SLAB_BLOCKS` row in `docs/environment-variables.md` is updated to match, and the `0` pin remains documented as unchanged.

## Technical Considerations

- Scope is small and local: one function in `src/execution/memory_estimate.rs`, plus a docs row only under option (b).
- The `Ok(0)` arm must keep returning `None` either way. That is the documented escape hatch for pinning pre-#899 allocation behavior and is relied on by benchmark sweeps.
- Related context: #899 introduced the derived slab sizing, #1104 and #1119 covered the environment-variable documentation for the paged decode v2 knobs.

rapsealk:
## Summary

`resolve_paged_slab_blocks` returned early on an unparseable `MLXCEL_PAGED_SLAB_BLOCKS`, which is the `0` pin (the historical 32-block slab), while its warning said the derived slab size was in use. This PR takes option (b) from the issue: the parse-error arm now falls through to the derivation, so a malformed value behaves exactly like an unset one, which is what the log line always claimed. `0` still pins the pool default.

## Related issues

Closes #1137

## Type of change

- [x] `fix` — bug fix
- [ ] `test` — tests only

## Decision recorded (acceptance criterion 1)

**(b), behavior change.** The two options cost the same one-arm edit, and (a) keeps a failure mode the issue itself names: a typo in the variable silently pins the 32-block slab, so a layer outgrows it on the first long prompt and the fused paged decode path (#899) becomes unreachable while the operator has been told the derived size is in use. Under (b) a typo degrades to the tuned default, which is the least surprising reading of "warn and continue". The `Ok(0)` arm is untouched, so the documented escape hatch for pre-#899 allocation behavior and the benchmark sweeps that rely on it are unchanged. If the maintainer prefers (a), the diff reduces to the message text plus the test's second assertion flipping to `None`.

## What changed

- `src/execution/memory_estimate.rs`: `Err(_)` no longer returns; it warns and falls through to the derivation. The warning no longer promises "the derived slab size", because the geometry probe right below it can still decline a model that is not pool-eligible, in which case nothing is derived and the pool default applies (same as unset). It now says the value is ignored and the slab is derived as if unset, in the structured `env_var = ..., value = ...` form the file's other env-var warnings use. The constant's rustdoc gains the same sentence. The single production call site (`src/server/model_worker.rs`, the `if let Some(blocks)` override) is unchanged.
- `paged_slab_blocks_malformed_env_derives_like_unset` pins the contract with all three arms in one test: unset and malformed return the same derived value (`Some(1024)` for ctx 8192, block 32, batch 4), `0` returns `None`, and a padded integer wins. The malformed and `0` arms are asserted against each other, so they cannot silently converge again (acceptance criterion 4). It also pins that an explicit value is verbatim: `8` resolves to `Some(8)`, below the 32-block floor the derived path applies, which was documented in the rustdoc but untested. It takes the shared `env_lock` and adds an `EnvRestore::unset` next to the existing `set`.
- `docs/environment-variables.md`: the `MLXCEL_PAGED_SLAB_BLOCKS` row now says a malformed value is ignored and the derived size applies as if unset (acceptance criterion 5), and its Values column says an explicit value is used verbatim, so the floor and cap in the Default column read as applying to the derived value only. `docs/CONTINUOUS_BATCHING.md` does not describe the malformed case, so it needed no change.
- `CHANGELOG.md`: entry under Unreleased / Fixed.

## Test plan

- [ ] `cargo deny check` (PR CI runs it; passed on this branch)

Run instead, on a CPU-only build (`MLXCEL_BUILD_METAL=0 MLXCEL_BUILD_ACCELERATE=0`), because this host has no Metal toolchain: `cargo clippy --profile test-fast -p mlxcel --all-targets -- -D warnings` (clean) and `cargo test --profile test-fast -p mlxcel --lib memory_estimate::tests` (39 passed, also at `--test-threads=8`). The two workspace gates above are left unticked deliberately.
- [ ] Real checkpoint: not applicable. The change is host-side slab sizing and moves no numerics; the pool allocation it selects is the same one an unset variable already selects.

The touched function has no Metal-dependent code.

## Notes for reviewers

- This PR takes `memory_estimate.rs` from 1,994 to 2,038 lines, across the 2,000-line row in `docs/code-guidelines.md`, with `mod tests` inline from line 1155. I left the split out to keep this to one logical change; happy to open a separate `refactor:` PR that moves the tests to `memory_estimate_tests.rs` first or after, whichever you prefer.
- Scope prefix is `server` because the observable effect is server startup; the file lives under `execution`, and precedent on it is mixed (`chore(execution)` #1573, `fix(core)` #1130, `feat(server)` #988).
The function is called once at worker startup, so the only observable behavior change is for operators who already had a malformed value in their environment: they were getting the 32-block slab and will now get the derived slab, which is the eager per-layer allocation the startup memory estimate already reports.

## Checklist

- [x] Updated `docs/` if user-facing behavior or supported models changed
- [x] Updated `// Used by: ...` comments on any shared function I modified (not applicable: single caller)


inureyes:

**Target PR:** #1602 (open, `fix/1137-paged-slab-blocks-parse`). **Linked issue:** #1137.

### Purpose fit

The issue asks a maintainer to choose between (a) fixing the message and (b) fixing the behaviour, and lists different follow-on work for each. This PR takes (b), which is the reading the existing warning text already promised, and does the follow-on work (b) requires: the `docs/environment-variables.md` row from #1119 said "warns and leaves that same 32-block default in place", which becomes false under (b), and it is updated. The `0` pin stays documented as unchanged, which the issue calls out as a hard constraint because benchmark sweeps rely on it.

Every acceptance bullet is satisfied, and the code matches. The `Err(_)` arm no longer returns; it warns and falls through to the derivation, so a malformed value now lands where the log line always claimed it did.

### Review findings

None requiring a code change. The things I checked and found correct:

- **The three arms are genuinely distinct now, and the test proves it rather than asserting it.** `paged_slab_blocks_malformed_env_derives_like_unset` computes the derived value with the variable unset, then asserts malformed equals that, `0` returns `None`, and an explicit `" 8 "` returns `Some(8)` verbatim (trimmed, and not floored at the 32-block pool default the derived path applies). That last assertion is the one that would catch a future "helpful" clamp on the explicit path.
- **Reversion check.** Restoring the pre-fix early return in the `Err(_)` arm fails the test with `left: None, right: Some(1024)`, which is exactly the defect #1137 describes: the parse-error arm silently returning the `0` pin instead of the derived size. Restored, the module is 43 passed. So the test is evidence, not decoration.
- **The env mutation is guarded.** The test holds `crate::test_support::env_lock::env_lock()` and the new `EnvRestore::unset` helper restores the previous value on drop, matching how the file's other environment tests behave, so it cannot leak state into a neighbour under `--test-threads=1` or otherwise.
- **No second reader to keep in sync.** `PAGED_SLAB_BLOCKS_ENV` is read only in `resolve_paged_slab_blocks`, so there is no other early return carrying the old behaviour.
- **The docs row and the doc comment now agree with each other and with the code**, including the detail that an explicit positive value is used verbatim, neither floored nor budget-capped.

### Security

No security surface. The variable is an operator-set sizing hint with no external input path, and the change narrows rather than widens what a malformed value can do: a typo now degrades to the tuned derivation instead of silently pinning an allocation policy the operator did not choose.

### Performance

This is a performance fix in effect, which is worth stating plainly since the diff looks like error handling. Under the old behaviour a typo in the variable silently kept the 32-block pool default, and since a layer whose rows outgrow its first slab falls back to the gather path, that quietly cost the fused v2 decode path on any long-context deployment while the log said the derived size was in use. Falling through to the derivation removes that trap. No hot path changes; `resolve_paged_slab_blocks` runs once at scheduler construction.

### Validation

- `cargo test --profile test-fast --features cuda -p mlxcel --lib execution::memory_estimate -- --test-threads=1`: 43 passed, 0 failed, including the new test.
- Reversion check as described above.
- `main` merged into the branch twice as it moved; the `CHANGELOG.md` conflicts (against the Youtu-VL entry #1600, then the burst-timings entry #1592) are resolved keeping every entry.

No follow-up code commits were needed; the branch carries only the merge resolutions.

Issue #1131, opened 2026-08-13
fix(cli): unreachable 2D-parallelism guard in validate_pipeline_parallel_args, decide whether to delete it or widen the early return


inureyes:
## Summary

`validate_pipeline_parallel_args` in `src/commands/generate.rs` contains an `ensure!` whose condition is the exact logical negation of an early return eight lines above it. The guard can never fail and its message can never reach a user. This was found while reviewing #1129 (which closes #1112) and deliberately left out of scope there, because #1112's acceptance criteria required the validation logic to be unchanged.

The fix is not obvious, because the dead guard is evidence of one of two different intents. This issue asks a maintainer to decide which reading is correct before any code is written.

### The tautology

`src/commands/generate.rs`, `validate_pipeline_parallel_args`:

```rust
if pp.pp_layers.is_none() && pp.pp_size <= 1 {
    return Ok(());
}

// ... comment block ...
let tp_size = args.tensor_parallel.tp_size;
if tp_size > 1 {
    ensure!(
        pp.pp_size >= 2 || pp.pp_layers.is_some(),
        "2D parallelism requires --pp-size >= 2 (or an explicit --pp-layers spec) \
         alongside --tp-size > 1"
    );
```

The early return fires when `pp.pp_layers.is_none() && pp.pp_size <= 1`. Past that point the negation holds: `pp.pp_layers.is_some() || pp.pp_size >= 2`, which is precisely the `ensure!` condition. The guard is therefore always satisfied whenever it is evaluated, so it is dead code and its error message is unreachable.

#1129 corrected the flag name inside that message (it previously said `--tensor-parallel-size`, which no binary accepts) but deliberately did not touch the logic, since #1112 scoped the change to the message text only.

### The test records the same tautology without naming it

`validate_pipeline_parallel_args_rejects_2d_without_pp_enabled` in `src/commands/generate_tests.rs` sets `pp_size = 1`, `tp_size = 2`, and asserts `is_ok()`. Its comment reads: \"The validator returns early when PP is disabled, so this case is harmless\". So the test name claims the validator *rejects* this input while the assertion confirms it *accepts* it. The name has to change whichever way the logic question is resolved.

## Proposed Solution

This is a decision to make first, not a patch to write first. The two readings lead to entirely different changes, and picking between them requires knowing the intended semantics of `--tp-size` without a pipeline topology.

[...]

Issue #1112, opened 2026-08-12
fix(cli): 2D-parallelism error tells the user to pass `--tensor-parallel-size`, which does not exist


rapsealk:
## Summary

The 2D (pipeline × tensor) parallelism validation in `src/commands/generate.rs:367-371` fails with a message telling the user to pass `--tensor-parallel-size > 1`. That flag does not exist; the real flag is `--tp-size`. A user who follows the instruction gets `error: unexpected argument '--tensor-parallel-size' found`. A comment three lines above points at two `docs/en/distributed/*.md` files that also do not exist.

## Background

`src/commands/generate.rs:367-371`:

```rust
ensure!(
    pp.pp_size >= 2 || pp.pp_layers.is_some(),
    "2D parallelism requires --pp-size >= 2 (or an explicit --pp-layers spec) \
     alongside --tensor-parallel-size > 1"
);
```

`grep -rn "tensor-parallel-size" src/` returns exactly one hit: this message string. It is never a clap flag. The tensor-parallel rank count is `--tp-size`, defined at `src/main.rs:728`, `src/main.rs:1378`, and `src/bin/mlx_server.rs:763`. The message is otherwise accurate (`--pp-size` and `--pp-layers` are both real), which makes the one wrong name more misleading than a vague message, because the other two names check out.

The comment at `src/commands/generate.rs:362-364` points at `docs/en/distributed/pipeline-parallelism.md` and `docs/en/distributed/tensor-parallelism.md`. Neither exists; `docs/en/` has never been part of this repository, and these are the only two `docs/en/` references in `src/`. The real operator document is `docs/distributed.md`.

## Proposed Solution

1. Change `--tensor-parallel-size` to `--tp-size` in the `ensure!` message.
2. Repoint the two comment references at `docs/distributed.md`, or drop them if that document does not cover 2D composition.

## Implementation Notes

- Message and comment text only; the validation logic is unchanged.
- Worth a quick sweep of the neighbouring `ensure!` messages in the same function while there: `--pp-micro-batch-size` (`src/commands/generate.rs:356`), `--pp-size`, and `--pp-layers` all resolve, but the file is the kind of place where one wrong name suggests checking the rest.

## Acceptance Criteria

- [ ] Every flag named in the message is one the binary accepts.
- [ ] No `docs/en/` reference remains in `src/`.
- [ ] `grep -rn -- "--tensor-parallel-size" src/` returns nothing.
- [ ] Message text only; the validation logic is unchanged.

---

[...]

PR #1604, opened 2026-09-03
test: assert every flag named in `src/commands` exists in some `--help`

rapsealk:
## Summary

Adds the machine check #1131 asked for as its separable follow-up: every `--flag` literal in `src/commands/*.rs` must be a flag some `mlxcel` subcommand or `mlxcel-server` actually accepts. This closes the class of defect behind #1112, where `validate_pipeline_parallel_args` told the user to pass `--tensor-parallel-size`, a flag no binary has, and the only sweep that could catch it was a shell pipeline in a PR description.

## Related issues

Refs #1131 (its last acceptance bullet: the flag-name invariant is added to `tests/cli_help_consistency.rs` rather than split into a new issue). Does not touch the `validate_pipeline_parallel_args` decision the issue is otherwise waiting on. Refs #1112, #1129.

## Type of change

- [ ] `fix` — bug fix
- [x] `test` — tests only

## How it works

One test, `every_flag_named_in_command_source_exists_in_some_help`, in the file that already owns the CLI-surface invariants.

- **Known flags** are collected from the built binaries, not from source: the signature lines and alias annotations of `--help` for every subcommand listed in `mlxcel --help`'s `Commands:` block (via the file's existing `all_documented_spellings`, so a flag named only in prose cannot whitelist itself), `mlxcel-server --help`, plus the `--dump-flag-surface` JSON of `mlxcel serve` and `mlxcel-server`. The dump is needed because the llama-server b10621 compatibility groups (`--spec-type`, `--mmproj`, `--numa`, `--rpc`, `--gpu-layers`, `--reasoning-format`, `--log-prompts-dir`, `--hf-file`, `--model-url`, `--docker-repo`, ...) are `hide = true`: with `--help` alone, all 13 mentions of them in `serve.rs` came back as unknown on the first run. Reusing the dump `tests/llama_compat_manifest.rs` already consumes means no hand-written allowlist.
- **Scanned text** is every `--[a-z][a-z0-9-]+` token in `src/commands/*.rs`, excluding `*_tests.rs`. Comments are scanned on purpose: a comment naming a flag that does not exist misleads the next reader the same way an error string misleads the user. The subcommand list is parsed from the name column of `mlxcel --help` at runtime, so a new subcommand is covered without editing the test. `long_flags` has the same kind of meta-test the file's other parsers carry.
- **Failure output** is `file:line: --flag`, one per line, so the fix is a jump.

Verified it catches the original defect: replacing one `--tp-size` in `src/commands/generate.rs` with `--tensor-parallel-size` fails the test with `src/commands/generate.rs:130: --tensor-parallel-size`; reverting makes it pass. On current `main` it passes clean, so no source change ships with it.

## Test plan

- [ ] `cargo deny check` (PR CI runs it)
- [ ] Validated with a real checkpoint: not applicable, no inference code is touched.

Run instead, on a CPU-only build (`MLXCEL_BUILD_METAL=0 MLXCEL_BUILD_ACCELERATE=0`), because this host has no Metal toolchain: `cargo test --profile test-fast --test cli_help_consistency` (27 passed, the two new tests included) and `cargo clippy --profile test-fast --test cli_help_consistency -- -D warnings` (clean). The two workspace gates above are left unticked deliberately. The test spawns the built binaries and reads source files only, so the Metal feature set does not change what it sees.

## Notes for reviewers

- Scope is `src/commands/` only, which is where the `--help`-facing subcommands live and where the #1112 defect was. `src/cli/*` and `src/server/cli_input.rs` are the compat translation layers and deliberately name llama-server spellings mlxcel rejects, so scanning them would need the allowlist this test avoids. Widening later is one line.
- No changelog entry, matching the repository's other `test:` commits.

## Checklist

- [x] Updated `docs/` if user-facing behavior or supported models changed (none)
- [x] Updated `// Used by: ...` comments on any shared function I modified (none)


inureyes:

**Target PR:** #1604 (open, `test/1131-flag-names-in-error-strings`). **Linked issue:** #1131 (its separable follow-up bullet; the `validate_pipeline_parallel_args` decision the issue is otherwise waiting on is untouched, correctly).

### Purpose fit

The issue asks for the `#1129` shell pipeline, which found the `--tensor-parallel-size` defect and then lived only in a PR description, to become a machine check. It is now one, in the file that already owns the CLI-surface invariants.

I did not take the PR body's word that it catches the original defect. Reproduced it: changing the one `--tp-size` at `src/commands/generate.rs:397` to `--tensor-parallel-size` fails the test with exactly `src/commands/generate.rs:397: --tensor-parallel-size`, and reverting returns the suite to 27 passed. So the gate catches the #1112 class, and its failure output is the promised jump-to-line.

The design choices hold up on inspection:

- Collecting known flags from the built binaries rather than from source is what makes this a real check rather than a restatement of the source.
- Consuming the `--dump-flag-surface` JSON is necessary, not incidental: the llama-server b10621 compatibility groups are `hide = true`, so `--help` alone would call every mention of them in `serve.rs` unknown, and reusing the dump `tests/llama_compat_manifest.rs` already consumes avoids a hand-written allowlist that would rot.
- Parsing the subcommand list from `mlxcel --help` at runtime means a new subcommand is covered without editing the test.
- `long_flags` has its own meta-test, matching how the file's other parsers are treated, and I traced it by hand for termination: `i` advances by at least two bytes per iteration in every branch, including the `--`, `---` and trailing-`--` cases the meta-test covers.

### Review findings

One, fixed in a follow-up commit on this branch.

**The source sweep was one `read_dir` deep.** `src/commands` is flat today, 24 files and no subdirectories, so the shallow scan covers everything and passes. The day a subcommand grows a submodule directory, the flags inside it stop being checked and the test stays green while covering less than its name claims. That is the same silent blindness this test exists to prevent, pointed at the test itself. The existing `!sources.is_empty()` assertion guards against the directory moving, and nothing guarded against it nesting.

The collection now walks the tree. Verified by construction rather than by passing: a throwaway `src/commands/nested_probe/probe.rs` naming `--definitely-not-a-real-flag` is reported as `src/commands/nested_probe/probe.rs:2: --definitely-not-a-real-flag` and fails the test; with the probe removed the suite is 27 passed.

### Noted, agreed as scoped

Restricting the sweep to `src/commands/` is right, and the PR's reasoning is correct: `src/cli/*` and `src/server/cli_input.rs` are the compat translation layers and deliberately name llama-server spellings mlxcel rejects, so scanning them would require the allowlist this design avoids. Scanning comments as well as error strings is also the right call, since a comment naming a flag that does not exist misleads the next contributor the same way an error string misleads the user.

### Security and performance

Not applicable to correctness or attack surface: the test adds no production code. On cost, it spawns `--help` once per subcommand plus two flag-surface dumps, which measured 2.4 seconds for the whole 27-test file, so it does not change the shape of the suite.

### Validation

- `cargo test --profile test-fast --features cuda --test cli_help_consistency -- --test-threads=1`: 27 passed, 0 failed.
- Defect reproduction and revert as described above.
- Nested-probe check for the recursion fix, then removed and re-run green.
- `cargo clippy --profile test-fast --features cuda --test cli_help_consistency -- -D warnings`: clean.
- `main` merged into the branch; no conflicts.

Issue #1441, opened 2026-08-26
fix(server): separate native completion and embedding routes from OpenAI routes


inureyes:
## Parent

- #1431

## Problem

The native `/completion` response and streaming envelope expose only a subset of b10621 fields. More seriously, `/completions` is routed to an OpenAI-shaped handler in mlxcel while b10621 treats it as a native llama-server alias. `/embeddings` has the same native-versus-OpenAI collision.

## Scope

- Align native `/completion` request parsing, default resolution, cache controls, token-return options, stop metadata, timings, and streaming termination.
- Route `/completions` to the native handler and retain OpenAI behavior exclusively under `/v1/completions`.
- Route `/embeddings` according to the b10621 native schema and retain OpenAI behavior under `/v1/embeddings`.
- Add compatible native fields including resolved generation settings, tokens/prompt when requested, cached/truncated state, stop type, stopping word, and progress reporting.
- Align reverse-prompt stopping and special-token output controls, including their interaction with streaming and returned token IDs.
- Match error status/envelopes and client-disconnect cancellation.
- Publish a migration note because the current non-`/v1` routes return different schemas.

## Acceptance criteria

- [x] Route method/path tests prove native and OpenAI handlers are no longer conflated.
- [ ] Non-streaming and streaming golden-schema tests cover normal length, EOS, stop-string, cancellation, cache hit, and truncation outcomes.
- [ ] Reverse-prompt and special-token tests distinguish emitted text, stop metadata, raw token IDs, and detokenization behavior.
- [x] Unknown native request fields are rejected or reported consistently rather than silently ignored when they affect semantics.
- [x] `/v1/completions` and `/v1/embeddings` remain OpenAI compatible.
- [x] Existing non-`/v1` behavior change is documented as breaking.

## Validation

- Compare the same prompt against b10621 and mlxcel using a real checkpoint, normalizing only backend-specific timing and model identifiers.

inureyes:
Partial implementation landed in #1467. This issue stays OPEN: closing it would strand manifest entries that are not honestly resolvable yet, and `scripts/ci/check_llama_compat_manifest.py --check-issues-open` fails the whole repository on a `deferred` entry whose issue is closed.

[...]

PR #1609, opened 2026-09-03
fix(server): stamp the speculative burst's first token at its prefill end so `timings` splits prompt from decode

### LINKED ISSUE #1592: fix(server): speculative decoding mis-attributes decode time to prompt_ms in native timings

rapsealk:
## Summary

On the speculative burst path the native `timings` block reported almost the whole request under `prompt_ms` and 0 to 1 ms under `predicted_ms`, because the first-token stamp was taken when the finished token vector was replayed after every verify round had already run. The burst now carries the instant its target prefill finished and stamps the sequence with it, so `prompt_ms` is queue wait plus prefill and `predicted_ms` is the verify rounds, as on classic decode. The drafter's once-per-process disk load is excluded from every `timings` field. Classic decode is untouched.

## Related issues

Closes #1592. Refs #1441 (streamed-equals-final invariant), #1588 (the `draft_*` counters that made the misattribution visible).

## Type of change

- [x] `fix` — bug fix
- [ ] `test` — tests only

## What changed

Implements the issue's four items, with one simplification to item 4.

- **`BurstSuccess` carries `prefill_end: Instant`** instead of the two `f64` millisecond fields that `finalize_burst_success` discarded (`src/server/batch/speculative_burst.rs`). DFlash B=1 derives it as `prefill_start + (total_burst - decode_elapsed)`, both timers it already kept. MTP B=1 stamps `generation_start` before the generator runs and adds `MtpGenerator`'s own `prefill_time_ms`. The batched window no longer passes `0.0, 0.0`: the DFlash wrapper stamps `Instant::now()` right after the batched target prefill and first-bonus sample, and the MTP wrapper derives it as `now - stats.decode_time_ms` from the batched round loop's decode slice. Every row of a window shares the one instant, since every row shared the one prefill.
- **`SequenceInfo::mark_first_token_at(Instant)`** (`src/server/batch/sequence.rs`) is the supplied-instant form; `mark_first_token()` delegates with `Instant::now()` and keeps its idempotent guard, which is what makes the replay's own call a no-op. It also stamps the new `StreamingDecodeState::stamp_first_token` (`src/server/model_worker.rs`), so the streamed `PrefillStats.prompt_ms` and the final `prompt_eval_ms` derive from one instant. `prompt_ms` clamps at zero if a driver ever hands in an instant earlier than the origin.
- **`finalize_burst_success` stamps `prefill_end` before streaming**, for both the B=1 and batched callers.
- **Cold-load exclusion without a new field.** `WorkerDrafterSlot::ensure_loaded` returns `Result<Option<Duration>, String>`: `Some(elapsed)` when this call loaded from disk, `None` when the drafter was resident. The four drivers and the scheduler's slice start (`src/server/batch/scheduler.rs`) then do `seq.created_at += load` (every row, in the batched window). That is the issue's `timing_origin_shift` applied to the origin itself: `created_at` feeds only `prompt_ms`, `prompt_eval_ms` and `generation_time_ms` (`sequence.rs`, and the zero-token result in `scheduler/prefill.rs`), so the effect is identical and it avoids adding a field to every `SequenceInfo` constructor, ten of which are test literals. The slice-grant reload in `scheduler/speculative_finalize.rs` keeps ignoring the value; it reloads a parked job's drafter mid-stream, where the first token is long gone.
- **Docs**: `docs/speculative-acceptance.md` "Per-request acceptance on the response itself" gains the paragraph on what `prompt_ms` and `predicted_ms` cover on a drafted request; the `timings` note in `docs/llama-server-compat.md` links back. `CHANGELOG.md` entry under Unreleased / Fixed.

## Tests

Four new tests in `src/server/batch/speculative_burst_tests.rs`, on `make_test_sequence()` with a backdated `created_at` so nothing sleeps:

- `mark_first_token_at_splits_prompt_time_from_generation_time`: a 300 ms-old sequence stamped 100 ms after its origin, then replayed through `on_token`, reports `prompt_eval_ms` in the prefill window and `generation_only_ms` covering the rest; a later `mark_first_token()` does not move it; the streamed `Prefill` snapshot's `prompt_ms` equals the final `prompt_eval_ms`.
- `shifting_created_at_by_a_cold_load_leaves_generation_time_alone`: a 50 ms origin shift reduces `prompt_eval_ms` by 50 and leaves `generation_only_ms` unchanged.
- `mark_first_token_at_clamps_an_instant_before_created_at`.
- `ensure_loaded_on_a_resident_drafter_reports_no_load`: `Ok(None)` on a slot holding `TrackingMockDrafter`.

Not written: the issue's sixth bullet, a sleeping `MockMtpTarget` through `drive_mtp_generator` asserting the derived `prefill_end`. The derivation happens in `run_mtp_burst`, which needs a `LoadedModel`, and what it adds is `generation_start + stats.prefill_time_ms`, arithmetic on a timer `MtpGenerator` already tests.

## Test plan

- [ ] `cargo deny check` (PR CI runs it)
- [ ] Validated with a real checkpoint: **not done here, see below.**

Run instead, on a CPU-only build (`MLXCEL_BUILD_METAL=0 MLXCEL_BUILD_ACCELERATE=0`), because this host has no Metal toolchain: `cargo test --profile test-fast -p mlxcel --lib server::batch::speculative_burst` (66 passed, the four new tests included) and `cargo clippy --profile test-fast -p mlxcel --lib --tests -- -D warnings` (clean). The two workspace gates above are left unticked deliberately.

## What is not verified

The issue's first four acceptance criteria are measurements on a real drafter pair (`qwen3.5-4b-4bit` + `qwen3.5-4b-dflash`, an MTP pairing, the batched window, streaming versus non-streaming), and the PR body is asked to carry before/after numbers for the two Evidence rows. None of that ran here: no checkpoints and no Metal toolchain on this host. The unit tests pin the bookkeeping the drivers feed; the `Evidence` request in the issue (`n_predict` 96, `temperature` 0, classic then drafted, first and second request) is the run to do on the M1 Ultra before merge. Expected after the fix: `prompt_ms` on the order of the classic 73 ms, `predicted_ms` on the order of the drafted wall time minus prefill, `predicted_per_second` in the low hundreds, `draft_rounds` unchanged at 48.

## Notes for reviewers

- The classic path's numbers do not move: `scheduler/prefill.rs` still stamps through `on_token` and then `mark_first_token()`, and no cold load shifts its origin. The one observable change there is that `mark_first_token()` now also stamps the decode state at the same instant it publishes the snapshot, which only tightens the #1441 invariant.
- `Instant + Duration` in the drivers cannot underflow; the batched MTP derivation uses `checked_sub` and falls back to `now`.

## Checklist

- [x] Updated `docs/` if user-facing behavior or supported models changed
- [x] Updated `// Used by: ...` comments on any shared function I modified (none are model-shared)


inureyes:

**Target PR:** #1609 (open, `fix/1592-burst-timings`). **Linked issue:** #1592.

The PR's own "What is not verified" section says the issue's first four acceptance criteria are real-hardware measurements that did not run, and that the `Evidence` request is "the run to do before merge". That run is done, on NVIDIA GB10 (CUDA, sm_121) with the exact checkpoint pair the issue names. It reproduces the bug and confirms the fix.

### Measurement

`mlxcel-server -m models/mlx/qwen3.5-4b-4bit [--model-draft models/mlx/qwen3.5-4b-dflash]`, release build with `--features cuda`, `POST /completion` with the issue's body (`n_predict` 96, `temperature` 0). Before and after are the same tree with only `src/` reverted between them, built into the same target directory, measured back to back on an otherwise idle box.

| Configuration | prompt_ms | predicted_ms | predicted_per_second | draft counters |
|---|---|---|---|---|
| Classic, warm | 66.0 | 1389.0 | 68.4 | absent |
| Drafted, **before**, 2nd request | 4860.0 | 1.0 | 95000.0 | 655 / 47 / 48 |
| Drafted, **before**, 3rd request | 4840.0 | 0.0 | 0.0 | 655 / 47 / 48 |
| Drafted, **after**, 2nd request | 69.0 | 4974.0 | 19.1 | 655 / 47 / 48 |
| Drafted, **after**, 3rd request | 67.0 | 4988.0 | 19.0 | 655 / 47 / 48 |

The before row lands on the issue's Apple Silicon figure exactly, `predicted_per_second` 95000.0, which is the arithmetic artefact of dividing 95 intervals by a 1 ms replay rather than anything machine-specific. `draft_n` 655, `draft_n_accepted` 47 and `draft_rounds` 48 are identical across all five drafted runs and identical to the issue's, so the fix moves the timings and nothing else.

Against the acceptance criteria:

- **`predicted_ms` and `prompt_ms`**: drafted warm `prompt_ms` is 67 to 69 against the classic path's 66, which is the "tens of ms, not thousands" the issue asked for, and `predicted_per_second` is 19, the same order as the classic 68 rather than 95000.
- **Streaming**: the first `timings` frame and the final frame both report `prompt_ms` 70.0 and `predicted_ms` 4910.0, so the #1441 streamed-equals-final invariant holds on the burst path, and the final frame matches the non-streaming value within jitter.
- **Cold load**: reported honestly as not decisively testable here. The `Drafter loaded (kind=dflash, 1 ms)` line shows the load costing 1 ms on this box, which is far below the first-request GPU warmup it would have to be visible against (first drafted request 1121 ms, first classic request 15945 ms). The exclusion mechanism is covered by the unit test instead.

### Review findings

**Design deviation from the issue, checked and endorsed.** The issue specifies a new `SequenceInfo::timing_origin_shift: Duration`; the PR instead does `seq.created_at += load`. That is only equivalent if `created_at` feeds nothing but the timings, so I checked every use rather than taking the PR body's word for it. In `src/server/batch/`, `created_at` is read at `sequence.rs:420` (the `Prefill` snapshot), `:441`, `:609` and `:615` (`finish_with_cache`), one log line at `scheduler/prefill.rs:1398`, and the four constructors. It is not used for admission order, priority, deadlines, or the decode watchdog, so shifting the origin cannot change scheduling. The simplification stands, and it avoids threading a field through ten test literals.

**One coverage gap, closed on this branch.** The fix changes the classic path in exactly one way: `mark_first_token()` now delegates to `mark_first_token_at`, which also stamps the decode state, so the streamed snapshot and the final frame share one origin instead of two calls a few dozen lines apart in `scheduler/prefill.rs` (the mark at `:1379`, `on_token` at `:1456`). All four of the PR's new tests drive the supplied-instant form, so the classic-path change had nothing pinning it, and a later reordering that let `on_token` win the stamp would move classic `prompt_eval_ms` silently. Added `classic_mark_first_token_stamps_the_decode_state_and_on_token_does_not_restamp`, which drives the classic ordering and asserts the snapshot equals the final frame, that `prompt_eval_ms` comes from the mark, and that the gap lands in `generation_only_ms`.

Verified by reversion rather than by passing: with `decode_state.stamp_first_token(at)` removed, the new test and `mark_first_token_at_splits_prompt_time_from_generation_time` both fail; restored, all 67 tests in the module pass.

**Checked and found correct, no change needed:**

- The batched MTP derivation `prefill_end = now - stats.decode_time_ms` depends on the batched round loop populating `decode_time_ms`, which the issue's own text suggests it might not. It does: `build_stats` in `speculative/mtp/round_loop_batched.rs` sets `decode_time_ms` from the real `decode_elapsed` and only the zero-round `zero_stats()` path leaves it at 0.0, which is the issue's explicitly legitimate "zero verify rounds" edge case rather than a residue of the bug.
- Event ordering is preserved. `finalize_burst_success` now emits the `Prefill` event before `begin_burst_stream`, ahead of every `Token` event, so no consumer sees tokens before the snapshot.
- Both new doc cross-links resolve: `llama-server-compat.md#completions-and-embeddings-are-native-routes-not-openai-aliases-1441` matches the `### /completions and /embeddings are native routes, not OpenAI aliases (#1441)` heading, and `speculative-acceptance.md#per-request-acceptance-on-the-response-itself` matches its heading.
- The issue's unwritten sixth test bullet (a sleeping `MockMtpTarget` through `drive_mtp_generator`) is a fair omission as the PR argues, and the measurement above covers the derivation end to end on real hardware, which is stronger than the synthetic version would have been.

### Security

No security surface. No route, auth path, input parser, or resource bound moves; the change is bookkeeping on instants already held in process. Worth noting the fix removes a small operator-facing misinformation hazard rather than adding one: `predicted_per_second` of 95000 on a drafted request is the kind of number a capacity plan gets built on.

### Performance

No runtime cost. `ensure_loaded` returns a `Duration` it already measured for its log line, and the drivers derive `prefill_end` by arithmetic on timers the round loops already keep. No new timer, allocation, or lock is introduced on the burst path. The measured decode rate is unchanged in substance: the 19 tokens per second the drafted path now reports is what it was always doing, previously hidden inside `prompt_ms`.

### Validation

- `cargo test --profile test-fast --features cuda -p mlxcel --lib server::batch::speculative_burst -- --test-threads=1`: 67 passed, 0 failed.
- Reversion check on the new test: 2 failed with the stamp removed, 67 passed with it restored.
- `cargo clippy --profile test-fast --features cuda -p mlxcel --lib --tests -- -D warnings`: clean.
- Real-checkpoint measurement above, non-streaming and streaming, classic and drafted, before and after.
- `main` merged into the branch; the `CHANGELOG.md` conflict against the Youtu-VL entry (#1600) is resolved keeping both.

The workspace-wide suite was not run: it aborts on this CUDA box for reasons that predate this branch and reproduce on `main`.

PR #1617, opened 2026-09-04
bench: refresh the 0.7.0 sweeps and fix the dtype, rotary and rope defects they exposed

inureyes:
Refreshes the 0.7.0 benchmark sweeps on M1 Ultra and M5 Max, and fixes the defects the sweeps exposed. The measuring turned out to be the smaller half: most of the diff is the runtime bugs that only showed up once two machines were compared row by row.

## Performance

**Gemma3n held two dtypes in one checkpoint.** The load policy kept the language MLP bf16 and converted everything else to f16, and MLX promotes f16 met with bf16 to f32, so every boundary between them promoted and the decode path read promoted weights the whole way down. Loading uniformly bf16 moves M5 Max decode from 38.02 to 47.48 tok/s and prefill from 2143 to 2371, taking decode from 78% of mlx-vlm to 97%. This is not a claim that bf16 beats f16: on families with no forced-bf16 subset the two land within noise, `qwen2.5-0.5b-bf16` at 401.8 against 397.3 and `llama-3.1-8b-bf16` at 33.07 against 33.02. The mixture is what costs, so every other family keeps converting.

**The 3D rotary dropped the input dtype.** `cos` and `sin` are built in f32 on purpose, and the rotated result left the function as f32, carried through attention and the output projection, and made every later layer promote its own weight to match. Restoring the dtype is worth up to 2.45x on M1 Ultra and 3.26x on M5 Max, largest on the non-quantized checkpoint and smallest on 4-bit dense, in proportion to the weight bytes the promotion reached.

**Two defects in `nemotron_nas`, and the family had never run.** `b429947c` fixed the rotation order and was insufficient on its own; the remaining defect read only `factor` out of `rope_scaling` and applied `1 / factor` as a uniform position scale, which is the `linear` scheme, to a checkpoint declaring `rope_type: "llama3"` whose rescaling is piecewise in the frequency. Before the first fix the model emitted a single "I"; after it, a fluent loop; only with both does it answer. It measures 19.37 against mlx-lm's 19.55 tok/s on the same machine, 99%.

## Method

Three of the day's mistakes came from reading a tool's output without checking its premise, so the guards are part of the diff.

`bench_decode.sh` stamps `mlxcel_commit` from git, not from the executable, so a sweep run without rebuilding after a rebase records provenance its binary does not have. That happened, and the rows read as "this change does nothing on this machine" for a change worth 3.26x there; both arms of the follow-up A/B were the same stale binary, so they agreed. The harness now refuses to run when the source is newer than the benchmark binary.

A verification at a length that cannot reach the changed branch is not evidence. The `nemotron_nas` rope fix first passed a 1300-token needle test, but `original_max_position_embeddings` is 8192 and the llama3 low-frequency band only carries positions past it. Re-run at 14737 tokens the fixed arm answers and the reverted arm does not, which is the test discriminating. `docs/benchmarks.md` gains the falsification condition.

## Catalog

Sixty directories held a name that disagreed with the checkpoint inside them: `llama-3.1-8b-4bit` held `Meta-Llama-3.1-8B-Instruct-4bit`, `nemotron-h-30b-4bit` held `NVIDIA-Nemotron-3-Nano-30B-A3B-4bit`, `glm4-flash-4bit` held GLM-4.7-Flash. A benchmark row is read by its name, and a table that shows a bf16 checkpoint beside a 4-bit one is only readable if the names say which is which. `docs/model-catalog.md` documents the rule and `docs/model-catalog.tsv` records what each name rests on, including the aliases that join a past-dated row to a current one.

Nine duplicate directories were byte-identical to a correctly named twin and were dropped, and two were names with no model behind them: a config, a README and no shards at all, tens of gigabytes unmaterialised in `.cache/` blobs, both listed for transfer to the other host. The catalog now carries a shard verdict per host so directory presence is no longer the only check.

## Known

`#1717` and `#997` track a sporadic NaN in the Granite 4 Vision text-only forward, about one process in six, present before this branch and unrelated to it. The gate is otherwise clean: fmt, `clippy --workspace --all-targets`, and 10532 tests.


inureyes:
Correction to the commit body of `c07826df`: the Falcon-H1 decode figure there reads `583.20`, and the right number is **597.84**.

`583.20` came from a run taken before the fix was committed, on a dirty tree (`fedf80c3-dirty`). The row that landed in `benchmarks/metal_m5max_2026-09-06.csv` is a re-measurement from the clean commit, and it is the only Falcon-H1 single-model CSV from that day:

```
metal_m5max_2026-09-07_single_falcon-h1-tiny-90m-instruct-4bit.csv
  decode=597.84  mlxcel_commit=c07826df
```

Both runs measured the same code, so the 2.5% gap is run-to-run variance rather than a code difference. The CSV is the source of truth per the benchmark workflow, so use `597.84`, `481.75` and `212.15`.

The commit message is left as it is: `c07826df` has three commits on top of it and the branch is shared with the M1 Ultra measurement session, so rewriting it would cost more than the wrong figure does. The underlying mistake was writing the commit body before the CSV was settled; the order should be measure, record, then write.

PR #1644, opened 2026-09-04
docs: point --distributed-config examples at the checked-in TOML templates

### LINKED ISSUE #1633: docs: point the six --distributed-config examples at files that exist in the repo
## Summary

Six sites (the distributed docs plus the `--help` text of both binaries) tell the user to pass `--distributed-config examples/distributed/generated_pipeline_remote_2node_tcp.toml`. That file is not in the repo and is not gitignored: it is written at runtime by a benchmark script. The files that are checked in are never mentioned.

## Background

Referencing sites:

- `docs/distributed.md:91,97`
- `src/main.rs:966,970`
- `src/bin/mlx_server.rs:137,141,146,151`

The referenced file is generated by `scripts/benchmark_pipeline_remote_rollout.sh:93`. The checked-in equivalents are `examples/distributed/pipeline_remote_2node_tcp.toml` and `examples/distributed/pipeline_remote_2node_thunderbolt.toml` (both confirmed present), with structure that matches what the script generates. Nothing tells the reader that, and `examples/distributed/` has no README.

## Proposed Solution

Either point the six example invocations at the checked-in templates, or add a short `examples/distributed/README.md` explaining the checked-in versus generated split (or both).

## Implementation Notes

- No distributed hardware is required to make or verify this change.
- If pointing invocations at the templates, keep the docs and both binaries' `--help` text consistent with one another.

## Acceptance Criteria

- [ ] Every `--distributed-config` example in the docs and in `--help` references a file that exists in the repo, or the generated-file workflow is explained where it is referenced.

---

### Title: docs: six sites point --distributed-config at a runtime-generated file that is not in the repo

Six sites — the distributed docs plus the `--help` text of both binaries — tell the user to pass `--distributed-config examples/distributed/generated_pipeline_remote_2node_tcp.toml`. That file is not in the repo and is not gitignored: it is written at runtime by a benchmark script. The files that *are* checked in are never mentioned.

## Evidence

Referencing sites:

- `docs/distributed.md:91,97`
- `src/main.rs:966,970`
- `src/bin/mlx_server.rs:137,141,146,151`

The referenced file is generated by `scripts/benchmark_pipeline_remote_rollout.sh:93`. The checked-in equivalents are `examples/distributed/pipeline_remote_2node_tcp.toml` and `examples/distributed/pipeline_remote_2node_thunderbolt.toml`, with byte-identical structure to what the script generates. Nothing tells the reader that, and `examples/distributed/` has no README.

## Suggested fix

Either point the six example invocations at the checked-in templates, or add a short `examples/distributed/README.md` explaining the checked-in-vs-generated split (or both). No distributed hardware is required.

## Acceptance criteria

- [ ] Every `--distributed-config` example in docs and `--help` references a file that exists in the repo, or the generated-file workflow is explained where it is referenced

agatha197:
## Summary

Points every `--distributed-config` example — `docs/distributed.md` plus the `after_help` string literals of `mlxcel serve` (`src/main.rs`) and `mlxcel-server` (`src/bin/mlx_server.rs`) — at the checked-in `examples/distributed/pipeline_remote_2node_tcp.toml` instead of the runtime-generated `generated_pipeline_remote_2node_tcp.toml`, and adds a short `examples/distributed/README.md` explaining the checked-in templates versus the `generated_*.toml` files written by `scripts/benchmark_pipeline_remote_rollout.sh`. The `.rs` changes touch only help-text string content, no code logic.

## Related issues

Closes #1633

## Type of change

- [ ] `fix` — bug fix
- [x] `docs` — documentation only
- [ ] `test` — tests only

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,7 @@ webpage/site/node_modules/
 # Raw qmv-wide-pin measurement runs written by scripts/bench_qmv_wide_pin.sh
 # (issue #1261). Curated results belong in docs/benchmark_results/.
 /bench-results/
+
+# Cluster TOMLs written at runtime by scripts/benchmark_pipeline_remote_rollout.sh write-config.
+# The checked-in templates beside them (pipeline_remote_2node_*.toml) are not covered.
+examples/distributed/generated_*.toml
--- a/docs/distributed.md
+++ b/docs/distributed.md
@@ -88,13 +88,13 @@ A minimal shape looks like this:
 ```bash
 # Stage process.
 mlxcel-server -m models/<checkpoint> \
-    --distributed-config examples/distributed/generated_pipeline_remote_2node_tcp.toml \
+    --distributed-config examples/distributed/pipeline_remote_2node_tcp.toml \
     --node-id stage-1 \
     --host 0.0.0.0 --port 18081 --no-warmup
 
 # Coordinator / serving process.
 mlxcel-server -m models/<checkpoint> \
-    --distributed-config examples/distributed/generated_pipeline_remote_2node_tcp.toml \
+    --distributed-config examples/distributed/pipeline_remote_2node_tcp.toml \
     --node-id coordinator \
     --host 0.0.0.0 --port 18080 \
     --parallel 2 --max-batch-size 2 --pp-micro-batch-size 2
--- a/examples/distributed/README.md
+++ b/examples/distributed/README.md
@@ -1,0 +1,13 @@
+# Distributed config examples
+
+`pipeline_remote_2node_tcp.toml` and `pipeline_remote_2node_thunderbolt.toml`
+are checked-in `--distributed-config` templates for a two-node remote pipeline
+(coordinator + two stages, over plain TCP and over a Thunderbolt Bridge
+network respectively). Copy or edit one to match your node addresses before
+starting the servers; see [`docs/distributed.md`](../../docs/distributed.md)
+for the full workflow.
+
+`scripts/benchmark_pipeline_remote_rollout.sh write-config` generates
+`generated_*.toml` variants of the same shape in this directory at runtime,
+filled from its `CLUSTER_NAME` / `TRANSPORT_BACKEND` / `*_ADDR` environment
+variables. Those generated files are not checked in.
--- a/src/bin/mlx_server.rs
+++ b/src/bin/mlx_server.rs
@@ -134,21 +134,21 @@ Remote Pipeline Parallel Example (TCP):
        STAGE0_ADDR=192.168.1.22:19001 \\
        STAGE1_ADDR=192.168.1.24:19001 \\
        scripts/benchmark_pipeline_remote_rollout.sh write-config \\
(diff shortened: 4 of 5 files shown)


inureyes:

**Target PR:** #1644 (open, `docs/distributed-config-examples`). **Linked issue:** #1633.

### Purpose fit

The issue reports that six sites point `--distributed-config` at a runtime-generated file not present in a fresh checkout. Both checked-in templates do exist (`examples/distributed/pipeline_remote_2node_tcp.toml` and `..._thunderbolt.toml`, tracked since June), and the new `examples/distributed/README.md` is the missing piece: nothing in the tree previously said which of those TOMLs were templates and which appear only at runtime.

But the sites are not all the same case, and the PR treated them as one.

### Review findings

One, and it inverted the PR's own intent. Fixed in a follow-up commit on this branch.

**The `--help` blocks' step 1 was retargeted to overwrite the checked-in template.** The `after_help` in `src/main.rs` and `src/bin/mlx_server.rs` is a self-contained four-step walkthrough: step 1 runs `scripts/benchmark_pipeline_remote_rollout.sh write-config <path>` and steps 2 through 4 start servers against what step 1 wrote. Renaming all four occurrences to `pipeline_remote_2node_tcp.toml` did not remove a dangling reference there, because step 1 creates the file; it redirected the generator's output onto the template the same PR introduces. Following the help as printed would clobber `examples/distributed/pipeline_remote_2node_tcp.toml` with cluster-specific addresses, leave the tree dirty, and silently revert an operator's edits on the next `git checkout`. It also contradicts the new README, which states that generated files are the `generated_*.toml` ones and are not checked in.

The eight sites in those two files go back to `generated_pipeline_remote_2node_tcp.toml`, which is additionally the script's own default output path (`$ROOT_DIR/examples/distributed/generated_pipeline_remote_2node_${TRANSPORT_BACKEND}.toml`), so the help and the script agree without an argument.

The two `docs/distributed.md` sites keep the PR's change. That page shows the stage and coordinator processes started with no generation step in front of them, so there the path really did dangle, and pointing it at the checked-in template is the correct fix. This is the case the issue actually reported.

Also added a `.gitignore` rule for `examples/distributed/generated_*.toml`, so following the help does not leave untracked files behind. Verified both directions with `git check-ignore`: the generated name matches, and neither checked-in template does.

### Security

No security surface. The cluster TOMLs carry node addresses and no credentials, and no example in the diff instructs a bind to a wider interface than before. Worth noting that the fix removes a small integrity footgun rather than adding one: an operator's edited cluster addresses can no longer be silently reverted by a routine `git checkout` of a tracked file.

### Performance

Not applicable. Help text, documentation, and a `.gitignore` entry; no runtime path changes.

### Validation

`git diff` against `main` is now: two `docs/distributed.md` sites repointed, the new `examples/distributed/README.md`, and the `.gitignore` rule. The two `.rs` files are byte-identical to `main` again, so the help text they compile in is unchanged. `grep` confirms no remaining `examples/distributed/` reference in `src/` outside the generated name. CI is green after the follow-up commit, including `cargo-fmt`, `cargo-clippy` and `cargo-deny`.

PR #1646, opened 2026-09-04
docs(python): document the LLM/AsyncLLM constructor parameters and server_kwargs

### LINKED ISSUE #1638: docs(python): document the LLM/AsyncLLM constructor parameters and the server_kwargs pass-through
## Summary

The `LLM` and `AsyncLLM` constructors are the only public API entry points in the Python bindings without parameter docstrings, and the `**server_kwargs` pass-through (the mechanism that forwards arbitrary flags to the spawned `mlxcel serve` process) is documented nowhere.

## Background

`LLM.__init__` (`python/src/mlxcel/_client.py:55`) and `AsyncLLM.__init__` (`python/src/mlxcel/_async_client.py:43`) both accept `model, base_url, socket, api_key, binary, host, port, timeout, startup_timeout, transport, **server_kwargs`. Every other public method in `python/src/mlxcel/` carries a docstring; these two constructors are the only gaps. `**server_kwargs` has zero mentions in `python/README.md` or `docs/python-client.md`.

## Proposed Solution

Add parameter docstrings to both constructors in the project's existing docstring style, and document `**server_kwargs` with at least one worked example in `python/README.md` and/or `docs/python-client.md`.

## Implementation Notes

- The parameter meanings are readable from `_spawn` and `_connect`, defined directly below the constructor in `_client.py`.
- The class is fully type-annotated, so this is documentation only, with no signature change.
- The two constructors mirror each other, so keep their docstrings consistent.

## Acceptance Criteria

- [ ] Both `__init__` methods have parameter docstrings in the existing style.
- [ ] `**server_kwargs` is documented with at least one example.

---

### Title: docs(python): LLM/AsyncLLM constructors document none of their parameters, and **server_kwargs is documented nowhere

The `LLM` and `AsyncLLM` constructors are the only public-API gaps in the Python bindings' docstring coverage, and the `**server_kwargs` pass-through — the mechanism that forwards arbitrary flags to the spawned `mlxcel serve` — is documented nowhere.

## Evidence

- `python/src/mlxcel/_client.py:55-69` — `LLM.__init__` takes `model, base_url, socket, api_key, binary, host, port, timeout, startup_timeout, transport, **server_kwargs` with no docstring
- `python/src/mlxcel/_async_client.py:43` — same for `AsyncLLM.__init__`
- An AST sweep over `python/src/mlxcel/*.py` shows every other public method has a docstring; these two `__init__`s are the only gaps
- `**server_kwargs` has zero hits in both `python/README.md` and `docs/python-client.md`

## Suggested fix

Pure docstring work on a small, fully type-annotated class: document each parameter (their meanings are readable from `_spawn`/`_connect` directly below), and add a `**server_kwargs` mention with an example to `python/README.md` and/or `docs/python-client.md`.

## Acceptance criteria

- [ ] Both `__init__`s have parameter docstrings in the project's existing docstring style
- [ ] `**server_kwargs` is documented with at least one example

agatha197:
## Summary

Adds Google-style `Args:` docstrings to `LLM.__init__` (`python/src/mlxcel/_client.py`) and `AsyncLLM.__init__` (`python/src/mlxcel/_async_client.py`) — the last two public-API docstring gaps in the Python bindings — with each parameter's meaning taken from `_connect`/`_spawn` and `ManagedServer`. Also documents the `**server_kwargs` pass-through (`ctx_size`, `n_predict`, `alias`, `warmup`, `extra_args`, `shutdown_grace`) with an example in `python/README.md` and in the managed-mode section of `docs/python-client.md`.

`ruff check` and `ruff format --check` pass on both touched modules.

## Related issues

Closes #1638

## Type of change

- [ ] `fix` — bug fix
- [x] `docs` — documentation only
- [ ] `test` — tests only

### DIFF
--- a/docs/python-client.md
+++ b/docs/python-client.md
@@ -37,6 +37,18 @@ The context manager (`with`) is the recommended form; it guarantees shutdown. Th
 
 Useful managed-mode arguments: `binary=`, `host=`/`port=` (forces TCP and binds there), `socket=` (an explicit Unix socket path), `api_key=`, `ctx_size=`, `n_predict=`, `alias=`, `warmup=`, `extra_args=[...]` (forwarded verbatim to `mlxcel serve`), and `startup_timeout=`.
 
+The last group (`ctx_size=`, `n_predict=`, `alias=`, `warmup=`, `extra_args=`) travels through `**server_kwargs`: any extra keyword argument configures the spawned `mlxcel serve` process rather than the client.
+
+```python
+with mlxcel.LLM(
+    "mlx-community/Qwen3-4B-4bit",
+    ctx_size=8192,
+    warmup=False,
+    extra_args=["--parallel", "1"],
+) as llm:
+    print(llm.generate("hello", max_tokens=32))
+```
+
 ## Connect mode
 
 Pass `base_url=` or `socket=` (but not a model) to talk to a server you started yourself. No subprocess is launched or managed.
--- a/python/README.md
+++ b/python/README.md
@@ -51,6 +51,20 @@ Async usage mirrors the sync API via `mlxcel.AsyncLLM` (`await llm.generate(...)
 
 On POSIX, managed mode defaults to a Unix domain socket for low-overhead local IPC. Keep socket paths short (`sun_path` is about 104 bytes on macOS, 108 on Linux); the default lives under `/tmp`. Pass `socket=` to override. Windows uses TCP.
 
+## Server options (`**server_kwargs`)
+
+In managed mode, extra keyword arguments to `LLM` / `AsyncLLM` are forwarded to the spawned `mlxcel serve` process: `ctx_size`, `n_predict`, `alias`, and `warmup` map to the matching CLI flags, `extra_args=[...]` appends raw CLI arguments verbatim, and `shutdown_grace` sets the termination grace period on close.
+
+```python
+with mlxcel.LLM(
+    "mlx-community/Qwen3-4B-4bit",
+    ctx_size=8192,
+    warmup=False,
+    extra_args=["--parallel", "1"],
+) as llm:
+    print(llm.generate("hello", max_tokens=32))
+```
+
 ## Sampling parameters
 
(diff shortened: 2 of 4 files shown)


inureyes:

**Target PR:** #1646 (open, `docs/python-client-docstrings`). **Linked issue:** #1638.

### Purpose fit

Both halves of the issue are covered: every constructor parameter on `LLM` and `AsyncLLM` now has an `Args:` entry, and `**server_kwargs` is documented in three places (both docstrings, `python/README.md`, and `docs/python-client.md`). I checked the substantive claims against the client sources rather than accepting the docstrings as self-evidently true.

- The `**server_kwargs` list (`ctx_size`, `n_predict`, `alias`, `warmup`, `extra_args`, `shutdown_grace`) is exactly the keyword set `ManagedServer.__init__` accepts in `python/src/mlxcel/_server.py`, with nothing invented and nothing left out.
- "Map to the matching CLI flags" holds: the spawn builder emits `--ctx-size`, `--n-predict`, `-a`, and `--warmup` or `--no-warmup` from the tri-state `warmup`, then appends `extra_args` verbatim, which is what "raw CLI arguments appended verbatim" claims.
- `shutdown_grace` is the timeout passed to `proc.wait` on close, so "sets the termination grace period" is accurate.
- `binary` resolution order is documented as explicit argument, then `MLXCEL_BIN`, then `PATH`, which matches the resolver's own docstring and its `binary or os.environ.get("MLXCEL_BIN")` fallback.
- The API key claim is the one worth checking, because getting it wrong would be a security statement rather than a style one: `api_key` is passed to the child in `env["LLAMA_API_KEY"]` and never appears in the spawn argv, so "via the `LLAMA_API_KEY` environment variable (never argv)" is correct as written.
- `timeout` defaults to `_common.DEFAULT_TIMEOUT`, which is `httpx.Timeout(connect=10.0, read=600.0, write=30.0, pool=10.0)`, so the "generous 600-second read" parenthetical names the real value.
- The async-only divergence is real: `AsyncLLM` resolves the model id through `_resolve_model` on first use while the sync client resolves in the constructor, which is what "unlike the sync client, the model id resolves lazily on the first request" describes.

### Review findings

None. The two docstrings stay in sync with each other and diverge only where the clients actually diverge (lazy resolution, the synchronous spawn inside an async `__init__`), which is the divergence a reader most needs flagged.

### Security and performance

No behaviour changes: the diff is one docstring per client plus prose in two Markdown files. The security-relevant statement in the new text (the API key never reaching argv) is verified above rather than asserted, which matters because a wrong claim there would invite users to pass secrets on a command line.

### Validation

PR CI is green, including the Python jobs that actually exercise this package: lint and type-check, and the test matrix on CPython 3.9, 3.11, and 3.13.

No follow-up commits were needed.

Issue #1324, opened 2026-08-22
fix(internlm3): stop doubling RoPE positions under dynamic NTK scaling and add a shared DynamicNTK RoPE helper


inureyes:
## Summary

`internlm3` passes `scale = 2.0` to `mlxcel_core::fast_rope` whenever `rope_scaling` is absent or its `rope_type` is `"dynamic"`. The checkpoint the family is validated on (`mlx-community/internlm3-8b-instruct-4bit`) ships `rope_scaling: {"factor": 6.0, "rope_type": "dynamic"}`, so every query and key is rotated at twice its true position on every request. The correct schedule for dynamic NTK is: positions unscaled (`scale = 1.0`), and only the rotary base grows once the sequence passes `max_position_embeddings`. This change fixes the scale, moves the dynamic NTK base computation into a shared `src/models/dynamic_ntk_rope.rs` helper that `internlm2` will reuse, and re-pins the InternLM3 greedy parity reference, which was captured against the same wrong schedule.

## Current behavior

`src/models/internlm3.rs:108-119`:

```rust
pub fn rope_scale(&self) -> f32 {
    self.rope_scaling
        .as_ref()
        .and_then(|s| {
            if s.rope_type == "linear" {
                Some(1.0 / s.factor)
            } else {
                None
            }
        })
        .unwrap_or(2.0)
}
```

`rope_scaling == None` and `rope_type == "dynamic"` both fall through to `unwrap_or(2.0)`. The value is stored as `Attention::rope_scale` (`src/models/internlm3.rs:297`) and handed to both rotary calls at `src/models/internlm3.rs:209-224`:

```rust
let q = mlxcel_core::fast_rope(&q, self.head_dim, self.rope_traditional, rope_base, self.rope_scale, offset);
let k = mlxcel_core::fast_rope(&k, self.head_dim, self.rope_traditional, rope_base, self.rope_scale, offset);
```

`fast_rope` multiplies every position by `scale` before computing the rotation angle, so with `scale = 2.0` the token at absolute position `p` is rotated as if it sat at `2p`. For `rope_theta = 5e7`, `head_dim = 128`, position 7, frequency pair `j = 64`: the correct angle is `7 * 5e7^(-64/128) = 9.899e-4`; the runtime applies `1.980e-3`. Relative angles between any two tokens are doubled as well, so attention patterns differ from what the weights were trained on at every context length, not only past 32768.

[...]

PR #1736, opened 2026-09-09
fix(internlm2): add family rope_scaling tests and NTK rescale log

### LINKED ISSUE #1320: fix(internlm2): honor rope_scaling (dynamic NTK, linear) and rope_traditional
## Summary

`internlm2` ignores `rope_scaling` entirely and ignores `rope_traditional` after parsing it. The public checkpoint `mlx-community/internlm2_5-7b-chat-4bit` ships `rope_scaling: {"type": "dynamic", "factor": 2.0}` with `max_position_embeddings: 32768`, so the rotary base must grow once a request passes 32768 tokens; today it stays at `rope_theta` and long contexts run with the wrong rotation frequencies. A `"linear"` entry would be ignored too. This change wires the family to the shared `DynamicNtkRope` helper (`src/models/dynamic_ntk_rope.rs`) introduced by #1324.

Below 32768 tokens the current output is already correct (positions unscaled, base unchanged), which is why this is medium rather than high.

## Current behavior

`src/models/internlm2.rs:28-58`: `ModelArgs` has no `rope_scaling` field, so the key is dropped at deserialization. `rope_traditional` is parsed at line 50-51 but never read.

`src/models/internlm2.rs:157-158`:

```rust
q = mlxcel_core::fast_rope(&q, self.rope_dims, false, self.rope_base, 1.0, offset);
k = mlxcel_core::fast_rope(&k, self.rope_dims, false, self.rope_base, 1.0, offset);
```

`traditional` is hard-coded `false`, `scale` is hard-coded `1.0`, and `rope_base` is the constant `args.rope_theta` set at `src/models/internlm2.rs:207`.

Checkpoint config (`models/internlm2-7b-4bit/config.json`): `rope_theta: 1000000`, `max_position_embeddings: 32768`, `rope_scaling: {"type": "dynamic", "factor": 2.0}`, `num_attention_heads: 32`, `hidden_size: 4096` (so `head_dim = 128`). Note the key is spelled `type`, not `rope_type`.

## Expected behavior

Same schedule as `internlm3` (see #1324), with `dims = head_dim = 128`, `base = rope_theta`, `max_pos = max_position_embeddings`:

- no `rope_scaling`, or `type`/`rope_type` in `{"default", absent}`: `fast_rope(x, dims, rope_traditional, base, 1.0, offset)`.
- `"linear"`, `factor = f`: `fast_rope(x, dims, rope_traditional, base, 1.0 / f, offset)`.
- `"dynamic"`, `factor = f`: `seq_len = max(L + offset, max_pos)`; `base_eff = base * ((f * seq_len / max_pos) - (f - 1)) ^ (dims / (dims - 2))`; `fast_rope(x, dims, rope_traditional, base_eff, 1.0, offset)`.
- any other type, or `linear`/`dynamic` without `factor`: load error.

Worked values for the checkpoint (`base = 1e6`, `f = 2.0`, `dims = 128`, `max_pos = 32768`):

| `L + offset` | `base_eff` |
|---|---|
| 56 | 1 000 000 (unchanged) |
| 32768 | 1 000 000 (unchanged) |
| 40000 | 1 449 795.74 |
| 65536 | 3 052 773.67 |

`rope_traditional` selects the rotation layout: `false` rotates the two halves of the head (`x[..., :d/2]`, `x[..., d/2:]`), `true` rotates adjacent pairs. The checkpoint's own `modeling_internlm2.py` uses the half-split layout, so `false` (the serde default) remains the value for every public checkpoint; the flag just has to reach the call.

## Implementation plan

1. `src/models/internlm2.rs` `ModelArgs`: add

   ```rust
   #[serde(default)]
   pub rope_scaling: Option<crate::models::config::RopeScaling>,
   ```

   `config::RopeScaling` (`src/models/config.rs:309`) already accepts both `type` and `rope_type`.

2. Replace the `rope_dims: i32, rope_base: f32` fields of `Attention` (`src/models/internlm2.rs:103-104`) with `rope: crate::models::dynamic_ntk_rope::DynamicNtkRope`. In `Attention::from_weights` build it with `DynamicNtkRope::from_scaling(head_dim, args.rope_theta, args.rope_traditional, args.max_position_embeddings, args.rope_scaling.as_ref())?`.

3. In `Attention::forward` replace the two hard-coded `fast_rope` calls with `self.rope.apply(&q, offset, l + offset)` and `self.rope.apply(&k, offset, l + offset)` (`offset = cache.offset`, `l = shape[1]`).

4. Extend the `// Used by:` comment on `DynamicNtkRope` to list `InternLM2`.

## Validation

(a) Unit tests, `src/models/internlm2_tests.rs` (new, attached with `#[cfg(test)] #[path = "internlm2_tests.rs"] mod tests;`):

- `internlm2_config_parses_type_spelled_rope_scaling`: deserialize the checkpoint config excerpt above into `ModelArgs`; `rope_scaling.scaling_type == Some("dynamic")`, `factor == Some(2.0)`.
- `internlm2_rope_is_unit_scale_dynamic_for_checkpoint_config`: the built `DynamicNtkRope` has `position_scale() == 1.0`, `base_for(56) == 1_000_000.0`, and `base_for(65536)` within `1e-3` relative of `3_052_773.67`.
- `internlm2_rope_linear_uses_inverse_factor`: `rope_scaling: {"type": "linear", "factor": 4.0}` gives `position_scale() == 0.25`.
- `internlm2_rope_traditional_reaches_the_helper`: `rope_traditional: true` yields `rope.traditional == true`.
- `internlm2_rejects_unknown_rope_type`: `{"type": "yarn", "factor": 2.0}` makes `Attention::from_weights` (or a thin `ModelArgs::rope()` accessor) return `Err`.

(b) Real checkpoint: `mlx-community/internlm2_5-7b-chat-4bit` (local `models/internlm2-7b-4bit`).

```
./target/release/mlxcel generate -m models/internlm2-7b-4bit -p "Write a short paragraph explaining why the sky is blue." -n 64
```

Acceptance: token-exact with the output produced before this change on a prompt of at least 64 tokens with `-n 64` (below 32768 tokens the schedule is unchanged, so any drift is a regression). Then a long-context smoke: a prompt of about 33000 tokens (for example a repeated paragraph) with `-n 32` must produce finite, fluent text; log the `base_eff` once at the first forward past `max_position_embeddings` (debug level) and confirm it matches the table within `1e-3` relative.

## Acceptance criteria

- [ ] `rope_scaling` with `type` or `rope_type` is parsed; `dynamic` rescales the base past 32768 tokens exactly as in the table; `linear` scales positions by `1 / factor`
- [ ] `rope_traditional` reaches `fast_rope`
- [ ] Output on prompts shorter than `max_position_embeddings` is token-identical to the current binary

## Dependencies

Blocked by #1324 (provides `src/models/dynamic_ntk_rope.rs`).

inureyes:
## Summary

PR #1389 already wired `internlm2::Attention` to the shared `DynamicNtkRope` schedule (issue #1324), so the `rope_scaling` and `rope_traditional` plumbing itself was already correct on main. This PR closes the rest of #1320: family-level tests proving that plumbing reaches the helper, a debug log of the rescaled base, and real-checkpoint validation.

## What changed

- `src/models/internlm2_tests.rs`: tests that `ModelArgs::rope()`, not just the helper, resolves a `linear` block to an inverse-factor scale, carries `rope_traditional` through in both states, reproduces the issue's dynamic table (`base_for(40000)`, `base_for(65536)`), accepts the `rope_type` spelling, and rejects a factor-less `dynamic` block with an error naming the family.
- `src/models/dynamic_ntk_rope.rs`: a `traditional()` accessor, and a debug-level log of `base_eff` the first time a dynamic schedule crosses `max_position_embeddings`. The dedup is keyed on the schedule's own parameters rather than a process-wide flag, following `rope_utils::report_unusable_rope_scaling_once`, so a second checkpoint in the same process (`--models-dir` routing, pipeline stages, TP ranks) still gets its line. `apply` asks `tracing::enabled!` before it takes the dedup lock, so a run with no debug subscriber never touches that shared lock; the rotary arithmetic is unchanged either way.
- Comments and `docs/supported-models.md` named a checkpoint directory that does not exist (`models/internlm2-7b-4bit`); corrected to `models/internlm2_5-7b-chat-4bit`, and the entry now records the long-context evidence rather than crediting the short run with it, at the figures the closed form actually gives.

## Validation

`--lib models::internlm2` (8 passed) and `--lib models::dynamic_ntk_rope` (17 passed) under `--profile test-fast --features metal,accelerate`; `cargo clippy --lib --tests` and `cargo fmt --all --check` clean, which is the scope `ci.yml` runs. The `--workspace` gate in issue #1320's acceptance criteria runs in `nightly-verify.yml`, not on this PR.

Real checkpoint `models/internlm2_5-7b-chat-4bit`, greedy. Against a binary built from this branch's parent commit, an 89-token prompt at `-n 64` and a 34021-token prompt at `-n 32` both produce byte-identical generated text (MD5 `5a6d6133` and `69a12bff`). That pair is a control rather than a parity result: the parent already carries #1389's wiring, so the two arms differ only by the debug diagnostic this PR adds, and byte-identity is what it should show.

The long-context run was repeated against the final binary, after the `tracing::enabled!` gate was added, so the diagnostic is confirmed to still reach a `RUST_LOG=debug` run rather than only having done so before that commit. A 34021-token prompt emits exactly one line, `seq_len=34021 base_eff=1077737.0`, and the generation answers the question the prompt ends with in fluent on-topic text (Rayleigh scattering). `1077737.0` is the issue's closed form at that length: f64 gives 1077736.99 there, so the f32 the code computes agrees to every digit the line prints, and a 34022-token prompt gives `base_eff=1077799.125`, which the same f32 evaluation reproduces bit for bit. The base moves about 62 per token at that length, which is why the entry in `docs/supported-models.md` now quotes the value with its length; its previous parenthetical read the closed form at 34020 as an f64-versus-f32 gap at 34021. The same form gives the issue's table at 40000 (1449795.74) and 65536 (3052773.67).

Closes #1320


inureyes:
## PR Finalization Check

Reviewed test coverage, documentation, and lint/format on `fix/issue-1320-internlm2-rope-scaling` at `1aba962d`. No gaps found, no changes made.

### Tests
`--lib models::internlm2` (8 passed) and `--lib models::dynamic_ntk_rope` (17 passed) under `--profile test-fast --features metal,accelerate` both pass locally, matching the counts in the PR description. Coverage already spans the linear/dynamic/absent `rope_scaling` cases, both `type` and `rope_type` spellings, `rope_traditional` true and the serde-default false, the factor-less `dynamic` load error naming the family, and a dedicated dedup test for the rescale log (first-crossing fires, repeat and per-layer-rebuilt schedules stay silent, a different base or factor still logs, and both the boundary and non-dynamic modes never log). No real gap to add tests for.

### Documentation
`docs/supported-models.md` already carries the corrected checkpoint path and states plainly what the long-context run does and does not establish. Checked `docs/llama-server-compat.md` and `docs/model-catalog.tsv` for stale references to the old `internlm2-7b-4bit` name; the TSV already tracks it as a documented alias of `internlm2_5-7b-chat-4bit`, and the benchmark result logs under `docs/benchmark_results/` are historical run records that correctly use whatever name was live at the time, so neither needed a change. No `docs/ko` tree exists in this repository (the mkdocs manual under `docs/en` and `docs/ko` is a separate documentation tree per the `Makefile` comments), so there is no Korean file to update.

### Lint / format
`cargo fmt --all --check` and `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` both pass clean. `cargo check --lib --tests` also clean.

All PR CI checks (`cargo-clippy`, `cargo-fmt`, `cargo-deny`, `OpenXLA feature compile`, `crate versions`, `cross-repo refs`, `kernel dtype keys`, `llama-compat manifest`) are green. Ready for merge from this angle.

Issue #1734, opened 2026-09-09
feat(distributed): Kimi K3 pipeline stage executor and full-model multi-node validation


inureyes:
Part of #1331. Follow-up to #1334.

## Problem / Background

#1334 adds the Kimi K3 text backbone (`model_type: "kimi_k3"`, `moonshotai/Kimi-K3`, 2.8T parameters, 93 layers: 69 Kimi Delta Attention layers and 24 gated NoPE-MLA layers, latent SiTU MoE with 896 experts in the `mxfp4-pack-quantized` layout, Attention Residuals every 12 layers). The checkpoint is about 1.56 TB on disk and about 1.4 TB resident at 4-bit, so the full model needs at least three 512 GB nodes through the pipeline in `docs/distributed.md`. No such nodes are reachable at the time of the #1334 PR, so that PR validates only a layer-truncated local copy (`num_hidden_layers` reduced to what fits on a 128 GB machine: load, shape, causality and finite-logits gate). Its acceptance criterion "fluent output on the full model through the distributed pipeline" is discharged by this issue.

The layer-truncated gate is not a substitute for the full run. Until this issue lands, Kimi K3 is single-process-only in mlxcel and cannot actually be served at full size on any hardware.

## Current Behavior

- The #1334 PR adds only a `build_per_layer_bytes` arm in `src/distributed/pipeline/partition_profile_heuristics.rs:32`, so the partitioner can size Kimi K3 stages. It adds no `StageFamily` variant (`src/distributed/pipeline/stage_executor/mod.rs:239`), no `resolve_stage_family` arm (`mod.rs:345`), and no executor under `src/distributed/pipeline/stage_executor/`, so `LoadedStageExecutor::load` bails on a `kimi_k3` model and `--pp-size`, `--pp-layers` and `--distributed-config` cannot run it.
- Stages exchange exactly one hidden-state tensor per micro-batch: `StageExecutionInput::HiddenStates(&MlxArray)` / `StageExecutionOutput::HiddenStates` (`mod.rs:77-88`), serialized by `serialize_mlx_array` (`src/distributed/pipeline/wire_tensor.rs:27`) into `ActivationMessage.tensor_data` (`src/distributed/pipeline/activation_transfer.rs:46`). Attention Residuals need more than that: layer `l` mixes its input against every frozen block stored by layers `0, 12, 24, ...` below it, so a stage that starts at layer 24 needs the two blocks stored by layers 0 and 12 on an earlier node.
[...]

Issue #1342, opened 2026-08-22
feat(vision): Kimi K3 MoonViT3D tower, temporal-pool patch merger and navit image preprocessing


inureyes:
## Summary

Kimi K3's vision path is MoonViT3D (`vision_config.model_type: "moonvit3d"`): a 14x14 patch embedding, a learnable 64x64 position grid bilinearly interpolated to each image's patch grid plus a sincos time embedding for multi-frame inputs, 27 pre-norm blocks with RMSNorm (eps `2^-7`), fused `wqkv`, 2D rotary embedding and a tanh-GELU MLP, a `sd2_tpool` merger (2x2 spatial groups averaged over time) and a `patchmergerv2` projector (`Linear(4096, 4096) -> GELU -> Linear(4096, 7168) -> RMSNorm`). Images are resized with the navit rule (patch budget 65536, at most 512 patches per side, padded to multiples of 28) and each image contributes `gh * gw / 4` `<|media_pad|>` tokens inside a `<|media_begin|>image {w}x{h}<|media_content|> ... <|media_end|>` prompt. This issue adds the tower, merger, projector, preprocessing and prompt expansion on top of the text backbone and the XTML renderer.

## Current behavior

`src/vision/encoders/kimi_vl.rs:170` (`KimiVLVisionConfig`, MoonViT for Kimi-VL / K2.5) has the interpolated 2D position embedding, 2D RoPE and a spatial merge, but uses LayerNorm-style eps, separate merge semantics (`spatial_merge_size`, no temporal mean), no sincos time embedding, a different projector (`pre_norm` then MLP) and different key names (`vision_tower.encoder.blocks.N.wqkv` / `wo` / `norm0` / `norm1` / `mlp.fc0` / `mlp.fc1` vs Kimi-VL's). `src/vision/kimi_vl.rs` binds MoonViT to the DeepSeek-V3 backbone, not to `kimi_k3`.

## Expected behavior

[...]

Issue #1338, opened 2026-08-22
feat(tokenizer): Kimi K3 tiktoken vocabulary with the K3 pre-tokenizer and a native XTML chat renderer


inureyes:
## Summary

Kimi K3 ships no `tokenizer.json` and no Jinja chat template. Its tokenizer is a tiktoken BPE (`tiktoken.model`, 163584 byte-level ranks) plus 256 control tokens and a K3-specific pre-tokenization regex; its chat format is XTML, a tag language built from the control tokens `<|open|>`, `<|close|>`, `<|sep|>`, `<|end_of_msg|>` that the provider renders in code. This issue extends mlxcel's native tiktoken tokenizer with a per-family pattern and control-token table, adds a native XTML renderer for OpenAI-style messages (system, user, assistant with reasoning and tool calls, tool results, tools, tool_choice, response_format, thinking effort), and wires the stop token and reasoning/tool-call stream markers.

## Current behavior

- `src/tokenizer/tiktoken.rs:27` hard-codes `HUNYUAN_PAT` and `src/tokenizer/tiktoken.rs:110` (`build_special_token_list`) hard-codes the HunYuan special token names; `find_tiktoken_file` (`src/tokenizer/mod.rs:565`) already finds `tiktoken.model`, so a K3 directory would load with the wrong pattern and wrong control ids.
- `src/tokenizer/tiktoken.rs:157` (`encode`) always matches special-token strings inside user text; XTML needs text segments encoded without special matching and control segments emitted as ids.
- `src/server/chat_template.rs:1464` (`builtin_chat_template`) provides built-in Jinja templates for Jina VLM and Florence-2 only; there is no code-rendered chat format and the generic `User:/Assistant:` fallback would be used.

### Tokenizer

[...]

PR #1741, opened 2026-09-10
feat(models): add the Kimi K3 text backbone

### LINKED ISSUE #1334: feat(models): add the Kimi K3 text backbone (fused-QKV KDA with full-rank gate, gated q-LoRA NoPE MLA, latent SiTU MoE, Attention Residuals, mxfp4 experts)
## Summary

Add the Kimi K3 text decoder (`text_config.model_type: "kimi_linear"` inside `model_type: "kimi_k3"`). Relative to `src/models/kimi_linear.rs` it needs: a fused `qkv_proj` and `qkv_conv` in the KDA layers, a full-rank KDA output gate `g_proj`, a gate lower bound in the gated delta recurrence, q-LoRA (`q_a_proj`, `q_a_layernorm`, `q_b_proj`) and a sigmoid output gate in the MLA layers, the SiTU-GLU activation everywhere, a latent MoE (down-project to 3584, experts in the latent, RMSNorm, up-project back) with 896 experts, Attention Residuals across 12-layer blocks with a final output mix, and the `mxfp4-pack-quantized` expert layout. It keeps the existing NoPE-MLA absorption and grouped sigmoid routing. Vision and the tokenizer are sibling sub-issues.

## Current behavior

- `src/models/detection.rs:306` maps `kimi_linear`, `kimi_vl`, `kimi_k25`; there is no `kimi_k3` arm.
- `src/models/kimi_linear.rs:623` (`KimiDeltaAttention::forward`) projects `q_proj`/`k_proj`/`v_proj` separately with three `ShortConv1d`s, uses the low-rank gate `g_b_proj(g_a_proj(x))` only, and calls `gated_delta_update` (`src/models/gated_delta.rs:692`) whose gate is `g = exp(-exp(A_log) * softplus(a + dt_bias))` with no lower bound.
- `src/models/kimi_linear.rs:458` (`KimiMLAAttention::forward`) has a plain `q_proj` (no `q_a_proj`/`q_b_proj`) and no output gate.
- `src/models/kimi_linear.rs:837` (`KimiSparseMoE::forward`) runs experts on the full hidden size with `SwitchGLU` (silu-SwiGLU only, `src/models/switch_layers.rs:775`), no latent projections, no SiTU.
- No module implements Attention Residuals.

### Config

Top level: `model_type: "kimi_k3"`, `media_placeholder_token_id: 163605`, `bos_token_id: 163584`, `eos_token_id: 163586`, `pad_token_id: 163839`, `vocab_size: 163840`, `tie_word_embeddings: false`, `text_config`, `vision_config`.

`text_config` (published values; types):

```
hidden_size 7168   num_hidden_layers 93   num_attention_heads 96   num_key_value_heads 96   intermediate_size 33792 (dense layer 0)
vocab_size 163840  rms_norm_eps 1e-5      max_position_embeddings 1048576   tie_word_embeddings false
hidden_act "situ" (reject anything else)   activation_situ_beta 4.0   activation_situ_linear_beta 25.0 (null means no input tanh)
attn_res_block_size 12 (null disables AttnRes)
linear_attn_config: {kda_layers: 1-based list (69 entries), full_attn_layers: 1-based list (24 entries: 4, 8, ..., 92, 93),
                     num_heads 96, head_dim 128, short_conv_kernel_size 4, gate_lower_bound -5.0, use_full_rank_gate true}
q_lora_rank 1536   kv_lora_rank 512   qk_nope_head_dim 128   qk_rope_head_dim 64   v_head_dim 128
mla_use_nope true (reject false)   mla_use_output_gate true
num_experts 896   num_experts_per_token 16   num_shared_experts 2   moe_intermediate_size 3072
routed_expert_hidden_size 3584 (null means experts run on hidden_size)   latent_moe_use_norm true
moe_router_activation_func "sigmoid" (reject others)   moe_renormalize true   routed_scaling_factor 1.0
first_k_dense_replace 1   moe_layer_freq 1   use_grouped_topk true   num_expert_group 1   topk_group 1
num_nextn_predict_layers 0
quantization_config: compressed-tensors, format "mxfp4-pack-quantized", group_size 32, num_bits 4, targets Linear,
                     ignore: self_attn.*, shared_experts.*, mlp.(gate|up|down)_proj, lm_head, vision_tower, mm_projector
```

Layer `i` (0-based) is KDA iff `(i + 1)` is in `kda_layers`, otherwise MLA. Layer `i` is MoE iff `num_experts > 0`, `i >= first_k_dense_replace` and `i % moe_layer_freq == 0` (so layer 0 is dense, 1..92 are MoE).

### SiTU activation

```
situ(x, gate) = (beta * tanh(gate / beta) * sigmoid(gate)) * (linear_beta * tanh(x / linear_beta))     # f32, cast back
                with x = up_proj(h), gate = gate_proj(h); when linear_beta is null the second factor is just x
MLP(h) = down_proj(situ(up_proj(h), gate_proj(h)))
```

### KDA layer (`P = 96 * 128 = 12288`, conv kernel `K = 4`)

```
qkv = qkv_proj(x)                                   # [B, T, 3P]  (rows: q, k, v)
qkv, conv_state = short_conv(qkv, conv_state)       # depthwise causal conv over 3P channels, kernel 4, then silu; state = last 3 inputs [B, 3, 3P]
q, k, v = split(qkv, [P, 2P]) each -> [B, T, 96, 128]
q = scale^2 * rms_norm(q, eps = 1e-6 / 128);  k = scale * rms_norm(k, same eps);  scale = 128^-0.5   # no learned weight
a = f_b_proj(f_a_proj(x)) -> [B, T, 96, 128];  b = b_proj(x) -> [B, T, 96]
beta = sigmoid(b);   g = exp( lower_bound * sigmoid( exp(A_log[h]) * (a + dt_bias[h, :]) ) )          # lower_bound = -5.0; g in (e^-5, 1)
out, ssm_state = gated_delta_recurrence(q, k, v, g, beta, ssm_state)                                    # existing gated_delta_ops, f32 state [B, 96, 128, 128]
gate = g_proj(x) -> [B, T, 96, 128]                  # full-rank gate (use_full_rank_gate); else g_b_proj(g_a_proj(x))
out = (RMSNorm_128(out; o_norm) * sigmoid(gate)).reshape(B, T, P);   y = o_proj(out)
```

`A_log` is stored as `[128]` with `num_heads = 96`: slice to the first 96 entries. `dt_bias` is `[12288]` = `[96, 128]`.

### MLA layer (NoPE, absorbed, with q-LoRA and output gate)

```
q = q_b_proj(RMSNorm_1536(q_a_proj(x); q_a_layernorm, eps 1e-6)) -> [B, 96, L, 192];  q_nope = q[..., :128];  q_pe = q[..., 128:]
c = kv_a_proj_with_mqa(x) -> [B, L, 576];  latent = RMSNorm_512(c[..., :512]; kv_a_layernorm, eps 1e-6) -> [B, 1, L, 512];  k_pe = c[..., 512:] -> [B, 1, L, 64]
cache stores (latent, k_pe); no rotation is applied to q_pe or k_pe (NoPE)
pe_scores = (q_pe * scale) @ k_pe^T, masked causal;  scale = 192^-0.5
decode (L == 1): q_lat = embed_q(q_nope) [B, 96, 1, 512]; scores = q_lat*scale @ latent^T + pe_scores; o = softmax(scores) @ latent; o = unembed_out(o) [B, 96, 1, 128]
prefill:         k = embed_q^T(latent) [B, 96, L, 128]; v = unembed_out(latent) [B, 96, L, 128]; o = softmax(q_nope*scale @ k^T + pe_scores) @ v
o -> [B, L, 12288];  o = o * sigmoid(g_proj(x));  y = o_proj(o)
```

`embed_q` `[96, 512, 128]` and `unembed_out` `[96, 128, 512]` are derived from `kv_b_proj.weight [24576, 512]` exactly as `src/models/kimi_linear.rs:1171` already does (`reshape [96, 256, 512]`, `wk = v[:, :128, :]^T`, `wv = v[:, 128:, :]`).

### Latent MoE

```
logits = gate(h)  [N, 896]  (bf16 matmul);  scores = sigmoid(logits.f32);  sel = scores + e_score_correction_bias
idx = top-16 of sel (num_expert_group == 1: plain argpartition);  w = take(scores, idx);  w = w / (sum(w) + 1e-20) (moe_renormalize);  w *= routed_scaling_factor (1.0)
z = routed_expert_down_proj(h)             # [N, 3584]
e = sum_k w_k * Expert_k(z)                # Expert: down(situ(up(z), gate(z))) with up/gate [3072, 3584], down [3584, 3072]; SwitchGLU with the SiTU activation
e = RMSNorm_3584(e; routed_expert_norm, eps)    # latent_moe_use_norm
e = routed_expert_up_proj(e)               # [N, 7168]
out = e + shared_experts(h)                # KimiK3MLP on the full hidden: gate/up [6144, 7168] (= 2 * 3072 stacked), down [7168, 6144], SiTU
```

### Attention Residuals (`B_res = attn_res_block_size = 12`)

Per layer `l` with input `x` (`[N, D]`), a list `blocks` of frozen residual states `raw_k` with precomputed `inv_rms_k = rsqrt(mean(raw_k^2, -1) + eps)`, and `w_eff = res_norm.weight.f32 * res_proj.weight.reshape(D)` (precomputed per layer for `self_attention_res_*`, `mlp_res_*`, and the model-level `output_attn_res_*`):

```
mix(blocks, partial, w_eff):
    if blocks is empty: return partial
    pf = partial.f32;  logit_p = (pf @ w_eff) * rsqrt(mean(pf^2, -1) + eps)             # [N]
    logit_k = (raw_k.f32 @ w_eff) * inv_rms_k                                           # [N] per stored block
    p = softmax([logit_0, ..., logit_{K-1}, logit_p], over the K+1 axis)                 # f32
    return (sum_k p_k * raw_k + p_p * partial).astype(partial.dtype)

layer forward (AttnRes enabled):
    partial = x
    h = mix(blocks, partial, w_attn)
    if l % B_res == 0: blocks.append(partial); partial = None                            # layer 0 stores the embeddings
    y = attention(input_layernorm(h))
    partial = y if partial is None else partial + y
    h = mix(blocks, partial, w_mlp)
    partial = partial + mlp(post_attention_layernorm(h))
    return partial
model: after the last layer h = mix(blocks, h, w_output) using output_attn_res_{proj,norm}; logits = lm_head(norm(h))
```

Without AttnRes (`attn_res_block_size` null) the layer is the ordinary `x + attn(norm(x))`, `+ mlp(norm(.))`.

Caches: KDA layers `(conv_state [B, 3, 36864] in the activation dtype, ssm_state [B, 96, 128, 128] f32)` (`KimiDeltaCache` with one fused conv slot); MLA layers `KVCache` holding `(latent, k_pe)`.

### Weight keys and sanitize

All text keys are prefixed `language_model.` (strip it). Drop `model.mtp*` and any `model.layers.N.*` with `N >= num_hidden_layers` (layer-truncated local copies).

```
model.embed_tokens.weight [163840, 7168]   lm_head.weight [163840, 7168]   model.norm.weight [7168]
model.output_attn_res_proj.weight [1, 7168]   model.output_attn_res_norm.weight [7168]
per layer: input_layernorm.weight, post_attention_layernorm.weight, self_attention_res_proj.weight [1, 7168], self_attention_res_norm.weight, mlp_res_proj.weight [1, 7168], mlp_res_norm.weight
           (also accept the legacy spellings self_attention_res.proj_weight / .norm_weight, mlp_res.*, output_attn_res.* and rename to the `_proj.weight` / `_norm.weight` forms)
KDA:  self_attn.{q,k,v}_proj.weight [12288, 7168]   self_attn.{q,k,v}_conv1d.weight F32 [12288, 1, 4]   self_attn.f_a_proj.weight [128, 7168]   self_attn.f_b_proj.weight [12288, 128]
      self_attn.b_proj.weight [96, 7168]   self_attn.g_proj.weight [12288, 7168]   self_attn.A_log F32 [128]   self_attn.dt_bias F32 [12288]   self_attn.o_norm.weight F32 [128]   self_attn.o_proj.weight [7168, 12288]
      sanitize: conv weights [C, 1, 4] -> moveaxis to [C, 4, 1]; concat q,k,v proj weights along axis 0 -> self_attn.qkv_proj.weight [36864, 7168]; concat the three conv weights -> self_attn.qkv_conv.conv.weight [36864, 4, 1]; A_log -> [:96]; keep A_log, dt_bias, o_norm f32
MLA:  self_attn.q_a_proj.weight [1536, 7168]   self_attn.q_a_layernorm.weight [1536]   self_attn.q_b_proj.weight [18432, 1536]   self_attn.kv_a_proj_with_mqa.weight [576, 7168]   self_attn.kv_a_layernorm.weight [512]
      self_attn.kv_b_proj.weight [24576, 512] -> embed_q / unembed_out   self_attn.g_proj.weight [12288, 7168]   self_attn.o_proj.weight [7168, 12288]
dense layer 0: mlp.{gate_proj,up_proj}.weight [33792, 7168]   mlp.down_proj.weight [7168, 33792]
MoE:  block_sparse_moe.gate.weight [896, 7168]   block_sparse_moe.gate.e_score_correction_bias F32 [896]
      block_sparse_moe.routed_expert_down_proj.weight [3584, 7168]   block_sparse_moe.routed_expert_norm.weight [3584]   block_sparse_moe.routed_expert_up_proj.weight [7168, 3584]
      block_sparse_moe.shared_experts.{gate_proj,up_proj}.weight [6144, 7168]   block_sparse_moe.shared_experts.down_proj.weight [7168, 6144]
      block_sparse_moe.experts.{e}.w1.weight_packed U8 [3072, 1792]   .w1.weight_scale U8 [3072, 112]     (w1 = gate, w3 = up: same shapes; w2 = down: packed [3584, 1536], scale [3584, 96])
      sanitize: rename block_sparse_moe.* -> mlp.*; e_score_correction_bias -> mlp.e_score_correction_bias (f32); stack experts e = 0..895:
                mlp.switch_mlp.gate_proj.weight = stack(w1.weight_packed).view(uint32) [896, 3072, 448]   .scales = stack(w1.weight_scale) uint8 [896, 3072, 112]   (E8M0 bytes, group 32)
                mlp.switch_mlp.up_proj.*   from w3;   mlp.switch_mlp.down_proj.weight uint32 [896, 3584, 192], .scales [896, 3584, 96]
                quantization for these = {group_size: 32, bits: 4, mode: "mxfp4"} (MLX's mxfp4 layout: 8 E2M1 codes per uint32, low nibble first, uint8 E8M0 block scales; no biases, no global scale)
```

`cast_predicate`: keep `e_score_correction_bias`, `A_log`, `dt_bias` in f32; everything else follows the checkpoint dtype (bf16). The fused decode MoE kernel is affine-only, so mxfp4 experts use the `SwitchGLU::forward` gather path; set `MLXCEL_FUSED_MOE=0` when reference-diffing anyway.

## Implementation plan

1. `src/models/kimi_k3.rs` (new; do not widen `kimi_linear.rs`, which stays the reference for Kimi Linear 48B): `KimiK3TextConfig` (serde; validation of `hidden_act`, `moe_router_activation_func`, `mla_use_nope`), `LinearAttnConfig` with `gate_lower_bound`, `use_full_rank_gate`.
2. `src/models/gated_delta.rs`: add `compute_g_lower_bounded(a_log, a, dt_bias, lower_bound)` and a `gated_delta_update_with_lower_bound(...)` entry (or an `Option<f32>` parameter on `gated_delta_update`; update the `// Used by:` comments for Qwen3Next, Qwen3.5, KimiLinear), keeping the Metal kernel path when `supports_metal_gated_delta_kernel` holds.
3. `src/models/switch_layers.rs`: add `SwitchGluActivation { SwiGlu, SiTU { beta, linear_beta } }` on `SwitchGLU` (default SwiGLU; update `// Used by:`), applied in both the sorted and unsorted branches; the fused kernel path is taken only for SwiGLU.
4. KDA: `KimiK3DeltaAttention` with `qkv_proj`, `qkv_conv` (reuse `ShortConv1d` from `kimi_linear.rs`, made `pub(crate)`, over `3P` channels; decode step via `src/models/conv_decode.rs`), `f_a/f_b/b_proj`, `g_proj` or `g_a/g_b`, `A_log`, `dt_bias`, `o_norm`, `o_proj`; forward as specified.
5. MLA: `KimiK3MLAAttention` = `KimiMLAAttention` plus `q_a_proj`/`q_a_layernorm`/`q_b_proj` (when `q_lora_rank` is set) and `g_proj` output gate.
6. `KimiK3MLP` (SiTU) and `KimiK3SparseMoE` (latent down/up, `routed_expert_norm`, `SwitchGLU` with SiTU, shared `KimiK3MLP` of width `2 * 3072`).
7. `ResidualBlocks { raw: Vec<MlxArray>, inv_rms: Vec<MlxArray> }` and `attn_res_mix(...)` as graph ops (f32 logits, bf16 outputs); `KimiK3DecoderLayer::forward(x, mask, cache, blocks) -> (x, blocks)`; model-level output mix.
8. Caches: `KimiK3LayerCache { Delta { conv: Option<MlxArray>, ssm: Option<MlxArray> }, Attn(KVCache) }` through the model-owned sequence state (`src/models/model_owned.rs`), snapshot/restore for `ModelStateSnapshot`.
9. Sanitize (`KimiK3Model::sanitize_weights`) implementing the rules above; idempotent.
10. Detection: `"kimi_k3" => ModelType::KimiK3` (text when no `vision_tower.*` weights are present, else the VLM type from #1342); `ModelType::KimiK3` in `src/models/mod.rs`; loader arm in `src/loading/mod.rs`. `eos_token_ids = [163586]`.
11. Distributed: register the family in `src/distributed/pipeline` partitioning (per-layer cost for KDA vs MLA layers); `docs/distributed.md` and `docs/supported-models.md` state the memory need (about 1.4 TB at 4-bit).

## Validation

(a) Unit tests, `src/models/kimi_k3_tests.rs`:

- `situ_matches_formula`: random `x`, `gate`; `beta = 4`, `linear_beta = 25`; compare to a scalar implementation, 1e-6 in f32; `linear_beta = None` reduces to `a * x`.
- `gated_delta_lower_bound_gate_range`: `g` stays within `(e^-5, 1)` for extreme `a`; `lower_bound = None` reproduces the existing gate.
- `kda_fused_qkv_equals_separate_projections`: build a layer from split q/k/v weights, sanitize into fused, assert the same output on a 16-token sequence, and `decode_matches_prefill` after 8 decode steps.
- `mla_q_lora_and_output_gate`: output equals `sigmoid(g_proj(x)) * o_proj(attn)` built from the unabsorbed path (f32, 1e-4).
- `latent_moe_forward_shapes_and_norm`: `routed_expert_hidden_size = 8`, 4 experts; the latent path matches a per-expert loop with the RMSNorm and up-projection.
- `attn_res_mix_matches_reference`: 3 stored blocks, random `w_eff`; compare to a scalar softmax implementation; `blocks` empty returns `partial` unchanged; with `attn_res_block_size = 2` and 4 layers the block list has 2 entries at the end.
- `sanitize_mxfp4_experts_and_fused_conv`: 2 experts with `weight_packed` `U8 [4, 16]` stack into `uint32 [2, 4, 4]`, `A_log [128]` slices to `[96]`-style (`[4]` in the tiny test), conv weights end `[C, 4, 1]`, legacy residual key spellings are renamed; idempotent.
- `prefill_is_causal` with a 96-token random prompt (KDA and MLA layers); `detection_kimi_k3` in `src/models/detection_tests.rs`.

(b) Real checkpoint: `moonshotai/Kimi-K3` (2.8T, 1.56 TB on disk, `modelsize:xlarge`). Two tiers:

- Layer-truncated local copy (single machine): download `config.json`, tokenizer files, and only the shards holding `language_model.model.layers.0..7.*`, `embed_tokens`, `norm`, `output_attn_res_*`, `lm_head` (about 260 GB), set `num_hidden_layers = 8` in a local `config.json` (the sanitize drops the missing layers). `./target/release/mlxcel generate -m models/Kimi-K3-8L -p "def fib(n):" -n 32` must produce finite logits and run decode without NaN. Output text is meaningless by construction; this tier is a load/shape/causality gate only.
- Full model (distributed pipeline, docs/distributed.md, at least three 512 GB nodes): `./target/release/mlxcel generate -m models/Kimi-K3 -p "Write a Python retry wrapper with exponential backoff." -n 64` with the XTML prompt rendered by #1338. Acceptance: fluent, correct Python; layer-by-layer mean-abs trace of the first 8 layers on the truncated copy within bf16 jitter of the same layers computed by a reference implementation of the formulas above in f32 (the reference is the unit-test scalar code run on the real weights, no Python).

## Acceptance criteria

- [ ] Kimi K3 text loads from the `compressed-tensors` mxfp4 checkpoint (full or layer-truncated) and runs prefill and decode with finite logits
- [ ] KDA (fused QKV, full-rank gate, lower-bounded gate), MLA (q-LoRA, output gate), latent SiTU MoE and Attention Residuals are implemented as specified and unit-tested against scalar references
- [ ] Fluent output on the full model through the distributed pipeline (or an explicit follow-up issue if no multi-node setup is reachable at merge time)
- [ ] docs/supported-models.md updated
- [ ] detection table in src/models/detection.rs updated with a test

## Out of scope

- Tokenizer and chat rendering (#1338), vision (#1342).
- Tensor-parallel sharding of KDA/MLA heads (pipeline parallel only).

## Dependencies
- Part of epic #1331

inureyes:
## Summary

Adds the Kimi K3 text decoder (`model_type: "kimi_k3"`, `text_config.model_type: "kimi_linear"`) as `src/models/kimi_k3.rs` with its sanitizer and tests beside it. It reuses Kimi Linear's `ShortConv1d`, `MultiLinear` and (now shared) `kv_b_proj` decomposition, and adds what that module lacks: fused-QKV KDA with a full-rank output gate and the `gate_lower_bound = -5.0` recurrence, NoPE MLA with q-LoRA and a sigmoid output gate, SiTU everywhere (including the routed experts through `SwitchGluActivation::SiTU`), the latent MoE (896 experts on a 3584-wide latent with `routed_expert_norm` and two shared experts), Attention Residuals over 12-layer blocks with the model-level output mix, and the compressed-tensors `mxfp4-pack-quantized` expert layout stacked into MLX's mxfp4 form at sanitize time. The tokenizer (#1338) and vision tower (#1342) are sibling sub-issues and are not part of this PR.

Two shared-function changes, both no-ops for existing callers: `gated_delta::compute_g_lower_bounded` / `gated_delta_update_with_lower_bound` (`None` is the existing softplus gate, same graph) and `SwitchGLU::with_activation` (default SwiGLU; the fused decode kernel declines any other activation). The pipeline partition profile prices KDA, MLA, dense and MoE layers separately for this family, and the tensor-parallel plan generator replicates it beside `kimi_linear` rather than letting it fall through to a generic q/k/v/o shard plan it has no projections for.

Kimi K3 opts out of chunked prefill. `forward_for_sequence` reads any multi-token forward carrying no sequence id as the start of a prompt and resets the model-owned KDA and MLA state, which is what makes the cache-less `forward` entry point safe across two unrelated prompts; a chunked prefill sends exactly that shape once per chunk, so every chunk after the first would discard what its predecessors built and the model would answer from the last `MLXCEL_PREFILL_CHUNK` tokens with nothing raised. `kimi_linear.rs` and five other families carry the same shape and are filed as #1749.

### Unit tests and lints

- `cargo test --profile test-fast --features metal,accelerate --lib models::kimi_k3` (25 passed): SiTU vs scalar, lower-bounded gate range, fused QKV vs separate projections, decode vs prefill, absorbed q-LoRA MLA vs the unabsorbed path, latent MoE vs a per-expert loop, AttnRes mix vs a scalar softmax, block accounting, sanitize incl. mxfp4 repack vs scalar dequant, causality, config validation, eos ids, sequence-state isolation, and negative tests for every load-time refusal added by the review commits (router bounds, six config-versus-checkpoint mismatches, mismatched expert planes, mismatched fused q/k/v planes, a `kv_b_proj` that disagrees with the config, the three Attention Residual widths, the five gate and output widths, and the chunked-prefill opt-out).
- `--lib models::gated_delta` (6), `--lib models::switch_layers` (24), `--lib models::detection` (61), `--lib distributed::tensor_parallel::plan_generator` (28, including a new `kimi_k3_is_replicated` pinning that the family gets a replicated plan rather than the generic q/k/v/o shard plan), `--lib distributed::pipeline::partition_profile` (7), `--lib cli::turbo_args` (19).
- `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` and `cargo fmt --all -- --check` clean.

### Real checkpoint

The full checkpoint is about 1.4 TB resident at 4 bits, so the gate runs on a layer-truncated local copy of `moonshotai/Kimi-K3` at `models/kimi-k3-8l-mxfp4`: `text_config.num_hidden_layers` lowered from 93 to 4 (`config.json.orig` kept beside it), `model.safetensors.index.json` rewritten to the 7 shards that hold layers 0 to 3 plus the embeddings, the final norm and `lm_head` (`model.safetensors.index.json.orig` kept beside it), and `metadata.total_size` corrected to the 54.46 GiB those shards actually contain so the preflight sizes the truncated copy rather than the full one.

Four layers is where the published schedule already contains one of every layer kind this family has: `[KimiK3] Config loaded: 4 layers (1 MLA, 3 KDA, 3 MoE), attn_res_block_size Some(12)`, with the dense MLP at layer 0 and the MLA layer at index 3 (1-based 4, the first entry of `full_attn_layers`). Each further layer is a repeat that costs about 17 GB of MoE experts, on a 137 GB host shared with seven concurrent build agents, so the memory went into a second variant instead of more layers.

| Gate | Command | Result |
|---|---|---|
| Pre-load budget | `mlxcel inspect -m models/kimi-k3-8l-mxfp4` | 54.46 GiB weights, 1.88 GiB KV at 8192 tokens, 67.69 GiB total, FITS with 53.91 GiB headroom |
| Load, prefill, decode | `mlxcel generate -m models/kimi-k3-8l-mxfp4 -p "def fib(n):" -n 32` | loaded in 5.3 s, resident 45.41 GB, peak 49.71 GB, 32 tokens in 0.60 s (53.66 tok/s), exit 0, no NaN and no abort |
| Finite logits | `examples/logit_trace models/kimi-k3-8l-mxfp4 <corpus> 32 2 8 0` | 64 teacher-forced positions, 512 top-k logits, 0 non-finite; NLL 10.35 to 17.23 (mean 13.65), top-k logits 6.66 to 10.38 |
| Multi-block Attention Residuals | same two commands on `models/kimi-k3-4l-attnres2` | 32 tokens at 54.61 tok/s, 64 positions traced, 0 non-finite; NLL 10.56 to 18.19 (mean 13.85) |

`models/kimi-k3-4l-attnres2` is the same 7 shards by symlink with `attn_res_block_size` lowered from 12 to 2, so layers 0 and 2 both freeze a block and layer 3's mix scores three entries against real `res_proj` / `res_norm` weights. At the published block size of 12 a copy small enough to load would only ever store one block, so the multi-entry softmax path would go untested on real data; the 13-layer copy that would exercise it naturally is about 220 GB.

Every row was re-run on the binary built from the final commit; both traces and both generations reproduce to the digit, which is also the determinism check. The generated text is multilingual noise in both runs. That is what 4 of 93 layers produces, compounded by the generic tiktoken fallback reading `tiktoken.model` with the wrong pre-tokenization regex until #1338 lands. The gate is the load, the shapes, the absence of NaN and the finiteness of the logits, not the text.

## Review

Two review passes (`pr-reviewer` with a Codex second opinion, and `pr-security-checker`) ran against this branch. Their findings are in https://github.com/lablup/mlxcel/pull/1741#issuecomment-5614031258 and are closed by the four `fix(kimi_k3)` commits: the Attention Residual mix no longer holds the whole window promoted to float32, the router top-k is bounded before `argpartition`, every `stack` / `view` / `concatenate` / `reshape` that takes checkpoint-derived arguments is checked first (each of those aborts the process instead of returning, so an unchecked one is an uncatchable `std::terminate` with no key named), 20-plus config-versus-checkpoint widths fail at load naming the tensor and the field, the chunked-prefill truncation above is fixed, and the two documentation paragraphs that claimed working pipeline parallelism are corrected.

## Acceptance criteria

- [x] Loads from the compressed-tensors mxfp4 checkpoint (layer-truncated) and runs prefill and decode with finite logits: see the table above.
- [x] KDA, MLA, latent SiTU MoE and Attention Residuals implemented as specified and unit-tested against scalar references.
- [ ] Fluent output on the full model through the distributed pipeline: not reachable on one machine (about 1.4 TB at 4 bits, three 512 GB nodes) and not reachable in code either, because this PR adds the partition profile but no pipeline stage executor. Discharged by the open follow-up #1734, which also carries the layer-by-layer f32 reference trace. `docs/distributed.md` and `docs/supported-models.md` now say so instead of describing PP as working.
- [x] `docs/supported-models.md` updated; detection table and test added.
- [ ] The workspace-wide `cargo test` / `cargo clippy --workspace --all-targets` gates were not run on this shared machine (concurrent `mlxcel-core` test binaries abort each other); the narrow scopes above plus CI cover them.

## Follow-ups filed

- #1734: Kimi K3 pipeline stage executor and full-model multi-node validation.
- #1749: chunked prefill discards the model-owned recurrent state in `kimi_linear` and five other families.

Closes #1334


inureyes:
## Security and performance re-check

Of the four HIGH findings claimed fixed in 4fcddfd6 and 9828b5ca, three are closed and one was only half closed. That one plus a new HIGH are fixed in 9c8ad3cd.

- **Attention Residuals transient (was half closed, now fixed).** The stacked `[B, T, n + 1, D]` value tensor is gone, but each block still went through one float32 array read twice, by the logit matmul and by the weighted term. Every consumer of the softmax runs after every logit, so all of those promotions stayed resident from the first loop until the fold: the same whole-window float32 residency the stack was removed to avoid, about 0.9 GB at hidden_size 7168 for a 4096-token prefill. Blocks now enter `matmul` and `multiply` in their stored dtype and MLX promotes inside each op, so each promotion is an op-local temporary. Bit-identical arithmetic (MLX's `matmul` and `multiply` both promote with the same `astype` the code was doing by hand).
- **Config versus checkpoint cross-checks: closed** for the cases named. `hidden_size` is covered transitively through the `model.norm.weight` numel check, so input-axis mismatches fail at load too.
- **Router bounds: closed.** `validate()` runs on both load paths, and `num_experts_per_token` in `[1, num_experts]` keeps `argpartition(kth = k - 1)` and the `[0, k)` slice inside the score axis.
- **`stack` / `view` / `concatenate` guards: closed.**

**New HIGH (fixed in 9c8ad3cd).** `decompose_kv_b_proj` reshaped checkpoint data with four `config.json` dimensions and no cross-check. A mismatched `text_config` aborted inside `reshape` during sanitize with no key to name, and a width that divides but is not `kv_lora_rank` survived sanitize and aborted inside the absorbed MLA matmul on the first forward pass. Kimi Linear shares the helper and now shares the guard.

**MEDIUM (reported, not fixed).** Remaining cross-check gaps that still surface as a mid-forward MLX abort rather than a load error: `lm_head.weight` rows against `vocab_size`, `routed_expert_up_proj` against hidden_size and the latent width, `moe_intermediate_size` against the stacked expert planes, `intermediate_size` against the dense MLP. `num_hidden_layers` is unbounded, so a hostile `config.json` turns `sanitize_weights`'s `0..num_hidden_layers` loop into a hang and `from_weights`'s `Vec::with_capacity` into a capacity-overflow panic. The residual window holds `ceil(num_hidden_layers / attn_res_block_size)` full-width copies of the residual stream for the whole forward (8 blocks at the published config, about 470 MB in bf16 for a 4096-token prefill) with no chunking for this family. Decode pays two mixes per layer, each one gemv plus one broadcast multiply per stored block, roughly 1.5k extra tiny ops per token at 93 layers and 8 blocks. `ResidualBlocks::push`, `KimiK3SparseMoE::forward` and `conv_decode_step` use `mlxcel_core::copy`, which is a real MLX Copy op; `astype` to the same dtype shares the handle for free, and the conv-state copy runs once per KDA layer per decode step.

**LOW.** `attention()` panics on a cache-type mismatch and `forward_for_sequence` panics on a missing prepared sequence state (the KimiLinear precedent, process-level in the server). `canonical_key` builds six format strings per key on the scope pass. Shape arithmetic such as `3 * delta_projection_dim()` is usize math on unbounded config fields and wraps in release before the shape comparison rejects it. `situ_activation` is `pub` and divides by `beta` with no guard of its own.

Also folded into 9c8ad3cd: only `.weight` was counted against `num_experts`, so a short `.scales` set stacked into a mismatched expert axis and aborted in `gather_qmm`; `Vec::with_capacity(num_experts)` reserved an unvalidated config field; the per-layer leftover scan built its prefix once per key.

Green after the change: `cargo fmt --all -- --check`, `cargo clippy --lib --tests --features metal,accelerate -- -D warnings`, 23 `models::kimi_k3` tests and 7 `models::kimi_linear` tests.

Issue #467, opened 2026-06-28
fix: externally quantized 4-bit Gemma 4 (OptiQ) decodes to degenerate output that is trapped entirely in reasoning_content


pjb7687:
## Summary

Serving an externally quantized 4-bit Gemma 4 model (`gemma-4-31B-it-qat-OptiQ-4bit`) on macOS / MLX produces degenerate decode output (a tiny set of tokens repeating, e.g. `//`, ` same`, `1`, even at `temperature: 0`), and every generated token is routed into `reasoning_content` while `content` is left empty. The failure reproduces on the raw `/v1/completions` endpoint and with speculative decoding disabled, so it is not a chat-template or sampling problem.

Two distinct mlxcel paths combine to produce exactly these symptoms: the 4-bit quantization loader silently reconciles quantization parameters with no validation, and the Gemma 4 thinking-channel filter classifies all output as reasoning whenever the closing channel marker never appears.

## Background

This triage is static. The project cannot be built or run in the triage environment (no Metal / MLX), so the analysis below comes from reading the checked-out source, not from reproducing the run.

One clarification on the original report first: `<|channel>` and `<channel|>` are not leaked gpt-oss / Harmony tokens. mlxcel deliberately uses that pair as Gemma 4's thinking-channel markers. See `src/server/thinking_budget.rs:223-249` (resolves token ids for `<|channel>` / `<channel|>`), `src/server/routes/chat.rs:930` (`OPEN_THINKING_SUFFIXES = ["<|channel>thought\n", ...]`) and `src/server/chat_template.rs:186-225` (primes `<|channel>thought\n` for Gemma 4 when `enable_thinking=true`). So those tokens appearing for a Gemma 4 model is expected; the real fault is the degenerate decode plus the way unclosed thinking is handled.

### Path A: degenerate decode (quantization reconciliation)

[...]

PR #1746, opened 2026-09-10
feat: surface reasoning-only truncation on the chat completions API

### LINKED ISSUE #1745: feat(server): surface reasoning-only truncation on /v1/chat/completions when a generation ends inside an open thinking block
## Summary

When a reasoning model's generation runs out of budget while still inside a `<think>...</think>` (or `<|channel|>`) block, the `/v1/chat/completions` response returns empty `content` with `finish_reason: "length"` and no signal that the answer never arrived. A caller cannot tell "the model intentionally returned nothing" apart from "the model was still thinking when it was cut off." This request adds an explicit, additive signal on the HTTP API for that condition, matching what the CLI already does after `#1721`.

## Background

The reasoning split and the existing partial coverage live in `src/server/routes/chat.rs`:

- `extract_reasoning_content` (around line 2199) routes the thinking scratchpad into `reasoning_content`, and `shape_response` in `src/server/reasoning_format.rs` decides whether thoughts land in `content` or `reasoning_content`.
- The only current detection of a never-closed thinking block is the `#467` guard at lines 819 to 827: `primed_thinking_unclosed(...)` emits a `tracing::warn!` on `target: "mlxcel::thinking"`. Two limits make it invisible to API callers:
  1. It is a server-side log line, never a response field.
  2. It is gated on `primed_open_thinking`, so it fires only when the prompt itself primed an open thinking channel, not for a plain unprimed request that simply exhausts `max_tokens` or `reasoning_budget` mid-thought.

So on an ordinary request to a reasoning model, when the budget ends inside the block, `content` is empty, `reasoning_content` holds the unfinished thinking, `finish_reason` is `"length"`, and nothing distinguishes this from a clean empty answer. `#1721` fixed the equivalent gap for the CLI `generate`/`chat` commands; that fix does not reach the HTTP API.

## Proposed Solution

Detect "generation ended while a thinking block was still open" for both primed and unprimed requests, and surface it additively on `/v1/chat/completions` for streaming and non-streaming responses. Keep it additive so existing clients are unaffected: existing fields (`content`, `reasoning_content`, `finish_reason`) keep their current values, and the new signal sits alongside them (for example, a boolean or a distinct field indicating the response is reasoning-only because it was truncated mid-thought).

## Implementation Notes

- The open-block detection already exists in narrower form. Generalize the condition used by `primed_thinking_unclosed` (`src/server/routes/chat.rs`) so it also covers the unprimed case: the generation stopped (notably `finish_reason == "length"`) with an unclosed thinking marker and empty user-facing `content`.
- Non-streaming path: the response is assembled around lines 890 to 961 via `ChatCompletionResponse::new(...)` / `new_with_tool_calls(...)` and `.with_reasoning_content_alias_field(...)`. Add the new signal on the same builder(s) in `src/server/types/response.rs`.
- Streaming path: the finish is computed around lines 1827 to 1910, where `finish_reason` is resolved and the final chunk emitted. Add the signal to the terminating chunk type in `src/server/types/stream.rs` (`ChatCompletionChunk`).
- Reuse `extract_reasoning_content` / `shape_response` rather than re-parsing; the signal is about detection state, not a change to how thoughts are placed.
- The existing `tracing::warn!` at lines 819 to 827 can stay as a server-side breadcrumb; the new field is the caller-facing counterpart.
- Cover both surfaces in the existing test modules (`streaming_tests.rs`, the `chat.rs` inline tests around the `extract_reasoning_content` / `shape_response` cases) with a truncated-mid-think fixture.

## Acceptance Criteria

- A non-streaming `/v1/chat/completions` response whose generation ends inside an unclosed thinking block (empty `content`, `finish_reason: "length"`, populated `reasoning_content`) carries an explicit signal that the answer was truncated while still reasoning.
- The streaming response surfaces the same signal on its terminating chunk.
- The signal fires for plain unprimed requests, not only when the prompt primed an open thinking channel.
- The change is additive: `content`, `reasoning_content`, and `finish_reason` keep their current values, and responses that did finish normally are unchanged (no false positive).
- Tests cover the truncated-mid-think case for both streaming and non-streaming.

---

## Summary

When a reasoning model's generation stays entirely inside `<think>...</think>` — because `max_tokens` or `reasoning_budget` runs out before the model closes the block — the `/v1/chat/completions` response (both streaming and non-streaming) returns `content: null`/`""` with `finish_reason: "length"` and no indication that anything else was coming. This is indistinguishable from a clean, intentionally empty response. `#467` covers a narrower version of this (server-side log only, only when the prompt itself primed an open thinking block); this happens for any plain, unprimed request too, and isn't surfaced to the API caller at all — only to server logs, and only in the primed case.

## Reproduction steps

```bash
mlxcel serve -m mlx-community/Qwen3.5-4B-MLX-4bit --port 8080
curl http://127.0.0.1:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"default","messages":[{"role":"user","content":"What is 2+2? Answer in one word."}],"max_tokens":60,"temperature":0}'
```

Observed: `content: ""`, `reasoning_content` full of unfinished thinking, `finish_reason: "length"`.
Expected: some field or signal indicating the answer never arrived because generation was still reasoning.

## Expected behavior

The response names this condition explicitly, the way the CLI's `generate`/`chat` now do (`#1721`) — e.g. an additive field alongside `content`/`reasoning_content`.

## Actual behavior

Silent. A client (or a person) sees an empty answer with no way to distinguish "the model has nothing to say" from "the model was still thinking."

## mlxcel version

v0.7.0 (main @ `13a0df44`)

## Install method

Built from source

## Platform

macOS (Apple Silicon)

## Model checkpoint

`mlx-community/Qwen3.5-4B-MLX-4bit`

## Additional context

Related to `#467` (narrower, primed-only, log-only) and `#1721` (CLI-only fix for the general case, doesn't reach the HTTP API). Confirmed reproducible even on trivial prompts — "What is 2+2?" didn't finish thinking within an 800-token budget at Qwen3.5's default `reasoning_budget=-1`.

## Pre-submission checks

- [x] I searched existing issues and did not find a duplicate
- [x] This is not a security vulnerability (those go to SECURITY.md instead)

bebekim:
## Summary

Adds an additive `reasoning_only: true` field to chat completion responses (both streaming and non-streaming) when a generation produced tokens but `content` came back empty because everything stayed in the reasoning channel — today that case is silent and indistinguishable from a clean, intentionally empty response.

## Related issues

Closes #1745
Refs #467
Refs #1721

(Neither #467 nor #1721 is a full duplicate: #467's fix only logs server-side, and only when the prompt itself primed an open thinking block; #1721 only fixed the CLI's `generate`/`chat` display. This generalizes past both to cover the far more common shape — a plain, unprimed request that opens its own `<think>` block and exhausts its token budget — and surfaces it to the actual API response, not just a log or a terminal.)

## Type of change

- [x] `feat` — new user-visible feature

## Test plan

- [x] `cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings`
- [x] `cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1` (full workspace: 10,605 passed, 0 failed)
- [x] `cargo deny check`
- [x] Validated with a real checkpoint: `mlx-community/Qwen3.5-4B-MLX-4bit`. Built baseline and patched release binaries from the same commit (stash/build/pop, mirroring `ab_output_equality.sh`'s own pattern) and compared live server responses:
  - Truncated mid-thought, non-streaming: before → `content:"", finish_reason:"length"`, no signal. After → same, plus `"reasoning_only": true`.
  - Truncated mid-thought, streaming: before → `delta:{}` on the finish chunk. After → `delta:{"reasoning_only":true}`.
  - Real completion (`--reasoning-budget 40`, reaches `finish_reason:"stop"`): field correctly absent in both paths, before and after — no regression on the common case.
- [ ] N/A — no arithmetic/generation-numerics touched (pure response-shaping), so the teacher-forced logit trace doesn't apply here.

## Notes for reviewers

- Design choice: an additive boolean field rather than overloading `finish_reason`, matching the precedent already set for `reasoning_content` — existing OpenAI-schema clients that validate `finish_reason` as an enum are unaffected.
- `reasoning_stream::is_reasoning_only` (added for #1721) is reused as-is with `show_reasoning: false` fixed at both server call sites — the server never suppresses reasoning into a hidden channel, so passing `false` correctly fires whenever `content` is empty and generation happened, independent of priming.
- Non-streaming: threaded through the two response arms where empty `content` is actually anomalous; deliberately *not* attached in the tool-calls-succeeded arm, where empty `content` is normal/expected by OpenAI convention.
- Streaming: added `StreamCallbackState::saw_content`, updated from a single check over the finalized per-token chunk batch (`chunk_carries_content`) plus the end-of-stream flush, read once at the terminal chunk.
- `primed_thinking_unclosed` keeps its original job (driving `strip_unclosed_primed_thinking`'s content-emptying) — only the log that used to key off it directly was generalized.

## Checklist

- [x] PR title uses a Conventional Commits prefix (`feat:`)
- [x] One logical change per PR
- [x] Updated `docs/`: N/A — no existing doc describes the chat completions response schema field-by-field
- [x] Updated `// Used by: ...` — extended `is_reasoning_only`'s doc comment to name the new server call site

---

https://claude.ai/code/session_016fmmt7cCRCmiwmTbWrEq8x

### DIFF
--- a/src/reasoning_stream.rs
+++ b/src/reasoning_stream.rs
@@ -311,6 +311,15 @@ pub fn render_full(
 /// fragment printed. True only when decoding produced text, none of it reached
 /// the content channel, and reasoning was suppressed; with `--show-reasoning` the
 /// text is already on screen.
+///
+/// Used by: the CLI `generate` and `chat` REPL paths above (#1721), and
+/// `src/server/routes/chat.rs`'s non-streaming and streaming chat completion
+/// handlers, which pass `show_reasoning: false` unconditionally — the server
+/// never suppresses reasoning into a hidden channel (`reasoning_content` is
+/// always additive alongside `content`), so the check there fires whenever
+/// `content` is empty and generation happened, regardless of priming. That
+/// generalizes past the narrower, primed-only condition the server logged
+/// under issue #467.
 pub fn is_reasoning_only(
     generated_text: &str,
     saw_visible_text: bool,
--- a/src/server/routes/chat.rs
+++ b/src/server/routes/chat.rs
@@ -888,6 +888,7 @@ pub(crate) async fn non_stream_chat_completion(
         let reasoning_only = log_if_reasoning_only(
             &result.text,
             &shaped.content,
+            shaped.reasoning_content.as_deref(),
             result.completion_tokens,
             &result.finish_reason,
         );
@@ -941,6 +942,7 @@ pub(crate) async fn non_stream_chat_completion(
     let reasoning_only = log_if_reasoning_only(
         &result.text,
         &shaped.content,
+        shaped.reasoning_content.as_deref(),
         result.completion_tokens,
         &result.finish_reason,
     );
@@ -1060,6 +1062,13 @@ struct StreamCallbackState {
     /// and no signal that nothing else is coming (see
     /// [`log_if_reasoning_only`] for the non-streaming counterpart).
     saw_content: bool,
+    /// Whether any non-empty text has reached `delta.reasoning_content` yet,
+    /// tracked exactly like `saw_content` over the same finalized chunk batch.
+    /// Read at finish time so `reasoning_only` can assert where the output
+    /// went, not merely that `delta.content` stayed empty: a tool-calling
+    /// stream and a structural-marker-only stream both leave `content` empty
+    /// without any reasoning having happened.
+    saw_reasoning_content: bool,
 }
 
 /// Re-emits the literal thinking delimiters into `delta.content` under
@@ -1588,6 +1597,7 @@ async fn stream_chat_completion(
             lp_buffer: std::collections::VecDeque::new(),
             thinking_echo: ThinkingDelimiterEcho::new(primed_close_marker.as_deref()),
             saw_content: false,
+            saw_reasoning_content: false,
         }));
         let cb_state_for_callback = cb_state.clone();
 
@@ -1786,6 +1796,11 @@ async fn stream_chat_completion(
                         if !cb.saw_content && pending.iter().any(chunk_carries_content) {
                             cb.saw_content = true;
                         }
+                        if !cb.saw_reasoning_content
+                            && pending.iter().any(chunk_carries_reasoning_content)
+                        {
+                            cb.saw_reasoning_content = true;
+                        }
                     }
 
                     for chunk in &pending {
@@ -1804,6 +1819,7 @@ async fn stream_chat_completion(
         // `cb.saw_content` (read at finish time, see its doc comment) reflects
         // this end-of-stream flush too, not just the main per-token loop.
         let mut remaining_had_content = false;
+        let mut remaining_had_reasoning = false;
         if let Some(text) = remaining.reasoning
             && !text.is_empty()
         {
@@ -1815,6 +1831,7 @@ async fn stream_chat_completion(
                     reasoning_alias_field,
                 );
                 let _ = finish_events.json(&chunk);
+                remaining_had_reasoning = true;
             }
             if reasoning_format.keeps_thoughts_in_content() {
                 let chunk = ChatCompletionChunk::content_with_logprobs(
@@ -1839,8 +1856,11 @@ async fn stream_chat_completion(
             let _ = finish_events.json(&chunk);
             remaining_had_content = true;
         }
-        if remaining_had_content && let Ok(mut cb) = cb_state.lock() {
-            cb.saw_content = true;
+        if (remaining_had_content || remaining_had_reasoning)
+            && let Ok(mut cb) = cb_state.lock()
+        {
+            cb.saw_content |= remaining_had_content;
+            cb.saw_reasoning_content |= remaining_had_reasoning;
         }
 
         if let Ok(r) = &result {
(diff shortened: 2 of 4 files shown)


inureyes:
## Hardening pass on this PR

Reviewed against #1745 (and the #467 / #1721 framing). The design holds: an additive `Option<bool>` that is never `Some(false)`, reusing `is_reasoning_only` so the CLI and the API agree, and excluding the non-streaming tool-calls arm were all the right calls. The type-level work, the `skip_serializing_if` omission, and the end-of-stream flush accounting are correct as authored. One commit on top, `88816432`, closes two false positives.

### Streaming tool calls were flagged as reasoning-only

`is_reasoning_only` reduces to "raw text non-empty and shaped content empty", which never checks that reasoning happened. `FilterState::ToolCall` suppresses the whole tool-call payload from `delta.content`, so a model answering with nothing but a tool call, the ordinary shape, reaches the finish chunk with `saw_content` false and a non-empty `result.text`. That chunk carried `reasoning_only: true` beside `finish_reason: "tool_calls"`. You excluded exactly this arm in the non-streaming path and called it out in the PR body; the streaming path had no equivalent, so the two surfaces disagreed on the same turn.

### Marker-only output claimed a channel it never used

`clean_structural_tokens` empties `content` for a generation containing no thinking block at all. A Gemma 4 turn that is nothing but `<turn|>` / `<channel|>` / `<|tool_call>` markers, which its own doc comment says happens when a request carries no tools, leaves `reasoning_content` `None` and still reported reasoning-only.

### Fix

Both paths now require that reasoning reached the client: non-streaming gates on the shaped `reasoning_content`, streaming tracks `saw_reasoning_content` over the same finalized chunk batch `saw_content` already used, flush included. The finish-time decision moved into `stream_reasoning_only`, so its three conditions are testable without a live model. This cannot suppress a true positive: `deepseek` / `auto` populate `reasoning_content`, and `none` / `deepseek-legacy` keep thoughts in `content`, which is then non-empty and fails the emptiness check anyway.

### Validation

31 targeted tests pass. Reverting each gate separately fails exactly its own test and no others. `cargo fmt --all -- --check` and `cargo clippy -p mlxcel --lib --tests -- -D warnings` both exit 0.

Not re-run here: your real-checkpoint check against `mlx-community/Qwen3.5-4B-MLX-4bit`. This box is Linux/CUDA, and the gates added are pure response-shaping with no generation numerics, so the live comparison you recorded still stands. Worth one streamed tool-call request against a tools-capable checkpoint before merge if you have one handy.

Security and performance: no auth, input-validation, or allocation behavior touched; the added work is one bool per chunk batch on a path that already scanned it.

Refs #1745

Issue #1343, opened 2026-08-22
feat(speculative): Muse Glimmer DFlash assistant drafter (muse_glimmer_assistant, 5 sliding-attention layers, bidirectional sliding mask) on the existing DFlash round loop


inureyes:
## Summary

`meta-models/Muse-Glimmer-30B-assistant` is a DFlash-style block-parallel drafter for `meta-models/Muse-Glimmer-30B` (`model_type: muse_glimmer_assistant`): five sliding-attention decoder layers with per-head q/k norms, fed with five of the target's residual streams through an `encoder.fc` projection, predicting a 16-token block in one non-causal forward under a bidirectional sliding-window mask, and borrowing the target's raw embedding table and untied `lm_head`. mlxcel serves Muse Glimmer (`src/models/muse_glimmer*.rs`) and has the DFlash round loop, but `mlxcel-server` rejects every `--draft-*` flag on a Muse target and the existing drafter model is the Qwen-shaped one (no q/k norms, no mask, plain growing cache, different key names). This change adds the assistant model, the Muse target adapter (hidden capture plus rotating-cache rollback), and the dispatch wiring so `--model-draft <assistant-dir>` runs the DFlash round loop on Muse Glimmer text requests with the exact-match verify rule.

## Current behavior

- `src/server/startup.rs:787-802` (`validate_muse_glimmer_unsupported_startup`): `ensure!(startup.draft_model_path.is_none() && startup.draft_kind.is_none() && startup.draft_block_size.is_none(), "Muse Glimmer VLM does not support speculative decoding or DFlash; ...")`, pinned by `src/server/muse_glimmer_startup_guard_tests.rs:86-97`. `docs/supported-models.md:562`: "Muse Glimmer is not a speculative or DFlash target in this baseline."
[...]

PR #1751, opened 2026-09-10
feat(speculative): LFM2 / LFM2.5 DSpark drafter on the DFlash loop

### LINKED ISSUE #1339: feat(speculative): LFM2 / LFM2.5 DSpark drafter (block-parallel DFlash backbone plus a low-rank Markov token-transition head) with exact greedy verify and short-conv rollback
## Summary

LiquidAI publishes "DSpark" drafters for the LFM2.5 family (`LiquidAI/LFM2.5-2.6B-DSpark` for `LiquidAI/LFM2.5-2.6B`, plus `LiquidAI/LFM2.5-8B-A1B-DSpark` for the `lfm2_moe` target and `LiquidAI/LFM2.5-1.2B-Instruct-DSpark`). A DSpark drafter is a 5-layer DFlash-style block-parallel transformer fed with five of the target's residual streams through an `fc` projection, plus a learned low-rank token-transition ("Markov") head that turns the block's parallel logits into a sequential chain. mlxcel serves `lfm2` and `lfm2_moe` (`src/models/lfm2.rs`) and already has the DFlash round loop, drafter trait and target trait (`src/lib/mlxcel-core/src/drafter/`), but no LFM2 target adapter, no Markov head, no DSpark config handling, and no short-conv rollback. This change adds all four so `--draft-model <dspark-dir>` on an LFM2 / LFM2.5 target runs the DFlash round loop with the greedy exact-match verify rule, at temperature 0 only, with output token-identical to classic decode.

## Current behavior

- No LFM2 drafter exists: `grep -rni "dspark\|lfm2.*draft" src/` is empty. `src/lib/mlxcel-core/src/drafter/mod.rs:211-230` maps only `gemma4_assistant`, `gemma4_unified_assistant` and `qwen3_5_mtp`; everything else resolves to `DEFAULT_DRAFTER_KIND = Dflash` (`:201`).
- Pointing `--draft-model` at a DSpark checkpoint today: `DFlashDrafter::load` (`src/lib/mlxcel-core/src/drafter/dflash/drafter.rs:98-128`) parses the config with `DFlashConfig::from_json` (`config.rs:220-243`, no `deny_unknown_fields`), so `markov_rank`, `rope_is_neox_style`, `runtime_block_size` and `enable_confidence_head` are dropped, the `markov_head.*` / `confidence_head.*` tensors are ignored, RoPE runs non-interleaved (`attention.rs:133-156` hard-codes `traditional = false`), and the server then declines the burst because the target is not Qwen 3.5 (`src/server/batch/speculative_burst.rs:1503-1516`).
- LFM2 cannot roll back a verify block: `Lfm2LayerCache::Conv` holds only the last `L_cache - 1` rows of `Bx` (`src/models/lfm2.rs:700-704`), `ShortConv::forward` overwrites it without keeping the previous state (`:393-395`), and the only trim hook resets the conv state to `None` (`trim_internal_caches`, `:1106-1124`). There is no per-layer hidden-state capture in `forward_embeds_with_caches` (`:784-819`).
- The DFlash infrastructure that exists: `trait Drafter` (`drafter/mod.rs:722-1076`: `bind`, `validate_target_compat`, `make_cache`, `reset`, `dflash_target_layer_ids`, `configured_block_size`, `prefer_requested_block_size`, `draft_block`, `sanitize`, `kind`), `DFlashDraftModel` (`drafter/dflash/model.rs`: `from_weights :117`, `forward :298-336`, `draft_block :352-386`), `DFlashAttention` (`drafter/dflash/attention.rs:125-183`), `trait SpeculativeTarget` and `DFlashGenerator::run` with `speculative_walk` (`drafter/dflash/round_loop.rs:136-270`, `:611-900`, `:407-447`), the Qwen 3.5 target impl (`src/models/qwen3_5.rs:2480-2560`, `rollback_speculative_cache :2133-2170`, conv-state recovery `:2262-2293`), dispatch (`src/server/speculative_dispatch.rs:96-154`, `src/server/batch/speculative_burst.rs:1463`), and the exactness gate (`src/models/speculative_exactness.rs`, `mtp_exactness_gate :295-366`).

### Drafter checkpoint

`config.json` of the published 2.6B drafter (the two other drafters share every value except `vocab_size`, `rope_theta`, `mask_token_id`, `target_layer_ids`, `num_target_layers`):

```
architectures: ["Lfm2DSparkDraftModel"], model_type: "qwen3", dtype: bfloat16
hidden_size 2048, num_hidden_layers 5, num_attention_heads 32, num_key_value_heads 8, head_dim 64,
intermediate_size 6144, hidden_act silu, rms_norm_eps 1e-5, vocab_size 128000, rope_theta 1e7,
max_position_embeddings 128000, layer_types ["full_attention"] x5,
block_size 9,
dflash_config: { mask_token_id 125017, target_layer_ids [2, 9, 17, 21, 27], num_target_layers 30 },
markov_rank 256, rope_is_neox_style false, enable_confidence_head true, markov_head_type "vanilla"
```

Detection: `dflash_config` present AND `markov_rank > 0` (or `architectures[0] == "Lfm2DSparkDraftModel"`). Semantics of `block_size`: it is the proposal count gamma (9); the verify width is `gamma + 1 = 10` rows (anchor plus 9 proposals). `runtime_block_size` (absent in the published config) defaults to 8 verify rows: 7 proposals plus the anchor; `--draft-block-size 10` restores the trained width. The drafter never backs the block size off on low acceptance (`prefer_requested_block_size() == true`).

Safetensors keys (no `model.` prefix; strip one if present, all bf16, 327.7M params):

```
fc.weight [2048, 10240]           hidden_norm.weight [2048]        norm.weight [2048]
layers.{0..4}.input_layernorm.weight [2048]       layers.{i}.post_attention_layernorm.weight [2048]
layers.{i}.self_attn.{q_proj [2048,2048], k_proj [512,2048], v_proj [512,2048], o_proj [2048,2048]}.weight
layers.{i}.self_attn.{q_norm, k_norm}.weight [64]
layers.{i}.mlp.{gate_proj [6144,2048], up_proj [6144,2048], down_proj [2048,6144]}.weight
markov_head.markov_w1.weight [128000, 256]    (embedding table: vocab -> rank)
markov_head.markov_w2.weight [128000, 256]    (linear rank -> vocab, stored [out, in])
confidence_head.proj.weight [1, 2304]  confidence_head.proj.bias [1]   (loaded, unused at decode)
```

No `embed_tokens`, no `lm_head`: both are borrowed from the target at `bind` (LFM2 ties them; `embed_scale = 1.0`).

### Draft step (one non-causal forward per round)

Inputs: `anchor` = the last committed token (the bonus of the previous round, or the last prompt token on the first round); `target_hidden` = `[1, S, 5 * 2048]`, the concatenation on the feature axis of the target's post-layer residual streams at layers `target_layer_ids` (captured before the final `embedding_norm`), for the `S` tokens committed since the last draft (the whole prompt on the first round, `accepted + 1` rows after a partial accept, all `bs` rows after a full accept).

```
gamma = bs - 1                                   # bs = verify width (8 by default)
draft_inputs = [anchor] ++ [mask_token_id] * (gamma - 1)        # [1, gamma]
h      = embed_tokens_target(draft_inputs)                      # [1, gamma, 2048]
h_ctx  = hidden_norm(fc(target_hidden))                         # [1, S, 2048]
for layer in layers:                                            # 5 x DFlash decoder layer
    x = input_layernorm(h)
    q = q_norm(q_proj(x));  k_prop = k_norm(k_proj(x));  v_prop = v_proj(x)
    k_ctx = k_norm(k_proj(h_ctx)); v_ctx = v_proj(h_ctx)
    q      = rope(q,      offset = cache.offset + S)            # traditional (interleaved) RoPE, theta 1e7
    k_ctx  = rope(k_ctx,  offset = cache.offset)
    k_prop = rope(k_prop, offset = cache.offset + S)
    K, V   = cache.update_and_fetch(k_ctx, v_ctx)               # only context K/V enter the cache; offset += S
    K = concat(K, k_prop, axis=2); V = concat(V, v_prop, axis=2)
    h = h + o_proj(sdpa(q, K, V, scale = 64^-0.5, mask = None)) # no mask: the block attends to itself bidirectionally
    h = h + mlp(post_attention_layernorm(h))                    # SwiGLU
base_logits = lm_head_target(norm(h))                           # [1, gamma, V]   (ALL positions; DFlash would drop position 0)
```

Markov chain over the block (vanilla head, rank 256):

```
prev = anchor
for step in 0..gamma:
    step_logits = base_logits[:, step] + markov_w2(markov_w1[prev])     # [1, V]
    prev = argmax(step_logits)
    draft[step] = prev
```

The drafter cache is one plain `KVCache` per drafter layer (the existing `drafter/dflash/cache.rs`), reset per generation, append-only (never trimmed): each round appends exactly the committed context rows.

### Verify rule and exactness contract

`verify_input = [anchor] ++ draft` (`[1, bs]`). One target forward over the block with causal attention, capturing the five residual streams, returns `[1, bs, V]` logits. `target_tokens = argmax(logits)` per row. Accept the longest prefix where `draft[i] == target_tokens[i]`; emit `draft[..accepted] ++ [target_tokens[accepted]]`, and `accepted == gamma` means all proposals were accepted and the last row's argmax is the bonus. This is the existing `speculative_walk` (`drafter/dflash/round_loop.rs:407-447`).

DSpark is greedy-only: any request with `temperature > 0` (or any sampler that is not argmax) declines to classic decode with a named log line; there is no stochastic acceptance rule.

Exactness contract at temperature 0: the emitted token sequence is identical to classic greedy decode. That holds if and only if every verify row's logits equal what a single-token decode step at that position would produce. On quantized checkpoints the `M = bs` matmuls and the `M = bs` MoE dispatch are not guaranteed bitwise equal to `M = 1`, so this change reuses the existing block-versus-chain exactness probe (`src/models/speculative_exactness.rs`, `mtp_exactness_gate`, with the `qmv_wide` retry) before enabling the burst, exactly as Qwen 3.5 MTP does, and declines to classic decode when the probe fails; `MLXCEL_MTP_ALLOW_INEXACT` keeps its meaning. #1289 (order-preserving streamed qmv) is the long-term route to passing the probe on every kernel; it is not a prerequisite here. The LFM2 MoE block must run the verify rows through the same path as single-token decode: `Lfm2MoeSparseMoeBlock::forward` takes the fused single-row kernel only when `x_flat.shape[0] == 1` (`src/models/lfm2.rs:553`), so the verify forward must either run the `bs` rows through the fused kernel one row at a time or the probe must cover the MoE layers (it does, because it compares full-model logits).

### Cache rollback

After a partial accept (`accepted < gamma`), with `n = accepted + 1` committed rows:

- Attention layers: `KVCache::trim(bs - n)` (`offset -= bs - n`; buffers untouched, the next append overwrites).
- Short-conv layers: the state is recomputed, not trimmed. During the verify forward each conv layer records `prev_state` (the `[1, L_cache - 1, H]` state before the block, zeros if `None`) and `bx_block` (the block's gated input `Bx = B * x`, `[1, bs, H]`). Rollback sets `state = concat(prev_state, bx_block[:, :n])[:, -(L_cache - 1):]`, i.e. the last `L_cache - 1` rows of the padded conv input truncated at the committed prefix (`L_cache = 3` on the published checkpoints, so a 2-row window).
- The drafter cache is not rolled back.

A full accept keeps every cache as the verify forward left it.

### Surfaces

`mlxcel generate -m <lfm2-target> --draft-model <dspark-dir> [--draft-block-size 8|10]` and `mlxcel-server -m <lfm2-target> --model-draft <dspark-dir>` (B = 1 burst; B > 1 declines to classic as the Qwen 3.5 DFlash path does today). `--draft-kind dflash` is accepted for the pairing; no new kind string. The per-request speculative stats already emitted for DFlash are reported unchanged.

## Implementation plan

1. `src/lib/mlxcel-core/src/drafter/dflash/config.rs`: add `markov_rank: usize` (default 0), `rope_is_neox_style: bool` (default `true`), `runtime_block_size: Option<usize>`, `enable_confidence_head: bool` to `DFlashConfig`; add `pub fn is_dspark(&self) -> bool { self.markov_rank > 0 }` and `pub fn verify_width(&self) -> usize` (`block_size + 1` when `is_dspark()`, `block_size` otherwise); extend `is_dflash_drafter_config` (`:274-290`) so `architectures: ["Lfm2DSparkDraftModel"]` also qualifies.
2. `src/lib/mlxcel-core/src/drafter/dflash/attention.rs:133-156`: thread a `traditional: bool` into the RoPE call (`mlxcel_core::fast_rope(..., traditional, ...)`), `true` when `rope_is_neox_style == false`.
3. New `src/lib/mlxcel-core/src/drafter/dflash/markov.rs`: `pub struct VanillaMarkovHead { w1: UnifiedEmbedding /* [V, rank] */, w2: UnifiedLinear /* rank -> V */ }` with `from_weights(weights, "markov_head")` (accept quantized `.scales/.biases` siblings through the existing `UnifiedEmbedding` / `UnifiedLinear` loaders) and `fn sample_block(&self, base_logits: &MlxArray /* [1, gamma, V] */, anchor: i32) -> Vec<i32>` implementing the chain above with `argmax` per step (`materialize_argmax_i32_vec`, `drafter/dflash/mod.rs:94`). Load `confidence_head.*` into an `Option` and never call it.
4. `src/lib/mlxcel-core/src/drafter/dflash/model.rs`: in `from_weights` (`:117`) build the Markov head when `config.is_dspark()`; in `draft_block` (`:352-386`) branch: DSpark builds `[anchor] ++ mask * (gamma - 1)` (gamma = `block_size_arg - 1`), keeps logits for all `gamma` positions, and returns `markov.sample_block(logits, anchor)`; DFlash keeps the existing `[bonus] ++ mask * (bs - 1)` and `logits[:, 1:]`. Override `prefer_requested_block_size() -> true` and `configured_block_size() -> min(verify_width, runtime_block_size.unwrap_or(8))` for DSpark.
5. `src/lib/mlxcel-core/src/drafter/mod.rs`: `load_drafter` (`:1097-1133`) `Dflash` arm already reaches `DFlashDrafter::load`; add `validate_target_compat` for DSpark: target `model_type in {lfm2, lfm2_moe}`, `hidden_size == 2048` (5 * hidden == `fc` input width), `num_layers == num_target_layers`, `vocab_size` equal, `target_layer_ids` strictly increasing and `< num_target_layers`. Reject `temperature > 0` at dispatch with `DrafterError::GreedyOnly` (new variant).
6. `src/models/lfm2.rs`:
   - `ShortConv::forward` (`:359-427`): add a `capture: Option<&mut Vec<ConvRollbackSnapshot>>` parameter (or a sibling `forward_capturing`) that pushes `ConvRollbackSnapshot { layer_idx, prev_state: Option<UniquePtr<MlxArray>>, bx_block: UniquePtr<MlxArray> }` (copy `prev_state` with `mlxcel_core::copy` before the overwrite; keep `bx` through `contiguous`).
   - New `pub struct VerifyOutput { logits, hidden_states: Vec<UniquePtr<MlxArray>>, conv_states: Vec<ConvRollbackSnapshot> }` and `fn forward_speculative(&self, verify_input, caches: &mut [Lfm2LayerCache], capture_layer_ids: &[usize]) -> VerifyOutput`: same loop as `forward_embeds_with_caches` (`:784-819`) with a causal mask anchored on the first attention layer's offset, pushing `h` after layer `i` for `i in capture_layer_ids` (before `embedding_norm`) and returning full-block logits (`embed_tokens.as_linear(embedding_norm(h))`, already `[1, bs, V]`).
   - `fn rollback_speculative_cache(&self, caches, conv_states, accepted: i32, block_size: i32)`: `n = accepted + 1; trim = block_size - n; for cache: Attention => kv.trim(trim); Conv(layer) => state = last n_keep rows of concat(prev_state_or_zeros, bx_block[:, :n])` with `n_keep = conv_L_cache - 1`.
   - `impl mlxcel_core::drafter::dflash::SpeculativeTarget for Lfm2Model` (and the `Lfm2Moe` / `Lfm2VL` text wrappers): `type Cache = Lfm2LayerCache`, `type VerifyOut = VerifyOutput`, `capture_layer_ids` from the bound drafter config, `verify_forward_with_capture_layers` -> `forward_speculative`, `rollback_partial` -> `rollback_speculative_cache`, `concat_hidden_for_drafter` = `concatenate(hidden_states, axis = -1)` (`[1, bs, 10240]`), `verify_logits`. Sequence-routed variants through `ModelOwnedSequenceState::with_sequence_state(Some(seq_id), ..)` as `src/models/qwen3_5.rs:2416-2450` does.
   - Exactness probe: implement `mtp_exactness_allows` / `probe_block_chain_exactness` for LFM2 following `src/models/qwen3_5.rs:1514-1600` (three probe draws, block of `bs` tokens versus the same tokens decoded one at a time, compare argmax per position and logits bitwise; the MoE path is exercised by the probe automatically).
7. Server: `src/server/batch/speculative_burst.rs:1503-1516` add `LoadedModel::Lfm2 | LoadedModel::Lfm2Moe | LoadedModel::Lfm2VL(text-only request)` to the DFlash target gate; generalize `Qwen35DFlashTarget` (`:209-229`) into a `DFlashTargetModel` trait with `make_dflash_caches` implemented by `Lfm2Model::make_caches`, and a `run_dflash_on_lfm2` (or make `run_dflash_on_qwen35` generic over the trait). The prefill feeds the whole prompt's captured hidden to the first draft (the current Qwen path slices to the last position only, `:1741-1745`; DSpark wants every prompt row, so make that a trait method `first_hidden_rows()`).
8. CLI: `src/commands/generate.rs:1486-1513` currently rejects `--draft-kind dflash` offline; either keep the rejection for LFM2 (server-only, like Qwen 3.5) or construct `DFlashGenerator` there. Minimum: a named error that points at `mlxcel-server`.
9. `src/cli/speculative_args.rs:217-233` (`resolve_draft_block_size`): peek the drafter config for DSpark and default to `min(block_size + 1, runtime_block_size.unwrap_or(8))`.
10. Docs: `docs/supported-models.md` "Speculative decoding" table gains a `DSpark | LFM2 / LFM2.5 text (B = 1)` row; `docs/speculative-acceptance.md` adds DSpark to the argmax-against-argmax list and states the greedy-only decline; README mention under speculative decoding.

## Validation

(a) Unit tests:
- `src/lib/mlxcel-core/src/drafter/dflash/markov_tests.rs`: `markov_chain_feeds_previous_token_into_next_step` (hand-built `w1`/`w2` where `w2(w1[t])` adds a large bias to token `t + 1`: with flat base logits the chain emits `anchor+1, anchor+2, ...`), `dspark_draft_block_uses_all_positions` (a synthetic drafter returning position-indexed logits yields `gamma` proposals with proposal 0 derived from the anchor row).
- `src/lib/mlxcel-core/src/drafter/dflash/config_tests.rs`: `dspark_config_lifts_dflash_config_and_markov_rank`, `dspark_verify_width_is_block_size_plus_one`, `rope_is_neox_style_false_selects_traditional_rope`.
- `src/models/lfm2_tests.rs`: `conv_rollback_matches_committed_prefix` (random 2-layer LFM2, prefill 5 tokens, verify 4, rollback `accepted = 1`; assert attention `offset == 7`, conv state `allclose 1e-5` and next-token logits `allclose 1e-4` against a reference model that consumed `prompt ++ verify[:2]` in one pass), `verify_forward_captures_requested_layers` (shapes `[1, bs, hidden]` for each id, before `embedding_norm`), `speculative_block_logits_match_single_token_decode` on the synthetic model (bitwise, the exactness premise).
- `src/lib/mlxcel-core/src/drafter/dflash/round_loop.rs` tests: `dspark_round_loop_greedy_parity_for_sixty_four_tokens` with a synthetic `SpeculativeTarget` (mirrors `greedy_parity_with_synthetic_target_for_thirty_two_tokens`, `:1749`).

(b) Real checkpoints (`LiquidAI/LFM2.5-2.6B` with `LiquidAI/LFM2.5-2.6B-DSpark`; the local `models/lfm2-350m-8bit` has no published drafter, and the MoE pair is `LiquidAI/LFM2.5-8B-A1B` with `LiquidAI/LFM2.5-8B-A1B-DSpark`; use the bf16 originals, or quantize the target locally with the checkpoint's own conversion tool, and keep the drafter bf16):

```
./target/release/mlxcel-server -m models/LFM2.5-2.6B --model-draft models/LFM2.5-2.6B-DSpark --port 8080
curl -s localhost:8080/v1/chat/completions -d '{"model":"m","messages":[{"role":"user","content":"Explain speculative decoding in three sentences."}],"max_tokens":256,"temperature":0}'
# baseline: same request with the server started without --model-draft
```

Acceptance: all 256 token ids identical to the baseline on both the dense and the MoE pair (`MLXCEL_FUSED_MOE=0` for the MoE reference diff); the response `timings` report a mean accepted length above 1; a `temperature: 0.7` request is served by classic decode with one warning in the log. Record tok/s for both pairs at `--draft-block-size 8` and `10` in `docs/benchmark_results/` (the published M5 numbers for the dense pair are about 2.9x over classic at 256 tokens on reasoning prompts and 1.24x on the MoE pair; M1 Ultra will differ).

## Acceptance criteria

- [x] A DSpark checkpoint is detected, loaded (Markov head included, confidence head ignored) and bound to an `lfm2` / `lfm2_moe` target; mismatched targets are rejected with a named error.
- [x] The DFlash round loop runs on LFM2 with `[anchor, mask...]` blocks, all-position logits, the Markov chain, and the exact-match walk.
- [x] Partial-accept rollback restores attention offsets and the short-conv state to the committed prefix (unit test against a one-pass reference).
- [x] Temperature 0 output is token-identical to classic decode on the dense and the MoE pair, gated by the block-versus-chain exactness probe; `temperature > 0` declines to classic with a log line.

## Out of scope

- Batched (B > 1) DSpark; the B = 1 burst declines to classic for concurrent requests like the Qwen 3.5 DFlash path.
- A sampled (non-greedy) acceptance rule for DSpark; the drafter is trained for greedy chains.
- The confidence head (loaded, unused) and any early-exit policy built on it.
- A Metal kernel reproducing the single-row reduction order for `M = bs` rows (#1289).

inureyes:
`--draft-model <dspark-dir>` on an `lfm2` / `lfm2_moe` target now runs the DFlash round loop with the greedy exact-match verify rule. A DSpark drafter is a 5-layer DFlash backbone fed five of the target's residual streams plus a rank-256 Markov token-transition head, so this adds the head, the DSpark config fields (`markov_rank`, `rope_is_neox_style`, `runtime_block_size`, `enable_confidence_head`, and `num_target_layers` lifted out of the nested `dflash_config`), the checkpoint's traditional RoPE pairing, the LFM2 verify forward with residual-stream capture, and the partial-accept rollback. Attention layers trim their KV cache; a short-conv state cannot be trimmed, so it is recomputed from the block's gated input at the committed prefix. `Qwen35DFlashTarget` generalizes into `DFlashTargetModel`, carrying `make_dflash_caches`, the first-draft prompt-row policy (Qwen keeps the last position, DSpark wants every prompt row) and the exactness gate, so one driver serves both families and #1343 can add a third with one `impl`.

Two LFM2 loader gaps that the bf16 originals expose are fixed alongside, because the pairing cannot load without them: `rope_theta` nested under `rope_parameters` was read as the 1e6 default, and unstacked per-expert `feed_forward.experts.{e}.w1/w2/w3` tensors were never renamed, so MoE expert stacking never fired.

DSpark is greedy-only: a request with `temperature > 0` and `top_k != 1` is served by classic decode with one named warning, because the Markov head proposes a chain trained against greedy targets and there is no stochastic acceptance rule.

## Real-checkpoint validation

Apple M5 Max (GPU generation 17), port 19339, against the release binaries built from this tree. The machine served other work during the session, so each throughput figure is one sample; the acceptance counters are deterministic and reproduced byte for byte across repeat runs and across a rebuild.

```
./target/release/mlxcel-server -m models/lfm2.5-2.6b-bf16 --model-draft models/lfm2.5-2.6b-dspark-bf16 --port 19339
curl -s localhost:19339/v1/chat/completions -H 'Content-Type: application/json' -d '{"model":"m","messages":[{"role":"user","content":"Write a detailed technical explanation of how speculative decoding works in large language model inference, covering the drafter, the verification step, and the acceptance rule."}],"max_tokens":256,"temperature":0,"logprobs":true}'
```

The baseline is the same request against a server started without `--model-draft`; identity is compared over the 256 `logprobs.content[].token` entries, both arms finishing on `length` at exactly 256 tokens.

**The block-versus-chain probe declines on this host, for both pairs.** That is the documented generation-15-and-newer condition and the shipped default: the burst falls back to classic decode, emits no `timings`, and serves the baseline's tokens (verified on the MoE pair with the drafter configured and the gate honored, 256 of 256 identical). The runs below therefore used `MLXCEL_MTP_ALLOW_INEXACT=1`, the existing research override.

- **Dense pair** (`models/lfm2.5-2.6b-bf16` + `models/lfm2.5-2.6b-dspark-bf16`): all 256 token ids identical to the baseline at the default verify width 8 and again at `--draft-block-size 10`. Mean accepted length 3.59 at width 8 (`draft_n` 489, `draft_n_accepted` 184, `draft_rounds` 71) and 3.75 at width 10 (599 / 187 / 68). 203.3 and 212.1 tok/s against 72.3 tok/s classic (256 tokens in 3594 ms of request wall clock including a 66 ms prefill), so about 2.8x.
- **MoE pair** (`models/lfm2.5-8b-a1b-bf16` + `models/lfm2.5-8b-a1b-dspark-bf16`, `MLXCEL_FUSED_MOE=0` for the reference diff): mean accepted length 2.61 at width 8 (674 / 158 / 98) and 2.88 at width 10; 115.5 tok/s against 104.3 tok/s classic (2525 ms wall, 81 ms prefill), so about 1.11x, and the fused default measured the same classic figure (2532 ms). Token ids match for 153 of 256 and then diverge on a tie the two arms resolve differently: the baseline takes `':'` at logprob -1.0312 and the block path `' formula'` at -1.0000, and on the issue's shorter prompt both candidates read -1.4375. That is the byte-identity the probe declined, not a semantic difference; it reproduces at the same position across both widths and across runs, while the fused and unfused classic arms agree with each other.
- **Temperature.** `"temperature": 0.7` on the dense pair is served by classic decode, with no `timings` and exactly one log line: `DFlash speculative dispatch declined for seq seq-2: the drafter is greedy-only (DSpark) and the request samples with temperature 0.7 / top_k 50; serving it with classic decode`.

**Re-validated after the review and security fixes**, against release binaries rebuilt from the final tree (no source file newer than the binary). The dense pair is 128 of 128 token ids identical to the same server without `--model-draft`, and the acceptance counters reproduce the pre-fix run digit for digit: 34 rounds, 234 proposed, 94 accepted, acceptance 0.402, 3.735 emitted per verify. That is the point of the re-run: the two allocation fixes below change what is kept, never what is computed, and identical counters are what says so. The temperature-0.7 decline reproduces with its one log line, and the mismatched pairing (`--model-draft models/qwen3.5-4b-dflash` on an LFM2 target, the case the new gate exists for) is refused by name at warmup and on the request, with the server still alive afterwards rather than aborted.

## Review fixes

The implementation review found no CRITICAL and no HIGH. One MEDIUM was a regression this branch introduced and is fixed here, along with two unbounded checkpoint-supplied numbers and two unchecked weight shapes found while fixing it, and three doc statements the review showed to be false.

**An LFM2 target paired with a non-DSpark DFlash drafter had no gate.** `Drafter::validate_target_compat` cannot make that call: it reads the target as a `LanguageModel`, which carries no architecture string, so a plain DFlash drafter has no way to tell an LFM2 target from the Qwen 3.5 one it was published for, and returns `Ok`. `DFlashDraftModel::forward` does not check its `fc` input width either, so the mismatch reached MLX as a shape throw, and an MLX C++ exception crossing the cxx bridge aborts the process rather than failing the request. Before this branch the pairing declined at the burst's variant gate, which no longer rejects LFM2. The target is the side that knows its own family, so the policy is now `DFlashTargetModel::requires_dspark_drafter()`, `true` for `Lfm2Model` and `Lfm2VlModel` and the permissive default everywhere else, read by both run arms before any forward and answered with one operator-facing message naming `--model-draft` and the published pairings. Qwen 3.5 keeps the default; widening it there would change which DFlash pairings that family accepts, which is a separate call.

**Two checkpoint-supplied numbers reached a buffer index or a round size unbounded**, and are now in the DSpark pairing gate beside the `fc`-width and `target_layer_ids` checks. `mask_token_id` indexes the target's embedding table, because a DSpark drafter ships none of its own, and MLX range-checks no positive gather index, so an id past the last row read whatever followed the table in the buffer and fed it to the logits; the target half of the same gate already pins the target's vocabulary, so bounding the id against `vocab_size` bounds the gather. A `runtime_verify_width()` below two rows proposes nothing and would emit one token per burst. `verify_width()` saturates rather than wrapping, since `usize::MAX` there panics in a debug build and wraps to a zero-row width in a release one.

**The Markov head's two factors were never measured against the config that sizes them.** `markov_w1` is gathered at a token id once per chain step, so a table with fewer rows than the vocabulary the chain draws from read past its own buffer for the same reason; a `markov_w2` of the wrong width threw inside MLX, which crosses the bridge as a process abort rather than a load error. `VanillaMarkovHead::from_weights` now takes the vocabulary as well as the rank and checks both factors before either becomes a layer, skipping the rank check for a quantized factor (which bit-packs `rank` along its last axis only, so its stored width is a function of the bit depth) and not the row check.

**Doc corrections, no behavior change.** `ShortConv::forward_with_capture` claimed the snapshot costs no copy while the code takes an explicit one, which it has to, since `conv_state` is reassigned at the end of the call. `configured_block_size` implied a DFlash round-loop block-size policy that does not exist: the only readers are the MTP generator and its batched round loop, and a DSpark run gets its width from `resolve_draft_block_size` before the drafter is loaded. The `dflash_target` module doc claimed a new family costs one `impl` block plus a match arm, when it is three arms in `speculative_burst` and one of them still carries the LFM2 B = 1 decline as per-family policy outside the trait. The `Drafter::is_dspark` impl also delegates to the inherent method by qualified name rather than repeating its body, so the two cannot drift.

The review's second MEDIUM, a `supports_batched()` trait method to move that last hardcoded arm into the trait, is left for #1343, which is the family that needs it. The LOW findings (the probe measures one width, `ProbeKey` has no family discriminator, the two block-size hooks are declarative, the benchmark numbers stayed in this body) are recorded in the technical report with the reasons they stay open.

## Security fixes

A second review pass looked specifically at what a downloaded checkpoint controls. Pointing `--model-draft` at a repository makes every field of its `config.json` and every weight shape in it a runtime input; MLX range-checks no positive gather index, and an MLX C++ exception crossing the cxx bridge aborts the process rather than failing the request. One HIGH, four MEDIUM and three LOW; six fixed, two left with the reason recorded in the technical report.

**HIGH: the verify width was bounded below and not above, and this branch is what made it a control input.** Before it the DFlash runtime block size came only from `--draft-block-size` or the flat constant 16, both operator-supplied. `resolve_draft_block_size` now peeks the drafter checkpoint and hands `runtime_verify_width()` to the scheduler as the server-wide block size, and a config setting `block_size` and `runtime_block_size` together escapes the eight-row default entirely. What that reaches first is the block-versus-chain exactness probe, which runs one single-token target forward per row, three draws over, on the scheduler thread, memoizes its verdict, and runs before the pairing gate that would otherwise refuse the config: at an absurd width that is a hang, not a slow request. `runtime_verify_width()` now clamps at `DSPARK_MAX_VERIFY_WIDTH` (32, against 8 to 10 on the published checkpoints), which is the bound that holds ahead of the probe because every caller goes through it; the gate separately refuses a config asking for more, so the operator is told rather than quietly served at 32.

**MEDIUM: the `mask_token_id` bound was vacuous for a self-contained drafter.** Its soundness argument was that the target half of the pairing gate pins the target's vocabulary, which holds only on the lazy-bind path every published DSpark checkpoint uses. A checkpoint shipping its own `embed_tokens` takes the other arm of `bind` and its table was never compared to anything. `LmHead::Own` had the same gap one step later, against the Markov head's `vocab_size`-shaped transition logits. Both are now measured at load for a DSpark config; a plain DFlash checkpoint is deliberately untouched.

**MEDIUM: the Markov rank check was skipped, not narrowed, for a quantized factor**, and every mlx-community conversion of these drafters is quantized, so the skipped case is the common one. The packed width is `rank * bits / 32`, so it is checkable: the check now accepts a width some supported bit depth explains, rather than trusting a declared `bits` a per-tensor override can contradict.

**MEDIUM: two prompt-sized allocations that nothing read.** `first_hidden_for` deep-copied the whole prompt hidden under the every-row policy, because `mlxcel_core::copy` is a real MLX `Copy` primitive and the signature took a borrow (168 MB at an 8k prompt on LFM2.5-2.6B, 671 MB at 32k); it now takes the array by value, which also ends the caller's retention of the original across the round loop. Separately, the short-conv rollback snapshots were captured on the prompt prefill, which is never rolled back, pinning roughly one prompt-sized buffer per conv layer alive through the eval that materialises the prefill (about 670 MB at 8k, about 2.7 GB at 32k on a 30-layer 2.6B checkpoint). `SpeculativeTarget` gains `prefill_forward_with_capture_layers`, defaulting to the verify hook so no other family changes, and LFM2 overrides it to skip the capture. The flag changes what is kept and never what is computed, which the new test pins on the logits, the captured hidden and the cache offsets at once.

**LOW: an empty prompt or an empty window** reached a negative slice start (a process abort) and an index panic respectively. Both were inherited unchanged from the Qwen arms and neither is reachable from the scheduler today; both are now request errors.

## Test plan

- [x] `cargo test --profile test-fast --features metal,accelerate`: `-p mlxcel-core drafter::dflash` 74 passed, and one `--lib` run over `models::lfm2`, `server::batch::speculative_burst`, `server::batch::dflash_target`, `cli::speculative_args` and `models::detection` 184 passed. Seven of those are new here: the verify-width ceiling including the `usize::MAX` saturating case, the self-contained DSpark vocabulary tables, the packed Markov rank check, the conv-capture flag, the mask-id and low-side width bounds, and the LFM2 non-DSpark pairing decline.
- [x] `cargo check` and `cargo clippy -- -D warnings` clean for the root package AND for `-p mlxcel-core` separately, plus `cargo fmt --all -- --check`. Checking `--lib --tests` at the workspace root resolves to `-p mlxcel` and never compiles `mlxcel-core`'s test target, which is how two test-only compile errors reached a suite run; both packages are checked from here on.
- [x] `--bin mlxcel offline_draft_model` 4 passed.

Closes #1339


inureyes:

### Intent

> `--draft-model <dspark-dir>` on an `lfm2` / `lfm2_moe` target runs the existing DFlash round loop with the greedy exact-match verify rule at temperature 0 only, emitting output token-identical to classic decode.

### Verified against the issue, line by line

- `block_size 9` is read as gamma and `verify_width()` as `block_size + 1`; `runtime_verify_width()` caps at the eight-row default and `--draft-block-size 10` restores the trained width. `resolve_draft_block_size` peeks the drafter config so the flat DFlash 16 never reaches a 9-proposal head, and returns `None` for a plain DFlash checkpoint so the Qwen default is untouched.
- The DSpark draft input is `[anchor] ++ mask * (gamma - 1)` at gamma rows and every one of the gamma logit positions is consumed, position 0 included. `dspark_draft_block_uses_all_positions` asserts that directly rather than by shape.
- The Markov chain is seeded with the anchor and feeds each step's argmax into the next step's gather, built as one lazy graph with a single host sync per round. `markov_chain_feeds_previous_token_into_next_step` is the test that a chain ignoring its own output would fail.
- `rope_is_neox_style: false` reaches all three `fast_rope` call sites as `traditional = true`, and the test asserts the attention output changes rather than just the struct field.
- Context K/V only in the drafter cache with proposal K/V concatenated after, `mask = None`, append-only and never trimmed: unchanged pre-existing `DFlashAttention` / `DFlashKVCache` behavior, correctly left alone.
- `prefer_requested_block_size() == true` and the target's tied `embed_tokens` borrowed as both input embedding and LM head at `bind`.

### Short-conv rollback

`n = accepted + 1`, attention `KVCache::trim(block_size - n)`, short-conv recomputed as the last `L_cache - 1` rows of `concat(prev_state_or_zeros, bx_block[:, :n])`. Two things I checked specifically: `prev_state` is `mlx::core::copy` (a real copy node, confirmed at `mlx_cxx_bridge.cpp:1078`) taken after `bx` is computed and before `*conv_state` is assigned, so it cannot be the post-block state; and the zeros fallback is `[bx.shape[0], L_cache - 1, bx.shape[2]]` in `bx`'s dtype, matching `conv_padding()`'s causal `(l_cache - 1, 0)`. `conv_rollback_matches_committed_prefix` compares offset, state and the next decode step's logits against a genuine one-pass reference, which is the right shape of test for this.

### `DFlashTargetModel` readiness for #1343

Ready. Every added hook defaults to the Qwen 3.5 behavior (`first_hidden_rows` to `LastPromptPosition`, `enable_speculative_buffers` to a no-op, `exactness_allows` to `true`), `concat_captured_hidden` and `first_hidden_for(LastPromptPosition, ..)` are equivalent to the code they replaced rather than merely similar, and no inherent method shadows a new trait method (`Qwen35Model` has `mtp_exactness_allows`, not `exactness_allows`, so the Qwen DFlash path does not begin gating on a probe it never gated on). `enable_speculative_buffers` is present and called by both drivers before the prefill, which is what the Muse Glimmer rotating cache needs; rotating-cache rollback is already per-family through `SpeculativeTarget::rollback_partial`.

### Findings, all left open

No CRITICAL and no HIGH. Nothing was changed on the branch except the technical report.

- **MEDIUM.** A mismatched non-DSpark pairing on an LFM2 target has no gate: `validate_target_compat` returns `Ok` immediately when the drafter is not DSpark and `DFlashDraftModel::forward` does not check its `fc` input width, so an LFM2 target with a genuine Qwen 3.5 DFlash drafter reaches an MLX shape throw through a non-`Result` cxx shim. Pre-existing in kind (a 4B Qwen target with a 27B DFlash drafter fails identically), but LFM2 targets used to decline at the variant gate. Narrow fix: decline in the burst gate when the target is LFM2 and the drafter is not DSpark. Broad fix: an `fc`-width check for every DFlash drafter, which changes which Qwen pairings are accepted and so is a maintainer call.
- **MEDIUM.** A third family touches four match arms, not the "one `impl` block plus a match arm" the module doc claims: the exactness gate, the `drive!` dispatch, the batched gate (which hardcodes the LFM2 B = 1 decline), and `model_variant_label`. A `supports_batched()` trait method would move the third into the trait; worth doing when #1343 lands.
- **LOW.** The probe measures one width. `dflash_exactness_allows(block_size)` probes at the configured verify width, but the round loop narrows `bs` near the token budget, and this project's benchmark guidance records that block-versus-chain disagreement varies strongly with width. Same shape as the MTP arms, so not a regression.
- **LOW.** `ProbeKey` is `{block_size, hidden_size, num_hidden_layers}` with no family discriminator and a docstring that assumes one target model per process. Router mode does not. Pre-existing, widened by a third family.
- **LOW.** `configured_block_size()` and `prefer_requested_block_size()` are implemented as the issue asked but only the MTP generator reads them; the DFlash round loop uses its constructed `block_size` directly. The no-backoff requirement holds vacuously, and the comment at `drafter.rs:459` implies a DFlash block-size policy that does not exist.
- **LOW.** `ShortConv::forward_with_capture`'s doc says "the state is replaced, never mutated in place, so holding the previous handle costs no copy" while the code takes a copy. The copy is correct and cheap; the comment contradicts it.
- **LOW.** The issue's Validation section asked for tok/s at both widths in `docs/benchmark_results/`; they are in the PR body instead. Given the host served other work and each throughput figure is one sample, keeping single-sample numbers out of the benchmark corpus is the right call, but the deviation is worth stating.

### Verification

- [x] No placeholder/mock code remaining (no `todo!`, `unimplemented!`, `FIXME` or stub in the diff)
- [x] Integrated into project code flow (burst gate, both generic drivers, CLI block-size resolution, offline rejection, detection)
- [x] Project conventions followed (no shared hotspot in `layers.rs` / `utils.rs` / `switch_layers.rs` / `gated_delta.rs` touched, no local `references/...` path, `docs/supported-models.md` and `docs/speculative-acceptance.md` updated)
- [x] Existing modules reused (`UnifiedEmbedding` / `UnifiedLinear` for the Markov head, `mtp_exactness_gate` for the probe, `materialize_argmax_i32_vec`, `create_causal_mask`, `slice_axis`, `concatenate_many`)
- [x] No unintended structural changes (the `Qwen35DFlashTarget` generalization is behavior-preserving)
- [x] Tests pass

Reproduced locally at review time: `cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings` and `cargo fmt --all -- --check` clean; `-p mlxcel-core drafter::dflash` 69 passed, `--lib models::lfm2` 29, `--lib server::batch::speculative_burst` 67, `--lib server::batch::dflash_target` 2, `--lib cli::speculative_args` 22, `--lib models::detection` 60 (the body says 58). Every count matches the recorded test plan.

A bilingual technical report is committed on the branch at `TECHNICAL_REPORTS/1751-lfm2-dspark-drafter-20260910.{en,ko}.md`.

Issue #1346, opened 2026-08-22
fix(server): the prompt cache donates and adopts KV-less shadow paged entries for model-owned families, so Gemma 3 and Llama 4 answer turn 2 without their history


inureyes:
## Summary

Under the default server configuration (`--parallel 4`, `--decode-storage-backend auto`, prompt cache on) a Gemma 3 or Llama 4 sequence is allocated on the paged backend purely for shadow block-table accounting; its real K/V lives in the model's `ModelOwnedSequenceState`. The prompt-cache donate path does not know that, detaches the shadow block table as a `DetachedKvSet::Paged` entry, and the next request that extends the conversation adopts it: the scheduler reports `cached=160/190` tokens, skips prefill for those 160 tokens, and the model, whose internal caches for the new sequence are fresh, decodes from the 30-token suffix alone. Reproduced on `gemma3-1b-4bit`: turn 2 answers "I really did cross the heliopause. It's a really fitting phrase..." with the cache on and "Voyager 1 crossed the heliopause in 2012. The Deep Space Network uses 70 meter antennas." with `--no-prompt-cache`. The fix gates donation, adoption and `detach_paged` on the model's natural sequence-state backend and refuses paged sets that carry no per-layer handles.

## Current behavior

Allocation: under the paged override a model-owned family gets backend `PagedKvCache` with `model.make_caches()` placeholders, which for Gemma 3 and Llama 4 is an empty vector (`src/models/gemma3.rs:1214-1216`, `src/models/llama4.rs:1713-1715`), plus a `PagedSequenceState` whose lengths are mirrored from the model-owned caches by `sync_paged_state_with_lengths` (`src/lib/mlxcel-core/src/cache.rs:6404-6440`, `:6745-6797`; the comment there says "For model-owned models (gemma3/llama4/qwen3_5) this is the authoritative length sync"). `pool.append_tokens` allocates real pool blocks for the shadow table, so `retained_block_count() > 0` and `seq_len() > 0`.

Donation (`src/server/batch/scheduler.rs:2976-3031`): the only model-owned exit is `supports_snapshot_reuse()` (Gemma 3, AFMoE and Llama 4 return the default `false`) followed by a check of the *allocated* backend, which is `PagedKvCache`, not `ModelOwned`:

```rust
if self.model.supports_snapshot_reuse() { ...snapshot...; return; }
let backend = self.cache_pool.get_mut(seq_id).map(|s| s.backend)
[...]

PR #1752, opened 2026-09-10
feat(core): share KV snapshot serialization across attention cache families

### LINKED ISSUE #1335: feat(core): shared snapshot serialization for KVCache / RotatingKVCache / ChunkedKVCache so Gemma 3, AFMoE and Llama 4 join exact-prefix prompt-cache reuse
## Summary

Prompt-cache reuse in the server has two producers: detached per-layer `KVCache` sets for families whose KV lives in the scheduler's `CachePool` (radix prefix matching, partial-block adoption), and exact-prefix `ModelStateSnapshot` copies for families that own their caches and opt in through `LanguageModel::supports_snapshot_reuse()`. The opt-in currently requires each family to hand-write per-cache-type serialization, and only Gemma 4 and Muse Glimmer have done so for the attention cache types. Gemma 3, AFMoE (which reuses Gemma 3's `Cache` enum) and Llama 4 hold ordinary `KVCache` / `RotatingKVCache` / `ChunkedKVCache` state in `ModelOwnedSequenceState` and are silently excluded from every prompt-cache path, so every multi-turn request on those families re-prefills the whole conversation. This change moves the KV-type serialization into one shared module and opts those three families in, with the same truncation rule Gemma 4 already uses.

## Current behavior

Donation is skipped for any model-owned family that does not declare snapshot reuse (`src/server/batch/scheduler.rs:2982-2996`):

```rust
if self.model.supports_snapshot_reuse() {
    self.insert_model_state_snapshot(seq_id, &ctx, tokens, SnapshotOrigin::Completion);
    return;
}
let backend = self.cache_pool.get_mut(seq_id).map(|s| s.backend)
    .unwrap_or(SequenceStateBackend::ModelOwned);
// Other `ModelOwned` families carry no detachable cross-request KV.
if backend == SequenceStateBackend::ModelOwned {
    return;
}
```

Lookup is gated the same way (`src/server/batch/scheduler.rs:2026-2036`): `supports_snapshot_reuse() == false` yields `SnapshotLookupOutcome::NoCandidate` before the store is consulted.

Families affected, with the cache types they own:

| Family | Layout | Cache types | Snapshot opt-in |
|---|---|---|---|
| gemma3 (`src/models/gemma3.rs:1222-1228`) | `SequenceStateLayout::model_owned`, batching on | `gemma3::Cache::{Standard(KVCache), Rotating(RotatingKVCache)}` (`src/models/gemma3.rs:644-647`) | none |
| afmoe (`src/models/afmoe.rs:114`, `:1149`) | model-owned, reuses `gemma3::Cache` | same | none |
| llama4 (`src/models/llama4.rs:38-44`, `:1660`, `:1721`) | model-owned | `Llama4Cache::{Chunked(ChunkedKVCache), Regular(KVCache)}` | none |

Gemma 4 already carries a complete, tested serializer for `Standard` and `Rotating` caches (`src/models/gemma4.rs:1196-1350`, truncation rule at `:1370-1440`), and Muse Glimmer carries a near-identical copy (`src/models/muse_glimmer_cache.rs:58-206`). The scalar tags `kv_cache_mode_to_i32` / `kv_cache_mode_from_i32` are private to `gemma4.rs:1139-1160`. No serializer exists for `ChunkedKVCache` anywhere (`grep -rn "chunked\." src/models/` is empty).

Cache types that are therefore not prefix-cacheable today, for the record:

- `KVCache` / `RotatingKVCache` inside a model-owned enum of gemma3, afmoe, gpt_oss, exaone4, ministral3, step3p5 (the last four hold them in `RefCell<Vec<Cache>>`, see `src/models/gpt_oss.rs:1107`, `src/models/exaone4.rs:685`, `src/models/ministral3.rs:645`, `src/models/step3p5.rs:1026`, and are single-stream).
- `ChunkedKVCache` (llama4 only).
- `KimiDeltaCache` (`src/models/kimi_linear.rs:395`), `RGLRUCache` (`src/models/recurrent_gemma.rs:127`), `Rwkv7Cache` (`src/models/rwkv7.rs:78`; has `snapshot_into` but the family never opts in). These three are blocked by #1220 (their state is rebuilt every forward) and stay out of scope here.
- `RingSlidingKVCache` (`src/lib/mlxcel-core/src/cache/ring.rs:54`, Unlimited-OCR only), by design.

`DetachedRotatingKVCache` (`src/lib/mlxcel-core/src/cache/detach.rs:621`) exists but has no caller outside its own module; the `CachePool` detach path only handles `Vec<KVCache>`.

## Expected behavior

1. A shared module `src/models/kv_snapshot.rs` exposes, for the three standard attention cache types, `snapshot_into(cache, snapshot, prefix) -> Result<(), String>`, `restore_from(cache, snapshot, prefix) -> Result<(), String>` and `snapshot_truncatable_to(snapshot, prefix, target_len) -> bool`, plus `kv_cache_mode_to_i32` / `kv_cache_mode_from_i32` moved out of `gemma4.rs`. Gemma 4 and Muse Glimmer call the shared functions; their behavior and tensor names do not change.
2. Serialization contract (all scalars stored through `recurrent_snapshot::push_i32`, all tensors copied through `ModelStateSnapshot::push_tensor`, which already deep-copies; only `KVCacheMode::Fp16` is accepted, any other mode returns `Err` so the donation is skipped, exactly as Gemma 4 does today):
   - `KVCache`: `{prefix}.standard.keys` `[B, H_kv, T_alloc, D]`, `{prefix}.standard.values` `[B, H_kv, T_alloc, D]`, `{prefix}.standard.offset` (i32, logical length; `T_alloc >= offset` because buffers grow by `step`), `{prefix}.standard.mode` (i32 tag). Restore sets `keys`, `values`, `offset`, `mode = Fp16`. Truncation to `target_len <= offset`: always allowed (restore then `KVCache::trim(offset - target_len)`).
   - `RotatingKVCache`: `{prefix}.rotating.keys`, `{prefix}.rotating.values` (physical ring buffers, `[B, H_kv, min(offset, max_size) or the over-window prefill length, D]`), `{prefix}.rotating.max_size`, `{prefix}.rotating.buffer_size`, `{prefix}.rotating.offset`, `{prefix}.rotating.start_position`, `{prefix}.rotating.idx`, `{prefix}.rotating.step`, `{prefix}.rotating.mode`, `{prefix}.rotating.turbo_seed`. Restore goes through `RotatingKVCache::restore_fp16_snapshot_state(RotatingKVCacheSnapshotState, keys, values)` (`src/lib/mlxcel-core/src/cache.rs:5260`). Truncation allowed only when `RotatingKVCacheSnapshotState::can_truncate_to(target_len)` (`cache.rs:4185`) holds, i.e. `buffer_size == 0 && idx == offset && offset <= max_size && target_len <= offset`: the ring has not wrapped so logical token `t` sits at slot `t`.
   - `ChunkedKVCache` (new): `{prefix}.chunked.keys`, `{prefix}.chunked.values` (`[B, H_kv, offset - start_position, D]` after `maybe_trim_front`), `{prefix}.chunked.chunk_size`, `{prefix}.chunked.offset`, `{prefix}.chunked.start_position`. Restore assigns the three scalars and the two tensors (the `step` field stays at its constructor default of 256; it only affects future growth). Truncation to `target_len` allowed only when `start_position == 0 && target_len <= offset`: once the front has been trimmed the window a fresh prefill would hold, `[target_len - chunk_size, target_len)`, is wider than `[start_position, target_len)`, so a truncated restore would attend over fewer tokens than a cold prefill and is not exact. When allowed, restore then slice `keys`/`values` to `[.., .., :target_len, ..]` and set `offset = target_len`.
3. Gemma 3 and AFMoE: `gemma3::Cache` gains `snapshot_into` / `restore_from` / `snapshot_truncatable_to` / `truncate_to` delegating to the shared module with the same `layer{idx}` prefix scheme Gemma 4 uses, and `Gemma3Model` / `AfmoeModel` implement `supports_snapshot_reuse() -> true`, `snapshot_sequence_state`, `restore_sequence_state`, `snapshot_truncatable_to`, `restore_sequence_state_truncated` following the Gemma 4 template at `src/models/gemma4.rs:5968-6070`. Family tags: `"gemma3"`, `"afmoe"`.
4. Llama 4: `Llama4Cache` gains the same four methods (`Chunked` arm uses the `chunked.*` contract, `Regular` arm the `standard.*` contract) and `Llama4Model` opts in with family tag `"llama4"`.
5. A second request whose token vector extends a finished request's `prompt + generated` vector on any of these families restores the snapshot and prefills only the appended tokens; the generated text is token-identical to a cold run at temperature 0.

## Implementation plan

1. Create `src/models/kv_snapshot.rs` (register in `src/models/mod.rs`). Move `kv_cache_mode_to_i32` / `kv_cache_mode_from_i32` from `src/models/gemma4.rs:1139-1160` into it (make the error message family-neutral). Add:
   - `pub(crate) fn snapshot_standard(cache: &KVCache, snapshot: &mut ModelStateSnapshot, prefix: &str) -> Result<(), String>` and `restore_standard(cache: &mut KVCache, snapshot, prefix) -> Result<(), String>`: lift the `Self::Standard` arms of `gemma4.rs:1204-1229` and `:1283-1308` verbatim.
   - `snapshot_rotating` / `restore_rotating`: lift the `Self::Rotating` arms of `gemma4.rs:1230-1273` and `:1309-1349`.
   - `snapshot_chunked(cache: &ChunkedKVCache, ...)` / `restore_chunked(cache: &mut ChunkedKVCache, ...)` per the contract above. `ChunkedKVCache.step` is private (`src/lib/mlxcel-core/src/cache.rs:5497`); leave it at the constructor value.
   - `standard_truncatable_to`, `rotating_truncatable_to` (lift `gemma4.rs:1370-1440`), `chunked_truncatable_to(snapshot, prefix, target_len) -> bool` returning `start_position == 0 && 0 <= target_len <= offset`.
   - `truncate_standard(cache, target_len)` (calls `KVCache::trim`), `truncate_rotating` (reuse the Gemma 4 `Cache::truncate_to` logic at the rotating arm), `truncate_chunked` (slice `keys`/`values` on axis 2 to `target_len`, set `offset = target_len`).
   - Use `recurrent_snapshot::{push_i32, push_optional, restore_i32, restore_optional}` (`src/models/recurrent_snapshot.rs:24-51`) for every field.
2. Replace the bodies of `gemma4::Cache::{snapshot_into, restore_from, snapshot_truncatable_to, truncate_to}` and `muse_glimmer_cache::MuseCache::{snapshot_into, restore_from, ...}` with calls into `kv_snapshot`. Keep the existing tensor names and error strings so `src/models/gemma4_tests.rs` and `src/models/muse_glimmer_tests.rs` snapshot tests pass unchanged.
3. `src/models/gemma3.rs`: add to `impl Cache` (`:644`) the four methods delegating to `kv_snapshot`. In `impl LanguageModel for Gemma3Model` (`:1210-1240`), add `supports_snapshot_reuse`, `snapshot_sequence_state` (iterate `self.sequence_state.with_sequence_state_ref(seq_id, ...)`, prefix `layer{idx}`, return `None` on any `Err` after a `tracing::warn!`), `restore_sequence_state` (check `snapshot.family() == "gemma3"`, build `self.model.make_caches()`, `restore_from` per layer, `self.sequence_state.replace_sequence_state(seq_id, state)`), `snapshot_truncatable_to` (all layers agree), `restore_sequence_state_truncated` (re-check, restore, then `truncate_to` per layer). Mirror `src/models/gemma4.rs:5968-6070`.
4. `src/models/afmoe.rs`: same five trait methods on `AfmoeModel` (`:1658-1690` region), family tag `"afmoe"`, using the `gemma3::Cache` methods from step 3.
5. `src/models/llama4.rs`: add `snapshot_into` / `restore_from` / `snapshot_truncatable_to` / `truncate_to` on `Llama4Cache` (`:38`), then the five trait methods on the model wrapper (`:1713-1740` region), family tag `"llama4"`. The VLM wrapper around Llama 4 (if it forwards `LanguageModel` calls) must forward the five new methods too; check `src/loaded_model.rs` dispatch macro at `:370` picks them up automatically (it forwards every trait method).
6. `src/server/prompt_cache/hybrid_ssm.rs`: no change; gemma3 / afmoe / llama4 are not in `HYBRID_SSM_MODEL_TYPES`, so APC block hashing stays enabled for them. The snapshot path does not use block hashes, so nothing else in `src/server/prompt_cache/` changes.
7. `docs/turbo-kv-cache.md`, section "Exact-prefix snapshots for recurrent state": list the snapshot-capable families and the per-type truncation rule (standard: always; rotating: unwrapped only; chunked: untrimmed only).

## Validation

(a) Unit tests, `src/models/kv_snapshot_tests.rs`:
- `standard_round_trip_preserves_keys_values_offset`: build a `KVCache`, push 37 tokens of random K/V, snapshot, restore into a fresh cache, assert `offset == 37` and `update_and_fetch` on one more token returns K/V whose first 37 positions are bitwise equal.
- `rotating_round_trip_preserves_ring_state_after_wrap`: `RotatingKVCache::new(16)`, push 40 tokens one at a time, snapshot, restore, assert `snapshot_state()` equals the original (`offset`, `idx`, `max_size`) and that the next `update_and_fetch` returns the same 16-token window as the original cache.
- `rotating_truncatable_only_while_unwrapped`: at 10 of 16 tokens `truncatable_to(7) == true`; at 40 tokens `== false`.
- `chunked_round_trip_after_front_trim`: `ChunkedKVCache::new(8)`, push 20 tokens, `maybe_trim_front`, snapshot, restore, assert `start_position == 12`, `offset == 20`, keys shape `[1, H, 8, D]`.
- `chunked_truncatable_only_before_front_trim`: at 6 of 8 tokens `truncatable_to(4) == true`; at 20 tokens `== false`.
- `non_fp16_mode_is_rejected`: a `KVCache::new_with_mode(KVCacheMode::Int8)` snapshot returns `Err`.
- `src/models/gemma3_tests.rs`: `snapshot_restore_matches_cold_decode`: run a 24-token prefill plus 8 decode steps through `Gemma3Model` with a synthetic config (`sliding_window = 8`, `sliding_window_pattern = 2`, 2 layers), snapshot at 32, restore, decode 4 more tokens, compare logits to a cold 36-token run at the same positions (allclose, atol 1e-3 in f16).
- `src/models/llama4_tests.rs`: same shape with `attention_chunk_size = 8`.

(b) Real checkpoint:

```bash
./target/release/mlxcel-server -m models/gemma-3-4b-it-4bit --prompt-cache --port 8080
# turn 1
curl -s localhost:8080/v1/chat/completions -d '{"model":"gemma-3-4b-it-4bit","messages":[{"role":"system","content":"<a 300-token system prompt>"},{"role":"user","content":"List three facts about Jupiter."}],"max_tokens":64,"temperature":0}'
# turn 2: same system prompt + turn-1 assistant text + a new user message
curl -s localhost:8080/v1/chat/completions -d '{... turn-1 messages + assistant reply + {"role":"user","content":"Now Saturn."} ...,"max_tokens":64,"temperature":0}'
curl -s localhost:8080/v1/cache/stats
```

Acceptance: `/v1/cache/stats` reports one snapshot hit for turn 2 with `matched_len` equal to the turn-1 `prompt + generated` length; turn-2 completion text is byte-identical to the same two requests served with `--prompt-cache` off; `prompt_tokens` time for turn 2 drops to the appended-tokens cost. Repeat with `mlx-community/Llama-4-Scout-17B-16E-Instruct-4bit` (chunked cache) and the AFMoE validation checkpoint `mlx-community/Trinity-Nano-Preview-4bit`. For Gemma 3 also run a third turn after the conversation exceeds `sliding_window` (1024) tokens and confirm the store still hits (exact restore) while `snapshot_truncatable_to` returns false for a shorter divergent prompt (a `SnapshotDiverged` reject, not a corrupt restore).

## Acceptance criteria

- [x] `src/models/kv_snapshot.rs` owns the standard / rotating / chunked serializers and truncation rules; `gemma4.rs` and `muse_glimmer_cache.rs` delegate to it with unchanged tensor names.
- [x] Gemma 3, AFMoE and Llama 4 return `supports_snapshot_reuse() == true` and implement snapshot / restore / truncatable / truncated-restore.
- [ ] Turn-2 snapshot hit on `mlx-community/gemma-3-4b-it-4bit` and `mlx-community/Llama-4-Scout-17B-16E-Instruct-4bit` with byte-identical output versus prompt cache off. Partly met in PR #1752: the turn-2 hit and exact restore are confirmed on both, including a wrapped-ring Gemma 3 turn over the sliding window, but comparing against `--no-prompt-cache` does not isolate the snapshot (the server defaults `preserve_thinking` differently and the cold prefill path itself differs), so the comparison was run with the cache on in both arms. Two gaps remain: `trinity-nano-preview-4bit` cannot be judged by token identity because it is a documented degenerate preview checkpoint, and Gemma 3 diverges reproducibly when the stored snapshot covers exactly one decoded token.
- [x] Non-Fp16 KV modes are refused at snapshot time with a warning, never restored with a mismatched layout.
- [x] `docs/turbo-kv-cache.md` lists the snapshot-capable families and per-type truncation rules.
- [x] cargo test --workspace --profile test-fast --features metal,accelerate passes
- [x] cargo clippy --workspace --all-targets -- -D warnings and cargo fmt --all -- --check pass

## Out of scope

- gpt_oss, exaone4, ministral3, step3p5: they keep caches in a `RefCell<Vec<Cache>>` that is not keyed by `SequenceId`; migrating them to `ModelOwnedSequenceState` is a separate refactor and a prerequisite for snapshot reuse.
- kimi_linear, recurrent_gemma, rwkv7: blocked by #1220.
- Block-level (radix / APC partial) reuse for rotating or chunked caches. Exact-prefix plus the unwrapped-only truncation rule is the correct contract for windowed caches.
- Snapshots of Int8 / Turbo KV modes (tracked separately in `prompt-cache-snapshot-non-fp16-kv-mode-gate`).

### LINKED ISSUE #1756: fix(prompt-cache): enforce require_whole_entry on the snapshot branch for multimodal requests
## Problem

`require_whole_entry` is enforced only on the KV branch of `try_adopt_cached_prefix` (`src/server/batch/scheduler/prompt_cache.rs:252`). The snapshot branch (`:126-230`) runs first and returns before the KV branch is reached, so it can hand back a truncated partial match for a multimodal request. `src/server/batch/scheduler/admission.rs:373-380` then drops the VLM embeddings once `prefill_start_offset > 0`, so a suffix that still holds image placeholder tokens is forwarded as raw token ids rather than as image embeddings.

This is pre-existing for the VLM wrappers that already declared snapshot reuse (`gemma4_vl`, `qwen3_5_vl` and similar), and it is gated behind `--enable-vlm-prefix-cache` (`vlm_prefix_sharing_allowed` at `src/server/batch/scheduler/mod.rs:207-213`), so it is off by default. PR #1752 widens the exposure by forwarding the five snapshot hooks through the shared `vision::VisionLanguageModel` (`src/vision/mod.rs:399-436`), which brings the Gemma 3 and Llama 4 VLM checkpoints onto the same path.

## Proposed solution

Apply the whole-entry requirement to the snapshot branch as well: when `require_whole_entry` is set and the snapshot lookup returns a truncated (partial) match rather than an exact full-entry match, decline and fall back to a cold prefill, recording the same `PromptCacheRejectReason::ModeMismatch` the KV branch records at `prompt_cache.rs:252-259`. The exact-match snapshot path stays available to multimodal requests, since a whole-entry match leaves no placeholder tokens in the suffix.

## Acceptance criteria

- [ ] With `--enable-vlm-prefix-cache` on, a multimodal request whose stored prefix is a proper prefix of the new request takes the cold-prefill fallback instead of adopting a truncated snapshot.
- [ ] A multimodal request that matches a stored entry exactly still adopts the snapshot, so the fix does not disable VLM prefix sharing outright.
- [ ] The reject is counted through the existing observability path, not swallowed silently.
- [ ] A regression test covers the multimodal proper-prefix case on the snapshot branch, mirroring the existing coverage for the KV branch.

## Verification

```bash
```

Context: PR #1752 and issue #1335. This is not a fix for #1335 itself.

inureyes:
## What this does

Moves the snapshot serialization for `KVCache`, `RotatingKVCache` and `ChunkedKVCache` into one shared module, `src/models/kv_snapshot.rs`, and opts Gemma 3, AFMoE and Llama 4 into exact-prefix prompt-cache reuse. Before this they held ordinary attention state in `ModelOwnedSequenceState` but declared no snapshot support, so every multi-turn request on them re-prefilled the whole conversation.

The module owns the truncation contract as well as the byte layout, because that is the part a per-family copy gets wrong. `KVCache` is always truncatable, since every token keeps its own slot. `RotatingKVCache` only while its ring is unwrapped, since a wrapped ring no longer holds logical token `t` at slot `t`. `ChunkedKVCache` only while its front is untrimmed, because a cold prefill of the shorter prompt would hold `[target_len - chunk_size, target_len)` while a front-trimmed buffer only offers `[start_position, offset)`. Exact-prefix restore is unaffected by all three; only the shorter-prefix variant declines.

A `KvSnapshotNames` parameter carries each family's tensor-name and error vocabulary, so Gemma 4 keeps writing `standard` / `rotating` and Muse Glimmer keeps writing `full` / `sliding`, and both still restore snapshots written by earlier builds. Only `KVCacheMode::Fp16` is accepted, at snapshot and at restore, and every restore path now checks the state it installs against the buffers it is handed: keys and values that disagree on rank, batch, heads or length, a declared length that does not fit its buffer, an invalid chunked `[start_position, offset)` range, and a rotating `max_size` that differs from the live cache are all refused rather than installed. `vision::VisionLanguageModel` forwards the five hooks to its text model, which is what lets the Gemma 3 and Llama 4 VLM checkpoints inherit the capability.

## Two defects fixed as causes

`RotatingKVCache`'s `CacheInterface::live_len` returned `seq_len()`, the physical buffer length, where the prefill mask has to be sized from `visible_len()`. `update_in_place` grows the ring by `step` (256) blocks during decode, so a cache that has generated even one token holds `physical > offset`, while `update_concat` concatenates only `visible_len()` prior keys. Sizing the mask from the physical length made it wider than the returned K/V and tripped `broadcast_shapes` on the first multi-token append after a decode, which is exactly a snapshot restore followed by an appended-token prefill. Live prefill-only sequences never hit it because concat leaves `physical == offset`. Guarded by `live_len_matches_the_keys_a_multi_token_append_returns`.

`restore_fp16_snapshot_state` rejected `idx > max_size` whenever `buffer_size == 0`, but `update_concat` stores more than one window after a prefill longer than `max_size` and pins `idx` to the stored length, so a snapshot captured between that prefill and the next single-token step legitimately carries `idx == physical_len > max_size`. The old check sent every over-window first turn down a cold prefill. The bound is now `idx <= physical_len`. Guarded by `rotating_round_trip_survives_an_over_window_prefill`.

## Scoped opt-outs

Gemma 3, AFMoE and Llama 4 answer `supports_padded_prefill()` with `false`, and Gemma 4's existing opt-out gains the same note. The reason is a missing hook, not a property of these models: the contract requires trimming the caches back to the real prompt length after an NA tile-aligned padded chunk, the scheduler honours it by trimming the `CachePool`'s `Vec<KVCache>`, and a `model_owned` family's pool entry has an empty `caches` vector. The trim reaches nothing, the pad positions stay in the model's own caches, and `offset` runs past the real token count (M5-only, gated by `should_align_prefill()`). The sequence-aware trim hook that would fix it properly is #1755.

`try_adopt_cached_prefix` enforced the multimodal whole-entry policy only on its KV branch, so a truncating snapshot restore was reachable for a multimodal request, and `admission.rs` drops the prepared VLM embeddings once `prefill_start_offset > 0`, leaving placeholder tokens in the suffix to forward as ordinary ids. That gap predates this branch: `vision::gemma4_vl` and `vision::gemma4_unified` have forwarded `snapshot_truncatable_to` since Gemma 4 first answered it, so this change widens the reach to `vision::VisionLanguageModel` rather than creating it. The gate now sits on both branches and declines before allocating a slot.

## Validation

Gate: `cargo fmt --all -- --check` clean, `cargo clippy --lib --tests --features metal,accelerate` and `cargo clippy -p mlxcel-core --lib` clean, and `models::kv_snapshot` (16), `models::gemma3` (25), `models::llama4` (5), `models::afmoe` (26), `models::gemma4_tests` (10) and `models::muse_glimmer` (13) pass. The 14 model-level snapshot tests carrying `#[ignore = "requires serial MLX execution"]` were each run alone by exact name; all pass.

Real checkpoints, Apple M5 Max, `mlxcel-server` on port 19335, prompt cache on, greedy (`temperature 0`, `seed 0`), `preserve_thinking` pinned in every arm.

Prompt cache on against `--no-prompt-cache` does not isolate the snapshot, so it is not the test used here. With the cache on, `capture_history_boundary_snapshot` (#1143) splits even a cold first turn at the conversation boundary: Gemma 3's 1323-token turn 1 is forwarded as 1320 then 3 where the cache-off arm forwards 512/512/299, and that width change reproducibly flips one near-tied token (a straight versus curly apostrophe) in an 89-token reply. Two controls pin the cause: reruns of the same arm in separate processes are bit-identical, and `models/qwen2.5-7b-instruct-4bit`, which has no snapshot support and so takes no boundary split, is byte-identical across the flag. The comparison below therefore holds the cache on in both arms and varies only whether the turn is served from a restored snapshot or computed cold in a fresh process.

Gemma 3 (`models/gemma-3-4b-it-4bit`, rotating 1024 plus standard). Turn 1 is 1323 prompt tokens over a 1024 window, the over-window shape the restore-bound fix admits, and generates 89. The completion donate stores `token_len=1412`, exactly turn-1 prompt plus generated, which the warm-up extends to the turn-2 boundary at 1417. Turn 2 adopts `restored 1417/1435, stored=1417, partial=false`; turn 3 adopts `restored 1547/1567, partial=false` with the ring long since wrapped. Against cold, turn 2 is byte-identical over 104 of 104 tokens, 103 with bitwise-equal logprobs and a minimum top-2 gap of 0.25, and turn 3 is byte-identical over 18 of 18 with every logprob bitwise equal.

Llama 4 Scout (`models/llama-4-scout-17b-16e-instruct-4bit`, chunked plus standard). Turn 1 is 1305 prompt tokens and 512 generated; the completion entry is `1817 = 1305 + 512`, extended to 1822. Turn 2 adopts `restored 1822/1840, partial=false` and is byte-identical to cold over all 512 tokens, 338 with bitwise-equal logprobs and a minimum top-2 gap of 0.0, so it agreed even at exactly tied positions.

AFMoE (`models/trinity-nano-preview-4bit`). Turn 1 is 1281 prompt tokens and 512 generated, and the completion entry is `1793 = 1281 + 512`. Turn 2 adopts `restored 1280/1755, stored=1793, partial=true`: the reply does not re-tokenize as a prefix of the stored tail, so the store truncates to the end of turn 1's prompt, exercising the truncation predicate on real weights. Output identity is not a usable signal for this checkpoint, which `docs/supported-models.md` records as emitting degenerate repetitive text under the mlx-lm reference as well as here, and both arms do exactly that. Its snapshot machinery is covered by unit tests instead, including a non-ignored `snapshot_restore_matches_cold_decode`.

The contract also declines correctly on real weights: replaying Gemma 3's 1323-token prompt finds the 1417-token entry, refuses to truncate into a wrapped ring, and rejects with `snapshot_diverged (context_len=1323, entry_len=1417)` instead of restoring a wrong window.

## Not fixed here

Replaying an identical prompt inside one process diverges from the first reply on every family tested, `models/gemma-4-12b-it-4bit` included, which has shipped this path since before this branch and is untouched by it. The adopt covers the whole prompt and `admission.rs` backs the prefill cursor off one token without rewinding the restored cache. Pre-existing and filed as #1760, with the Gemma 4 control as the evidence that this branch does not cause it.

The snapshot bucket is still sized by the fixed 512 MiB default for these three families, because `snapshot_family_is_model_aware` lists only recurrent ones, whose state is O(1) in context where an attention cache is O(context). Llama 4 Scout stores about 192 KiB per token, so a conversation of a few thousand tokens is rejected `Oversized` and the feature silently stops working. The Llama 4 entries measured above were 402,653,712 bytes for the 1817-token turn and 503,317,008 for the 2352-token one, which is 384 and 480 MiB of step-aligned buffer, so this two-turn validation sits just under the cap and a third turn would not fit. Filed as #1761, deliberately outside this diff because it changes a default memory policy and needs its own measurement.

#1754 claims completion-origin snapshots claim one more token than their caches hold. Llama 4's turn 1 above finished on `length`, the finish class that issue names, and the restored turn 2 was still byte-identical to cold, so the premise needs confirming before the donate arithmetic changes; noted on the issue.

Closes #1335, closes #1756


inureyes:
## Implementation review

### Root cause of the "What is not validated" Gemma 3 divergence

It is a pre-existing scheduler defect this PR exposes, not one it introduces. `donate_finished_sequence_cache` donates `tokens = prompt_tokens ++ generated_tokens` with `token_len = tokens.len()` (`src/server/batch/scheduler/prompt_cache.rs:1112-1149`), but the decode loop forwards the *previous* sampled token (`decode_tick.rs:501`, `:744`, `:1273-1293`), so the final sampled token never gets a forward pass and the model-owned caches hold `tokens.len() - 1` tokens of KV. The mismatch is conditional on the finish reason: a merged-EOS `Stop` returns before pushing the token (`decode_tick.rs:1448-1456`) and the eos-at-prefill donate passes `&[]` (`prefill.rs:1418`), so both are exact, while `Length`, `StopSequence`, `RepetitionLoop`, `Cancelled` and the finished-during-prefill donate (`prefill.rs:1443`, `:1461-1470`, `:1515`) push first and stop after, leaving `offset == tokens.len() - 1`.

The restore side resumes from `matched_len`, not from the restored offset: `try_adopt_cached_prefix` returns `matched_len` (`prompt_cache.rs:214`), `admission.rs:341-344` assigns it to `prefill_start_offset`, and `prefill.rs:727` forwards `prompt_tokens[prefill_start_offset..]`, while `restore_standard` installs the stored `.offset` scalar (`src/models/kv_snapshot.rs:178`). An exact-full-entry restore of such an entry therefore drops the last generated token's KV and puts every suffix token one RoPE position low, which is exactly the P prompt plus 1 decoded token shape at restore length 333.

The other two insert sites are exact by construction, which is why 334, 335, 339 and 344 are clean: the warm-up forwards `tokens[matched_len..]` (`prompt_cache.rs:838-881`) and the boundary capture forwards `prompt_tokens[start..boundary]` (`prompt_cache.rs:1002-1053`). Truncating restores are safe too, since `check_truncate_target` and `standard_truncatable_to` both demand `target_len <= offset` (`kv_snapshot.rs:214`, `:628-640`) and the store consults the truncate predicate only when no exact-prefix candidate exists (`src/server/prompt_cache/store.rs:1252-1273`).

No family compensates, and none ever did: `origin/main`'s `gemma4.rs:1221` and `:1314-1315` stored and restored `.offset` identically, so Gemma 4, Muse Glimmer and the recurrent families have carried this since exact-prefix reuse landed, and this diff touches no file under `src/server/`. Left unfixed here deliberately: the repair belongs in the scheduler, changes behavior for every snapshot family, and needs real-checkpoint validation. Worth its own issue alongside the sequence-aware trim hook the PR body already names.

### Other findings

- MEDIUM. `require_whole_entry` is enforced only on the KV branch of `try_adopt_cached_prefix` (`prompt_cache.rs:252`); the snapshot branch above it (`:126-230`) can return a truncated partial match for a multimodal request, and `admission.rs:373-380` drops the VLM embeddings once `prefill_start_offset > 0`, so a suffix still holding image placeholders would forward as raw ids. Pre-existing for the VLM wrappers that already opted in, gated behind `--enable-vlm-prefix-cache` (`scheduler/mod.rs:207-213`), widened here to `vision::VisionLanguageModel`.
- MEDIUM. `src/models/gemma3.rs:1887` and `:1905` still document and assert that `live_len()` returns `seq_len()`, which this PR changed to `visible_len()`.
- LOW. Muse Glimmer's mode-error strings are not preserved: the removed `validate_fp16_mode` said "Muse Glimmer restore {prefix}.full.mode: snapshot cache mode {n} is not supported", the shared path says "Muse Glimmer restore {prefix}: full snapshot mode {mode:?} is not supported", and an unknown tag now reports "unknown serialized KV cache mode tag {n}" (`recurrent_snapshot.rs:99`). Tensor names are byte-identical, which is the part that matters for restoring older snapshots.
- LOW. `restore_standard` no longer force-sets `cache.mode = Fp16` the way Muse Glimmer's copy did; it refuses a non-Fp16 live cache instead. Safer and consistent with the issue's fourth criterion, but a behavior change.
- LOW. `supports_padded_prefill() == false` costs only the M5 single-sequence NA tile alignment for these three families, not batched cold prefill: all four inherit `supports_batched_prefill() == false` (`generate.rs:704-706`), so `plan_prefill_cohorts`'s equal-length restriction never reaches them.

### Verified

Shared module factoring, `KvSnapshotNames` name preservation, the five hooks across the three families, the three truncation predicates, both core fixes and the `src/vision/mod.rs` forwarding all check out. `live_len()` returning `visible_len()` is right because `update_concat` keeps `min(visible_len, max_size - 1)` prior keys (`cache.rs:4686-4740`) and `create_sliding_window_prefill_mask` clamps to the same value (`utils.rs:840-853`). The `idx <= physical_len` bound equals the old check whenever the ring has wrapped and is looser only for the over-window shape it is meant to admit. The padded-prefill reasoning holds: the trim iterates `CachePool::get_caches_mut` (`cache.rs:6558-6560`), empty for a `model_owned` set.

Gate re-run locally: fmt and the narrow clippy clean; `models::kv_snapshot` 16, `gemma4_tests` 10, `muse_glimmer` 13, `gemma3` 25, `llama4` 5, `afmoe` 26 pass, plus the four ignored Gemma 3 snapshot tests individually by exact name.

### Acceptance criteria

All met except the turn-2 byte-identity criterion, which the PR body correctly reports it could not run as written and substituted with a warm-versus-cold comparison, and which still fails at the one-decoded-token Gemma 3 boundary for the reason above. No changes made in this pass.

inureyes:

Pushed 983b8bae and e46eae1e.

### Fixed

- HIGH, `src/server/batch/scheduler_model_owned_cache_tests.rs`. `model_owned_paged_family_never_donates_or_adopts` fails on this branch. I confirmed it against the pristine branch HEAD with my own changes reverted, so it is the PR's own regression, not mine: the test asserts `store.len() == 0` after a Gemma 3 request finishes, and Gemma 3 now donates one snapshot entry, which `len()` and `PromptCacheStats::entries` both count. Rewritten to assert #1346's actual guarantee on the K/V counters (`entries - snapshot_entries`, `inserts`, `lookups`); turn 2 now skips the shared prefix through the snapshot restore, which is what #1335 exists to produce.
- MEDIUM, `src/models/kv_snapshot.rs` `restore_standard` and `restore_chunked`. Both assigned a snapshot's `offset`, `start_position`, keys and values into a live cache with no agreement check, while the sliding-window path validates that class through `RotatingKVCache::restore_fp16_snapshot_state`. A declared length running past its own buffer installs cleanly, and the next append then slices past the sequence axis (`KVCache::update_fp16` normalizes to `buffer_idx()` before it grows), reaching the caller as an MLX throw at the FFI boundary rather than as a declined restore. Now refused: keys and values disagreeing on rank, batch, heads or length; a length that does not fit the buffer; an invalid chunked `[start_position, offset)`; and a rotating `max_size` that differs from the live cache, since the window width is model configuration and `restore_chunked` already refused the same for `chunk_size`.
- LOW, `src/models/kv_snapshot.rs` `truncate_chunked`. It installed the key slice before building the value slice, so a failure on the second left a shortened key buffer beside a full-length one and a stale `offset`. Both slices are built before either is installed. A negative truncate target is also refused by name instead of falling through to a trim-count mismatch.
- Comment accuracy, `src/server/batch/scheduler/prompt_cache.rs`. The whole-entry gate was not first reachable through `vision::VisionLanguageModel`: `vision::gemma4_vl` and `vision::gemma4_unified` have forwarded `snapshot_truncatable_to` since Gemma 4 first answered it, so a multimodal partial snapshot restore could already bypass the rule on main. This PR widens the reach rather than creating it, and the comments now say so.

### Found, not fixed

- HIGH performance. `snapshot_family_is_model_aware` (`src/server/prompt_cache/snapshot_sizing.rs`) does not list gemma3, llama4 or afmoe, and none of them classify as `Hybrid` or `PureSsm`, so all three keep the fixed 512 MiB `DEFAULT_SNAPSHOT_CAPACITY_BYTES`. Llama 4 Scout stores 192 KiB per token across 48 layers (8 KV heads, head dim 128, fp16, K and V), so every conversation past 2731 tokens is rejected `Oversized` and the feature silently does nothing; the 417-token real-checkpoint validation in the PR body sits below that line. Gemma 3 4B is 121.6 MB of sliding window plus 20 KiB per token, so 289 MB at 8192 tokens, and one conversation stores two entries (boundary and completion), which already overflows the bucket. Separately, the entries that list does have never match: `model_type()` prefers `text_config.model_type`, and the checkpoints in `models/` report `gemma4_text`, `gemma4_unified_text` and `muse_glimmer_text`, none of which equal `gemma4` or `muse_glimmer`. This deserves its own issue: it changes a default memory policy and should be measured on real checkpoints rather than landed inside this PR.

### Checked, no finding

- No CRITICAL. Snapshots never leave the process: `ModelStateSnapshot` lives only in the in-memory store and there is no disk persistence, so every tensor shape and scalar is model-produced rather than attacker-supplied. The only attacker-controlled input is the token vector, which selects `matched_len`; that is floored at `min_prefix_tokens` and re-checked by `restore_sequence_state_truncated` before anything is installed.
- No half-initialized state on a restore failure. Each family builds a local `Vec<Cache>` and calls `replace_sequence_state` only after every layer succeeds, and the scheduler calls `release_sequence_caches(seq_id)` on the error path, which releases the model-owned slot as well as the pool entry. The `require_whole_entry` decline correctly returns before `allocate_sequence_state`, so it allocates nothing.
- No hidden GPU cost on the capture or restore hot path. Measured here: reading a snapshot scalar costs 17.3 us the first time, because `ffi::copy` is lazy and `item_i32` evaluates it, and 0.2 us on every read after that. A 34-layer Gemma 3 restore therefore pays about 3 ms once per stored entry, against a saved prefill of hundreds of ms. Not worth changing.
- `CacheInterface::live_len` moving from `seq_len()` to `visible_len()` is correct. The value is used only to size prefill masks, and `create_sliding_window_prefill_mask` clamps its kept-prior count to `min(sliding_offset, window - 1)`, which is exactly `update_concat`'s `kept_len`.

### Gate

`cargo fmt --all -- --check`, `cargo clippy --lib --tests --features metal,accelerate -- -D warnings` and `cargo clippy -p mlxcel-core --lib` are clean. `models::` (1662), `server::` (2883), `vision::` (493) and `mlxcel-core cache::` (533) pass, as do the fourteen `#[ignore]`d model-level snapshot tests run one at a time.

The push moved the remote branch from 4c1c3b3f to e46eae1e, so adcecb3b (`fix(server): apply the whole-entry policy to the snapshot adopt branch`) had not reached GitHub before now and is included.

PR #1772, opened 2026-09-10
fix(mlx): bump MLX pin to 81ba1c6a so mxfp8 block scales round up

### LINKED ISSUE #1769: test(fp8_block): fp8_block_requantize_matches_direct_path fails deterministically on M5 Max with macOS 27 / Xcode 27
## Problem / Background

The CI-faithful gate on main `58dc9693`, `cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1`, reports 10981 passed, 3 failed, 359 ignored. One of the three is `models::fp8_block::fp8_block_tests::fp8_block_requantize_matches_direct_path`, panicking at `src/models/fp8_block_tests.rs:259:9` with `element 4: mxfp8 error 0.50390625 exceeded 0.3002931 (group max 4.8046875)`. Re-running that single test three times in a row reproduces the identical message every time, so it is deterministic, not a flake.

Host: M5 Max (137 GB, 18 cores), macOS 27.0 (26A428), Xcode 27.0 (27A266a), Metal 32023.921, `air64-apple-darwin27.0.0`. The machine was upgraded to macOS 27 / Xcode 27 on 2026-09-10, after #1742 merged. GitHub CI does not run the Metal test suite, so this test has never been exercised on Apple GPU generation 17 in CI.

## Current Behavior

`fp8_block_requantize_matches_direct_path` at `src/models/fp8_block_tests.rs:165` builds a seeded fixture, `block_fp8_fixture("model.layers.0.mlp.down_proj", 130, 160, 0x51D3_9E11)` at `:170`, whose input bytes and bf16 block scales are therefore identical on every host. It then checks three things in order: the packed mxfp8 plane is byte-identical to `quantize_weights_with_mode(from_slice_f32(expanded), 32, 8, "mxfp8")` (`:224`), the scale plane likewise (`:229`), and the dequantized round trip lands within a per-group error bound (`:252-262`).

Only the third check fails. Both byte-identity assertions pass, so the requantize path (`requantize_block_fp8_weights` at `src/models/fp8_block.rs:346`, which calls `reconstruct_block_fp8_tensor` at `:244` and then `quantize_weights_with_mode` at `:398`) still agrees exactly with the direct path. What fails is the accuracy bound on MLX's own mxfp8 quantize/dequantize round trip, which the test name does not describe.

The bound is `let bound = group_max / 16.0 + f32::EPSILON;` at `:257`, justified by the comment at `:235-237`: "E4M3 carries four significant bits, so a value sharing its group's binade with the group maximum rounds by at most 2^-4 of that maximum". That is a round-to-nearest half-ulp assumption.

## Diagnosis so far

The reported numbers are internally consistent: `4.8046875 / 16 = 0.30029296875`, plus `f32::EPSILON` gives `0.3002931`, matching the message exactly, so the formula and the fixture are not in question.

The observed error over the group max is `0.50390625 / 4.8046875 = 0.10488`, which sits between half an E4M3 ulp (`2^-4 = 0.0625`) and one full ulp (`2^-3 = 0.125`). One full ulp, `group_max / 8 = 0.6005859375`, covers the observed error; the failure margin against the current bound is 1.678x. That is the signature of a quantizer that rounds toward zero rather than to nearest, not of a broken reconstruction.

Element index 4 lies in row 0 of a 130x160 tensor, so it is not in one of the padded trailing blocks that the fixture shape was chosen to exercise.

## Proposed Solution

**Decision criterion, run first.** Execute the same test on an M1 Ultra (Apple GPU generation 13). The fixture is seeded, so both hosts consume identical input. If M1 Ultra passes and M5 Max fails, this is a backend or toolchain divergence and the investigation belongs in the reconstruct/requantize path and the pinned MLX commit. If both fail with the same numbers, the bound never held for MLX's rounding and the test has simply not been run on this arithmetic before.

In the second case, re-derive the bound to one full E4M3 ulp (`group_max / 8`) and rewrite the comment at `:235-237` to state which rounding mode MLX's `fp_quantize` implements and why one ulp is the right ceiling. Do not widen the constant until it goes green without recording that derivation; the comment is the only thing that keeps this assertion meaningful.

## Scope

**In scope:** `src/models/fp8_block_tests.rs` (the bound at `:257` and the derivation comment at `:235-237`), `src/models/fp8_block.rs` only if the M1 Ultra comparison shows the reconstruction is at fault, and a `docs/benchmarks.md` note if a gen-13 vs gen-17 difference is confirmed.

**Out of scope:** the other two failures in the same gate run, which need separate issues.

## Implementation Notes

- **Reuse**: the existing `mlxcel_core::quantize_weights_with_mode` and `mlxcel_core::dequantize` bridge declarations at `src/lib/mlxcel-core/src/lib.rs:2739` and `:1145`. Do not add a host-side mxfp8 emulator to sidestep the comparison.
- **Constraint**: keep the fixture seed `0x51D3_9E11` and the 130x160 shape. Changing either hides the failing draw rather than explaining it, and the shape is chosen so both axes need padding.
- **Sibling test**: `fp8_block_requantize_matches_direct_path_when_block_aligned` at `:270` asserts byte identity only and passes. Keep the two tests' claims distinct; the round-trip accuracy assertion may belong in its own test with an accurate name.
- **Precedent**: mxfp4 on M5 Max already diverges from the graph reference at generation 17 (`docs/benchmarks.md:904` and the CLAUDE.md "Changes that move the numbers" rules), so an E4M3 group-scaled path missing a half-ulp bound on gen-17 hardware is not a novel class of result.
- **Edge case**: dump the element value, the group's shared E8M0 exponent and the E4M3 code for group 0 element 4 before changing anything, and compare the decoded value against both round-to-nearest and round-toward-zero. That single print distinguishes the two candidate causes without a second host.

## Acceptance Criteria

- [ ] The root cause is named in the PR body as either a requantize-path defect or a bound-derivation defect, with the M5 Max and M1 Ultra results that decided it.
- [ ] `cargo test --profile test-fast --features metal,accelerate -p mlxcel models::fp8_block` passes on an M5 Max (gen 17) and on an M1 Ultra (gen 13).
- [ ] The workspace gate reports no `fp8_block` failure on both hosts, and the failure count drops from 3 to 2 (green once the other two are resolved separately).
- [ ] If the bound moved, the comment at `src/models/fp8_block_tests.rs:235-237` states the new derivation and the rounding mode it assumes.
- [ ] The fixture seed and shape are unchanged.
- [ ] The PR body records the measured maximum error and its ratio to the group max on both hosts.

## Verification

```bash
cargo test --profile test-fast --features metal,accelerate -p mlxcel models::fp8_block -- --test-threads=1 --nocapture
cargo test --workspace --profile test-fast --features metal,accelerate --no-fail-fast -- --test-threads=1
cargo fmt --check
```

A pass is both `fp8_block` tests green on both hosts with the recorded error ratios attached, and the workspace run no longer listing this test among its failures.

No vendor fine-grained FP8 checkpoint is present in `models/` as of 2026-09-11 (227 entries, none matching `fp8`), so the measurement vehicle is the test's synthetic fixture. If a vendor FP8 checkpoint is added to the store, additionally run `examples/logit_trace` through the FP8 loader per `docs/benchmarks.md:407` "Judging a change that moves the numbers" and compare the arms with `scripts/compare_logit_traces.py`, reporting disagreement at decided positions rather than byte-identity.

## Technical Considerations

Follow-up of #1742; surfaced by the post-merge workspace gate after the merges of 2026-09-10. `src/models/fp8_block.rs` and `src/models/fp8_block_tests.rs` were introduced by #1742 (`6ad3febc`, "feat(qwen3_5): load vendor fine-grained FP8 checkpoints") earlier on 2026-09-10, and none of the twelve PRs merged after it touch `fp8_block*`, so the change under suspicion is the toolchain upgrade rather than in-tree code.

inureyes:
## Why

The fp8 round-trip bound failed on every Metal host, M1 Ultra byte-identically to M5 Max, because the pinned MLX `9a795735` predates ml-explore/mlx#4353: Metal and CPU encoded the mxfp8 E8M0 block scale as `round(log2(amax / 448))`, so about half the blocks saturated their maxima, losing up to `1 - 2^-1/2`. CUDA rounds up, which is why #1742 passed on GB10. The test was right, and `requantize_block_fp8_weights`, the only E8M0 quantize caller, was clipping vendor FP8 checkpoints on Metal. Widening the bound, as the issue proposed, was rejected.

## What changed

- MLX pin `9a795735` to upstream main `81ba1c6a` (99 commits). The seven overlays whose targets upstream touched are three-way merged and keep their deltas; the other 21 are unchanged upstream. `metal/compiled.cpp` also drops a leftover `elem_to_loc_1<uint>` that undid part of ml-explore/mlx#3720, so its only delta is the mixed-dtype cast.
- Adaptations to what the new pin changes under the bridge:
  - `gather_qmm` gained `global_scale` ahead of `sorted_indices` (ml-explore/mlx#4458), so all 13 calls pass `std::nullopt`.
  - CUDA now needs cuSOLVER (ml-explore/mlx#4208). `link_cuda()` names it, `docs/installation.md` lists it, and CI's link job now runs on pin and CUDA link-list changes, which is how this slipped past CI.
  - `array::is_available()` now throws and clears a failed launch's error (ml-explore/mlx#3742). The rejection sampler's drain now reads status, signal and error pointer instead, so a GPU fault no longer terminates the process from inside `fused_sample` or hides the error from its owner. A regression test aborts the process with the old drain and passes with the new one.
- The round-trip check moves into its own test, `fp8_block_requantize_round_trip_stays_within_half_an_e4m3_step`, with the same seed and shape. It states the derivation and asserts `group_max <= 448 * scale` per block.

## Validation (M1 Ultra, macOS 27.0)

- fp8: the old pin fails the new test at block 0 (the maximum 4.8046875 scales to 615 and saturates). The new pin saturates 0 of 650 blocks, with a worst error of 0.0489 of the group max against 0.2928 before.
- Turbo launchers pass (max RMS 1.7263e-4 and 1.5259e-4).
- Teacher-forced logit traces, old pin vs new, on five checkpoints at widths 1, 8 and 256: 0 disagreements on decided positions. Over 4,096 positions on qwen3-30b-a3b, 1 of 1,720 decided positions differs, at the reference's rank 2, with perplexity -0.20%.
- Branches the short runs missed:
  - head-dim-512 decode past 1,024 keys: identical text.
  - GQA-8 decode past 8,192 keys: identical text, decode 55.6 to 60.8 tok/s.
  - head-dim-72 vision towers: image prefill 239 to 279 tok/s. Descriptions diverge into equally faithful text; decided answers are unchanged.
- Short-context throughput is within 0.6% on three checkpoints.
- Workspace gate: 10986 passed, 0 failed, 359 ignored. Clippy, fmt, both pin parsers and the cross-repo reference check are clean.
- CUDA was compiled and linked in CI but not run: `OpenXLA feature link` linked a `--features cuda,xla-iree` release binary on GB10 at the new pin, which covers the overlays and the cuSOLVER link. The green `CUDA sm_70 compile` check is a skip, because CUDA 13.0 cannot target sm_70.

The per-checkpoint numbers, method and derivations are in `TECHNICAL_REPORTS/1772-mlx-pin-mxfp8-round-up-20260911.en.md`.

## Not validated, or reported and not fixed

- No CUDA execution. No M5 Max run: the generation-17 NAX paths changed upstream in this range are unmeasured.
- `array_evaluated_bytes`, the server's lookahead read, is another non-`Result` bridge function that now throws on a failed launch. It needs routing through the scheduler's step-failure path.
- The mixed-dtype cast in `metal/compiled.cpp` would cast a comparison's inputs to `bool`. This is latent, since no compiled function contains a comparison.

Closes #1769

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,11 @@ jobs:
               - 'scripts/iree/**'
               - 'rust-toolchain.toml'
               - '.github/workflows/ci.yml'
+              # The CUDA half of the same link line: mlxcel-core's build
+              # script names the CUDA libraries, and the MLX pin decides which
+              # of them libmlx.a needs. See the `xla-link` job comment.
+              - 'src/lib/mlxcel-core/build.rs'
+              - 'src/lib/mlx-cpp/CMakeLists.txt'
             # Everything that can compile differently per CUDA architecture: the
             # MLX overlays and turbo kernels, the pinned MLX commit, and the
             # build scripts that hand CMake the architecture list. Rust sources
@@ -519,6 +524,15 @@ jobs:
   # #1274 (found by hand later, not by CI). Running on every Rust PR was
   # rejected because the release link is minutes of shared-runner time and the
   # overwhelming majority of Rust PRs cannot affect the link line.
+  #
+  # `src/lib/mlxcel-core/build.rs` and `src/lib/mlx-cpp/CMakeLists.txt` are on
+  # the causal path as well. The former lists the CUDA libraries every
+  # `--features cuda` link resolves `libmlx.a` against; the latter holds the MLX
+  # pin, and a pin bump can add one without touching any build script. That is
+  # how ml-explore/mlx#4208's cuSOLVER dependency arrived with the 81ba1c6a bump
+  # (#1769): MLX links it PRIVATE, `cargo check` never links, and no job on the
+  # PR linked CUDA. A pin bump invalidates `mlxcel-core`'s MLX build, so those
+  # runs pay the cold figure below rather than the warm one.
   #
   # Measured cost of this job's link command on the GB10 runner: 15m40s from a
   # purged `CARGO_TARGET_DIR`, most of which is MLX's CUDA sources compiling
--- a/TECHNICAL_REPORTS/1772-mlx-pin-mxfp8-round-up-20260911.en.md
+++ b/TECHNICAL_REPORTS/1772-mlx-pin-mxfp8-round-up-20260911.en.md
@@ -1,0 +1,199 @@
+# Technical Report: PR #1772 - fix(mlx): bump MLX pin to 81ba1c6a so mxfp8 block scales round up
+
+**Date**: 2026-09-11
+**Author**: mlxcel maintainers
+**Reviewer**: implementation review cycle (implementation review, security and performance review, finalization)
+**Status**: Completed on Metal (M1 Ultra). CUDA compiled and linked in CI (GB10, sm_121); no CUDA test or inference ran. M5 Max not measured.
+**Languages**: C++, Rust, CMake, YAML
+**Risk Level**: Medium-High (every MLX-owned kernel and host path moves by 99 upstream commits; measured on one GPU generation)
+
+---
+
+## Executive Summary
+
+Issue #1769 reported that `fp8_block_requantize_matches_direct_path` failed deterministically on an M5 Max after its macOS 27 upgrade, and suspected the toolchain or the GPU generation. Running the issue's own decision criterion on an M1 Ultra, also on macOS 27, produced the byte-identical failure message. The test had simply never run on Metal: #1742 was validated on Linux/CUDA only.
+
+The cause is in MLX, not in mlxcel's requantize path and not in the test. At the pinned commit `9a795735`, the Metal and CPU backends choose the mxfp8 E8M0 block exponent as `round(log2(amax / 448))`. Rounded to nearest in log2 space, the scale lands below `amax / 448` for about half the blocks, those blocks' largest values scale past E4M3's 448 ceiling, and they saturate, losing up to `1 - 2^-1/2` (29.3%) of the block maximum. CUDA rounds the exponent up, which is why the test passed where it was written. Upstream fixed Metal and CPU in ml-explore/mlx#4353, six days after the pin.
+
+So the test's bound was correct and the quantizer was wrong. The same quantizer runs in production: `requantize_block_fp8_weights` is the only E8M0 quantize caller in the tree, and every vendor FP8 checkpoint loaded on Metal went through it.
+
+The maintainer chose to move the pin to upstream main `81ba1c6a` rather than patch around it. That brought 98 further commits, and three of them change behaviour underneath the bridge in ways that compile or pass tests while being wrong: a parameter inserted mid-signature, a new link dependency, and an accessor that started throwing. All three are handled here. The first surfaced at compile time; review found the other two.
+
+---
+
+## 1. The failure, and why the issue's hypothesis did not hold
+
+The issue asked for one experiment before any change: run the same seeded test on generation-13 hardware. The result on M1 Ultra:
+
+```
+element 4: mxfp8 error 0.50390625 exceeded 0.3002931 (group max 4.8046875)
+```
+
+That is the M5 Max message character for character. Both hosts ran macOS 27 and Xcode 27, so the experiment ruled out a generation difference but could not by itself rule out a toolchain one. What ruled that out was the history: #1742's validation section lists only CUDA runs, so there was never a macOS 26 pass to regress from.
+
+The mechanism reads directly off the pinned kernel (`mlx/backend/metal/kernels/fp8.h`, `struct fp8_e8m0`):
+
+```cpp
+float le = metal::log2(x);
+int n = int(metal::round(le));
+```
+
+For block 0 of the fixture, `log2(4.8046875 / 448) = -6.54` rounds to `-7`. The block maximum then scales to `4.8046875 * 128 = 615`, and E4M3's conversion saturates anything at or above 448. Element 4, `4.00390625`, scales to 512.5, saturates, and decodes as `448 / 128 = 3.5`, an error of 0.50390625. A host emulation of the pinned kernel over the whole seeded fixture reproduces that first failure exactly and gives the full picture: 301 of 650 blocks saturate, with a worst loss of 0.2928 of the group maximum. The theoretical ceiling for round-to-nearest in log2 space is `1 - 2^-1/2 = 0.2929`.
+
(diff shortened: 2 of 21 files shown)


inureyes:
## PR finalization

### Test coverage
Added a regression test for the deferred drain's Failed branch: `a_failed_stashed_launch_is_dropped_without_throwing_or_consuming_its_error` in `src/lib/mlxcel-core/src/sampling_rejection_tests.rs`, backed by a new test-only cxx bridge hook (`sampling_dispatch_stash_failed_launch_for_test` plus two accessors) in `mlx_cxx_bridge.cpp`/`.h` and `lib.rs`, following the existing `sampling_dispatch_drain_pending`/`sampling_dispatch_reset` hook pattern. It builds a valid, signalled event carrying an error through MLX's public `Event`/`Error` API (no GPU JIT needed), stashes it into pending-verification slot 0, then asserts `drain_pending_verification` neither throws nor consumes the error and drops the slot. Proved by temporarily reverting the drain to call `is_available()`: the reverted build crashed the whole test process with `libc++abi: terminating due to uncaught exception of type std::runtime_error: mlxcel test: synthetic launch failure`, the exact ml-explore/mlx#3742 failure mode this PR works around, then restored. `cargo test --profile test-fast --features metal,accelerate -p mlxcel-core --lib sampling_rejection_tests::` reports 28 passed with the fix restored. Commit a2db7934.

### Documentation completeness
Checked `docs/supported-models.md` (Qwen3.5/3.8 FP8 entry), `docs/architecture.md`, `CONTRIBUTING.md`, `README.md`, `docs/environment-variables.md`, `docs/turbo-kv-cache.md`, and `docs/upstream/mlx-cuda-quantized-matmul-nondeterminism.md` for stale MLX pin, CUDA link library, or fp8-accuracy claims. None name the old pin as a current-state fact or make an accuracy claim the pin bump invalidates; mentions are either mechanism descriptions that hold regardless of the pinned commit, or historical citations (CHANGELOG.md's `9a795735` entry, CONTRIBUTING.md's `2c46b953` CUDA-threading citation, the upstream bug report pinned to `b7c3dd6d`) that are provenance and correctly left alone, matching how CHANGELOG.md is only touched at release-prep commits rather than per-PR. `docs/installation.md` remains the one doc that needed the cuSOLVER update, already in place. No changes made beyond the test coverage commit.

PR #1780, opened 2026-09-11
chore(ci): add a concurrency group so superseded runs stop stacking

### LINKED ISSUE #1774: chore(ci): add a concurrency group so superseded CI runs do not stack on GB10
## Problem / Background

`.github/workflows/ci.yml` has no top-level `concurrency:` key, so every push queues a full CI run and nothing supersedes anything. Four of its jobs run on `GB10`, a single self-hosted runner: `clippy` (`ci.yml:168`), `xla-compile` (`ci.yml:421`), `xla-link` (`ci.yml:580`), and `cuda-sm70-compile` (`ci.yml:661`). Superseded runs therefore do not just sit idle, they hold a place in line ahead of the runs that matter.

Measured over 2026-09-10 KST (2026-09-09T15:00Z to 2026-09-10T15:00Z), from `gh run list`: 113 CI runs were created, and at 2026-09-10T03:18:40Z **14 runs were open at once**, 11 of them stacked on four branches (5 on `feature/issue-1359-got-ocr2`, 2 each on `fix/issue-1320-internlm2-rope-scaling`, `feature/issue-1361-fp8-block-requantize`, and `feature/issue-1347-laguna-family`). Six runs in that window ended `cancelled` rather than completing, every one of them superseded by a newer push on its own branch and cancelled by hand to free the runner: `34381499951`, `34429102450`, `34429978657`, `34430290450`, `34431841890`, `34431968023`. As of this writing two more CI runs (`34507516364`, `34508402981`) sit queued on the same branch, `feature/issue-1351-laguna-dflash-drafter`, neither superseding the other.

Two jobs in this same file already solved this locally and prove the pattern is accepted here: `xla-link` (`ci.yml:588-590`) and `cuda-sm70-compile` (`ci.yml:667-669`) each carry a per-PR `concurrency` group with `cancel-in-progress: true`, with the reasoning that they are the jobs long enough to hold GB10. `pipeline-parallel-ci.yml:64-66` uses the same shape at workflow level. Nothing covers `clippy` or `xla-compile`, and nothing covers the run as a whole.

## Aggravating factor

The GB10 runner shares hardware with local development builds, so a starved runner does not fail the way a lint failure does. On 2026-09-10 at 02:15:54 KST the runner service was OOM-killed mid-job. From `journalctl -u actions.runner.lablup.lablup-dgxspark21.service`:

```
Sep 10 02:15:54 spark-101 systemd[1]: actions.runner.lablup.lablup-dgxspark21.service: A process of this unit has been killed by the OOM killer.
Sep 10 02:15:54 spark-101 systemd[1]: actions.runner.lablup.lablup-dgxspark21.service: Main process exited, code=killed, status=9/KILL
Sep 10 02:15:54 spark-101 systemd[1]: actions.runner.lablup.lablup-dgxspark21.service: Failed with result 'oom-kill'.
Sep 10 02:15:54 spark-101 systemd[1]: actions.runner.lablup.lablup-dgxspark21.service: Consumed 4h 56min 37.217s CPU time, 28.4G memory peak, 68.1M memory swap peak.
```

The `cargo-clippy` job of run `34381499951` is recorded by GitHub with zero steps and no logs (`gh api repos/lablup/mlxcel/actions/runs/34381499951/jobs`), which is indistinguishable from a real lint result until someone reads the service journal. That run was created at 2026-09-09T17:12:39Z and its clippy job did not start until 2026-09-10T01:50:55Z, 8h38m later, with no CI run created anywhere in the repository across an 8h56m gap. Four branches were waiting behind it: `feature/issue-1334-kimi-k3-text-backbone`, `fix/issue-1320-internlm2-rope-scaling`, `feature/issue-1338-kimi-k3-tokenizer`, `feature/issue-1349-video-with-audio`.

Note what I could not confirm and left out: GitHub's API records that job's conclusion as `cancelled`, not `failure`, so the "stepless job looks like a lint failure" framing above is stated as "no steps and no logs" rather than as a recorded failure.

## Proposed Solution

Add a workflow-level `concurrency` block to `.github/workflows/ci.yml`, immediately after the `on:` block and before `permissions:`, keyed on workflow plus ref so a newer push on a branch supersedes the in-flight run for that branch and PRs never cancel each other:

```yaml
concurrency:
  group: ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

**The one decision to make explicitly, not by default.** `cancel-in-progress` on pushes to `main` is a real trade-off and the expression above takes one side of it: a post-merge run on `main` is the run that catches what a stale PR base hid, and cancelling it because the next merge landed 90 seconds later loses that signal. Against that, this repository merges to `main` frequently enough that back-to-back merges will queue GB10 jobs. Decide it, then record the reason in a comment above the block, in the style the rest of this file already uses. If the decision is to cancel on `main` too, `cancel-in-progress: true` is the whole change.

The existing per-job groups in `xla-link` and `cuda-sm70-compile` stay: a workflow-level group and a job-level group compose, and the job-level ones are keyed differently on purpose. Reconciling them into one scheme is a separate change and not required here.

Optionally, bound the runner service so a starved runner fails visibly instead of vanishing mid-job. A `MemoryMax=` (and `MemoryAccounting=yes`) in a systemd drop-in for `actions.runner.lablup.lablup-dgxspark21.service` turns the failure into a job that dies inside the runner, with logs, rather than the whole agent being reaped. This is host configuration, not a repository file, so it belongs in the runner runbook rather than in this workflow; call it out but do not block the concurrency fix on it.

## Scope

**In scope:** `.github/workflows/ci.yml` (add the top-level `concurrency` block plus its rationale comment).

**Out of scope:** `release.yml` (its GB10 job is a release build that must not be cancelled by a push), `nightly-verify.yml` (already grouped with `cancel-in-progress: false`, deliberately), `pipeline-parallel-ci.yml` (already grouped), the per-job groups in `xla-link` and `cuda-sm70-compile`, adding a second self-hosted runner, and the systemd `MemoryMax` change.

## Implementation Notes

- **Reuse**: copy the group-expression shape from `pipeline-parallel-ci.yml:64-66` and `ci.yml:588-590` rather than inventing a third form. `github.event.pull_request.number || github.ref` is what both already use, and it keeps a PR run and a `main` push run in distinct groups.
- **Constraints**: this file is heavily commented and every non-obvious decision in it carries a rationale comment. A bare `concurrency:` block with no comment would be out of place, and the `main` decision above is exactly the kind of thing a future reader will otherwise re-litigate.
- **Edge cases**: fork PRs run under the base repository, so they land in the same groups; a `workflow_dispatch` trigger does not exist on this workflow today, so there is no manual-run-cancels-CI hazard; the `changes` job's path filter is unaffected.
- **Error handling**: none. A cancelled run reports as `cancelled`, which GitHub does not treat as a required-check failure.

## Acceptance Criteria

- [x] `.github/workflows/ci.yml` has a top-level `concurrency` block with a group keyed on workflow plus ref and an explicit `cancel-in-progress` value.
- [x] A comment above the block records the `main` decision and why, in the style of the file's existing rationale comments.
- [x] Two pushes in quick succession to the same PR branch leave exactly one CI run not `cancelled`, verified from `gh run list` on the PR that makes this change. (Observed on PR #1780 without staging a race: run `34562655380` went `completed`/`cancelled` twenty seconds after the second push, run `34562815627` survived, and no `gh run cancel` was issued. `OpenXLA feature link` was cancelled mid-run on `lablup-dgxspark21` and `CUDA sm_70 compile` was released before claiming a runner.)
- [ ] A `main` push run behaves as the recorded decision says (cancelled by a following merge, or not). **Unverified, and not verifiable before merge.** The decision recorded in the comment is that an in-flight `main` run is never cancelled (`cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`). Confirming it requires a merge to `main` followed by a second merge while the first run is still going, so it can only be checked after this PR lands.
- [x] No change to `release.yml`, `nightly-verify.yml`, or `pipeline-parallel-ci.yml`. (`git diff --name-only origin/main` returns only `.github/workflows/ci.yml`.)

## Verification

```
python3 -c "import sys,yaml; d=yaml.safe_load(open('.github/workflows/ci.yml')); print(d['concurrency'])"

git push && git commit --allow-empty -m 'chore: supersede check' && git push
gh run list --repo lablup/mlxcel --workflow ci.yml --branch "$(git rev-parse --abbrev-ref HEAD)" \
  --limit 10 --json databaseId,status,conclusion,createdAt
```

A pass is: the YAML parses and prints the expected group; and after the second push the older run for that branch reads `completed / cancelled` while exactly one run remains `queued` or `in_progress`, with no manual cancellation.

## Technical Considerations

The runner-choice comment at `ci.yml:191-208` already anticipated the contention half of this: "a release in flight queues this job behind it and delays a merge... It is a delay and not a failure: a queued job consumes nothing." That is right as far as it goes. What it did not cover is the case measured here, where the queued jobs are for commits nobody is waiting on any more and they still sit ahead of the ones that are. Related: #1283 (why `clippy` runs on GB10 at PR time at all).

inureyes:
## Summary

`.github/workflows/ci.yml` had no top-level `concurrency` key, so every push queued a full run and nothing superseded anything. Four of its jobs land on `GB10`, the single self-hosted Linux runner that also serves the release build, so a superseded run did not merely sit idle, it held a place in line ahead of the runs that mattered. This adds one workflow-level `concurrency` block plus its rationale comment. No other workflow is touched.

## The `main` decision, made explicitly

`cancel-in-progress` is the expression `${{ github.ref != 'refs/heads/main' }}`, not a bare `true`. **In-flight runs on the default branch are never cancelled.** A push-to-main run is the one that catches whatever a stale PR base hid, and it is the last gate before a release builds from that commit, so killing it because the next merge landed 90 seconds later would discard the only signal nothing else produces. The queueing cost of protecting `main` is bounded rather than unbounded: cancelling a *pending* run is inherent to grouping and `cancel-in-progress` governs only the *running* one, so `main` holds one in-flight plus one pending regardless of merge rate. The residual trade, stated in the comment so nobody rediscovers it: an intermediate `main` commit can lose its run while still pending, while the commit at the head of `main` is always newest in the group and always gets one.

The group is `ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}`, the shape `pipeline-parallel-ci.yml` and the two per-job groups in this file already use, which keeps a PR run and a `main` push run in distinct groups.

## Workflow inventory

Every file under `.github/workflows/` was enumerated:

| Workflow | Had `concurrency`? | Touches `GB10`? |
|---|---|---|
| `ci.yml` | no, workflow-level (job-level in `xla-link`, `cuda-sm70-compile`) | yes, 4 jobs: `clippy`, `xla-compile`, `xla-link`, `cuda-sm70-compile` |
| `release.yml` | no | yes, `build-linux-cuda` |
| `nightly-verify.yml` | yes, `cancel-in-progress: false` | no (self-hosted macOS) |
| `pipeline-parallel-ci.yml` | yes, `cancel-in-progress: true` | no (`pp-three-host`) |
| `python.yml` | no | no (ubuntu-latest) |
| `update_homebrew_formula.yml` | no | no (macos-latest) |

`release.yml` is the only other contributor to the GB10 queue. **Deliberately not changed, and flagging rather than silently taking the decision:** it triggers only on `release: published` and `workflow_dispatch`, so it does not stack from pushes, and a grouped release build risks cancelling an artifact-producing run. It is not starved by this change either, because cancellation only ever removes CI work from the queue ahead of it. `nightly-verify.yml` is unaffected (different runner, already grouped `false` on purpose). `python.yml` and `update_homebrew_formula.yml` have no group but never reach GB10, so they are outside the problem this issue describes; grouping them would be a hosted-minutes change on its own merits, not part of this fix.

The per-job groups in `xla-link` and `cuda-sm70-compile` stay: workflow-level and job-level groups compose, and those two are keyed differently on purpose.

## What was verified, and what was not

Verified locally:

- The YAML parses and the block resolves to the intended group and expression; the four `GB10` jobs and both surviving per-job groups were re-read out of the parsed tree.
- `actionlint` 1.7.7 reports **byte-identical findings before and after**: 8 pre-existing (unknown self-hosted runner labels, shellcheck infos), 0 introduced.
- That clean result is load-bearing rather than vacuous: a negative control with `github.reff` and a bogus context field makes `actionlint` fail on this block, so its type-checker does inspect these expressions. `cancel-in-progress` accepting an expression is documented behavior.

**Verified live, without staging a race.** A deliberate double-push was considered and rejected: `.github/workflows/ci.yml` appears in all four `changes` path filters (`rust`, `mlx_pin`, `xla_link`, `cuda_arch`), so every push to this branch starts all four GB10 jobs, two with `timeout-minutes: 120`. Manufacturing the heaviest available load on the single shared runner to prove a point about relieving load on it is the wrong trade. The evidence arrived anyway, because committing the technical report was a second push as ordinary required work:

| Run | Head | Created | Final state |
|---|---|---|---|
| `34562655380` | `ab7fd234` (first push) | 04:33:41Z | `completed` / **`cancelled`** at 04:36:33Z |
| `34562815627` | `bc4ac8c8` (report push) | 04:36:13Z | `queued`, not cancelled |

Exactly one run for the branch is not `cancelled`, and no `gh run cancel` was issued at any point; the group cancelled the older run twenty seconds after the newer push registered. The cancelled run's job breakdown is the value proposition in miniature: `OpenXLA feature link` was **cancelled while running on `lablup-dgxspark21`**, and `CUDA sm_70 compile` was cancelled with an empty `runner_name`, released before it ever claimed the runner. Those two are the pair carrying `timeout-minutes: 120`.

**Still not verified, and not claimed: the `main` half of the expression.** That `cancel-in-progress` evaluates false on `refs/heads/main` and leaves an in-flight post-merge run alone can only be observed after a merge, followed by a second merge while the first run is still going. The run above exercised only the pull-request branch, where `github.ref` is `refs/pull/1780/merge`. The `main` behavior rests on the documented semantics of the expression and on reading it, not on an observation.

Closes #1774


inureyes:
Post-merge check for whoever lands this, since it is the one acceptance criterion that cannot be exercised from a PR.

The PR-ref half of the expression is now confirmed twice on this branch: pushes `bc4ac8c8` and `7c33f6f9` each auto-cancelled their predecessor (`34562655380` and `34562815627` both `completed`/`cancelled`), leaving exactly one live run each time, with no `gh run cancel` issued. The first cancellation caught `OpenXLA feature link` mid-run on `lablup-dgxspark21` and released `CUDA sm_70 compile` before it claimed the runner.

The `main` half is untested by construction. After this merges, the next back-to-back pair of merges is the check: the in-flight `main` run should keep running rather than going `cancelled`, because `github.ref` is `refs/heads/main` and the expression resolves false. If it does get cancelled, the expression is wrong and should be reverted to a literal `cancel-in-progress: false` for the default branch.

Issue #1798, opened 2026-09-11
perf(cuda): evaluate raising the MLX graph ops budget on GB10 (sm_121)


inureyes:
## Problem / Background

Follow-up to #1782 (PR #1795, merged as `e391ae9c`), and deliberately not changed there: its blast radius and its regression profile are different from that PR's, which touched drafter load dtype and the gated-delta ops fallback only.

During #1782's control sweep, `MLX_MAX_OPS_PER_BUFFER=100` improved the **classic single-token decode arm**, which has no drafter in it at all: 60.20 tok/s (59.78 to 60.80) against 58.42 (58.14 to 58.62), n = 3 each, ranges disjoint, about +3%. On the speculative arms of the same sweep it gave nothing at block 2 (26.03 against 25.82) and about +10% at block 8 (31.46 against 28.55, with the round device sync falling from 81.6 ms to 71.9 ms). Host: GB10 (sm_121, DGX Spark), MLX pin `81ba1c6a`, CUDA release build, `models/mlx/qwen3.5-4b-4bit`, one `mlxcel-server` per configuration, greedy 200-token streaming completions on a fixed 165-token prompt, GPU held exclusively on an otherwise idle host, same binary on every arm. Full record: `docs/benchmark_results/dflash-verify-fixed-cost-gb10-2026-09-11.md`, including the separate finding section at its end.

The classic arm is the reason this is its own issue rather than a note on #1782. A gain there is a property of the CUDA graph budget, so if it holds it applies to every CUDA decode path in the tree, not to speculative decoding. It is also one model, one prompt, at batch 1, which is not enough evidence to change a default.

## Current Behavior

The budget is MLX's per-architecture table and mlxcel never overrides it on CUDA. `get_graph_limits` in the pinned MLX (commit `81ba1c6a0e50a9268b931579c2d4f1158b9aab5a`, `mlx/backend/cuda/device.cpp:181-203`, stock upstream, not one of the in-tree overlays under `src/lib/mlx-cpp/patches/`) computes `cc = major * 100 + minor * 10` and switches on it:

| case | ops | MB |
|---|---|---|
| 800 (A100) | 20 | 400 |
| 900 (H100), 1000 (B200), 1200 (consumer Blackwell) | 100 | 1000 |
| 1210 (DGX Spark) | 20 | 25 |
| unenumerated fallback (for example V100 at 700) | 20 | 100 |

[...]

Issue #1782, opened 2026-09-11
perf(cuda): the multi-row speculative verify forward is launch-bound on GB10


inureyes:
## Problem / Background

Measured under #1351 (PR #1771, still open) on GB10 sm_121: the Laguna XS 2.1 NVFP4 target with its 0.9 GB bf16 DFlash drafter, greedy, raw code prompts, 200 tokens, idle host, GPU held exclusively for the whole sweep, n=3 per configuration against an n=5 kill-switch-off baseline on the same binary. No speculative configuration wins.

| Configuration | n | tok/s mean (min to max) | Mean accepted | vs off |
|---|---|---|---|---|
| off (baseline) | 5 | 32.50 (31.24 to 33.58) | | |
| block 2 | 3 | 16.57 (16.38 to 16.87) | 0.84 | 0.51x |
| block 4 | 3 | 27.23 (27.01 to 27.52) | 2.21 | 0.84x |
| block 6 | 3 | 30.90 (30.55 to 31.38) | 2.90 | 0.95x |
| block 8 | 3 | 31.47 (31.10 to 31.98) | 3.33 | 0.97x |
| block 10 | 3 | 31.03 (30.86 to 31.18) | 3.42 | 0.95x |
| block 16 (checkpoint default) | 3 | 27.92 (27.36 to 28.45) | 3.55 | 0.86x |

Within-configuration spread is about 2 percent, so this is not the GB10 single-run decode bimodality from #755 hiding a win.

Acceptance is not the binding constraint, the verify block's device cost is. PR #1771 reports, per round at block 8, about 32 ms of host-side drafter graph construction, 3 ms of target graph construction and 100 ms of synchronized device work for 4.33 emitted tokens, against 30.8 ms per token on the classic path. Across widths the verify block's device cost fits a fixed **77 ms plus 3.3 ms per row** (83 ms at 2 rows, 130 ms at 16), so even a 2-row verify costs 2.7x a single-token step. With a fixed term that size, perfect acceptance at block 8 caps the reachable speedup near 1.8x, and the measured 3.3 to 3.6 accepted tokens land at 0.86x to 0.97x.

A per-block cost nearly independent of row count is a property of how the CUDA backend executes a multi-row forward, not of any one drafter. Every speculative family in the tree makes the same call, so fixing it once is the lever that makes speculative decoding worth enabling on this hardware.

## Current Behavior

[...]

Issue #821, opened 2026-07-19
chore(deps): report the MLX CUDA graph-cache lifetime-miss abort upstream and track dropping our local default


inureyes:
## Summary

Report the CUDA graph-cache "Cache thrashing" abort semantics to upstream ml-explore/mlx, and track the upstream fix so mlxcel can eventually drop or lower the `MLX_CUDA_GRAPH_CACHE_SIZE=2000` default it now sets (#818, merged in bac0fca). This is #818 proposal 3, split out as its own tracking task.

## The upstream problem

In vendored MLX 0.32.1 (pin `57c66cac`), `mlx/backend/cuda/lru_cache.h` throws a fatal `std::runtime_error("Cache thrashing is happening, please set the environment variable MLX_CUDA_GRAPH_CACHE_SIZE to a larger value than {} ...")` based on a lifetime miss counter:

- `lru_cache.h` line ~91: `if (env_name_ && ++cache_misses_ > 2 * capacity_) throw ...`.
- `lru_cache.h` line ~131: `size_t cache_misses_{0};` is a lifetime counter. It is incremented on every insert-miss and is never reset, not on a cache hit and not on trim.
- `mlx/backend/cuda/device.cpp` line ~210: `graph_cache_("MLX_CUDA_GRAPH_CACHE_SIZE", /* default_capacity */ 400)`, so the throw fires at 800 lifetime misses.
- `mlx/backend/cuda/device.cpp` line ~490: the cache key is `graph_nodes_key_ + ":" + graph_deps_key_`, i.e. one entry per distinct captured graph shape.

Because the counter measures lifetime misses rather than a recent miss rate, any long-lived, shape-diverse CUDA workload trips it eventually even when the cache is behaving well. Speculative or batched decode reaches it fastest (draft and verify phases across varying batch sizes and sequence-length buckets multiply the number of distinct graph shapes). "Thrashing" is a misdiagnosis of a legitimately large lifetime working set, and the response is a whole-process `terminate`, not a recoverable signal.

## What to file upstream

An issue (and ideally a PR) proposing one of:

1. Measure a *windowed* or *recent* miss rate (e.g. misses within the last K operations, or a miss/hit ratio over a sliding window) instead of a monotonic lifetime count, so genuine thrashing (high steady-state miss rate against a fixed working set) is distinguished from a working set that simply grew past the initial capacity over a long run.
[...]

PR #1817, opened 2026-09-11
fix(cuda): keep small masked verify blocks off cuDNN SDPA (#1799)

### LINKED ISSUE #1799: perf(speculative): attribute the Laguna DFlash verify round's fixed and per-row cost on GB10
## Problem / Background

On the Laguna NVFP4 DFlash pairing (`models/mlx/laguna-xs-2.1-nvfp4` target with `models/mlx/laguna-xs-2.1-dflash` drafter), a multi-row speculative verify round still costs a fixed floor plus a per-row term that neither PR #1795 fix removes, so speculative decoding has no net win at any block width on GB10. PR #1771 (issue #1351) is blocked on this issue by maintainer decision: the Laguna DFlash drafter is implemented, reviewed and CI-green at `dab18fdf` (every check SUCCESS or SKIPPED) behind a fail-closed exactness gate, and is held unmerged until the pairing shows a net win here. The hold is not free: the branch has already taken two merge commits to follow main (`d64c73af`, `08e09036`) and will take more the longer it waits.

Post-rebase sweep recorded on PR #1771 and in `TECHNICAL_REPORTS/1351-laguna-dflash-drafter-20260911.en.md:59-83` (branch-only). GB10 sm_121, MLX pin `81ba1c6a` (`src/lib/mlx-cpp/CMakeLists.txt:122`), CUDA release build, same binary with the feature off as the control arm, GPU held exclusively, n = 3 per width, 200-token raw code completions through `mlxcel generate`, decode tok/s as the CLI reports it.

| width | tok/s mean (min to max) | mean accepted | device ms/round | drafter host build ms/round | vs off |
|---|---|---|---|---|---|
| off | 31.18 (30.41 to 31.98) | | | | |
| 2 | 16.19 (16.12 to 16.25) | 0.84 | 80 | 30 | 0.52x |
| 4 | 27.05 (26.61 to 27.53) | 2.21 | 85 | 31 | 0.87x |
| 6 | 30.46 (30.05 to 31.26) | 2.90 | 92 | 33 | 0.98x |
| 8 | 31.26 (31.07 to 31.36) | 3.33 | 101 | 34 | 1.00x |
| 10 | 30.02 (29.68 to 30.19) | 3.42 | 109 | 35 | 0.96x |
| 16 (default) | 27.41 (26.90 to 27.90) | 3.55 | 127 | 35 | 0.88x |

Best width is 8 at 1.00x with its whole range inside the off arm's, so it is not a demonstrated win, and the checkpoint default of 16 is 0.88x. The PR text fits the device time as a fixed 77 ms plus about 3.1 ms per row; an ordinary least squares fit over all eleven measured widths gives 72 ms plus 3.55 ms per row. Either way the floor alone is more than two classic tokens: 80 ms at 2 rows against a 31 to 32 ms graph-replayed classic token is 2.5x a single-token step. Including the 34 ms of host-side drafter graph construction, perfect acceptance at width 8 would reach about (101 + 34) / 8 = 17 ms per token, capping near 1.8x to 1.9x; at the measured 3.33 accepted it lands at 1.00x.

## Why neither #1795 fix reaches this pairing

PR #1795 removed two causes on the Qwen 3.5 pairing: an unconditional bf16 to f16 drafter conversion that promoted the whole drafter to float32, and a row-independent 64-row gated-delta chunked scan. Both statements below were checked in the code, not inherited from the PR text. First, the Laguna drafter was already bf16 on this host before #1795: the loader carried its own `drafter_weights_prefer_f16()` with the same Ampere-and-later rule, removed in `aad08e1c` in favor of the shared policy, so `src/lib/mlxcel-core/src/drafter/laguna_dflash/drafter.rs:65` now calls `apply_drafter_load_dtype_policy`, which returns false on compute capability 12.1 unless `MLXCEL_CUDA_F16_NORMALIZE` is set (`src/lib/mlxcel-core/src/drafter/dflash/drafter.rs:187-191,241-270`). Second, Laguna has no gated-delta layers: it is mixed full and sliding softmax attention with optional per-head sinks (`src/models/laguna_layers.rs:110-199` on the PR branch), and `gated_delta` appears nowhere in the Laguna sources. The measured numbers duly did not move across the rebase.

## The mechanism is NOT established, and that is the point of this issue

The phrase used so far, "the classic step is graph-replayed while the multi-row verify runs eagerly and launch-bound", is an inference from the fixed-plus-per-row arithmetic fit. It is not a profiled result: no nsys kernel attribution has been run on the Laguna pairing at any width. Do not implement against it. The Qwen 3.5 investigation in #1782 disproved exactly that intuition on the pairing where it was actually measured: CUDA graph capture is on for every eval unless `MLX_USE_CUDA_GRAPHS=0` (MLX pin, `mlx/backend/cuda/device.cpp:17-18`, default true) and nothing in the capture path branches on row count; replay was healthy at block 2 (260 `cudaGraphInstantiate` against 71089 `cudaGraphExecUpdate`); turning capture off made block 8 6 percent faster; and the classic step turned out GPU-bound on the quantized projections (12.4 ms of about 15.5 ms of per-token kernel time in `docs/benchmark_results/dflash-verify-fixed-cost-gb10-2026-09-11.md`), not launch-bound. The real causes there were a dtype bug and an algorithmic one, and neither was visible in the arithmetic fit.

## Proposed Solution

Attribution first, implementation second, the shape that worked for #1782.

**Phase 0, attribution (required, and the only phase with a fixed deliverable).** Profile the Laguna verify round per block width with `nsys profile -t cuda,nvtx --cuda-graph-trace=node` (mandatory: without it graph-captured work lands on `event_signal_kernel`, rule 4 of `docs/benchmark_results/volta-sm70-baseline-2026-08-31.md:64-88`). Report kernel names, absolute kernel time and per-round launch counts at 2, 8 and 16 rows, and split them into the kernels whose counts and time stay flat across widths (the fixed floor) and those that scale with rows (the per-row term). Reconcile the profiled run against an unprofiled one and state the inflation ratio before drawing any kernel-level conclusion. Name the mechanism, with the profile as the evidence.

Candidates to check. Each needs evidence; none may be assumed.

- **Is the multi-row verify replayed or re-instantiated on this pairing?** Measure it as the `cudaGraphInstantiate` to `cudaGraphExecUpdate` ratio per arm, the same way #1782 did. MLX keys an LRU of instantiated graphs (`graph_cache_`, default capacity 400, `mlx/backend/cuda/device.cpp:210`) and reuses through `cudaGraphExecUpdate` (`device.cpp:527-539`), so a round whose shapes vary can miss the cache here even though it did not on Qwen 3.5. A partial accept changes the next round's row count, so this is a live possibility rather than a formality.
- **The NVFP4 dispatch path.** `supports_fp_qmv` refuses when `vec_batch > 8`, and also refuses affine mode and compute capability 9 or below (MLX at the pin, `mlx/backend/cuda/quantized/qmm/qmm.cu:127-157`), while `quantized.cpp` selects `qmv` or `fp_qmv` when `M * B < 8` and `qmm_sm80` otherwise (`src/lib/mlx-cpp/patches/mlx/backend/cuda/quantized/quantized.cpp:282-292` dense, `:525-534` gather). So the family switch at 8 rows is real on this pairing, but the 5-to-7 accumulator cliff seen on the affine path does not exist here: `fp_qmv.cu` instantiates only `fp_qmv_single` and `fp_qmv_batched`, its `rows_per_block = 8` (`fp_qmv.cu:17`) tiles output rows (`blocks_y = (N + rows_per_block - 1) / rows_per_block`), and the only per-launch template variation is `n_per_thread` in {1, 2, 4} chosen by pointer alignment, not by input rows. The measured curve agrees: 5, 6 and 7 sit on a smooth line. Note also that Laguna is MoE and its experts go through `SwitchGLU` and `gather_qmm` (`src/models/laguna_layers.rs:23,451-469`), where there is no `fp_qmv` arm at all (`quantized.cpp:495-534`) and the fused decode-MoE kernel does not apply to a non-affine expert set (`src/models/switch_layers.rs:79-81`), so the expert matmuls are a separate dispatch question from the attention projections and the profile must separate them.
- **The Laguna attention shape itself**, including the multi-row mask. A verify with `l > 1` materializes an additive array mask per layer, a causal band on full layers and a window-clipped band on sliding layers, while a single-token decode passes no mask at all (`src/models/laguna_layers.rs:165-199`). On Qwen 3.5 the equivalent materialized-mask term measured under 2 ms per round at 16 rows and was ruled out; here it is unmeasured.
- **Drafter host-side graph construction**, 30 to 35 ms per round at every width, roughly one classic token, and a serial link separate from device time (`diagnostics.draft_time_ms`, built at `src/lib/mlxcel-core/src/drafter/dflash/round_loop.rs:754-767` and logged as `draft_ms` by `src/server/batch/dflash_target.rs:706-718`).

**Phase 1, change, only if phase 0 supports one.** A round is a serial chain of host draft build, host verify build and device sync (`round_loop.rs:751-818`), which is why the two #1795 fixes were superlinear together: each removed a different link and either alone left the others dominating. Measure any proposed change against the whole round and against end-to-end throughput, never against the link it targets.

**Phase 2, re-measurement.** Same method as the sweep above and as #1782: same binary on every arm, feature off as the control, GPU held exclusively, host load stated, n >= 3 per arm with min and max reported, and a width range wide enough to find a crossover if one exists (2 through 16 at least). Commit the record under `docs/benchmark_results/` and the attribution under `TECHNICAL_REPORTS/`.

## Scope

**In scope:** the per-round cost of the Laguna DFlash verify on CUDA, its attribution, and any change the attribution justifies in `src/lib/mlxcel-core/src/drafter/dflash/round_loop.rs`, `src/models/laguna_speculative.rs`, `src/models/laguna_layers.rs` or the MLX patches under `src/lib/mlx-cpp/patches/`; the committed benchmark record and technical report.

**Out of scope:**

- The default draft block width and any hardware-gated width policy: that is #1797, which already names this pairing as unmeasurable until #1771 lands, and already puts the `qmm_sm80` term at 8 rows and above outside its own scope.
- The second unmet criterion on PR #1771, the exactness probe, which belongs in its own issue and must not be folded into a performance negative. Recorded here so it is not lost: the probe declines this host by default at width 16 (107246 of 200704 logit bytes differ at the first verify position) and at width 8 (107466), the `qmv_wide` retry does not change it, and with `MLXCEL_MTP_ALLOW_INEXACT=1` the divergences are bf16 tie flips, 1 of 6 prompts byte-identical and 1 to 2 tie positions per 128 tokens on the other 5, every one at a chain-arm top-2 margin of 0.0 or 0.125. It is kept separate because its mechanism (kernel rounding at logit ties) and its fix surface (the gate in `src/models/laguna_speculative.rs:98-135` and the kernel numerics) share nothing with device-time attribution, and because folding it in would let a measured performance negative close an open correctness question. One concrete lead for that issue: both recorded declines are at widths where `M * B >= 8` sends the projections to `qmm_sm80`, while widths 2 to 7 stay on `fp_qmv`, whose per-row arithmetic is row-independent by construction; the gate probes before it consults the override (`src/models/speculative_exactness.rs:295-312`), so per-width verdicts at 2 to 7 exist in the sweep logs or can be re-run cheaply.

## Implementation Notes

- **Reuse**: the per-round split is already instrumented. `DFlashDiagnostics` carries `draft_time_ms`, `verify_time_ms` and `target_argmax_time_ms` (`round_loop.rs:530-538`) and the server logs them as `draft_ms`, `verify_ms` and `target_argmax_sync_ms` on the `DFlash diagnostics` line (`dflash_target.rs:706-718`). Do not add a parallel timing path. `scripts/bench_block_width.sh` already interleaves widths so drift spreads across the table, but it hardcodes `--draft-kind mtp` (lines 86 and 112) and drives `mlxcel generate`; extend it rather than writing a new harness, and say which harness produced each number.
- **Constraints**: the fix must not change greedy output on any pairing, must keep a same-binary kill switch for the A/B as #1795 did, and must not regress the classic arm or the Qwen 3.5 speculative arm (1.24x at width 2 and 1.32x at 4 after #1795). Anything touching the MLX patches under `src/lib/mlx-cpp/patches/` reaches every quantized decode on CUDA, so state the blast radius.
- **Measurement hazards**: `MLXCEL_TRACE_ARCH` prints the quantized path taken on the process's first call only (`quantized.cpp:58-74`), which is prefill, so it cannot separate verify widths; use nsys kernel names instead. GB10 single-run decode is bimodal by up to 25 percent even at pinned clocks (#755), hence n >= 3 and reported spreads. The #1771 sweep ran with the GPU exclusive but the host not idle (another session's `cargo build` at load 2.6 to 3.9), which is a real hazard for a chain whose drafter graph build is host-side, so re-measurement must run on an idle host and say so. nsys inflated the #1782 runs by 15 to 46 percent, so compare proportions and launch counts across arms, not absolute times.
- **Edge cases**: a partial accept makes the next round's row count vary, which is what could churn graph keys; rollback replays accepted rows and can add launches the width alone does not predict; the sliding caches run with speculative slack (`src/models/laguna_speculative.rs:64-82`), so the key count a rotating layer exposes changes as the window fills.

## Acceptance Criteria

- [x] Phase 0 names the mechanism of the fixed floor and of the per-row term, each backed by a committed nsys table with kernel names and per-round launch counts at 2, 8 and 16 rows, stating which kernels are flat across widths and which scale.
- [x] The graph replay question is answered with a number: the `cudaGraphInstantiate` to `cudaGraphExecUpdate` ratio on the Laguna verify arm at at least two widths.
- [x] Every candidate above is either confirmed with evidence or ruled out with a measured bound, including the ones that turn out not to matter.
- [x] Any code change ships with a same-binary kill switch, and its A/B is reported against the whole round and against end-to-end throughput, not against the link it targets.
- [x] Re-measurement across widths 2 to 16 on an idle host with the GPU held exclusively, n >= 3, min and max reported, same binary and feature-off control, with the classic arm re-measured in the same session.
- [x] Greedy output is unchanged on the Qwen 3.5 pairing (still byte-identical to classic at every width after #1795) and that pairing's speculative win is not regressed.
- [x] A benchmark record under `docs/benchmark_results/` and a technical report under `TECHNICAL_REPORTS/` (`.en.md` and `.ko.md`), following the #1782 pair.
- [x] **A measured negative closes this issue successfully.** (Not needed: the pairing reaches a net win, 1.33x at block 4. Kept ticked as the clause is satisfied by the positive result.)  If the evidence says the cost is irreducible on this pairing and this host, record that with the profile, say what would have to change (hardware, kernel, or drafter shape) for the conclusion to flip, and hand the merge decision on PR #1771 back to the maintainer with that evidence. Shipping a change that does not move the measurement is not an acceptable outcome.

## Verification

```
# Attribution (per width; repeat at 2, 8, 16)
nsys profile -t cuda,nvtx --cuda-graph-trace=node -o laguna_w8 \
  mlxcel generate -m models/mlx/laguna-xs-2.1-nvfp4 \
    --draft-model models/mlx/laguna-xs-2.1-dflash --draft-kind dflash \
    --draft-block-size 8 -n 200 --temp 0

# Per-round split from the same runs
RUST_LOG=info mlxcel generate ... 2>&1 | grep "DFlash diagnostics"

# Graph replay ratio
nsys stats --report cuda_api_sum laguna_w8.nsys-rep | grep -E "cudaGraphInstantiate|cudaGraphExecUpdate"

# Gate (this host)
cargo fmt --all -- --check
cargo clippy --profile test-fast --features cuda --lib --bins --tests -- -D warnings
make verify-test-cuda   # cargo test --workspace --profile test-fast --features cuda -- --test-threads=1
```

A pass is: the profile tables committed, the fixed and per-row terms each attributed to named kernels, and either a re-measured throughput table showing a net win at some width or a recorded negative with the evidence behind it.

## Technical Considerations

Provenance: issue #1351 and PR #1771 (the drafter, blocked on this), issue #1782 and PR #1795 (the Qwen 3.5 attribution and fix, merged as `e391ae9c`), issue #1797 (the width policy follow-up). Prior art to read first: `docs/benchmark_results/dflash-verify-fixed-cost-gb10-2026-09-11.md` and `TECHNICAL_REPORTS/1782-dflash-verify-fixed-cost-cuda-20260911.en.md`. The Laguna-side numbers quoted above live on the PR branch in `TECHNICAL_REPORTS/1351-laguna-dflash-drafter-20260911.en.md` and in `docs/supported-models.md:714`, which already documents the pairing as measured and not winning on GB10.

## Resolution log (2026-09-11)

Attribution on an idle host, tree `2deae307`: the fixed floor is host time in `ScaledDotProductAttention::eval_gpu` (67 to 76 ms per round against 2.8 ms per classic token, nsys NVTX ranges differenced per round): a multi-row masked SDPA with head_dim 128 is cuDNN-eligible, MLX caches the cuDNN plan by exact `k_len` and mask shape, and a verify round's key length is new every round, so three shape classes (target full, target sliding with sinks, drafter) rebuild a plan each round at about 22 ms; the same LRU's lifetime miss counter aborted every block-2 run past about 170 rounds. The per-row term is the routed experts reading each selected expert per row (`qmm_sm80`, 2.5 ms per row). Ruled out with bounds: byte and op budgets, capture, re-instantiation (under 1 instantiate per 1000 exec updates), the mask (under 1 ms of GPU time per round), and the NVFP4 family switch on the projections, which does not exist here because this checkpoint's attention projections are bf16 and `gather_qmm` sees `M = 1, B = 8n` so the experts take `qmm_sm80` at every width. Fix: a patched `supports_sdpa_cudnn` that keeps 2 to 32 query rows over a longer key sequence off cuDNN (`MLXCEL_SDPA_FALLBACK_MAX_QUERIES`, 0 restores upstream), plus `MLX_CUDA_SDPA_CACHE_SIZE=2000` by default on CUDA. After, same binary kill-switch A/B, n = 3: block 4 1.33x, block 6 1.32x, block 2 1.13x, block 8 1.12x, crossover between 10 and 12, checkpoint default 16 at 0.83x; Qwen 3.5 unchanged (1.26x, 1.35x) and byte-identical. Record: `docs/benchmark_results/laguna-dflash-verify-cost-gb10-2026-09-11.md`; report `TECHNICAL_REPORTS/1799-laguna-dflash-verify-cost-cuda-20260911.{en,ko}.md`. Also recorded for #1798: GB10's default graph budgets make capture a net loss on this MoE classic arm (graphs off +12%, both budgets raised +16%), opposite sign to Qwen 3.5.

inureyes:
## Summary

The Laguna DFlash verify round's fixed floor on GB10 was host time, not GPU work: a multi-row masked SDPA with head_dim 128 is cuDNN-eligible, MLX caches the cuDNN execution plan by the exact key length and mask shape, and a verify round's key length is new every round, so three shape classes (target full layers, target sliding layers with sinks, drafter layers) rebuilt a plan on the host each round at about 22 ms each: 67 to 76 ms per round by nsys NVTX ranges against 2.8 ms per classic token. The same LRU's lifetime miss counter aborted every block-2 generation past about 170 rounds. The per-row term is the routed experts reading each selected expert per row, inherent to a MoE verify. Full attribution, including the graph-side candidates ruled out with bounds and two corrected premises (this checkpoint's attention projections are bf16, and the experts take `qmm_sm80` at every width), is in `docs/benchmark_results/laguna-dflash-verify-cost-gb10-2026-09-11.md`; the report pair is `TECHNICAL_REPORTS/1799-laguna-dflash-verify-cost-cuda-20260911.{en,ko}.md`.

- A patched `supports_sdpa_cudnn` (the pinned `81ba1c6a` file plus one gate) keeps a masked call with 2 to 32 query rows over a longer key sequence off cuDNN, so it takes MLX's own ops fallback with no per-shape build; the one-row decode step and prefill are untouched. `MLXCEL_SDPA_FALLBACK_MAX_QUERIES=0` restores upstream dispatch without a rebuild.
- `MLX_CUDA_SDPA_CACHE_SIZE` defaults to 2000 on CUDA builds (`hardware::apply_cuda_sdpa_cache_default`, the #818 sibling, wired from all four binaries), because prefill prompt-length diversity alone crosses the 512-miss abort on a long-lived server.
- `materializing_sdpa_query_chunk` mirrors the new gate (`cuda_sdpa_small_query_fallback`) so the score-matrix chunk budget still applies to the calls the gate diverts (review finding; inert below the 1 GiB budget).

## Measured

Same binary, kill-switch A/B on an idle host (load1 gated and recorded, `ps` checked), n = 3 with ranges, tree `2deae307` for the baseline and the rebuilt tree for the after arm (the rebase onto `bec64748` changed no file of this PR): Laguna DFlash goes from a best of 0.72x to 1.33x at block 4 and 1.32x at block 6 (1.13x at 2, 1.12x at 8, crossover between 10 and 12, checkpoint default 16 still 0.83x); the kill switch reproduces the baseline within 2%; the device sync fit moves from `71.6 + 3.79/row` to `36.0 + 3.6/row` ms; the after profile has no `ScaledDotProductAttention` host range at all and the round is GPU-bound. Qwen 3.5 through the #1782 server harness on the same binary: 1.26x and 1.35x (#1782: 1.24x, 1.32x), greedy text byte-identical to classic at both widths; its head_dim 256 never enters cuDNN, so that number shows the rebuild changed nothing else.

## Blast radius

Measured: Laguna, and Qwen 3.5 as the cross-family control. Sharing the new path without a number: every other CUDA model whose masked SDPA calls with 2 to 32 query rows and prior context reach `scaled_dot_product_attention.cpp` on Ampere or later, which is the speculative verify of any head_dim-128-or-less pairing (Muse Glimmer DFlash, LFM2 and LFM2.5 DSpark, Inkling MTP; none re-probed here, and a flipped exactness probe declines speculation rather than corrupting output), a short incremental prefill over a reused prefix-cache prefix, and the trailing short chunk of a chunked prefill. The fallback's cost grows with the key length while the plan build it replaces is constant (about 10 ms more GPU per round at block 2 with 350 keys; the crossover is in the tens of thousands of keys at block 16, unmeasured). Metal and CPU are unaffected. The Laguna exactness gate is not touched. The general fix, bucketing the key length so the cuDNN plan cache can hit and keep the flash kernel, is deliberately not attempted here and is being filed as its own issue. Also recorded for #1798: GB10's default graph budgets make capture a net loss on this MoE classic arm (graphs off +12%, both budgets raised +16%), opposite sign to Qwen 3.5.

## Test plan

- [x] Baseline, zero-code controls, nsys at 2, 8 and 16 rows, cuDNN-off control, after-fix A/B and the Qwen 3.5 check on GB10, all in the record with n and ranges
- [x] `cargo test --profile test-fast --features cuda -p mlxcel-core -- hardware:: small_query_fallback materializing_sdpa chunked_query_attention` (35 passed, on the rebased tree)
- [x] `cargo clippy --profile test-fast --features cuda --lib --bins --tests -- -D warnings` (clean, on the rebased tree)
- [ ] `metal,accelerate` gate: not runnable on this Linux/CUDA host, unrun; the patched file compiles only into the CUDA backend and the cache default is `None` without the `cuda` feature

Closes #1799


inureyes:

Reviewed against the #1799 deliverables. The attribution, the kill-switch A/B, the blast-radius statement, the record and the bilingual report are all present, and the gate itself is sound: every call it refuses reaches either `sdpa_vector` (2 or 3 unmasked rows, whose `i <= kL - qL + q_seq_idx` is the same bottom-right alignment as cuDNN's `set_causal_mask_bottom_right`) or the `mlx/fast.cpp` ops fallback (array mask or 4 to 32 rows), which honors array and bool masks, GQA repeats, sinks and a `k_len - q_len` causal offset. One-row decode never reaches the gate, backward never consults `supports_sdpa_cudnn`, and `use_fallback` and `eval_gpu` evaluate the gate on identical shapes so they cannot desync. The patch file is pinned upstream `81ba1c6a` plus exactly two hunks.

### Fixed and pushed as `38f2bb57`

- **HIGH** `apply_cuda_sdpa_cache_default` was wired into three binaries but not `src/bin/speculative_bench.rs`, the fourth that raises the graph-cache sibling and the one its own comment calls the canonical workload for that abort. Now wired, with the SAFETY note naming all four callers.
- **HIGH** Both `TECHNICAL_REPORTS/1799-*.{en,ko}.md` shipped a literal `VERIFICATION_PLACEHOLDER` as their verification section. Filled in with the patch provenance, the dispatch paths, the measurement protocol and the gates; the entry-point count in both goes from three to four.

### Left open

- **MEDIUM** The blast-radius examples (PR body, both reports, `docs/environment-variables.md`) name only the trailing chunk of a chunked prefill. With the prefix cache adopting the longest common prefix, a short new turn over a long reused prefix is also 2 to 32 rows over a much longer `k_len`, and that regime is unmeasured here (every arm used a 152-token prompt). Narrowing the gate without a measurement would defeat the point, so this is a follow-up.
- **MEDIUM** `docs/upstream/mlx-cuda-graph-cache-lifetime-miss-abort.md`, the #821 upstream draft, says the graph cache "is the only one whose key space grows with the shape of the host's op stream" and that the SDPA caches are keyed by "a comparatively fixed set of layer configurations". This PR is the counterexample and the draft should be corrected before it is filed.
- **LOW** The two new helpers sit after the `── Detection ──` divider in `hardware.rs`, away from the sibling they mirror.
- **LOW** MLX resolves the variable with `atoi`, so an unparseable `MLXCEL_SDPA_FALLBACK_MAX_QUERIES` silently disables the gate; the docs row promises "non-negative integer". Consistent with the file's existing switches.
- **LOW** The record gives block 6's baseline round wall as 126.7 ms in one table and 126.6 ms in another.

`cargo fmt --all -- --check` is clean after the push. Cargo test and clippy were in flight when the review started and the push lands after them, so both need a re-run; `speculative_bench` is covered by the `--bins` selector.

inureyes:
## Follow-up: second-opinion review, adjudicated

A second reviewer raised three items. One is a real bug and is now fixed, one is half right, one duplicates a fix already pushed.

**Fixed as `473e2d7a` (HIGH).** `cuda_sdpa_materializes_scores` in `src/lib/mlxcel-core/src/layers.rs` states in its own doc comment that it mirrors `supports_sdpa_cudnn`, and this PR changed that function without updating the mirror. The guard therefore still classified a masked 2-to-32-row call as flash-eligible and skipped chunking, so MLX's now-forced ops fallback allocated the whole `[B, heads, q_len, k_len]` score matrix with no budget check. `materializing_sdpa_query_chunk` now takes whether the call is masked or causal and consults a `cuda_sdpa_small_query_fallback` mirror that reads `MLXCEL_SDPA_FALLBACK_MAX_QUERIES` with MLX's `atoi` semantics. Below the 1 GiB budget `query_chunk_len` still returns `None`, so no configuration measured for #1799 changes; only calls over the budget begin to chunk. This also bounds what happens if an operator raises the variable, which previously widened the unguarded range without limit. A unit test pins the mirror to the gate.

**Half right, left open (MEDIUM).** The claim that the DFlash drafter is maskless is false for Laguna: `LagunaDFlashAttention::forward` passes an explicit sliding-window causal mask, which is why the record shows its host build dropping from 30 ms to 9 to 11 ms. It is true for the generic `DFlashAttention::forward`, which calls `fast_scaled_dot_product_attention` with a null mask and no causal flag, so a non-Laguna DFlash pairing with head_dim 128 or less still rebuilds one plan per round. That is outside this PR's measured workload and extending the gate to maskless shapes is a dispatch change no measurement here covers, so it belongs in a follow-up.

**Already fixed.** The `speculative_bench` wiring gap was pushed as `38f2bb57`.

inureyes:
Security and performance review of the SDPA gate. No CRITICAL or HIGH findings, so nothing was auto-fixed and the branch is untouched. `cargo fmt --all -- --check` is clean, the patched file differs from the pinned upstream in exactly the header comment and the gate, and no commit or the PR body carries AI attribution. Two MEDIUM items for the author to decide on.

MEDIUM, performance: the gate bounds query rows but not key length. The cuDNN plan build it replaces is a constant (about 22 ms times three shape classes), while the ops fallback's cost grows linearly in `k_len`. Every arm in the record ran at `k_len` under about 350 (152-token prompt, 200 generated), and the record itself notes the fallback is slower than cuDNN's flash kernels on long prompts. Scaling the recorded "about 10 ms more GPU time per round at block 2" by the ten full-attention layers times 48 heads puts the crossover in the tens of thousands of key tokens at block 16, inside the pairing's usable context. Consider a score-element or `k_len` bound in the gate, or record the unmeasured regime in the report and in `docs/environment-variables.md`.

MEDIUM, memory: `materializing_sdpa_query_chunk` is reached only from `layers::attention` (`src/lib/mlxcel-core/src/layers.rs:4349`) and `causal_attention` (`src/lib/mlxcel-core/src/lib.rs:3386`). The four in-tree sinks call sites call `fast_scaled_dot_product_attention_with_sinks` directly and never consult it: `src/models/laguna_layers.rs:183` (the target sliding layers, one of the three diverted shape classes), `src/models/gpt_oss.rs:764`, `src/models/deepseek_v4_attention.rs:346`, `src/models/falcon_ocr.rs:435`. On those paths the newly diverted call materializes `[B, heads, q_len, k_len]` with no budget at all. Laguna's own exposure stays bounded because its sinks layers are window-rotating, but gpt_oss carries sinks on its full-attention layers too, so a 2 to 32 row masked call over a long cache materializes there unguarded. The doc comment on `cuda_sdpa_small_query_fallback` describes protection those call sites do not have.

PR #1821, opened 2026-09-11
fix: track live slot occupancy during prefill

### LINKED ISSUE #1800: fix(server): make GET /slots track live slot occupancy on all routes
## Summary

Backend.AI GO wants to show how full each loaded model's context is by polling `GET /slots` on both llama-server and mlxcel-server (occupancy = `n_prompt_tokens / n_ctx`, busy = `is_processing`). mlxcel has served `/slots` in b10621's shape since #1440, but on the OpenAI-compatible and Anthropic routes a slot's fields do not follow a running request: the slot reads idle for the whole prefill, `n_prompt_tokens` stays 0 until the request ends, and it never counts generated tokens. On b10621 the same fields track the slot's token stream live. Downstream context: lablup/backend.ai-go#5153 removed a hardcoded usage bar because no engine data backed it.

## Measured

mlxcel-server 0.7.0 (the build Backend.AI GO installs) on macOS Apple Silicon. The slot code is unchanged between `v0.7.0` and `main` `2deae307` apart from a copyright line: `git diff v0.7.0 origin/main` over `slots_state.rs`, `routes/slots.rs`, and the drains in `model_provider.rs` shows no slot or Prefill change. Model `mlx-community/gemma-4-12B-it-qat-4bit`, `--ctx-size 20480 --parallel 1`, one streaming `/v1/chat/completions` request with a 15,021-token prompt and `max_tokens: 40`, polling `GET /slots` every 0.2 s.

| t (s) | is_processing | n_prompt_tokens | n_prompt_tokens_processed | n_prompt_tokens_cache | n_decoded |
|---|---|---|---|---|---|
| 0.0 to 64.8 (prefill) | false | absent | absent | absent | absent |
| 64.84 (first token) | true | 0 | 0 | 0 | 2 |
| 65.46 | true | 0 | 0 | 0 | 19 |
| 65.67 (after finish) | false | 15021 | 15021 | 0 | 23 |

"absent" means the slot had carried no task yet, so the task fields were not in the object. To reproduce, poll while the request runs:

```bash
while sleep 0.2; do curl -s "$BASE/slots" | jq -c '.[0] | {is_processing, n_prompt_tokens, n_prompt_tokens_processed, n_prompt_tokens_cache, n_decoded: .next_token[0].n_decoded}'; done
```

## Expected (b10621)

Read from the source at tag `b10621`, `tools/server/server-context.cpp`, and not yet measured on the pinned binary:

- `is_processing()` is `state != SLOT_STATE_IDLE` (l.420-422), so it is true from task assignment.
- `n_prompt_tokens` is `prompt.tokens.size()` (l.654). That vector is trimmed to the cached prefix when a task starts (`keep_first(n_past)`, l.3298), grows per processed prompt token (l.3431), and grows per sampled token (`prompt.tokens.push_back(sampled)`, l.494).
- `n_prompt_tokens_processed` grows per batch (l.3391).

For the run above, b10621 would therefore report the slot busy during prefill with both counters climbing, `15021 + n_decoded` during decode, and 15,044 after the task.

## Root cause

On origin/main `2deae307`:

1. Binding waits for the first token on the chat routes. `SlotHandle` binds on its first update (`src/server/slots_state.rs`: `begin` at l.146 with the "Deliberately NOT bound here (#1440)" rationale at l.165, `try_bind` at l.319). The chat, completions, responses, and anthropic routes call only `slot.on_token` and `slot.finish` (for example `src/server/routes/chat.rs:1302` and `:1333`), so the first update is the first decoded token.
2. The prefill events exist and are dropped on those routes. The scheduler publishes `GenerateEvent::Prefill` at the Queued to Prefilling transition (`src/server/batch/scheduler/handoff.rs`, `begin_prefill`, l.111-121), once per prefill chunk (`src/server/batch/scheduler/prefill.rs:1024` and `:1207` through `report_prefill_progress`, `src/server/batch/sequence.rs:448`), and at the first token (`mark_first_token_at`, `sequence.rs:422`). Only `drain_generation_events_with_prefill` (`src/server/model_provider.rs:2301`), used by the native `/completion` streaming arm, forwards them; the other drains discard them (`model_provider.rs:2248-2250`, whose comment says "every other consumer skips it", and `:2287`).
3. Even the native arm updates the slot only at the first token: `src/server/routes/native_completion.rs:996-999` returns early unless `stats.first_token`, then calls `slot.on_prefill`.
4. `n_prompt_tokens` is the prompt length only. `on_prefill` (`slots_state.rs:359`) and `finish` (`:387`) set it to the prompt token count, and `on_token` (`:368`) increments only `n_decoded`. The `GET /slots` entry in `compat/llama-server/b10621/observability-and-slots.json` records only the `params` subset as a divergence, so this one is unrecorded.

## Secondary: n_ctx under `--ctx-size 0`

`routes/slots.rs:146` passes `state.config.context_size`, which `resolve_parallel_context_size` (`src/server/startup.rs:793`) leaves at 0 when `--ctx-size` is 0; `/props` `default_generation_settings.n_ctx` (`routes/props.rs:208`) and `/v1/models` `meta.n_ctx` (`routes/models.rs:73`) report the same 0. b10621 turns 0 into the model's window (`llama_n_ctx_seq`) and never reports 0. The `--ctx-size` manifest entry (`compat/llama-server/b10621/runtime-and-context.json`) says the resolved per-slot window is reported as `n_ctx` in `/props`, which holds only for a non-zero `--ctx-size`, while the comment at `routes/models.rs:71-72` documents 0 as "the checkpoint's own trained context applies unclamped". A client dividing by `n_ctx` divides by zero. In the same measurement session `/v1/models` also reported `meta.n_ctx_train: 0` for this Gemma 4 checkpoint, so whatever source resolves the window must work for it. Backend.AI GO always passes an explicit `--ctx-size`, so this does not affect it today.

## Proposed change

1. Forward `GenerateEvent::Prefill` to the `SlotHandle` on every generation route (chat, completions, responses, anthropic, native non-streaming), for example through an optional prefill observer on the provider's streaming and non-streaming entry points, or by moving those routes onto the prefill-aware drain. The native `return_progress` framing stays as it is.
2. Bind on the first Prefill observation (the `begin_prefill` frame) instead of the first token. This keeps #1440's rule that only a request the worker is serving holds a slot, because `begin_prefill` runs at scheduler admission, not while the request waits in the queue. A path that never emits Prefill keeps binding at the first token.
3. Keep the counters live and in b10621's meaning: `n_prompt_tokens_cache` from the first observation; `n_prompt_tokens_processed` = `processed` minus the cached count (mlxcel's `processed` counts cache-supplied tokens, per the `report_prefill_progress` doc comment); `n_prompt_tokens` = `processed` during prefill, prompt tokens + `n_decoded` during decode (accepted tokens only), prompt tokens + completion tokens after finish.
4. Under `--ctx-size 0`, report the effective per-slot window as `n_ctx` in `/slots`, `/props`, and `/v1/models` `meta.n_ctx`, using the same resolution `resolve_default_max_tokens` already applies (`startup.rs:1029`: the per-slot size when set, else the model's `max_position_embeddings` from `config.json`, else the 4096 fallback). If 0 is intended, record it in the `--ctx-size` and `GET /slots` manifest entries instead.
5. Measure against the pinned b10621 binary before pinning values (a long prompt, `--parallel 1` and `2`, polling during prefill, during decode, and after finish), as #1440 did, and update the manifest entries.

## Acceptance criteria

- [ ] During the prefill of a streaming `/v1/chat/completions` request, `GET /slots` shows the serving slot with `is_processing: true` and `n_prompt_tokens_processed` increasing across polls.
- [ ] During decode `n_prompt_tokens` equals prompt tokens + `n_decoded`; after finish it equals prompt + completion tokens.
- [ ] The same holds for `/v1/completions`, `/v1/responses`, `/v1/messages`, and native `/completion`, streaming and non-streaming.
- [ ] A request still waiting in the scheduler queue holds no slot: `slots_state_tests.rs::a_queued_request_does_not_hold_a_slot_away_from_a_serving_one` still passes, and a new test covers a queued request behind a prefilling one.
- [ ] With `--ctx-size 0`, `n_ctx` is positive in `/slots` and `/props`, or the manifest records why it is not.
- [ ] New tests fail when the Prefill forwarding is removed.
- [ ] The `GET /slots` and `--ctx-size` manifest entries are updated.

Verification: `cargo test --lib slots_state` for the registry, and the poll loop under Measured against each route for the live behavior.

## Out of scope

- KV-image save and restore, and `id_slot` request routing: the token-stream persistence divergence is recorded by design under `POST /slots/:id_slot`.
- Server-wide KV pool usage: `/metrics` already exports `mlxcel_cache_pool_paged_blocks_live`, `mlxcel_cache_pool_paged_blocks_free`, `mlxcel_cache_pool_paged_bytes_in_use`, `mlxcel_prompt_cache_bytes`, and `mlxcel_slots_available` behind `--metrics`.
- The auto `--parallel` per-slot division (auto becomes 4 slots that split `--ctx-size`, where b10621's auto also enables `kv_unified` so one request can use the whole context). It is a recorded by_design divergence (#1473), and `/slots` already reports the per-slot share correctly.

inureyes:
## Summary

Make `/slots` consume every prefill progress event from the generation routes, so slots bind at scheduler admission, stay isolated from queued requests, and report b10621 prompt/cache/decoded counters live through finish. Also resolve `--ctx-size 0` to the same positive effective context window on `/slots`, `/props`, and `/v1/models`.

## What changed

- Added prefill-aware provider drains and wired chat completions, completions, Responses, Anthropic messages, native completion, and ASR streaming into their route-owned `SlotHandle` in streaming and non-streaming paths.
- Updated live and retained slot counter semantics while preserving queued-request isolation and native `return_progress` framing.
- Resolved the effective context window once during `AppState` construction, avoiding repeated checkpoint configuration I/O on polled metadata endpoints.
- Updated focused tests, b10621 compatibility manifests, directly affected docs, and the bilingual technical report.

## Test plan

- [x] `cargo test --workspace --profile test-fast --features metal,accelerate`
- [x] `cargo test --profile test-fast slots_state`
- [x] `cargo test --profile test-fast logprobs_drain_forwards_every_prefill_observation`
- [x] `cargo test --profile test-fast ctx_size_zero`
- [x] `cargo test --profile test-fast --test llama_compat_manifest`
- [x] `cargo clippy --all-targets --features metal,accelerate -- -D warnings`
- [x] `cargo fmt --check`
- [x] `jq empty compat/llama-server/b10621/observability-and-slots.json compat/llama-server/b10621/runtime-and-context.json`
- [x] Real-checkpoint smoke on `models/gemma-3-1b-it-4bit`: `/props`, `/v1/models`, and `/slots` reported `n_ctx: 1024`; `/v1/completions` retained `n_prompt_tokens: 8` for 6 prompt plus 2 decoded tokens; chat, completions, Responses, Anthropic messages, and native completion streaming/non-streaming one-token requests completed.

The pinned b10621 executable was not present locally, so an exact pinned-binary measurement is not included; available b10883 artifacts were not substituted as the reference.

Closes #1800


inureyes:

### Intent
> Make `/slots` report live b10621-style occupancy from scheduler prefill admission through decode/finish on every generation route, while reporting positive effective `n_ctx` when `--ctx-size 0`.

### Findings Addressed
- [x] No correctness, completeness, integration, or convention defects found in review.

### Remaining Items
- [ ] Pinned b10621 binary measurement remains unavailable locally; the PR body already records this limitation.

### Verification
- [x] Tests pass

inureyes:
## Security and Performance Review Summary

Reviewed PR #1821 for request-boundary data exposure, slot-state isolation, synchronization and cleanup, counter safety, and hot-path cost.

### Finding addressed

- **MEDIUM — repeated checkpoint I/O on observability routes:** with `--ctx-size 0`, each request to `/slots`, `/props`, or `/v1/models` reread and reparsed the model configuration. Frequent polling could turn a read-only observability surface into avoidable filesystem work. Commit `ac8162bf` now resolves the effective per-slot context once during `AppState` construction and serves the cached value thereafter.

### Remaining review result

- No CRITICAL or HIGH security or performance findings.
- Slot updates retain the existing prompt-text redaction policy, use bounded server-owned slot state, and keep queued requests unbound until scheduler progress.
- Prefill callbacks perform constant-size counter updates; lock ordering does not introduce a new inversion, and RAII drop still releases bound slots on cancellation or error.
- Counter inputs are clamped or use saturating arithmetic where request-derived totals are combined.

### Validation

- `RUSTC_WRAPPER= cargo test --profile test-fast ctx_size_zero`
- `cargo fmt --check`
- `git diff --check`

Issue #1803, opened 2026-09-11
refactor(core): route custom kernels by GPU backend kind instead of treating every non-Metal GPU as CUDA


inureyes:
Part of #1801. Phase 1. Depends on #1802.

## Context

mlxcel's fused kernels pick an implementation with the same idiom in ten places:

```cpp
const bool use_cuda = !mlx::core::metal::is_available();
```

Sites: `src/lib/mlxcel-core/cpp/mlx_cxx_kernels.cpp:166,1483,1991` and `src/lib/mlx-cpp/turbo/fused_rope_append.cpp:421`, `sampling.cpp:412`, `fused_norm.cpp:370`, `paged_attention.cpp:467`, `paged_attention_v2.cpp:600`, `paged_attention_v2_merge.cpp:203`, `sampling_rejection.cpp:779`.

On a ROCm build Metal is unavailable, so every site takes the CUDA branch and calls `fast::cuda_kernel`. The upstream no-CUDA stub throws `[cuda_kernel] No CUDA back-end.` Some callers are gated and fall back to an MLX graph (sampling, rejection sampling, fused add+RMSNorm, fused RoPE+append, SSM kernels via `ssm_kernel_available`, paged decode). Two are not:

- fused MoE (`MLXCEL_FUSED_MOE`, on by default for `mixtral.rs`, `qwen2_moe.rs`, `qwen3_moe.rs` and others), whose bridge function does not return `Result`, so the C++ exception is likely to reach `std::terminate`;
- BitNet `bitlinear_matmul`, which has no fallback.

## Scope

Replace the Metal-or-CUDA boolean with an explicit backend kind and make every custom-kernel site correct on ROCm, using the existing graph fallbacks. Porting kernels to `fast::hip_kernel` is out of scope (#1814).

## Implementation plan

1. Add a single bridge helper, e.g. `enum class GpuKernelBackend { Metal, Cuda, Rocm, None }`, resolved from the build (`MLX_USE_ROCM` / CUDA / Metal compile definitions) and the runtime device. Expose it to Rust through the cxx bridge next to `gpu_backend_available` (`src/lib/mlxcel-core/src/lib.rs:2078`).
2. Convert the ten sites to use it. `Rocm` and `None` take the existing graph fallback wherever one exists.
[...]

PR #1867, opened 2026-09-12
fix: refuse byte-identical arms in the A/B output gate

### LINKED ISSUE #1866: fix: the A/B output-equality gate accepts a baseline and arm that are byte-identical
## Problem / Background

`scripts/ab_output_equality.sh` is the output-equality half of an A/B: it runs the baseline and arm binaries over the same (model, prompt) pairs at `--temp 0 --show-reasoning` and requires byte-identical generated text. It validates that `--baseline` and `--arm` are not the same *path* (`scripts/ab_output_equality.sh:78-82`), never that they differ in *content*, so two paths holding identical bytes pass and every pair reports EQUAL for comparing the unchanged binary against itself.

That is the form the mistake actually takes: the recipe in the script's own header says to build the baseline, copy the binary aside, patch, and rebuild. A `cargo build` that no-ops leaves the copy and the "rebuilt" arm identical. A bridge `.cpp` edit produced that 0.12s no-op on an M5 Max on 2026-09-09 while an M1 Ultra rebuilt the same edit normally, so whether the gate is honest depends on the host it runs on.

Second defect, same file: `usage()` prints the header through a hard-coded `sed -n '2,45p'` window (`scripts/ab_output_equality.sh:53-54`), so a header one line past it silently truncates the documented exit statuses out of `--help`.

## Proposed Solution

Refuse byte-identical arms by default (`cmp -s`, before any generation runs), with an explicit opt-out for a source change that genuinely compiles to the same output, and make the usage window track the real header length instead of a fixed range.

## Acceptance Criteria

- [ ] Two paths with identical bytes exit non-zero before any model loads, naming the likely cargo no-op.
- [ ] The opt-out flag proceeds to the runs; the existing same-path check still rejects.
- [ ] `--help` renders every header line, exit statuses included, with no hard-coded range in `usage()`.

## Verification

```bash
bash -n scripts/ab_output_equality.sh && shellcheck scripts/ab_output_equality.sh
/bin/cp target/release/mlxcel /tmp/mlxcel.copy
./scripts/ab_output_equality.sh --baseline /tmp/mlxcel.copy --arm target/release/mlxcel --model models/mlx/granite-4.0-h-tiny-4bit; echo "exit=$?"
./scripts/ab_output_equality.sh --help | tail -5
```

## Technical Considerations

The Metal half of the last acceptance criterion of #1803 is the next user of this gate; that change touches `src/lib/mlxcel-core/cpp/mlx_cxx_kernels.cpp` and `src/lib/mlx-cpp/turbo/*.cpp`, the same bridge-`.cpp` class that triggered the no-op. Refs #1803

inureyes:
Closes #1866

`scripts/ab_output_equality.sh` rejected the same path for both arms but not the same bytes, and identical bytes at two paths is the form this mistake actually takes. The recipe in the script's own header builds the baseline, copies the binary aside, patches, and rebuilds; when that rebuild no-ops the arm is a byte-identical copy of the baseline, and every pair then reports EQUAL for comparing the unchanged binary against itself. A bridge `.cpp` edit produced exactly that 0.12s no-op on an M5 Max on 2026-09-09 while an M1 Ultra rebuilt the same edit normally, so whether the gate is honest depended on the host it ran on.

Identical bytes are now refused before any generation runs. `--allow-identical-binaries` covers a source change that genuinely compiles to the same output, so the escape hatch exists but has to be asked for.

Second defect in the same file: `usage()` printed the header through a hard-coded `sed -n '2,45p'` window, so a header one line past that range silently truncated the documented exit statuses out of `--help`. It now prints the leading comment block at whatever length it has.

Validated by hand, since this file is tooling and no suite covers it: the guard exits 1 before any generation runs, `--allow-identical-binaries` proceeds to the runs, `--help` renders through the last header line, and `bash -n` plus `shellcheck -S warning` are clean.

Refs #1803, whose last acceptance criterion needs this gate on a Metal host.

### DIFF
--- a/scripts/ab_output_equality.sh
+++ b/scripts/ab_output_equality.sh
@@ -31,6 +31,10 @@
 #
 #   --model / --prompt repeat, and every (model, prompt) pair is checked.
 #
+#   --allow-identical-binaries lets the two arms hold the same bytes. Without it
+#   that is refused, because a `cargo build` that no-ops leaves the copied-aside
+#   baseline and the "rebuilt" arm identical and every pair reports EQUAL.
+#
 # Producing the baseline binary: build at the unpatched commit, copy the binary
 # aside, apply the patch, rebuild.
 #
@@ -46,12 +50,17 @@ set -uo pipefail
 BASELINE_BIN=""
 ARM_BIN="./target/release/mlxcel"
 MAX_TOKENS=128
+ALLOW_IDENTICAL_BINS=0
 OUT_DIR=""
 MODELS=()
 PROMPTS=()
 
 usage() {
-  sed -n '2,45p' "$0" | sed 's/^# \{0,1\}//'
+  # Print the whole header comment block rather than a hard-coded line range.
+  # The range spelling (`sed -n '2,45p'`) silently truncated the last line as
+  # soon as the header grew, which is how the documented exit statuses went
+  # missing from `--help`.
+  awk 'NR == 1 { next } /^#/ { sub(/^# ?/, ""); print; next } { exit }' "$0"
   exit "${1:-0}"
 }
 
@@ -63,6 +72,7 @@ while [[ $# -gt 0 ]]; do
     --prompt) PROMPTS+=("$2"); shift 2 ;;
     -n|--max-tokens) MAX_TOKENS="$2"; shift 2 ;;
     --out-dir) OUT_DIR="$2"; shift 2 ;;
+    --allow-identical-binaries) ALLOW_IDENTICAL_BINS=1; shift ;;
     -h|--help) usage 0 ;;
     *) echo "unknown argument: $1" >&2; usage 1 ;;
   esac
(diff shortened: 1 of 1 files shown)


inureyes:
Reviewed, and it looks right to me. GitHub will not take a formal approval because both PRs go up under the same account, so this is a comment. I reproduced the checks on the branch rather than taking them on trust: `bash -n` is clean, `--help` now renders through the exit-status line (it stopped at the line above before), and the header block ends at line 46 with the first non-comment line at 47, so the `awk` replacement prints exactly the block and nothing past it. I could not run shellcheck; it is not installed here.

The ordering is right in a way that is easy to get wrong. The executability loop runs before the new `cmp`, so the guard always compares two real files rather than failing open on a missing arm. And the guard happens to close a second hole the same-path check misses: that check canonicalises the directory but not the file, so a symlinked or hard-linked arm slips through it, and byte-identity catches both.

The recipe in the header is the one that produces this failure, which makes the guard sit in the right file.

One thing worth saying in the header, not a change request. This bounds one failure mode, not the class. A build that recompiles the Rust but reuses stale C++ objects gives two binaries that differ in bytes, so the guard passes while the change under test is still absent from the arm. The build-log check is what covers that, and it is the part a reader will skip precisely because the script now looks like it guarantees the arms differ. A line saying the guard catches the total no-op and the log is what catches a partial one would stop someone reading more safety into it than it provides.

On the label question: `area:benchmark` reads right to me. The script is the output half of an A/B, and it lives next to the benchmark scripts.

Thank you for the `usage()` find. That one only surfaced because you grew the header, and it would have kept silently dropping the last line every time anyone added to it.

inureyes:
Added the header note in `099afe77`. The wording says the refusal catches a build that no-opped entirely and that the build log is what catches a partial one, so a reader does not take the guard for a guarantee that the arms differ.

Your point about the same-path check is the sharper half of this: it canonicalises the directory but not the file, so a symlinked or hard-linked arm was slipping past it, and byte-identity closes that too. I had only been thinking about the no-op.

`bash -n`, `shellcheck -S warning`, `--help` through the last header line, and the guard exiting 1 before any generation all still hold after the addition.

Issue #1862, opened 2026-09-12
feat(rocm): port the BitNet bitlinear_matmul kernel to mlxcelverse


inureyes:
Part of #1801

## Problem / Background

`bitlinear_matmul` is the only mlxcel fused kernel with no graph fallback. `src/lib/mlxcel-core/cpp/mlx_cxx_kernels.cpp` holds a Metal port (`get_bitlinear_kernel`, around line 76) and a CUDA port (`get_bitlinear_kernel_cuda`, around line 141) and picks between them at `mlx_cxx_kernels.cpp:167`, with nothing to fall back to. On ROCm neither exists, so a BitNet checkpoint used to reach `fast::cuda_kernel`, whose no-CUDA stub throws, and the throw crosses the cxx bridge into a `noexcept` extern and ends the process. The ROCm gate showed this as an abort in `models::bitnet::tests::bitlinear_matmul_known_ternary_case` (`src/models/bitnet.rs:510`).

#1803 does not fix that by adding a kernel. It rejects BitNet checkpoints at load on backends without the kernel, with a message naming the backend, so the failure is a clear refusal instead of a process abort. This issue is the follow-up that makes BitNet actually run on ROCm.

## Why this is separate from #1814

#1814 ports the fused kernels that matter for decode throughput (sampling, rejection sampling, fused add and RMSNorm, RoPE and append, paged attention, fused MoE, SSM, gated delta), each of which already has a working graph fallback, so it is a performance issue gated by profiling. `bitlinear_matmul` is a capability gap: without the kernel the model family does not run at all, and there is nothing to profile against. Ordering it by decode impact alongside the others would rank it last, which is the wrong call for the only path that is refused rather than slow.

## Scope

Add a `fast::hip_kernel` port of `bitlinear_matmul` inside `src/lib/mlx-cpp/patches-rocm/`, or as an mlxcel-side kernel next to the Metal and CUDA ports in `src/lib/mlxcel-core/cpp/mlx_cxx_kernels.cpp`, and remove the load-time rejection that #1803 adds once it lands. `fast::hip_kernel` is already available in the overlay (`src/lib/mlx-cpp/patches-rocm/mlx/fast.h:96`), the counterpart of `fast::metal_kernel` and `fast::cuda_kernel`.

## Implementation plan

[...]

Issue #1806, opened 2026-09-11
feat(quant): backend quantization capability table and load-time convert-or-reject policy


inureyes:
Part of #1801. Phase 1. Depends on #1802. #1807 and #1808 build on the table introduced here.

## Context

The ROCm backend handles quantization modes unevenly (measured on `gfx1151`, see #1801): affine 4/8-bit correct, mxfp8 correct after the dispatch fix in the mlxcelverse ROCm overlay, mxfp4 broken (qmm hang, GPU quantize launch failure at large shapes), NVFP4 unsupported (no group-16 dispatch, no E4M3 scale path).

mlxcel loads these modes through several paths that assume the backend can run them:

- NVFP4: `nvfp4_repack_strategy` (`src/models/sanitize.rs:533-547`) defaults to `DirectTranscode`, which produces native NVFP4 (group 16). The `DenseAffine` route already exists (re-quantizes to affine 4-bit, `sanitize.rs` around 850-868) but only when `MLXCEL_NVFP4_DENSE_REPACK=1` is set. Layers whose `in_dim` does not match the target group are skipped with a "Skipping NVFP4 repack" message (`sanitize.rs:662-709`) rather than failing.
- Vendor FP8 block checkpoints: `src/models/fp8_block.rs` always re-quantizes to mxfp8 (`MXFP8_MODE`, group 32, bits 8).
- mxfp4 checkpoints (for example gpt-oss): loaded as-is; no conversion path.

The only backend-capability policy today is the pre-Ampere CUDA one (`pre_ampere_cuda()` and `bf16_to_f16_at_load()`, `sanitize.rs:1803-1879`).

Direction for this epic: where the backend cannot run a mode, convert it to one it can run; reject at load with an actionable message only when conversion is impossible.

## Scope

A single place that answers "can this backend run quantization mode M natively?", and load-time policy that uses it to convert or reject. The conversions for mxfp8 and mxfp4 themselves live in #1807 and #1808; this issue covers the table, the policy hook, and NVFP4.

## Implementation plan

1. Add a backend quantization capability query (Rust side, fed by the backend kind from #1803 or the build features): per mode, `Native`, `ConvertTo(mode)` or `Unsupported`. Initial ROCm table: affine native; mxfp8 native (after #1807 validates it); mxfp4 `ConvertTo(affine 4-bit)` until #1808 fixes the kernels; NVFP4 `ConvertTo(affine 4-bit)`. Metal and CUDA report exactly today's behavior.
[...]

Issue #1802, opened 2026-09-11
build(rocm): vendor the ROCm backend into mlxcelverse and add the rocm cargo feature


inureyes:
Part of #1801. Phase 0. Gates every other sub-issue.

## Context

The feasibility spike in #1801 built the ROCm backend from `NripeshN/mlx@75915908` on top of mlxcel's pin `ml-explore/mlx@81ba1c6a`, fixed six API-drift breaks, and validated it on `gfx1151` (42/42 op checks, mlx-lm decode at fork parity). The result is a 121-file overlay: the fork's `mlx/backend/rocm/` directory (106 files, already including the drift fixes and the mxfp4/mxfp8 scale-type dispatch fix) and 15 MLX core files merged at `81ba1c6a`:

```
CMakeLists.txt
mlx/CMakeLists.txt
mlx/backend/common/buffer_cache.h
mlx/backend/common/compiled.cpp
mlx/backend/common/compiled.h
mlx/backend/gpu/primitives.cpp
mlx/compile.cpp
mlx/device.cpp
mlx/fast.cpp
mlx/fast.h
mlx/fast_primitives.h
mlx/io/safetensors.cpp
mlx/ops.cpp
mlx/primitives.cpp
mlx/stream.cpp
```

Today `src/lib/mlx-cpp/CMakeLists.txt` fetches MLX at the pin (line 122), applies whole-file overlays in `mlx_apply_source_overlays` (lines 18-82; `patches-cuda/` only when `MLX_BUILD_CUDA`, lines 70-81), then `add_subdirectory`. `src/lib/mlxcel-core/build.rs` picks the backend by `target_os` plus the `cuda` feature, links CUDA in `link_cuda()` (line 492) and resolves architectures in `detect_cuda_arch()` (line 444). Nothing knows about ROCm.

## Scope

Land the overlay and the build glue so that `cargo build --features rocm` produces a working GPU binary on a Linux AMD host. Nothing else in the runtime changes here; kernel routing, memory policy and quantization policy are Phase 1 sub-issues.

## Implementation plan

[...]

Issue #634, opened 2026-07-02
perf(cuda/attn): port the native paged-attention decode kernel to CUDA


inureyes:
Part of #623.

## Context

The native fused paged-attention decode kernel exists only as a Metal kernel: `src/lib/mlx-cpp/turbo/paged_attention.cpp` uses `mlx::core::fast::metal_kernel` (line ~211) with no `cuda_kernel` counterpart. The runtime gate (`src/lib/mlxcel-core/src/layers.rs:2815-2854`, env `MLXCEL_PAGED_ATTENTION_NATIVE`) refuses the native path off-Metal, so CUDA always takes `paged_decode_attention_pooled_fallback`: gather the paged KV blocks into contiguous tensors, then run regular SDPA. That adds a gather pass plus extra memory traffic per step, precisely where long-context decode is bandwidth-bound.

This matters for the server's paged decode storage (`--decode-storage paged`, `DecodeStorageBackend::Paged`, block budget `--kv-cache-budget`), which is the memory-efficient multi-tenant configuration.

A proven porting pattern exists in this codebase: the fused decode-MoE kernel was ported from `metal_kernel` to `fast::cuda_kernel` in `src/lib/mlxcel-core/cpp/mlx_cxx_kernels.cpp` (selection via `use_cuda = !metal::is_available()`, `simd_sum` -> `__shfl_down_sync`, header-source split; see lines ~103-160 and ~1400-1650).

## Scope

CUDA port of the paged-attention decode kernel, gated the same way as Metal, with the gather fallback retained.

## Implementation plan

1. Read the Metal kernel in `src/lib/mlx-cpp/turbo/paged_attention.cpp`: block table indexing, head layout, GQA handling, softmax accumulation scheme, and the pooled-fallback contract (`layers.rs` fallback call) that defines expected outputs.
2. Port to `fast::cuda_kernel` following the #319 MoE pattern: one thread block per (sequence, head-group), warp-parallel over KV block pages, online-softmax accumulation in registers/shared memory, `__shfl_down_sync` reductions replacing `simd_sum`. Respect the existing quant modes if the Metal kernel supports quantized KV pages; if it is fp16-only, keep the port fp16-only and note the interaction with #635.
3. Wire the gate: `layers.rs:2849-2854` currently allows native only on Metal; extend to CUDA when the kernel is compiled in, still honoring `MLXCEL_PAGED_ATTENTION_NATIVE` (default: enabled where available, env=0 forces fallback).
[...]

PR #1869, opened 2026-09-12
refactor(core): route custom kernels by GPU backend kind

### LINKED ISSUE #1803: refactor(core): route custom kernels by GPU backend kind instead of treating every non-Metal GPU as CUDA

inureyes:
Ten sites picked a fused kernel with `const bool use_cuda = !mlx::core::metal::is_available()`, reading "not Metal" as "CUDA". That held while Metal and CUDA were the only GPU backends. With ROCm (#1802) it stopped holding: Metal is unavailable there too, so every site took the CUDA branch and called `fast::cuda_kernel`, whose no-CUDA stub throws. Most callers catch that, but the throw crosses the cxx bridge into a `noexcept` extern wherever they do not, which ends the process in `std::terminate`. On the ROCm gate that was 53 aborts.

`src/lib/mlx-cpp/turbo/gpu_backend.h` replaces the boolean with `GpuKernelBackend` (None, Metal, Cuda, Rocm) and `custom_kernels_available()`. On Metal and CUDA builds the enum resolves to exactly what the boolean did, which is the point: only a ROCm build sees a new value. Four availability predicates spelled `metal::is_available() || cu::is_available()` become the same call.

Each family already had a fallback; what was missing was a way to reach it. `split_kv_enabled()`, the `paged_decode_batched` planner and `fused_moe_enabled()` now decline on a backend without ports, so the MLA absorbed path, the gather-then-SDPA path and SwitchGLU serve those steps. An eleventh site, `paged_decode_backend()` in `layers.rs`, made the same decision in Rust where the issue's grep would not have found it; it now asks the predicate first, so correctness no longer depends on `cuda_is_available()` staying false on ROCm, and its `Other` variant stops describing ROCm as a CPU.

Tests reach three launchers directly, past every gate, so the launchers refuse too. `paged_attention_decode`, `paged_attention_decode_v2_partial` and `paged_attention_merge_states` throw a message naming the backend rather than falling through to whichever port was tried, and their bridge declarations return `Result` so cxx turns the throw into an `Err`. The C++ signatures are unchanged. `paged.rs` already carried a comment from the #634 review saying a bare `UniquePtr` return meant a failed launch would abort rather than error; that is the condition this removes.

`bitlinear_matmul` is the one fused kernel with no graph fallback at all, so a CPU-only build terminated on a BitNet checkpoint just as ROCm did. BitNet checkpoints are now refused at load with a message naming the reason, the op returns `Result`, and its test skips with a printed reason where no port exists. The ROCm kernel port is #1862.

Measured on gfx1151: `make verify-test-rocm` goes from 53 aborts to 2, neither from this issue (nvfp4 group 16 is #1806, and one sampling kill-switch test predates this). 9359 tests pass. `paged_v2`, `mla::split_kv` and `cache::paged_batch_decode` report ordinary failures naming the backend instead of killing the binary.

Not verified here: that Metal and CUDA select the same kernels as before. The enum is equivalent by construction on those backends, but the runners were down and neither host is reachable from this one. The A/B is set up on the Metal side and the CUDA arm is described in a PR comment.

Closes #1803

Validated on gfx1151: `cargo fmt --all -- --check`, `scripts/ci/check_cross_repo_refs.py`, `scripts/insert_apache_header.py --check`, `scripts/ci/check_kernel_dtype_keys.py`, and `make verify-test-rocm`.

On that last checker, one thing is worth writing down because the next refactor here will meet it. It scopes itself by whether a file contains the literal `cuda_kernel(`, so moving a launch call out of these files would take all eight out of scope silently. This PR moves only the selection boolean and leaves the launches where they were, so the scope is unchanged (verified: the eight files still carry the token, and it reports OK). Its glob is `*.cpp` only, so a launch that ends up in a header is invisible to it whether or not the token is there.

### DIFF
--- a/examples/page_gather_microbench.rs
+++ b/examples/page_gather_microbench.rs
@@ -669,6 +669,7 @@ fn run_config(
             scale,
             0,
         )
+        .expect("paged_attention_decode: this microbench needs a backend with the fused kernel")
     });
     let mut rot = Rotation::new(rot_count);
     let fused_v2 = time_body(warmup, iters, || {
--- a/src/lib/mlx-cpp/turbo/fused_norm.cpp
+++ b/src/lib/mlx-cpp/turbo/fused_norm.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "fused_norm.h"
+#include "gpu_backend.h"
 
 #include <mlx/fast.h>
 #include <mlx/ops.h>
@@ -325,7 +326,7 @@ inline FusedNormKernelHolderCuda& get_fused_norm_kernel_cuda() {
 } // namespace
 
 bool fused_add_rms_norm_available() {
-    return mlx::core::metal::is_available() || mlx::core::cu::is_available();
+    return mlxcel::custom_kernels_available();
 }
 
 std::vector<mlx::core::array> fused_add_rms_norm(
@@ -367,7 +368,8 @@ std::vector<mlx::core::array> fused_add_rms_norm(
     // Metal kernel on Apple, CUDA port elsewhere. `fast::metal_kernel` throws
     // "[metal_kernel] No Metal back-end" on the CUDA backend and vice versa;
     // `metal::is_available()` is false on a CUDA-only build.
-    const bool use_cuda = !mlx::core::metal::is_available();
+    const bool use_cuda =
+        mlxcel::gpu_kernel_backend() == mlxcel::GpuKernelBackend::Cuda;
     auto& kernel =
         use_cuda ? get_fused_norm_kernel_cuda().get() : get_fused_norm_kernel().get();
 
--- a/src/lib/mlx-cpp/turbo/fused_rope_append.cpp
+++ b/src/lib/mlx-cpp/turbo/fused_rope_append.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "fused_rope_append.h"
+#include "gpu_backend.h"
 
 #include <mlx/fast.h>
 #include <mlx/ops.h>
@@ -372,7 +373,7 @@ inline FusedRopeKernelHolderCuda& get_fused_rope_kernel_cuda() {
 } // namespace
 
 bool fused_rope_qk_append_available() {
-    return mlx::core::metal::is_available() || mlx::core::cu::is_available();
+    return mlxcel::custom_kernels_available();
 }
 
 std::vector<mlx::core::array> fused_rope_qk_append(
@@ -418,7 +419,8 @@ std::vector<mlx::core::array> fused_rope_qk_append(
             "[fused_rope_qk_append] qkv trailing dim does not match the head geometry.");
     }
 
-    const bool use_cuda = !mlx::core::metal::is_available();
+    const bool use_cuda =
+        mlxcel::gpu_kernel_backend() == mlxcel::GpuKernelBackend::Cuda;
     auto& kernel = use_cuda ? get_fused_rope_kernel_cuda().get()
                             : get_fused_rope_kernel().get();
 
--- a/src/lib/mlx-cpp/turbo/gpu_backend.cpp
+++ b/src/lib/mlx-cpp/turbo/gpu_backend.cpp
@@ -1,0 +1,91 @@
+// Copyright 2025-2026 Lablup Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "gpu_backend.h"
+
+#include <mlx/backend/cuda/cuda.h>
+#include <mlx/backend/metal/metal.h>
+
+// `mlx/backend/rocm/rocm.h` ships in the ROCm overlay
+// (`src/lib/mlx-cpp/patches-rocm/`), which CMake copies into the MLX source
+// tree only for a ROCm build, so the include has to be gated the same way the
+// Metal-backend-only symbols are. build.rs defines this alongside
+// MLXCEL_BRIDGE_METAL_BACKEND.
+#ifdef MLXCEL_BRIDGE_ROCM_BACKEND
+#include <mlx/backend/rocm/rocm.h>
+#endif
+
+#include <cstdio>
+#include <cstdlib>
+
+namespace mlxcel {
+
+namespace {
+
+GpuKernelBackend resolve_gpu_kernel_backend() {
+  if (mlx::core::metal::is_available()) {
+    return GpuKernelBackend::Metal;
+  }
+#ifdef MLXCEL_BRIDGE_ROCM_BACKEND
+  // Checked before CUDA: on a ROCm build `cu::is_available()` is the no-CUDA
+  // stub and returns false, but ordering it this way keeps the answer right if
+  // a future build ever links both.
+  if (mlx::core::rocm::is_available()) {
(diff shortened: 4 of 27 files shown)


inureyes:
## Running the CUDA arm on a GB10 node

This is the half of the last acceptance criterion that cannot be checked from the machine this branch was written on. It needs a CUDA host, and neither that host nor a GB10 node is reachable from here or from the Metal host running the other arm. Everything below is self-contained: it assumes only a GB10 node with a CUDA toolchain and a clone of this repository, and no context from this PR.

What it is checking: that this branch selects exactly the same kernels on CUDA as `main` does. The change replaces `const bool use_cuda = !mlx::core::metal::is_available()` with an enum that resolves to `Cuda` on a CUDA build, so the two should be equivalent by construction. "Should be" is the reason for the check.

### 1. Build both arms

`main` is the baseline, this branch is the arm. The merge base is `9ead2b7d`, so `main` at or after that commit is the right baseline.

```bash
git fetch origin
git checkout 9ead2b7d
cargo build --release --features cuda --bin mlxcel
/bin/cp target/release/mlxcel target/release/mlxcel.before

git checkout feature/issue-1803-kernel-backend-kind
cargo build --release --features cuda --bin mlxcel
```

Check the second build's log before going on. It must show `mlx_cxx_kernels.cpp` and the files under `src/lib/mlx-cpp/turbo/` actually recompiling. A `cargo build` that reports a fraction of a second and rebuilds nothing after a bridge `.cpp` edit has been observed on this project, and it makes the arm binary secretly identical to the baseline, at which point every comparison below passes for the wrong reason.

### 2. The test gate

This is the stronger of the two checks, because the tests that exercise these launchers call them directly.

```bash
git checkout 9ead2b7d && make verify-test-cuda 2>&1 | tee /tmp/cuda-before.log
git checkout feature/issue-1803-kernel-backend-kind && make verify-test-cuda 2>&1 | tee /tmp/cuda-after.log
grep '^test result:' /tmp/cuda-before.log | tail -5
grep '^test result:' /tmp/cuda-after.log  | tail -5
```

Pass condition: the same tests pass in both, and no test that passed before fails after. A new failure in `paged_v2::*`, `mla::split_kv` or `cache::paged_batch_decode` is the signal that matters; those are the paths this PR changed.

### 3. Greedy output equality

```bash
./scripts/ab_output_equality.sh \
    --baseline target/release/mlxcel.before \
    --arm target/release/mlxcel \
    --model models/mlx/Qwen3-30B-A3B-4bit \
    --model models/mlx/Meta-Llama-3.1-8B-Instruct-4bit \
    --prompt "Explain mixture-of-experts routing in two sentences." \
    -n 64
```

The script pins `--temp 0` and `--show-reasoning` on both arms and runs the baseline twice as a self-consistency control, so it reports INCONCLUSIVE rather than FAIL when the baseline does not reproduce itself. Exit 0 means every pair matched.

If it refuses to start saying the two binaries are byte-identical, that is the no-op build from step 1, not a script problem. Rebuild and check the log.

### 4. The paged launchers, if you have time for one more

Steps 2 and 3 leave a gap: `mlxcel generate` has no paged-cache flags, and the fused paged decode kernel is off by default, so the generate arm never reaches `paged_attention.cpp` or `paged_attention_v2*.cpp`. The test gate in step 2 does cover them, which is why it is listed first. To exercise them through a server instead:

```bash
MLXCEL_PAGED_ATTENTION_NATIVE=1 ./target/release/mlxcel-server \
    -m models/mlx/Meta-Llama-3.1-8B-Instruct-4bit --kv-unified --port 8080
```

Then send a few chat completions long enough to build context, and compare against the same run on the baseline binary. `MLXCEL_PAGED_ATTENTION_NATIVE=1` forces the fused path past the token floors that would otherwise decline it.

Worth watching rather than assuming: that the fused kernel actually ran. This PR makes the paged path ask a new predicate before dispatching, and if that predicate ever answered wrongly on a host that has the kernels, the path would fall back to gather. The output would be identical and only the speed would change, so the equality check in step 3 cannot see it. `/slots` and the decode-path counters report which path served each step.

### What to report back

The two `test result:` summaries from step 2, the script's exit status and verdict lines from step 3, and the `cargo build` timing from step 1 that shows the C++ actually recompiled. A plain "all equal" is not enough on its own, because the most likely failure mode here produces exactly that.

inureyes:
Superseded by #1877. GitHub closed this automatically when the FFT merge deleted its base branch, and a closed PR whose base is gone cannot be reopened or retargeted. Same branch, rebased onto the new main so it carries only the routing commit.

Issue #1834, opened 2026-09-12
epic: ship a bundled macOS-style model control interface


inureyes:
## Goal

Ship a self-contained, local-first WebUI that feels at home on macOS 27 while remaining a browser application on supported platforms. Balance convenience (discover → load → chat), control (explicit model lifecycle and scoped settings), and information (honest readiness, resource use, context, and request timings). This issue defines the product and integration contract; child issues define bounded ownership and executable acceptance criteria. No native wrapper is required.

## Reviewed baseline and reuse decisions

Reviewed on 2026-09-12: `lablup/mlxcel` main `1f1077101151f454450a67962ae07f06362a68f3`; reference `lablup/continuum-router` local checkout `c79e54d0f9920953c417acd57b0a60421e485c93`. Findings are source review, not a new runtime benchmark. Rebase the implementation survey when work starts; issue closure alone is not evidence that code landed.

| Existing surface | Evidence | Decision / gap |
|---|---|---|
| Both server entry points | `src/bin/mlx_server.rs`, `src/main.rs`, `src/commands/serve.rs`, `src/server/cli_input.rs`, `startup.rs` | Primary command is `mlxcel-server --webui`; `mlxcel serve --webui` uses identical shared resolution, not a second server. |
| UI flag compatibility | `src/cli/ui_compat_args.rs`, closed #1435, `compat/llama-server/b10621/` | `--webui` currently aliases hidden `--ui` and is intentionally rejected. Replace that classification, not add a conflicting Clap argument. |
| Model-free router and management | `router_server.rs`, `router_models.rs`, `router_cache.rs`, `router_presets.rs` | Reuse the in-process RouterPool, cache, presets, load/unload/download/remove, and model SSE. Do not build a process supervisor or a second model registry. `router_front.rs` is a different, distributed tokenizer-bearing service and is NOT this foundation. |
| Discovery | `discover_models`, downloader store, `src/models/registry.rs`, detection, completeness helpers | `--models-dir` scans immediate model children, cache uses owner/name, source precedence is cache < models-dir < preset. Downloaded, architecturally supported, complete, runnable on this backend, and loaded are different facts. `mlxcel list` is not a support catalog. |
[...]

PR #1872, opened 2026-09-12
feat: add safe WebUI model library operations

### LINKED ISSUE #1841: feat(webui): track safe model downloads and cache deletion
## Parent / dependencies

Parent: #1834. Depends on #1835, #1839, #1840.

## Ownership

Own the downloader/UI adapter, `router_cache.rs` integration, bounded download operation progress and cache deletion tests. Existing downloader owns transport, filters, completeness and store containment; no separate shell downloader. Coordinator owner retains lifecycle/event types and shared locks.

## Required behavior

- Accept explicit public HuggingFace repo ID and optional revision, not arbitrary URL/local path. Resolve metadata/revision once and pin the download to that immutable commit; carry revision through existing downloader rather than inventing a different cache layout. Display expected destination and known bytes before transfer; unknown size remains unknown. Never execute remote modeling code.
- Use fixed configured store and current offline policy; no requests on catalog browsing. Honor configured offline mode with a typed disabled reason. Private/gated repos return actionable unsupported/auth-required guidance without a token entry form in v1. Do not silently read/browser-persist Hub credentials.
- Reuse download hooks but expose the common operation ID, monotonic received bytes, optional total/rate and real terminal result. Start with one active transfer and a bounded queue per server; no busy polling or network under pool locks. Duplicate repo/revision requests deduplicate; immutable revision mismatch must not overwrite a loaded model.
- Cancel is cooperative and becomes terminal only after writer exit; preserve or clean partial data under an explicit policy. Interrupted/incomplete snapshots never become runnable because config.json happened to arrive first. Publish completion atomically after existing completeness validation; define retry/resume and server-restart reconciliation.
- Cache deletion is a separate confirmed operation, never ordinary unload. Enforce managed-cache source, recheck canonical containment and revision immediately before deletion, refuse busy/loading/downloading models until drained/cancelled, and report actual disk deletion result. Preserve user models-dir/preset files, shared/external artifacts and active download targets.

## Acceptance criteria

- [ ] Fake HTTP/downloader tests cover valid download, bad repo/revision, gated/404, offline, timeout, disconnect, disk full, checksum/completeness failure, cancel/retry, restart, duplicate action and bounded queue saturation.
- [ ] Test symlink swap/traversal/owner-name sanitization and removal races; cancellation cannot publish ready and deletion cannot race a writer or streaming provider.
- [ ] One opt-in real public small-checkpoint download completes into the configured store, appears once, loads and generates; record repo+revision. CI hermetic tests do not require external network or GPU.
- [ ] Existing compatibility download/removal endpoints call the same primitives and preserve their wire schema; UI-specific 202/progress semantics stay in the versioned adapter.

inureyes:
## Summary
Adds authenticated WebUI downloads and managed-cache removal using the existing pool/coordinator: anonymous immutable-revision downloads, bounded admission/progress/cancellation, descriptor-anchored publication/deletion and shared bootstrap/admission policy. Single-model mode remains read-only; external/user-root/preset aliases and subtree overlaps are rejected. Includes the unload owned-drain revision fix and Linux device-identity portability repair.

## Verified acceptance
Runtime source `84f1aeb2b4f5706bb0d63866b20ca48d74bf340a` is based on main `e9ee5f04`; the final report commit changes no source or bundle.

- Full workspace: 11,313 unique top-level tests passed, zero failed, 361 ignored; two nested restart-child passes counted separately. Workspace all-target clippy passed default and feature-disabled configurations with warnings denied.
- All 46 contracts, structural/format checks, both binaries' production relocated empty/offline/TTY checks, 76 frontend tests/type/lint and deterministic bundle verification passed.
- Unchanged strict fresh-cache driver passed pinned public SmolLM download/hash, production restart/new identity/old-operation 404/stable catalog ID, real load and nonempty 32-token generation, observed worker-exit unload, API deletion of only the new checkpoint and empty catalog. Repetitive output is not a quality claim. Both owned servers exited 0 without force.
- Actual hosted CI run 34844128193 passed, including Linux clippy and feature-off checks; no GB10 exception was needed. Workflow-skipped CUDA/OpenXLA link jobs are not execution claims.

## Evidence boundaries
The original `3cb4817d` unload failure and `857937ca` firmware timeout remain preserved. Later success after user-confirmed GPU availability establishes neither their external cause nor a reboot. ENOSPC coverage is injected, not physical disk exhaustion. Safari/VoiceOver/native 200% checks remain user-deferred until the whole implementation is ready, not passed.

Bilingual technical reports record architecture, reviews, failures and final evidence. PR remains in review pending central merge.

Refs #1834; Closes #1841


inureyes:
## Rebase needed onto `main` after #1889

#1889 has merged as `b395a9e5` and restores the `OpenXLA feature compile` job, which had failed on `main` since 2026-09-12 16:12 and is the failing check on this PR too.

The job's second step runs `cargo check --no-default-features --features xla-diagnostics --all-targets` under `RUSTFLAGS="-D warnings"`. Because `default = ["surgery", "webui"]`, it builds with the WebUI off, and the WebUI series left the handlers, validators, constants, request types and imports behind its routes outside `#[cfg(feature = "webui")]`, so each became a dead-code error. #1889 gates all of them on `webui`.

This PR touches `src/server/router_server.rs`, which #1889 edited in its import block and in front of 35 items, so expect conflicts there. Two things to carry through the rebase so the check stays green:

- Keep the new gates on the items #1889 gated, including the split imports (`HeaderMap`, `Uri`, `AxumPath`, `RouterModelAction`, and the whole `router_lifecycle` import).
- Gate anything this PR adds that only webui routes use. The quickest way to find leftovers is `cargo check --no-default-features --features cuda --lib --tests`, which reports the same set of dead-code warnings that `-D warnings` turns into errors in CI, without needing the IREE build. A clean result there plus `cargo check --features cuda --lib --tests` with defaults on covers both sides of the gate.

Issue #1873, opened 2026-09-13
test(cuda): run the deferred CUDA arm for the merged ROCm routing work on a GB10 node


inureyes:
Part of #1801

## Problem / Background

The changes listed below are already merged to `main`. This issue is a verification pass, not a merge gate: it can find a regression that already shipped, and a failure here means opening a fix issue against the specific change, not reverting the set blindly.

The CUDA arm was deferred at merge time because no CUDA host was reachable from the ROCm development machine or from the Metal host, and the self-hosted CUDA CI runner has been down (its jobs sat queued and expired at the 24 hour timeout). Rather than leaving an implied "someone should check CUDA" note in each PR, the whole deferred set is collected here to be run in one pass when a GB10 node is available.

## What needs the CUDA arm

| Change | Why CUDA is at risk | What would go wrong |
|---|---|---|
| `refactor(core): route custom kernels by GPU backend kind` (#1803, PR #1869) | Ten call sites that chose a fused kernel with `!mlx::core::metal::is_available()` now ask a `GpuKernelBackend` enum, and four availability predicates now call `custom_kernels_available()`. On a CUDA build the enum resolves to `Cuda` and the predicates to true, so the selection is equivalent by construction. | If it is not, CUDA silently takes a Metal kernel or a graph fallback. A fallback keeps the output identical and only changes speed, so an output comparison alone cannot see it. |
| The same PR's `Result` change | `paged_attention_decode`, `paged_attention_decode_v2_partial` and `paged_attention_merge_states` now return `Result` through cxx instead of a bare value, and their launchers refuse when the backend has no port. | The refusal branch is unreachable on CUDA, but the `Result` wrapping is new code on a hot path. |
| `paged_decode_backend()` in `src/lib/mlxcel-core/src/layers.rs` | It now asks `custom_kernels_available()` before Metal and CUDA. | If that predicate ever answered false on a CUDA host, every fused paged decode would fall back to gather with identical output and lower throughput. |
| `feat(rocm): port the BitNet bitlinear_matmul kernel` (#1862, PR #1870) | The kernel selection gained a third arm and a new `bitlinear_kernel_available()` predicate. | CUDA must still pick the CUDA kernel, and the BitNet tests must run rather than skip. |
[...]

Issue #1805, opened 2026-09-11
feat(rocm): memory estimation, device info and hardware detection on AMD UMA hosts


inureyes:
Part of #1801. Phase 1. Depends on #1802.

**Measured 2026-09-14** on the spike host (Radeon 8060S, `gfx1151`, ROCm 10.0.0), at `d67b64d3` on `main`. The measurements below replace what this issue originally guessed; start from them, not from the guess.

## Context

`resolve_available_memory` (`src/execution/memory_estimate.rs:725`) walks `MLXCEL_MEMORY_LIMIT`, then the MLX allocator cap `memory_limit()`, then `HardwareCapabilities::unified_memory_gb`, then `/proc/meminfo`. What that walk actually sees on this host:

| Call | Value |
|---|---|
| `mlxcel_core::memory::memory_limit()` | 82463372083 bytes (76.80 GiB) |
| `gpu_max_memory_size()` | 103079215104 bytes (96.00 GiB) |
| `get_wired_limit()` | 0 |
| `HardwareCapabilities.unified_memory_gb` | 0 |
| `HardwareCapabilities.silicon_gen` | `Unknown` |

The ROCm allocator reports a nonzero `memory_limit()`, so the walk returns at its second step with 76.80 GiB and never reaches `/proc/meminfo`. The 28 GiB figure this issue was filed to worry about is not what the estimator sees, and nothing that fits in the carve-out is refused for that reason. `gpu_max_memory_size()` falls through to `total_memory` and reports the full 96.00 GiB as predicted (`src/lib/mlxcel-core/cpp/mlx_cxx_bridge.cpp:4630`), and the wired limit is a no-op exactly as on CUDA (`:4588`). Resolving available memory from `device_info()` ahead of `/proc/meminfo` is therefore **not needed for correctness**; anyone who still wants it has to argue it as a policy choice (the allocator's 76.80 GiB default cap against the device's 96.00 GiB) and not as a fix.

Two failures are real and live instead. Both are in the `mlxcel-core` library test binary and both occur under `make verify-test-rocm`, but that run aborts on an unrelated nvfp4 case (`terminate called ... Unsupported group_size for QuantizedMatmul: 16`, issue #1806) before the harness prints its failure list, which is why reading the gate output misses them. Run each test directly to see it:

[...]

PR #1881, opened 2026-09-14
docs: record #1803 Metal gate, add paged decode counter A/B

inureyes:
Records the Metal regression gate for #1803 and lands the harness for its paged arm, so #1873 can run the CUDA half without rediscovering how.

`docs/benchmark_results/kernel-backend-kind-metal-m1ultra-2026-09-13.md` covers four arms: the backend probe #1803 itself ships, greedy byte-identity across six configurations, the batched paged decode counters, and the workspace gate (11119 passed, 0 failed). Verdict is that Metal selects the same kernels. It also names what was not measured, including the sampler's two-arm diff, which greedy decode cannot reach.

The reusable part is the two runs that passed for the wrong reason. Five checkpoints reported 5/5 EQUAL while three never reached a changed decision point, because `FUSED_ADD_RMSNORM_DEFAULT` and `FUSED_ROPE_APPEND_DEFAULT` are opt-in, `falcon-mamba` does not use `ssm_update_kernel`, and `split_kv_enabled()` short-circuits on `absorbed_enabled()`. And the paged arm compared gather against gather because the dispatch floor declined both sides. Each cost a re-run; each recurs unchanged on CUDA.

`scripts/paged_decode_counter_ab.sh` encodes that. Output equality cannot judge the batched paged path, since every gate on it chooses between the fused kernel and a gather fallback computing the same thing, so a disabled fused path is byte-identical and silent. The discriminator is `gather_fallbacks`, not `declines`, which is never exported to the gauges. The script also refuses the vacuous run where neither side launched the fused kernel.

Validated end to end against the two binaries the record describes, reproducing its counters exactly. shellcheck clean, `insert_apache_header.py --check` and `check_cross_repo_refs.py` pass. No product code changes.

Refs #1803, #1873, #1877


inureyes:
Read the script and the report. The shape is right, and the part I would have got wrong is the part you got right: refusing a run where neither side took the fused path, instead of reporting EQUAL. That is the failure your own first measurement produced, and a comparison tool that cannot fail for the right reason is worth less than no tool.

Checked here: `bash -n` clean, the four traps are enforced rather than documented, `basename "$MODEL"` matches what `/v1/models` reports as the id so the required `model` field is populated, and the discriminator is `gather_fallbacks`.

Three things from the ROCm and GB10 side, none blocking.

`jq` is a hard dependency and there is no preflight for it. It is not installed by default on a minimal Debian or on the ROCm container images I have seen, and the failure lands mid-run after the server is up, so the operator pays a model load before finding out. Either check for it next to the other argument validation, or build the request with a here-doc, since the body is four fixed fields and one prompt string. The prompt would need escaping, which is the reason to prefer the check.

The requests are sequential, so the launch is a batch of one even with `--max-batch-size 4`. That is enough for `paged_decode_batched` and it is what produced your numbers, so the run is sound. It does mean the cascade path is never exercised: `cascade=0` in the report is not evidence that cascade is unchanged, only that nothing asked for it. Worth a line in the report, since a reader looking at three counters will take all three as measured. If you want cascade covered, the requests have to overlap in time and share a prefix.

The 180 second readiness wait is generous for an 8B on a warm page cache and thin for a first load from cold storage on a machine nobody has touched, which is the GB10 case. A knob would do; so would saying in the usage text that a cold first load may need it raised.

One question rather than a finding: on a failure path the script kills the server and returns, and the `trap` cleanup runs on exit. If the server is mid-load when the kill lands, does it leave the port held long enough that an immediate second run collides? I ask because `PORT` is a fixed 18803 and the natural next action after a failure is to rerun.

On the vendor design, agreed, and the invariant test is the part I would not have thought to add. Writing the lesson in a comment and fixing it in a test are different things: I have put the reasoning in comments twice today and both times the next person who needs it is me, later, with the comment scrolled off. A test that fails when someone adds a vendor and forgets is the version that survives.

I will take `is_apple_silicon()` plus the `vendor` field into #1805 as you describe, with the eleven sites converted mechanically so the diff shows the predicate is the same comparison, and the AMD arm added after.

Issue #1710, opened 2026-09-08
fix(phi,stablelm): port the upstream f32 attention guard


inureyes:
Upstream promotes the attention inputs to f32 in phi, phixtral and stablelm; mlxcel had ported that guard only into `phixtral.rs`. `phi.rs` and `stablelm.rs` ran their scores at the input dtype.

Without it `phi-2-4bit` goes NaN at layer 29 of 32 in f16, and every token after that is the argmax of a NaN row, which decodes as `!`. It is reproducible in bare MLX with no mlxcel in the loop: the same forward with a f32 residual stays finite through all 32 layers and picks `' a'` for "Paris is", and with a f16 residual it goes NaN. `phixtral.rs`'s own comment already records the same failure ("NaN from layer 30 of 32"), so the family had been diagnosed once and only one of its files was fixed.

Nothing caught it because a second defect was masking it. `gelu_approx` in the cxx bridge returned f32 for a half-precision input, which widened the whole residual stream from the first MLP onward and kept the scores out of f16 range by accident. Fixing that exposed this.

Upstream reference: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/phi.py and https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/stablelm.py, `queries.astype(mx.float32)` into the attention call and `.astype(values.dtype)` out of it; stablelm widens the keys as well.

The guard costs throughput, and the honest accounting is that `phi-2-4bit`'s faster number was the speed of a broken model. On M1 Ultra it reads 164.10 tok/s with the widened attention removed and 129.93 with it, against 128.55 before any of this work. `stablelm-1.6b-4bit` goes 245.40 to 209.32, which is 98% of mlx-lm on the same host, the reference paying the same cost; its previous 115% was the guard not being paid. Measured headroom on stablelm is 11.8x against f16's ceiling on a 16-token prompt, worst `|q·k^T|` 5534.5, so its guard is precautionary. Phi's is not.

Two follow-ups. The other families upstream guards this way should be checked against their mlxcel ports: mlx-vlm applies it in minimax_m3_vl, paligemma, phi, phixtral and nemotron_labs_diffusion. And a f16 range assertion would catch the next one, since the symptom is a silent NaN and every parity check passes right up to the layer that overflows.

[...]

PR #1882, opened 2026-09-14
test(core): assert attention scores stay inside the f16 range

### LINKED ISSUE #1830: test(core): assert attention scores stay inside the f16 range
## Problem / Background

lablup/mlxcel#1710's second follow-up. `phi-2-4bit` ran its attention scores in f16 for months, overflowed 65504 at layer 29 of 32, and emitted `!` for every token after that, and nothing in the tree failed. Layers 0 through 29 tracked the reference to three decimal places first (`src/models/phixtral.rs:490-494`), so a parity check that samples an early layer or a short prompt passes right up to the layer that overflows. The defect was additionally masked by lablup/mlxcel#1709: `gelu_approx` returned f32 for a half-precision input, which widened the residual stream from the first MLP onward and held the scores in range by accident. Fixing that exposed it. This issue asks for a guard that fails loudly instead.

## Current Behavior

No check on attention scores or on decode logits exists in the library. Every `is_finite` under `src/` is either host-side scalar validation of CLI and server settings (`src/main.rs:811`, `src/server/startup.rs:890`, `src/server/runtime_settings.rs:832`) or an assertion inside a test (`src/embeddings/pooling_tests.rs:67`, `src/lib/mlxcel-core/src/layers.rs:7717-7728`). The closest existing thing is whole-model and lives in `tests/`: `tests/mamba2_hybrid_finite.rs` decodes a real checkpoint 250 times and asserts the max logit is finite, and `tests/mamba2_hybrid_decode_finite.rs` does the same for the M5-only SSM graph. Neither covers attention scores and neither runs for the phi or stablelm families.

The scores themselves are never materialized in mlxcel. `src/lib/mlxcel-core/src/layers.rs:4336` (`attention`), `:4704` (`attention_from_ptr`) and `metal4_causal_attention` all hand `q`, `k` and `v` to MLX's fused SDPA, so there is no `q @ k^T` array in mlxcel's own graph to assert on. Anything that inspects the scores has to recompute them.

## Proposed Solution

Two pieces, because the thing that discriminates and the thing the title asks for are not the same artifact.

**A. A real-checkpoint finiteness test that fails when the guard is gone.** New `tests/attention_f16_range.rs`, built on `tests/mamba2_hybrid_finite.rs` and `tests/common/mod.rs`'s `repo_model_dir`. One `#[test]` per f16 family that has a checkpoint in the store: `phi-2-hf-4bit-mlx`, `stablelm-2-1_6b-chat-4bit`, `phixtral-4x2_8-4bit` (all three confirmed present under `models/mlx/` and all three F16 in their safetensors headers). Each prefills a fixed prompt, decodes at least 32 tokens, and after every step asserts the max logit of the sampled row is finite, naming the family and the step index in the failure message. A missing checkpoint prints a loud skip, exactly as `tests/mamba2_hybrid_finite.rs` does. These tests are **not** `#[ignore]`d: unlike the M5-only SSM case, #1710's overflow reproduces on any Apple Silicon host and in bare MLX, so a CI run without the checkpoints is the only reason to skip.

**B. A score-range probe so "inside the f16 range" is measurable.** New `pub fn max_abs_attention_score(q: &MlxArray, k: &MlxArray, scale: f32) -> f32` in `src/lib/mlxcel-core/src/layers.rs`, placed beside `attention` and documented as a diagnostic that no decode path may call, plus `pub const F16_MAX: f32 = 65504.0` in the same module. It computes the scores at the dtype it is handed, deliberately unwidened, as `item_f32(max_all(abs(matmul(q, transpose_axes(k, &[0, 1, 3, 2])))))` scaled by `scale`. All five helpers already exist in the bridge: `abs` (`src/lib/mlxcel-core/src/lib.rs:251`), `matmul` (`:427`), `transpose_axes` (`:445`), `max_all` (`:384`), `item_f32` (`:171`). A unit test in `layers.rs`'s existing `mod tests` (`:5769`) pins it on synthetic `q` and `k` whose worst product is known in closed form, with one case below and one above 65504, so the helper's own arithmetic is not taken on faith. This is what makes #1710's headroom numbers (stablelm 11.8x against the ceiling, worst `|q·k^T|` 5534.5 on a 16-token prompt) reproducible instead of a one-off measurement.

Rejected, with the reason, so it is not relitigated: **a runtime guard on the decode path, in any form, including `debug_assert!`**. Reading a score or a logit requires `mlxcel_core::eval` and a host readback, which is a per-layer GPU sync. That boundary already existed once and was removed: `falcon_h1` and `granitemoehybrid` forced an eval at the end of every mixer forward and it cost 3.31x and 2.24x of decode on M5 Max (`tests/mamba2_hybrid_decode_finite.rs:14-21`). A `debug_assert!` pays the same sync in every debug build, including the `dev`-profile test targets. If an always-on check is ever wanted, the only acceptable shape is an env-gated graph-only side computation that costs nothing when unset, the way `MLXCEL_SPARSE_V_COUNT` is built (`docs/environment-variables.md:375`); that is explicitly not this issue.

## Scope

**In scope:** `tests/attention_f16_range.rs` (new), `src/lib/mlxcel-core/src/layers.rs` (the probe helper, the constant, and its unit test).

**Out of scope:** any change to the guard in `src/models/phi.rs`, `src/models/stablelm.rs` or `src/models/phixtral.rs`; the per-family audit of which families need the guard, which is lablup/mlxcel#1829; an env-gated runtime diagnostic.

## Implementation Notes

- **Reuse.** `tests/common/mod.rs`'s `repo_model_dir` resolves a checkpoint across `models/`, `models/mlx/`, `models/mlx-big/` and the sibling internal tree. Do not hand-roll a path. Its doc comment (`:18-34`) records why: a wrong root turns every real-model test into a no-op that still reports `ok`, which is how the NaN guard from #1718 came to cover nothing on either machine.
- **Reuse.** The decode loop shape in `tests/mamba2_hybrid_finite.rs:30-52` (`load_model`, `LanguageModel::make_caches`, `forward`, `eval`, `slice` the last row, `max_all`, `item_f32`). Copy it rather than writing a second one.
- **Constraint.** Nothing added here may be reachable from `Generator`'s decode loop. `max_abs_attention_score` must have no caller under `src/models/` or in `generate.rs`.
- **Edge case.** bf16 checkpoints. bf16 carries f32's exponent range, so `F16_MAX` is the wrong threshold there and the test asserts finiteness, not a 65504 bound. Say so in the test's module doc so nobody later extends the bound to a bf16 family.
- **Edge case.** A checkpoint absent from the host. Skip loudly with the resolved path in the message, and verify the test actually runs at least once with `MLXCEL_REQUIRE_MODELS=1`, which turns an unresolvable name into a panic.
- **Edge case.** One test function per model. `tests/mamba2_hybrid_finite.rs:66-69` records that running another checkpoint first changed the outcome and made a combined test pass without the fix.
- **Error handling.** The assertion message must carry the observed value and the family and step, not just "not finite": the whole point is that the failure is otherwise indistinguishable from garbage output.

## Acceptance Criteria

- [ ] `tests/attention_f16_range.rs` exists with one `#[test]` per family, passes on a host holding `models/mlx/phi-2-hf-4bit-mlx`, `models/mlx/stablelm-2-1_6b-chat-4bit` and `models/mlx/phixtral-4x2_8-4bit`, and skips loudly where a checkpoint is absent.
- [ ] Deleting the f32 widen from `src/models/phi.rs:199` makes `tests/attention_f16_range.rs`'s phi test **fail**, with a message naming phi and the step. This is the only proof the test discriminates, and it must be reported in the PR with the failing output quoted. Note that reverting commit `3576d734` is **not** this experiment: `3576d734` narrowed an already-present guard from `{q, k, v}` to `{q}`, so reverting it restores a wider guard and the model stays finite. The guard entered in `f4ecc926` (lablup/mlxcel#1709); removing the single `astype` line is the discriminating change.
- [ ] The same deletion applied to `src/models/stablelm.rs:197-198` is also reported, whether or not it fails: #1710 measured stablelm's headroom at 11.8x, so a pass there is the expected result and is itself information about what the test does and does not cover.
- [ ] `mlxcel_core::layers::max_abs_attention_score` and `mlxcel_core::layers::F16_MAX` exist, are documented as diagnostics, and have no caller under `src/models/` or in `src/lib/mlxcel-core/src/generate.rs`.
- [ ] A unit test in `layers.rs`'s `mod tests` pins `max_abs_attention_score` against a hand-computed value in both a below-ceiling and an above-ceiling case.
- [ ] Decode throughput on `phi-2-4bit` is unchanged by this issue, confirmed by a before/after measurement, because nothing here may reach the decode path.
- [ ] The full workspace gate passes.

## Verification

```bash
cargo fmt --all -- --check
cargo clippy --workspace --all-targets --features metal,accelerate -- -D warnings

# The new test, forced to resolve its checkpoints rather than skip them.
MLXCEL_REQUIRE_MODELS=1 cargo test --profile test-fast --features metal,accelerate \
    --test attention_f16_range -- --test-threads=1 --nocapture

# The discrimination proof. Delete the `astype(&q, ... FLOAT32)` line at
# src/models/phi.rs:199 by hand, run the test, then restore that one file.
MLXCEL_REQUIRE_MODELS=1 cargo test --profile test-fast --features metal,accelerate \
    --test attention_f16_range -- --test-threads=1
git checkout HEAD -- src/models/phi.rs
```

A pass is the three family tests green with the guard present, the phi test red with the widen deleted, and the restore confirmed by `git status` showing no modified file. Restore by path only; do not use `git stash`, `git reset --hard` or `git checkout -- .`, which would also discard unrelated work in the tree.

## Technical Considerations

Priority is medium rather than high because the three families known to need the guard already carry it after `3576d734`: this is a regression guard and a measurement tool, not a live defect. What raises it above low is that the failure mode is silent and late, and lablup/mlxcel#1829 may add the guard to further families, each of which then needs a test in this file.

Related: lablup/mlxcel#1710 (the fix and the request for this assertion), lablup/mlxcel#1709 (`f4ecc926`, which both masked and exposed the defect), lablup/mlxcel#1829 (the per-family audit), `tests/mamba2_hybrid_finite.rs` and `tests/mamba2_hybrid_decode_finite.rs` (the precedent for the test shape).

inureyes:
`tests/attention_f16_range.rs` decodes 32 steps on each f16 family carrying the f32 attention widen (phi, stablelm, phixtral) and checks every step, so a regression in that guard fails the build instead of emitting garbage. `max_abs_attention_score` and `F16_MAX` in `layers.rs` make the headroom measurable: mlxcel hands q, k and v to MLX's fused SDPA and never forms the scores, so anything inspecting them has to recompute them. Both are diagnostics with no caller under `src/models/` or in `generate.rs`.

## The discrimination experiment

Required by #1830: a test that cannot fail proves nothing.

**phi, widen deleted: RED.** Removing the single `astype(&q, FLOAT32)` line at `src/models/phi.rs:199`:

```
phi-2-hf-4bit-mlx: steps 1..=8 all sampled token 0, with max logits
[4.52e-41, 4.52e-41, 4.52e-41, 4.52e-41, 4.52e-41, 4.52e-41, 4.52e-41, 4.52e-41].
The row is finite, so this is a collapsed logit distribution rather than a NaN one;
the f32 widen on this family's attention scores is the guard for it.
```

**stablelm, both widen lines deleted: GREEN.** 25 distinct tokens and ordinary text, matching the 11.8x headroom #1710 measured for it. #1830 asks for it either way, and it bounds what this file covers.

Guards restored by path; `git status` clean after each run.

## The failure shape is not the one #1710 described

This is why the file checks two things. Deleting phi's widen produces no NaN row on this prompt: every step's max logit stayed finite at 4.52e-41, a denormal, and argmax returned token 0 at all 33 steps. A finiteness assertion alone would have passed a model emitting nothing but `!`. `max_all` does propagate NaN, checked directly against a row containing one, so that assertion is sound for the NaN shape; this prompt simply collapses before reaching it. The collapse check therefore runs inside the decode loop, so its failure names the step where the run began rather than only reporting that the whole generation was degenerate.

## Decode throughput

Timing could not settle this and the binaries can. Built at this commit and with `layers.rs` reverted to `origin/main`, the two release binaries have **byte-identical `__text` sections**; 99 of 53,488,176 bytes differ, all outside the code (load commands and `__LINKEDIT`), and the new symbol survives into neither binary because nothing calls it. Identical code cannot decode at a different speed.

Timing is reported as a caution. Position-balanced over 32 runs of `-n 128 --temp 0`, after/before is 0.9799 by median. That 2.0% gap is smaller than the 2.6% effect of *which binary ran first within a rep*, which was measured by reversing the order and watching the sign flip, and both sit inside a 17.3% spread on a box at load ~3.3. Before the order was reversed the first pass looked like a consistent 4.9% regression in 6 of 7 reps.

## Validation

M1 Ultra, macOS 27.0, all three checkpoints resolved under `MLXCEL_REQUIRE_MODELS=1`. Workspace gate 11247 passed / 0 failed / 124 suites; `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean.

Closes #1830


inureyes:
`OpenXLA feature compile` is red here and it is not this branch. It fails on `main` too: red at `7e4577b1` and at `b8d10fb1`, green at `d67b64d3`, so it entered with #1871 (`7e4577b1`, "enable model-free WebUI startup"), which touches 30 Rust files including every file the errors name.

The errors are all `unused import` and `never used` under the XLA feature set, in `src/server/app.rs`, `src/server/router_server.rs`, `src/server/auth.rs` and `src/models/mod.rs`: `ui_model_actions`, `ui_operation_cancel`, `invalid_webui_field`, `is_webui_token`, `request_id` and others. Nothing in them names `layers.rs`, `tests/attention_f16_range.rs`, `max_abs_attention_score` or `F16_MAX`.

This branch adds one `pub const`, one uncalled `pub fn` and one test file. The other 11 checks pass and the default-feature `cargo clippy --workspace --all-targets -- -D warnings` is clean locally.

PR #1892, opened 2026-09-14
feat: add scoped WebUI settings and next-load profiles

### LINKED ISSUE #1846: feat(webui): expose scoped settings and validated load profiles
## Parent / dependencies

Parent: #1834. Depends on #1835, #1839, #1840, #1842, #1843, #1838, #1837.

## Ownership

Own `webui/src/features/settings/`, request/load-profile form components and the narrow next-load profile validator/adapter in `src/server/webui/`. Reuse `runtime_settings.rs`, existing startup/preset resolution and actual server schema. Central startup/config changes are coordinated with their owner.

## Required behavior

- Separate Appearance/Privacy (browser only), Generation (per request), Loaded model (live server values), Next load (profile) and Server startup (read-only restart flags). Labels state scope and timing; show current versus pending values. UI does not silently enable `--settings`, `--props`, `--metrics` or slots when the operator disabled them.
- Render live controls from `/settings` schema/current/fingerprint: type/range/allowed values/mutable/reason. PATCH is partial success; reflect applied fields, preserve rejected input with its field error, then refetch effective values. Detect stale external edits by fingerprint and require refresh/reconfirmation; do not claim an atomic compare-and-swap unless the server actually enforces one. Reset is scoped and confirmed.
- Generation basics: max output, temperature, top-p/top-k, seed and system prompt, with advanced repetition/reasoning/structured-output options only if the selected endpoint/model supports them. Unset means omit and inherit, not send null or zero. Validate context budget from actual `/props` and tokenization when available; an estimate never replaces server validation.
- Implement an explicit next-load profile attached to the operation, NOT hot mutation: initial allowed keys `ctx_size`, `n_parallel`, and `kv_cache_mode` (serialized from the existing KVCacheMode choices, not a new quantizer) only when existing CLI/startup validators and selected model/backend support them. Reuse conversion, accepted enum names and range logic; expose effective resolved values including CLI/preset overrides. No raw argv, shell text, model path, token, network or distributed fields. Loaded model requires explicit unload/reload to apply.
- Profiles persist only in browser preferences (no secrets); carry schema version, validate import, identify model-specific versus reusable values, and show a sanitized reproducible CLI command without executing it. Single-model mode offers restart instructions, not fake next-load Apply. Keep draft/adapters/distributed/worker-geometry fields read-only in v1 unless added through a reviewed contract change.

## Acceptance criteria

- [ ] Test disabled settings endpoint, schema read-only reason, invalid range/enum, mixed applied/rejected PATCH, another client changing values, reset and unsupported model knob. No successful toast for rejected fields.
- [ ] Test precedence startup CLI/preset/profile and omission semantics, plus profile version migration and rejected unsafe keys; API and UI share validators/fixtures.
- [ ] Real checkpoint reload with a supported context profile reports the actual changed `/props`; no hot worker-setting mutation occurs. Pending changes survive a failed load without being represented as active.
- [ ] Context displays honor actual per-slot/shared-pool facts, handle n_ctx=0/unknown safely and reference #1815 as external work rather than implementing unified KV here.

inureyes:
Closes #1846. Part of #1834.

## Changes

- Add scoped browser/privacy, in-memory generation defaults, live settings and explicit next-load profiles using the published `@lablup/ui-common@0.1.0-alpha.19` adapters. No implicit credential or system-prompt persistence.
- Preserve CLI/environment > profile > preset precedence, existing context/parallel/KV validation, operation identity and no hot worker mutation. Live edits detect observed fingerprint drift, retain rejected drafts and refetch effective values.
- Fix shared authentication-session fencing, modal-unmount focus, field accessible names and base64url asset-hash caching; preserve feature-off compilation and exact browser assertions.

## Validation

At source `8308a79f`, both workspace/all-target Clippy configurations, formatting, 47 contract fixtures/structural checks, both binary builds and both model-free entrypoints with relocated assets, network denial and TTY checks passed. Hosted run 34859248743 passed all executable checks; three jobs were skipped, not passed.

The runtime/UI-equivalent `153d3c73` passed 119 frontend tests, 21 local/22 hosted browser cases, deterministic bundle verification and the serialized workspace suite (11,323 passed, zero failed, 361 ignored, plus two nested tests). A subsequent GPU recovery event has unknown root cause; the maintainer explicitly deferred that investigation to finish WebUI development. This is not a claimed firmware diagnosis or a waiver of actual feature failures.

Actual Settings acceptance passed at `8308a79f`: Llama 2048/2/int8 and Granite 1024/1/fp16 reloads matched `/props` and generated responses; conflicting profiles respected all three CLI pins. Partial live PATCH/refetch, unchanged hot geometry and observed worker exit passed. Actual production-CSP browser checks at 1440/light and 390/dark passed with zero axe violations, no secret persistence and profile save without server mutation. Both temporary servers exited gracefully. Native Safari/VoiceOver and actual 200% zoom are deferred to the final integrated manual session. No CUDA runtime validation is claimed.

Final security rebase `c8a78b7c` includes #1896. Both binaries were rebuilt; 47 contracts, the complete actual Settings acceptance above, and all executable hosted checks passed again.


inureyes:
Central verification at `8308a79f`: both WebUI-on/off workspace all-target Clippy, 47 contract fixtures, formatting, binary builds, and both production CLI entry points' relocated model-free/offline/TTY security checks passed. Hosted run 34859248743 passed every executable check; three jobs were skipped, not runtime evidence.

The cfg-off constructor defect is repaired without lint suppression. The earlier `153d3c73` full suite remains 11,323 passes, zero failures, 361 ignored, plus two nested subprocess passes, but its independent Metal recovery prevents claiming GPU stability. Actual Settings profile application and inference acceptance remain unexecuted, so this PR stays in review.

A subsequent bounded diagnostic was stopped immediately by a monitoring-harness header-matching bug, not by a new GPU event. That attempt supplies no test result. Evidence is retained separately; no further GPU run is scheduled until monitoring is reliable.

Issue #1849, opened 2026-09-12
docs(webui): document secure workflows and bundled distribution


inureyes:
## Parent / dependencies

Parent: #1834. Depends on #1836, #1838, #1837, #1844, #1845, #1846, #1847, #1848. Documentation may be drafted earlier but must reflect measured final behavior.

## Ownership

Own user/developer WebUI docs and translations in the project's existing layout, README/CLI help links, `docs/webui/README.md`, compatibility manifest final review, NOTICE/release guidance and packaging verification documentation. Coordinate CLI help edits with startup owner; no new website framework or independent package version.

## Required content

- Copy/paste quickstarts for `mlxcel-server --webui`, equivalent `mlxcel serve --webui`, explicit `--models-dir models/mlx`, managed model-store root, single-model mode, offline use, enabled live settings/slots/metrics and safe authenticated remote/proxy access. Explain generated local key/no-terminal behavior and memory-only browser login; never show a real secret or token-bearing URL.
- Explain supported versus downloaded versus loaded, source precedence/containment, explicit preload exception, model/task capability gates, load/drain/memory-release distinction, cancellation limits, cache removal versus unload and capacity/eviction confirmation. Document restart/failed download recovery and non-chat API task examples.
- Separate browser preferences/history, request sampling, live settings partial application, next-load profiles and restart-only controls. Describe actual context/metric scopes and unknown states; link related #1815 without promising unmerged behavior. State privacy defaults, local history opt-in/export/clear, attachment limits, no remote telemetry and no tool execution.
- Contributor workflow: contract-first changes, shared module ownership, frontend source versus generated assets, pinned pnpm install/build, asset regeneration/drift check, offline Rust builds, UI Cargo feature graph and browser/hardware test evidence. Explicitly distinguish `webui/` from project marketing `webpage/`.
- Verify supported release packaging includes embedded UI through the chosen default feature in Metal/CUDA artifacts and no-default-features builds omit it. Retain all frontend/adapted-code licenses in NOTICE and per-file derivation notices; do not redistribute Apple design-kit or system font assets without permission.

## Acceptance criteria

- [ ] English and Korean user docs/help examples are exercised against the installed artifact, not only reviewed as prose; bad root, unauthenticated access and UI-not-compiled diagnostics match docs.
- [ ] b10621 UI alias/disable/environment/static-path/config/tool/proxy classifications accurately describe implemented versus rejected behavior and preserve root-health divergence notes.
[...]

PR #1898, opened 2026-09-15
update: add bundled WebUI integration verification

### LINKED ISSUE #1848: test(webui): gate bundled UX, security, and real-model integration
## Parent / dependencies

Parent: #1834. Depends on #1838, #1837, #1839, #1840, #1841, #1842, #1843, #1844, #1845, #1846, #1847. Final integration owner; individual children remain responsible for their own unit/contract tests.

## Ownership

Own `tests/webui_*`, frontend Playwright/accessibility/visual integration suite, shared verification script and CI integration, final cross-surface regression fixes coordinated with module owners. Browser mock fixtures do not replace a real bundled Rust server.

## Test matrix and release gate

- Build and launch BOTH server commands in UI-on/UI-off and single-model/empty-router modes, feature-on/off, nonempty API prefix, loopback generated key, explicit key/TLS and supported reverse proxy configuration. Verify root health, public shell/private data, asset/cache/CSP behavior and offline operation outside source tree. Keep required MLX runtime resources available; single bundled WebUI does not imply static linking of MLX.
- Hermetic browser flow against the real router and fake model/downloader traits: login → empty library → rescan/download → inspect → load → chat → cancel → unload/drain → remove. Exercise bad state transitions, stale multi-tab changes, unknown POST result, event gaps, auth expiration, server restart, download error/cancel, and corrupted/missing assets. Security attacks target legacy routes as well as UI adapters.
- Hardware gate with actual checkpoints: one dense plus one hybrid/MoE family from the project's available regression set, one supported VLM image input, and an opt-in small public download. Record binary SHA/features, checkpoint path and revision, hardware/OS, command, generated output, operation order and before/after resource observations. Set `MLXCEL_REQUIRE_MODELS=1` or equivalent required-fixture assertion; unavailable hardware/checkpoint is a reported blocker, not a green skip.
- Run Rust workspace/test-fast gate with Metal+accelerate and relevant workspace/all-target clippy/fmt; feature-disabled build and CUDA server smoke on its supported host. Do not parallelize GPU test binaries. No quantization/kernel arithmetic changes belong here; if unavoidable follow the teacher-forced logit trace conditions.
- Playwright Chromium/Firefox/WebKit plus actual Safari on macOS 27 (WebKit CI alone is not Safari validation); 390/1024/1440 widths, light/dark/tinted/opaque/high-contrast, 200% zoom, reduced motion, keyboard/IME and manual VoiceOver. Capture approved deterministic screenshots; failed baselines need review rather than blanket regeneration.
- Initial JS ≤200 KiB gzip, total JS ≤700 KiB gzip, bundle ≤5 MiB; local cold shell usable ≤2 s, action feedback ≤100 ms excluding backend work, no layout shifts from status updates. Record reference host/browser/dataset and startup RSS/time separately from model load. Test 1,000 catalog entries and 10,000-token transcript; observation overhead follows the Activity child's paired-run budget.

## Acceptance criteria

- [ ] One `make verify-webui` (or documented equivalent) orchestrates lint/type/unit/contract/bundle/browser checks; actual hardware checks are clearly separate required release evidence, never masked by mock success.
- [ ] Compatibility endpoints/CLI manifests and UI-disabled regression pass. CI tests the installed artifact, not only the Vite dev server.
- [ ] Shared requirement-to-test matrix accounts for every epic acceptance criterion, including security, state completion and no-model/no-network guarantees. No enabled placeholder or misleading metric survives final review.
- [ ] Publish test evidence and remaining limitations; do not close the epic while a required gate remains unverified.

inureyes:
Closes #1848. Integration gate for epic #1834; evidence is tracked in [the requirement matrix](docs/webui-integration-matrix.md).

- Adds secured-router browser journeys with fake provider leaves, including restart/gap recovery, failed/cancelled downloads, stale revisions, lost POST responses, and Stop/drain.
- Adds installed-artifact, startup/RSS, single-model, TLS proxy, and Activity verifiers; fixes prefixed UI URL derivation and public root health.
- Uses published `@lablup/ui-common@0.1.0-alpha.19`. Keeps strict performance budgets and separates mock, installed, real-model, and manual evidence.

## Defects this gate found in the product

Two were found by using the bundled UI rather than by any automated check, which is what the gate exists for.

- The catalog judged completeness from the shard index alone, so a repackaged mlx-community quant whose stale index names shards that were never written was reported as "required shard ... is missing or empty" with every task unavailable, while `mlxcel generate` loaded and ran it. Nine of two hundred entries on the M1 Ultra store were blocked. The downloader already had the rule that distinguishes this from an interrupted download; it is now shared instead of duplicated, and the nine drop to zero.
- Every 2 s tick re-read server identity and the whole inventory, so a 200-entry library cost seven requests every two seconds, five of them catalog pages fetched back to back. The contract makes polling a fallback to the event stream. The tick now splits: a steady-state tick reads only operations and the selected model's runtime, and a full snapshot is taken when there is not one yet, when the stream is not connected, when an event says the inventory moved, on a forced resnapshot, when the tab becomes visible, on an explicit refresh, and on a 60 s safety net. Measured the same way before and after: 70 requests per 20 s becomes 16.
- `performanceCompletion` derived completeness from whether the browser was headed, so a headed run that skipped the hidden mode would have summarized as a complete acceptance. It now derives it from whether the hidden mode actually ran.

## Validation

Metal workspace gate at `aa5a4041`: 126 test binaries, 11,363 passed, 0 failed, 362 ignored. 316 frontend tests, 18 Node harness-policy tests, 49 contract fixtures, 29 Chromium baseline tests, 21 cross-engine tests. Real secured-router browser harness passes. Both-command HTTP/TLS/generated-key/UI-off/feature-off checks, source-read/network-denied relocation, startup 12 runs, four single-model greetings, and the public SmolLM download with an actual process restart, generation, unload and isolated-cache removal. Workspace clippy in both feature configurations, fmt, and warning-denied checks with and without the `webui` feature.

Native Safari on the installed artifact was verified by the maintainer on 2026-09-16: screens and layout across all four, keyboard Tab and Shift-Tab order with Escape restoring focus, and native 200 percent zoom with no horizontal page scroll at 390 px with increased contrast readable.

## Still open

Manual VoiceOver is NOT RUN and is a distinct evidence row; the axe and DOM checks assert markup, not what a screen reader announces.

The native hidden Activity acceptance is user-deferred as of 2026-09-16, scoped to the minimize-to-`document.hidden` measurement only. On the GB10 runner with Xvfb and openbox provisioned, headed Chromium gets a real 700x900 content area and the browser reports the window as `minimized`, but `document.visibilityState` stays `visible`; foregrounding a sibling tab also leaves `document.hidden` false, and both visibility-suppressing launch arguments were removed with the removal verified against a live browser process. That is a property of the automation browser, not of the runner or the product. One step in that chain is unverified: whether Playwright's sibling page is a tab in the same window. The headed visible-mode observation overhead still runs in CI with every numeric budget applied, and the visibility backoff stays covered by two unit tests and a `visibilitychange` browser test.

The earlier full-Metal failure in the DFlash configured-depth test did not reproduce in a serialized single run; that is a non-reproduction, not a root cause, and the investigation remains user-deferred. Single-model SIGINT follows existing `-2` semantics, not worker-drain proof. Raw logs remain local; CI publishes reviewed summaries only.


inureyes:
Progress checkpoint for epic #1834: 13 of 15 child issues are closed; #1848 (this draft PR) and #1849 (documentation drafts prepared) remain open.

Pushed `6472f5d4` (canonical artifact paths; CPU regression and real-router relative-path browser test passed) and `3d65d1b8` (headed WebKit performance under Xvfb; typecheck, lint, YAML, and whitespace checks passed). New CI results are pending; performance thresholds and browser case coverage are unchanged.

Local acceptance passed for both-command installation/TLS/offline/proxy checks, startup and single-model runs, prefixed Llama/Granite/Qwen3-VL journeys, and public download/process-restart/unload/removal. Workspace Clippy passed in both feature configurations at c61ff73b.

Remaining: cross-engine and GB10 installed-artifact CI, native-hidden Activity performance, final Safari/VoiceOver/native 200% verification, and #1849 final documentation. The full Metal workspace run failed with GPU error/recovery; investigation is user-deferred, not a PASS or confirmed firmware cause. No merge or epic closure is claimed.

PR #1919, opened 2026-09-19
fix(server): record streamed chat completions in Metrics

inureyes:
## Summary

The streaming branch of `POST /v1/chat/completions` never called `Metrics::record_request`, so `requests_total` and `completion_tokens_total` (and the prompt-token and generation-time totals) only moved for non-streaming chat. The WebUI chat always streams, so the Activity tiles read 0, and `mlxcel_requests_total` undercounted every streaming client.

- `stream_chat_completion` records the request once, right after `slot.finish`, using the result's own counts. It does not depend on `stream_options.include_usage`. `Err` records nothing, as on the non-streaming path.
- Streams the client cancelled are not counted. That covers a disconnect without `X-Conversation-Id`, `DELETE /v1/stream`, and session replacement. After an abort the worker still returns `Ok` with the partial count, so the route checks the cancellation flag. The flag is now cloned before it moves into the provider call. A resumable stream whose client only disconnects keeps generating and is counted.
- The `mlxcel_requests_total` and `mlxcel_completion_tokens_total` HELP lines, the two `counter_source` reasons (and `runtime.ready-shared.json`, which pins them) and `docs/webui/runtime.md` now state these semantics. Metric names, values and both response shapes are unchanged.

The scope is chat completions only. Streaming `/v1/completions` and native `/completion` still never record, and `/v1/messages` and `/v1/responses` streams still record even when cancelled. That is why the HELP line says "chat completion streams". Both are left for follow-ups.

## Tests

New tests in `server::routes::stream_route_tests::metrics_tests`, a child module of `stream_route_tests` that reuses its helpers. `stream_route_tests.rs` was already 698 lines. `scripted_app_with_state` also returns the `AppState`, and `scripted_app` delegates to it. The cancelled and resumable cases wait for the generation task to exit before asserting, because its completion-control registration drops after the record point. Without that wait, a zero count would prove nothing.

- `completed_stream_records_one_request_with_its_counts`: 3 tokens, `[DONE]`, then +1 request, +3 completion tokens, still 1 after task exit.
- `client_cancelled_plain_stream_is_not_recorded`: disconnect, flag flips, `finish()`, task exits, `requests_total == 0`.
- `resumable_stream_is_recorded_after_a_client_disconnect`: the flag stays clear, and the request counts with all 3 tokens.
- `metrics_endpoint_reports_a_completed_stream`: whole-line `mlxcel_requests_total 1` and `mlxcel_completion_tokens_total 3`.
[...]

Issue #1910, opened 2026-09-18
epic: make the bundled WebUI usable, on the shared ui-common foundation


inureyes:
## Problem / Background
Epic #1834 shipped a bundled WebUI whose contracts, security and accessibility gates pass, but whose screens are not usable: the Settings page is one 5,720 px scroll of 129 inputs, the Models page shows four rows per viewport, the Chat transcript has no role distinction, Korean mode is mostly English, and the Activity counters read zero for every UI chat because the server never records streaming completions. The four follow-ups filed on 2026-09-16 (#1900, #1901, #1902, #1903) and the ROCm catalog bug #1886 become children of this epic rather than standalone work.

## Goal
Turn the bundled WebUI into a product surface: dense, task-first screens on the shared `@lablup/ui-common` foundation, every user-facing string in the catalog, honest metrics measured on the path the UI actually uses, and no internal state leaking into copy.

## Evidence (review of 2026-09-17, release build of main `0ef0a1a4`)
The review ran the real binary (`mlxcel-server --webui --models-dir models/mlx --api-key-file ... --no-models-autoload --settings --props --metrics --slots`) against 212 local checkpoints and drove login, load, streaming chat and unload with Playwright at 1440x900 and 390x844. Measured:

- Settings page: 5,720 px tall, 129 form inputs, 763 words, 11 `h2` headings, one uninterrupted scroll (`webui/src/app.tsx` SettingsScreen plus `webui/src/features/settings/server-settings.tsx`).
- Models page: 3,132 px tall at 1440x900 with four table rows visible in the viewport; each row is about 100 px because the name cell stacks a ghost button, "Selected", and a `source · quantization` line (`webui/src/features/models/screen.tsx:186-242`). At 390 px the page is 5,936 px tall.
- Chat page renders prompt, an "Edit and regenerate" button, the response, "Copy response" and a "Response details" disclosure as one flat article with no role distinction (`webui/src/features/chat/transcript.tsx`); conversations are a `<select>`; the composer is not pinned; the transcript is a separate 60vh scroll box (`webui/src/features/chat/chat.css`).
[...]

Issue #1913, opened 2026-09-18
refactor(webui): move every user-facing string into the i18n catalog and remove native confirm dialogs


inureyes:
Part of #1910.

## Problem / Background

`docs/webui/ux-contract.md:13` requires every user-facing string to have a stable key, an English value, a Korean value and a test ID in `tests/fixtures/webui/strings.json`, resolved through `t(locale, key)` in `webui/src/i18n/catalog.ts:354`, and `docs/webui/ux-contract.md:33` says page work must use those keys. The Chat, Settings and Activity screens bypass this. In Korean mode the Chat screen shows only its title in Korean (verified by screenshot on 2026-09-17 against the bundled binary), and even that title is wrong: `webui/src/features/chat/chat.tsx:131` hardcodes `locale === 'ko' ? '채팅' : 'Chat'` while the catalog entry `chat.title` (and `nav.chat`) says `대화`. The CJK layout tests, which are release gates per the contract, never see any of the bypassed strings because they are not in the fixture.

Four confirmations use `window.confirm`, which is unstyled, unlocalized, blocks the event loop, has to be special-cased in Playwright and ignores the contract's destructive-confirmation rules (`docs/webui/ux-contract.md:15` lists `settings.clear_history.confirm.body` as a required key; it is declared in `catalog.ts` and `strings.json` but no component renders it).

## Current Behavior

- `webui/src/features/chat/chat.tsx` renders its copy from literals: 20 quoted literals and 9 JSX text nodes of 12+ characters (headings, banner text, the composer hint "Enter sends · Shift+Enter inserts a line.", errors such as "Conversation limit reached. Delete or export older conversations first.", announcements such as "Stopped locally; checking server activity."). `t(` is never called; `locale` is used only at line 131. `transcript.tsx`, `parameters.tsx` and `privacy.tsx` in the same directory do not import `Locale` at all and are entirely English.
- `webui/src/features/settings/context-check.tsx:20`, `generation-settings.tsx:12`, `live-settings.tsx:16`, `profile-settings.tsx:13` and `server-settings.tsx:17` each define `const words = (en: string, ko: string): string => locale === 'ko' ? ko : en` and call it 9, 13, 16, 24 and 21 times respectively (83 inline en/ko pairs that exist nowhere in the fixture).
[...]

PR #1920, opened 2026-09-19
fix(webui): print the checkpoint's real name as its display name

### LINKED ISSUE #1912: fix(webui): catalog display names must keep the checkpoint's real name
Part of #1910.

## Problem / Background

`display_name` is the label every WebUI surface prints for a model, and it is produced by rewriting the checkpoint name: `src/server/webui/catalog_metadata.rs:465-470` takes the last path segment and replaces every `-` and `_` with a space, so `qwen3-0.6b-4bit` renders as `qwen3 0.6b 4bit` and `Meta-Llama-3.1-8B-Instruct-4bit` as `Meta Llama 3.1 8B Instruct 4bit`. Observed on 2026-09-17 against 212 local checkpoints: the user never sees the directory or inference id they pass to `-m` and the API; two directories that differ only by `-` versus `_` collapse to one label; typing the real name `qwen3-0.6b-4bit` into the library search finds nothing, because `inventory()` at `webui/src/features/models/policy.ts:83-92` matches `display_name`, `inference_id` and `architecture` case-insensitively but the hyphenated form only matches `inference_id`, which no surface shows; and Playwright specs have to target the transformed name (`webui/tests/models.spec.ts:178`).

## Current Behavior

The helper is called at `catalog_metadata.rs:61` (router entries; `model.name` is the directory name for models-dir entries and `owner/name` such as `mlx-community/retry-once` for cache entries, see `src/server/router_models_tests.rs:856`) and at `catalog_metadata.rs:177` (single-model mode). Consumers: the library row `webui/src/features/models/screen.tsx:194-197`, the inspector heading `webui/src/features/models/inspector.tsx:56`, the toolbar pill `webui/src/provider-surfaces.tsx:25`, the Chat picker `webui/src/features/chat/chat.tsx:132`, the Activity selector `webui/src/features/activity/index.tsx:25`, operation titles `webui/src/features/models/operations.tsx:121`, the settings selector `webui/src/features/settings/server-settings.tsx:27`, and the dialogs at `webui/src/features/models/dialogs.tsx:34,88,100`. Sorting uses it at `policy.ts:104` and `webui/src/state/reducer.ts:151`; server-side `q` search concatenates it with `inference_id` at `src/server/webui/catalog.rs:358-370`. The delete dialog (`dialogs.tsx:62-78`) prints the opaque `identity.id` in a `<code>` block and only accepts that id as the token, while the body string `models.library.delete_body` (`webui/src/features/models/strings.ts:245`) says "Type the model name to confirm" and the field label `models.library.confirm_name` (`strings.ts:250-253`) says "Exact cache model ID to confirm", so the two strings disagree about what to type.

## Proposed Solution

Make `display_name` the inference id with no character rewriting: models-dir and single-model entries get the last path segment of the directory (`qwen3-0.6b-4bit`), cache entries get the full `owner/name` (`mlx-community/Qwen3-4B-4bit`). Concretely, in `catalog_metadata.rs:465-470` drop the `.replace(['-', '_'], " ")`, and for `RouterModelSource::Cache` return `model.name` unchanged instead of the `rsplit('/')` segment. Do that by giving the helper the source (`fn display_name(name: &str, source: RouterModelSource) -> String`) or by branching at the call site at `:61`; the `:177` single-model path keeps the last-segment rule. Do not add a humanized variant in this issue; if one is ever wanted it goes in as a separate optional field, never as the primary label. Rejected alternative: keep the space-rewritten label and print `inference_id` next to it, because that leaves two names for one thing and fixes neither search nor the `-`/`_` collision.

Schema: this is a semantic change to an existing string field, so `schema_version` stays `webui.ui-api.v1` (`src/server/router_lifecycle.rs:29`). The only `docs/webui/api.yaml` edit is widening `ModelIdentity.display_name.maxLength` at `docs/webui/api.yaml:1518-1522` from 128 to 256 to match `inference_id` at `api.yaml:1513-1517`, because a cache id can reach 193 characters (`validRepo` at `policy.ts:110` allows 96 per part) and a directory name can be 255 bytes; leave the root-level `display_name` at `api.yaml:1340-1344` alone. Add a one-line `description` on the field stating it is the inference id verbatim, then regenerate `docs/webui/generated/ui-api.d.ts` with `python3 scripts/ci/check_webui_contract.py --fix` (the TS type stays `string`, so expect no diff there).

Delete dialog: keep `identity.id` as the token, because the reviewed identity fence at `webui/src/features/models/screen.test.tsx:312-322` exists precisely so a duplicate display name cannot authorize a delete (`tests/fixtures/webui/scenarios/identity-collision.json:9,97`). Fix the wording instead: `models.library.delete_body` (en and ko at `strings.ts:245-246`) must say to type the ID shown below, and `models.library.confirm_name` must name it as the opaque model ID printed in the dialog, so the label tells the user exactly what to type and the visible name is no longer implied.

## Scope

**In scope:** `src/server/webui/catalog_metadata.rs` (helper and both call sites); `src/server/webui/catalog_contract_tests.rs` (new test: a models-dir name containing both `-` and `_` and a cache `owner/name` both come back verbatim); `docs/webui/api.yaml` plus regenerated `docs/webui/generated/ui-api.d.ts`; `docs/webui/catalog.md:23` (state that `display_name` is the inference id verbatim, not a humanized form); `webui/src/features/models/strings.ts` delete strings (en and ko); `tests/fixtures/webui/scenarios/identity-collision.json:97` (`alpha models dir` becomes `alpha`; the scenario still collides on `inference_id`, which is its purpose); and the frontend tests and fixtures that encode space-separated names, rewritten to hyphenated ids: `webui/src/state/reducer.test.ts:10`, `webui/src/features/models/screen.test.tsx:195,216,221`, `webui/src/features/models/load-profile-integration.test.tsx:19`, `webui/tests/performance-fixtures.ts:41`, `webui/tests/models.spec.ts:235`. Keep the CJK long-name string `models.long_name` (`webui/src/i18n/catalog.ts:222`, `tests/fixtures/webui/strings.json:262-265`) exercising wrapping, but as a long hyphenated id such as `Qwen3-Very-Long-Local-Checkpoint-Name-With-Mixed-English-and-한국어-모델-이름`, and keep the `models-wrap` treatment on the row, inspector heading and dialog so a 200-character id wraps instead of overflowing.

**Out of scope:** any other inspector content change; the model library redesign is a separate child of #1910 that depends on this issue.

## Implementation Notes

- **Reuse**: `model.name` and `inference_id` are already the verbatim ids at `:61` and `:177`; no new derivation is needed. Leave `matches()` at `catalog.rs:358-370` and `inventory()` as they are; once `display_name == inference_id` they match the real name by construction.
- **Edge cases**: a name ending in `/` must still fall back to the whole name via the existing `unwrap_or(name)`; a name that already contains spaces is kept verbatim; a `-` directory and a `_` directory now produce distinct labels and distinct sort positions (`localeCompare` at `policy.ts:105`).
- **Error handling**: nothing new. The server never truncates; a name over 256 characters is a schema violation that `scripts/ci/check_webui_contract.py` reports on fixtures.

## Acceptance Criteria

- [ ] With `models/mlx/qwen3-0.6b-4bit` in the store, `GET /ui-api/v1/catalog` returns `identity.display_name == "qwen3-0.6b-4bit"`, and the library row, inspector heading and toolbar pill print exactly that.
- [ ] A cache entry prints `mlx-community/Qwen3-4B-4bit` in those three places and in the Chat and Activity selectors.
- [ ] Typing `qwen3-0.6b-4bit` in the library search box returns that row; the underscore spelling does not match it.
- [ ] The delete confirmation body and label both say to type the opaque model ID printed in the dialog, and the fence test at `screen.test.tsx:312-322` passes unchanged.
- [ ] `catalog_contract_tests.rs` asserts verbatim `display_name` for a models-dir name with `-` and `_` and for a cache `owner/name`; the test fails on `main` and passes with the fix.
- [ ] `schema_version` is still `webui.ui-api.v1`, and `python3 scripts/ci/check_webui_contract.py` passes with the widened `maxLength`.
- [ ] Verified against the running bundled binary with a store that holds both hyphenated and underscored names, not only fixtures.

## Verification

```
cargo clippy --workspace --all-targets -- -D warnings && cargo fmt --all -- --check
python3 scripts/ci/check_webui_contract.py
pnpm --dir webui run unit
pnpm --dir webui run browser
```

Manual: `cargo build --release --features metal,accelerate`, then `./target/release/mlxcel-server --webui --models-dir models/mlx --api-key-file <file> --no-models-autoload` (`--models-dir` is the router-mode discovery root that yields `models_dir` entries; `--model-store-root` is the managed cache root and yields `cache` entries, so test both kinds), open the WebUI, and confirm the three surfaces above, the search-by-real-name result, and the delete dialog wording. A pass is the unmodified directory name on every surface.

inureyes:
Part of epic #1910.

## Summary

- `identity.display_name` is now the model's inference id with no character rewriting. The catalog no longer turns `-` and `_` into spaces. Cache and preset entries keep their full name (`mlx-community/Qwen3-4B-4bit`), models-dir entries keep the directory name, and single-model entries keep the last `/` segment of the served id, as the issue decided. `qwen3-0.6b-4bit` and `qwen3_0.6b_4bit` are two distinct rows again. Every surface that prints `display_name` (library row, inspector heading, toolbar pill, Chat and Activity selectors, operation titles, settings selectors, dialogs) now shows the name passed to `-m` and the API. The Rust change is the `display_name` helper in `src/server/webui/catalog_metadata.rs` and its two call sites.
- `ModelIdentity.display_name.maxLength` in `docs/webui/api.yaml` goes from 128 to 256 to match `inference_id`, because a cache `owner/name` can be 193 characters. The field's new description gives the rule for each source. `schema_version` stays `webui.ui-api.v1`, and the generated `ui-api.d.ts` is unchanged.
- The delete dialog still accepts only `identity.id` as its token, and the fence test is byte-identical to main. The body now says to type the model ID shown below, not the model name. The field label reads "Opaque model ID shown above (starts with mdl_)" in English and "위에 표시된 모델 ID (mdl_로 시작)" in Korean. The `tests/fixtures/webui/strings.json` copies of both strings, which had drifted, now match `strings.ts`.
- Long ids wrap. The dialog body paragraph and the model names listed in the capacity-recovery dialog now carry `models-wrap`, like the id already did. Without it, a verbatim underscore name has no break point: at 390px the capacity dialog measured scrollWidth 523 against clientWidth 356.
- Fixtures and tests that used space-separated names now use hyphenated ids. That includes `identity-collision.json`, which still collides on `inference_id`, and the performance fixtures and specs. The `models.long_name` gallery string is now a long hyphenated id, the same in en and ko. `docs/webui/catalog.md` and `docs/webui/catalog.ko.md` describe the new meaning. The committed bundle is regenerated.

## Premises checked

- Search already found the real name before this change: `inventory()` and the server `q` filter both match `inference_id`, which already held `qwen3-0.6b-4bit`. What changes is that the matched row now prints that name.
- The issue expected a trailing `/` to fall back to the whole name through the existing `unwrap_or(name)`, but it never did. `"x/".rsplit('/').next()` is `Some("")`, so the old helper returned an empty string. The new helper skips the empty segment, and the test covers that case.
- The issue did not say how to treat presets; this PR keeps their names whole. A preset's inference id is its section name (`config.model_alias = name`), and an overlay-only preset takes over a cache `owner/name`, which the last-segment rule would shorten.
- Single-model mode keeps the last-segment rule, as the issue specified. The id there is either `--alias` or the checkpoint directory's `file_name()`, never a path, so the rule only changes an alias that contains `/` (for example `--alias Qwen/Qwen3-8B` displays as `Qwen3-8B`). The docs and the api.yaml description say so. Showing the alias whole would be a one-line change if that is preferred.

## Changes during review

- Implementation review: the capacity-dialog list items got the wrap class. The capacity-recovery unit test now checks the listed names and the class, and fails without them.
- Implementation review: the api.yaml description and both catalog docs now give the rule for each source. The earlier wording said "verbatim" everywhere, which is not true for single-model aliases. The Korean confirm label dropped "불투명", a word Settings already uses for the visual Opaque material.
- Security and performance review: `performance.spec.ts` searched the 1000-entry catalog for `'0999'`, which only matched the old padded label, so the `WebUI bundle` engine smoke timed out on every engine. It now searches for `perf-model-999`. The review found no injection or path exposure: names render as React text or escaped attributes, and `display_name` is always equal to `inference_id` or a substring of it. The delete fence still requires the exact id and revision.

## Revert-and-fail

With the `catalog_metadata.rs` change alone reverse-applied, the new contract test fails:

```
test server::webui::catalog::tests::catalog_contract_tests::display_name_is_the_inference_id_verbatim ... FAILED
panicked at src/server/webui/catalog_contract_tests.rs:543:5:
assertion `left == right` failed
  left: [("Meta-Llama-3.1-8B-Instruct_4bit", "Meta Llama 3.1 8B Instruct 4bit"), ("mlx-community/Qwen3-4B-4bit", "Qwen3 4B 4bit"), ("qwen3-0.6b-4bit", "qwen3 0.6b 4bit"), ("qwen3_0.6b_4bit", "qwen3 0.6b 4bit"), ("team/qwen3-preset", "qwen3 preset")]
 right: [("Meta-Llama-3.1-8B-Instruct_4bit", "Meta-Llama-3.1-8B-Instruct_4bit"), ("mlx-community/Qwen3-4B-4bit", "mlx-community/Qwen3-4B-4bit"), ("qwen3-0.6b-4bit", "qwen3-0.6b-4bit"), ("qwen3_0.6b_4bit", "qwen3_0.6b_4bit"), ("team/qwen3-preset", "team/qwen3-preset")]
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 8596 filtered out
```

With the patch re-applied, all 34 catalog tests pass. The check was run twice, once by the implementer and once independently. The frontend check was done the same way. Reverting `api.yaml`, `strings.ts` and `dialogs.tsx` makes the new 193-character schema-bound test and the delete-dialog wording test fail. Removing the capacity-list wrap class makes the capacity-recovery test fail.

## Test plan

This branch is rebased onto origin/main at fd33ff15, which includes #1921 (catalog backend column) and #1922 (i18n catalog consolidation). The committed bundle was rebuilt by `build_bundle.py` at each rebased commit rather than merged by hand. One rebase conflict needed both sides: the performance spec's Chat option now reads `perf-model-0 · Ready`, combining this PR's name with #1922's localized state label. No test or fixture added by #1922 uses a space-separated model name. Results on the rebased head:

- [x] `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib server::webui`: 78 passed, 0 failed, including `display_name_is_the_inference_id_verbatim`
- [x] `cargo clippy -p mlxcel --lib --tests --features metal,accelerate -- -D warnings` and `cargo fmt --all -- --check`: clean
- [x] `pnpm --dir webui run typecheck` and eslint: clean
- [x] `pnpm --dir webui run unit`: node 18/18, vitest 340/340 in 39 files
- [x] Playwright chromium default suite: 30/30. Engines config (`engines.spec.ts` and `performance.spec.ts`), chromium: 7/7. All runs used a private port with `--strictPort`, never the preview server shared with other worktrees
- [x] `check_webui_contract.py` (49 fixtures), `build_bundle.py --verify`, `check_binary_assets.py`, `check_cross_repo_refs.py`: pass

Before the rebase, the same gates also passed `server::router_server` (63 passed, 1 ignored) and the engines config in firefox (7/7).

## Real-binary acceptance

The orchestrator built `mlxcel-server` from 3082e1d8, the pre-rebase head carrying the same change (`--profile test-fast --features metal,accelerate`), and ran it with `--webui --models-dir models/mlx --no-models-autoload` against the real checkpoint store and the default managed cache. I drove the bundled WebUI in Chromium at 1440px with a read-only Playwright script that aborts any load, unload, download, delete or refresh request. It recorded no mutating request.

- [x] Hyphenated models-dir entry: `qwen3-0.6b-4bit` prints as `qwen3-0.6b-4bit` in the library row, the inspector heading and the toolbar pill (`qwen3-0.6b-4bit · Unloaded`).
- [x] Underscored models-dir entry: `stablelm-2-1_6b-chat-4bit` prints unchanged in the same three places.
- [x] Cache entry: `mlx-community/Qwen3-0.6B-4bit` prints as the full `owner/name` in the row, the inspector heading and the pill. It also appears as `mlx-community/Qwen3-0.6B-4bit · unloaded` in the Chat "Model for next turn" picker and as `mlx-community/Qwen3-0.6B-4bit` in the Activity selector (225 options).
- [x] Search: `qwen3-0.6b-4bit` returns the `qwen3-0.6b-4bit` row, plus `mlx-community/Qwen3-0.6B-4bit`, which contains the same text in a case-insensitive match. `qwen3_0.6b_4bit` returns no rows ("No matching local models").
- [x] Delete dialog, opened on the cache entry and cancelled without typing: the body ends "To confirm, type the model ID shown below, not the model name.", the `mdl_` id is printed below it, the field label reads "Opaque model ID shown above (starts with mdl_)", and Confirm stays disabled while the field is empty. The long cache name wraps inside the dialog.

## Merge with main and CI

- `origin/main` 5616f86b (#1924, ui-common components) was merged into this branch as 7e40c01c by the epic #1910 orchestrator. Only the generated bundle conflicted; it was regenerated with `scripts/webui/build_bundle.py` (`--verify` digest 711cbf2a) and not hand-merged. On the merged tree: `typecheck`, `eslint`, `unit` (Vitest 362 in 47 files), and Chromium Playwright on a private port (`models`, `ui-common-shell`, `ui-common-components`, `browser`, `chat`, `activity`, `settings`: 56 passed) all pass. The merged Rust tree was checked earlier: `server::webui` 78 passed, `models::registry` 10 passed.
- The GB10 `WebUI installed artifact` job's Activity performance gate reported `investigate` on four consecutive attempts at the pre-merge heads while every other step passed: `two-visible` median decode degradation -0.35 (baseline CV 5.63), +0.20 (CV 5.45), +2.74 (CV 3.38), and `one-visible` +3.38 (CV 4.22). The gate flags a median above 2 percent or a baseline CV above 5 percent over five pairs. On the same runner, `main` after #1922 failed the same gate (`two-visible` -6.76, CV 7.45) and passed on rerun, and #1919 failed it once (+3.19) and passed on its next head. This change only alters the catalog label string and a few UI strings, which do not touch decode or the Activity polling path, so these are read as runner noise. Per the maintainer's decision for this epic, the PR merges with this evidence recorded rather than with the gate relaxed; stabilizing the gate is tracked in #1925.

Closes #1912


inureyes:

### Intent
> `identity.display_name` is the inference id verbatim on every WebUI surface, with the delete dialog wording pointing at the opaque model ID.

### Findings Addressed
- [x] Capacity recovery dialog listed loaded models in `<li>` items with no wrap rule. Verbatim underscore names have no break points, so at 390px the modal scrolled sideways (Chromium: dialog scrollWidth 523 vs clientWidth 356). The items now carry `models-wrap` (356 vs 356), the capacity unit test asserts it and fails without it, and the bundle is rebuilt. Commit 69b3ba73. (HIGH)

### Remaining Items
- [ ] Single-model mode with a `/` in `--alias` (for example `--alias Qwen/Qwen3-8B`) still prints only `Qwen3-8B`, while the new `api.yaml` description says the field is the inference id verbatim. Single-model inference ids are `--alias` or the checkpoint basename, never a path, so the last-segment rule only ever strips an operator-chosen owner. Left as is because the issue and the brief keep that rule; either make SingleModel verbatim or qualify the description. (MEDIUM)
- [ ] Korean confirm label `위에 표시된 불투명 모델 ID (mdl_로 시작)`: `불투명` is already the Settings translation of the visual "Opaque" material, so it reads oddly for an identifier. `위에 표시된 모델 ID (mdl_로 시작)` is clearer. Left because it is wording only and `strings.json` is shared with #1913. (LOW)

### Verification
- [x] All stated requirements implemented (real-binary acceptance still pending orchestrator verification)
- [x] Tests pass

Also checked: no client code rewrites names; the preset rule matches `model_alias = name` in `router_models.rs`; the fence test is byte-identical to main; the `catalog_metadata.rs` diff is only the helper and its two call sites. After the fix: typecheck, eslint, unit (node 18/18, vitest 319/319), `build_bundle.py --verify`, `check_binary_assets.py`, the contract check, and Playwright `models.spec.ts` 3/3 all pass.

inureyes:
Security and performance review of f6a1db70. One fix pushed in 3082e1d8.

**Fixed (HIGH, merge-blocking):** `webui/tests/performance.spec.ts:332` searched the 1000-entry catalog for `0999`, which only matched the old padded label. After the fixture rename to `perf-model-${index}` it matched no row, and the `WebUI bundle` job failed that test on chromium, firefox and webkit-headed-performance on every retry. It now searches `perf-model-999`. Locally on chromium all four performance tests pass (search-to-paint 14.4 ms, budget 100 ms).

**Checked, no finding:**
- Injection: every `display_name` sink is React text or a React-set attribute (`aria-label`, toolbar `title`, option labels). Neither `webui/src` nor `@lablup/ui-common` uses `dangerouslySetInnerHTML` or `innerHTML`, and `t()` interpolates through a replacer function. A browser probe rendered a markup-shaped name as literal text.
- Paths: `display_name` never carries more than `inference_id` in the same object. Models-dir names come from `file_name()`; single-model ids from `file_name()` or the operator's alias.
- Delete fence: still `token === identity.id` plus instance and revision checks. The fence test is unchanged and passes.
- Length: `display_name` is schema-valid exactly when `inference_id` is, so 256 adds no new rejection and removes the old one for 129-255 character directory names. A 256-character unbroken name at 390 and 1440 px caused no horizontal scroll in the library, inspector, delete dialog, Chat, Activity or Settings. Without `models-wrap` the dialog body overflows, so the class is load-bearing.

**Left (pre-existing, out of scope):** operator-supplied names are not bidi-isolated, so U+202E can reorder the rest of the delete sentence (the id fence still holds). One inference id over 256 code points still rejects the whole catalog page.

`WebUI installed artifact` failed at link time on the self-hosted runner's shared target (`undefined reference` in a stale rlib), unrelated to this diff.

Issue #1814, opened 2026-09-11
perf(rocm): allocator footprint and ROCm ports of mlxcel fused kernels


inureyes:
Part of #1801. Phase 4. Depends on #1803, #1809.

## Context

Correctness comes first in this epic; this issue collects the performance work the spike exposed.

- **Memory footprint.** mlx-lm reported a peak of 20.2 GB for `Meta-Llama-3.1-8B-Instruct-4bit` (about 4.5 GB of weights) and 21.7 GB for `Qwen3-30B-A3B-4bit` (17 GB) on ROCm. The fork's allocator (arena, decode free list, GTT/managed fallbacks) may be caching aggressively; on a UMA host this competes with the OS and other GPU tenants.
- **Fused kernels.** After #1803, ROCm runs mlxcel's fused paths (sampling, rejection sampling, fused add+RMSNorm, RoPE+append, paged attention, fused MoE, SSM, gated delta) as MLX graph fallbacks. The ROCm backend exposes `fast::hip_kernel`, the counterpart of `fast::metal_kernel` and `fast::cuda_kernel`, so these can be ported.
- **Decode bandwidth.** Decode-shaped q4 GEMV reached about 150 GB/s including per-call sync. Llama-3.1-8B-4bit decodes at 32 tok/s; the ceiling from weight traffic alone at that bandwidth is higher.

## Scope

Measure first, then port the fused kernels with the largest decode impact and tune the allocator policy. Each change is gated by the #1809 correctness matrix.

## Implementation plan

1. Profile decode on the #1809 models with `rocprofv3` and attribute time to kernels, graph fallbacks and host overhead.
2. Explain and bound the allocator peak (compare `MLX_*` cache limits and the fork's allocator knobs); set sensible defaults for UMA hosts.
3. Port the top fused kernels to `fast::hip_kernel`, following mlxcel's JIT kernel rules (include every input dtype in the template arguments so cached kernels are not reused across dtypes; key the JIT cache by MLX commit as with `MLX_PTX_CACHE_DIR` on CUDA, and check whether hipRTC has the same stale-cache issue).
4. Report before/after with the #1810 harness.

## Acceptance criteria

- [ ] A profile-based breakdown of decode time on ROCm is published.
- [ ] Peak memory for an 8B 4-bit model is explained, and reduced or bounded by a documented default.
- [ ] At least the top two fused kernels by decode impact run natively on ROCm, with the correctness matrix unchanged and measured speedups.

## References

[...]

Issue #1801, opened 2026-09-11
epic: AMD GPU (ROCm) backend on Linux via mlxcelverse


inureyes:
## Summary

Program of work to run mlxcel model inference with GPU acceleration on Linux hosts with AMD GPUs (ROCm/HIP). The first target is RDNA 3.5 (`gfx1151`, Strix Halo), where a feasibility spike already produced correct results and usable 4-bit decode speed.

The approach is to vendor an existing ROCm backend for MLX into mlxcel as a set of source overlays and apply it on top of the same pinned `ml-explore/mlx` commit that the Metal and CUDA builds use. It becomes the ROCm part of **mlxcelverse**, the name for mlxcel's MLX-side layer (see Design decisions). The ROCm overlay is copied into the MLX tree only when building with a new `rocm` cargo feature, so Apple Silicon and CUDA builds are untouched.

The ROCm backend comes from the `rocm-support` branch of [NripeshN/mlx](https://github.com/NripeshN/mlx/tree/rocm-support) (MIT), which is the head of the upstream draft [ml-explore/mlx#2300](https://github.com/ml-explore/mlx/pull/2300) and the only actively developed ROCm line for MLX today (tracking request: [ml-explore/mlx#2556](https://github.com/ml-explore/mlx/issues/2556)). Downstream projects such as [lemonade-sdk/lemon-mlx-engine](https://github.com/lemonade-sdk/lemon-mlx-engine) already build against it.

## Why an in-repo overlay and not a separate MLX fork repository

The fork was measured against its upstream merge base (`39886de4` to `75915908`):

| Part | Size | Nature |
|---|---|---|
| `mlx/backend/rocm/` | 106 new files, +43,904 lines | Purely additive. Upstream never touches this directory. |
| MLX core glue | 17 files, about +345/-65 lines | Mostly `#ifdef MLX_USE_ROCM` hooks (default device, `fast::hip_kernel` declaration, custom-kernel stubs, CMake option) plus a few general tweaks (buffer donation, buffer cache, a quantized-matmul fallback). |
| Everything else | Python bindings, tests, docs, bench scripts | Not needed by mlxcel. |

[...]

PR #1921, opened 2026-09-19
fix(rocm): select the catalog backend column from the resolved backend

### LINKED ISSUE #1886: fix(rocm): the WebUI catalog reports every architecture unsupported on ROCm
Part of #1910.

Phase 1 ROCm work under epic #1801. Found while validating PR #1883 (issue #1805) on the ROCm spike host: AMD Ryzen AI MAX+ 395 / Radeon 8060S, `gfx1151`, RDNA 3.5, wave32, ROCm 10.0.0, Debian 13, branch `feature/issue-1805-gpu-vendor`. Pre-existing defect, not a regression from that PR.

## Context

Five tests fail in the `-p mlxcel --lib` target on a ROCm build:

- `server::webui::catalog::tests::catalog_cache_tests::catalog_cache_uses_epoch_and_projects_fresh_provider_lifecycle`, `src/server/webui/catalog_cache_tests.rs:56`
- `server::webui::catalog::tests::catalog_contract_tests::actual_catalog_producer_matches_whole_expected_shape`, `src/server/webui/catalog_contract_tests.rs:263`
- `server::webui::catalog::tests::catalog_contract_tests::cached_single_model_catalog_accessor_projects_fresh_app_state_provider`, `src/server/webui/catalog_contract_tests.rs:318`
- `server::webui::catalog::tests::catalog_contract_tests::pooling_parent_symlink_is_not_embedding_layout_evidence`, `src/server/webui/catalog_contract_tests.rs:496`
- `server::webui::catalog::tests::metadata_scan_reports_supported_complete_model_without_loading_provider`, `src/server/webui/catalog_tests.rs:117`

The contract test's diff shows a qwen3 fixture coming back with `"runnable_on_backend": false`, `"supported": false`, and every capability `"available": false` with reason `"architecture is not runnable on the current backend"` and `"runnable_on_backend_reason": "current backend does not advertise this architecture as runnable"`, where the expected shape has all of those true or null.

The defect is `runnable_on_backend` at `src/server/webui/catalog_metadata.rs:298-307`:

```rust
fn runnable_on_backend(family: &ArchitectureFamily) -> bool {
    let status = if cfg!(all(target_os = "macos", feature = "metal")) {
        family.backends.metal
    } else if cfg!(feature = "cuda") {
        family.backends.cuda
    } else {
        BackendStatus::Unsupported
    };
    matches!(status, BackendStatus::Supported | BackendStatus::Partial)
}
```

This is the same "there are two GPU backends and everything else is unsupported" assumption that #1803 removed in the kernel router and #1805 removed in hardware detection, surfacing one layer higher in the catalog. On a ROCm build both `cfg!` arms are false, so every architecture resolves to `BackendStatus::Unsupported` and the WebUI catalog tells a user that no model on the host can run, on a host where models do run. The five failing tests are the symptom; the user-visible defect is the catalog, which is why this is filed as a bug rather than a test fix.

The data the function reads has no ROCm column at all. `BackendSupport` at `src/models/registry.rs:111-114` is `{ metal: BackendStatus, cuda: BackendStatus }` and is the only consumer of the two-element arrays: `src/models/registry.rs:605-608` fills it from `caps.backends[0..1]` and `src/models/registry.rs:629-632` from `family.backends[0..1]`, both `[BackendStatus; 2]` (`src/models/registry.rs:80` and `src/models/registry.rs:159`). The source values are two sites only: `ModelType::capabilities()` at `src/models/registry.rs:469-476`, where `backends: [BackendStatus::Supported, cuda]` and `cuda` is `Partial` for `Kokoro`, `Whisper`, and `Gemma4` and `Supported` otherwise, and the single `STANDALONE_ARCHITECTURE_FAMILIES` entry (`rt_detr_v2`) at `src/models/registry.rs:176`. So the registry survey is small: one match arm and one literal, not a per-family audit. `src/server/webui/catalog_metadata.rs:300` and `:302` are the only reads of `family.backends` outside the registry, so the blast radius of widening the type is confined to these files.

Note that `BackendSupport` derives `Serialize` (`src/models/registry.rs:110`), so adding a field changes any JSON that carries it; confirm whether any endpoint or CLI output emits it before choosing the field name.

## Scope

**In scope:** `src/server/webui/catalog_metadata.rs` (the `runnable_on_backend` function), `src/models/registry.rs` (`BackendSupport`, the two `[BackendStatus; 2]` declarations, `ModelType::capabilities()`, and the `rt_detr_v2` standalone entry), and the five failing tests plus the catalog fixture they compare against.

**Out of scope:** the accuracy of per-architecture ROCm support claims beyond a first-pass value; treat the initial ROCm column as a starting point to be refined as families are actually validated on AMD, and say so in the code comment. Any other `cfg!(feature = "cuda")` two-backend assumption outside the catalog; file those separately as they are found.

## Proposed solution

Add a ROCm column to the registry data and a ROCm arm to `runnable_on_backend`, and select the arm from the runtime-resolved backend rather than from a `cfg!` chain where practical. PR #1883 adds `mlxcel_core::hardware::gpu_backend_kind()` (`src/lib/mlxcel-core/src/hardware.rs:164` on `feature/issue-1805-gpu-vendor`), which reports the backend at runtime, so the selection no longer has to be a compile-time guess. A `cfg!` chain also cannot express a build compiled with more than one GPU feature, which is the shape this bug keeps reappearing in. Keep a compile-time fallback only for the case where no GPU backend is compiled in at all.

**Coordinate before landing.** WebUI is under active concurrent development, so the catalog fixture and the contract test's expected shape are moving targets. Rebase on the current WebUI work and confirm the fixture owner's intent before changing `runnable_on_backend`'s output shape, rather than landing blind against a fixture that is about to be regenerated.

## Implementation plan

1. Add a `rocm: BackendStatus` field to `BackendSupport` at `src/models/registry.rs:111-114`, and widen `[BackendStatus; 2]` to `[BackendStatus; 3]` at `src/models/registry.rs:80` and `src/models/registry.rs:159` (or replace the positional array with a named struct, which removes the index-position coupling that made this easy to miss).
2. Populate the new column at the two source sites: the `backends:` literal in `ModelType::capabilities()` (`src/models/registry.rs:476`) and the `rt_detr_v2` entry (`src/models/registry.rs:176`). Decide the first-pass ROCm value per family and record the reasoning in a comment next to it.
3. Wire the field through the two projections at `src/models/registry.rs:605-608` and `src/models/registry.rs:629-632`.
4. Rewrite `runnable_on_backend` (`src/server/webui/catalog_metadata.rs:298`) to select the column from `mlxcel_core::hardware::gpu_backend_kind()`, mapping `Metal`, `Cuda`, and `Rocm` to their columns and falling back to `BackendStatus::Unsupported` only when no GPU backend is resolved. Keep the `matches!(status, Supported | Partial)` predicate unchanged.
5. Confirm whether `BackendSupport`'s serialized form reaches any HTTP response or CLI output; if it does, note the added field in the corresponding docs and treat it as an additive API change.
6. Regenerate or update the catalog fixture and the contract test's expected shape so the five tests assert the corrected behavior, after coordinating with in-flight WebUI work.
7. Add a regression test that pins `runnable_on_backend` per backend kind, so a future backend addition fails a test rather than silently reporting every architecture unsupported.

## Acceptance criteria

- [ ] On a ROCm build on the gfx1151 host, the WebUI catalog reports `runnable_on_backend: true` and `supported: true` for architectures the host can actually run, with `runnable_on_backend_reason` null.
- [ ] All five listed tests pass under `cargo test --workspace --profile test-fast --features rocm -p mlxcel --lib`.
- [x] `runnable_on_backend` selects its column from the resolved backend, and a build with no GPU backend still resolves to `Unsupported`. (#1921)
- [x] `BackendSupport` carries a ROCm column populated at every construction site, with no `[BackendStatus; 2]` left. (#1921)
- [x] A test pins the per-backend-kind mapping so adding a fourth backend without a column fails. (#1921)
- [ ] Metal and CUDA catalog output is byte-identical to before, verified by the unchanged contract fixture on those backends.
- [x] The change is integrated into the catalog the WebUI actually serves, not confined to the registry type. (#1921)

## Validation

```bash
cargo test --workspace --profile test-fast --features rocm -p mlxcel --lib server::webui::catalog
cargo test --workspace --profile test-fast --features rocm -p mlxcel --lib
cargo clippy --workspace --all-targets --features rocm -- -D warnings
cargo fmt --all -- --check
```

Regression guard on a CUDA host, which must stay unchanged per #1805:

```bash
cargo test --workspace --profile test-fast --features cuda -p mlxcel --lib server::webui::catalog
```

Manual check on the ROCm host: start the server, open the WebUI model catalog, and confirm a qwen3 model is listed as runnable with its capabilities available rather than every entry greyed out with "architecture is not runnable on the current backend".

## References

- Epic #1801 (AMD GPU (ROCm) backend on Linux via mlxcelverse), phase 1
- #1805 and PR #1883, where this was found and where `gpu_backend_kind()` is added
- #1803, which removed the same two-backend assumption from the kernel router

inureyes:
## Summary

The WebUI catalog picked its registry column with a `cfg!(metal) / cfg!(cuda) / else` chain, so every ROCm build reported every architecture as not runnable. It now reads the column for `mlxcel_core::hardware::gpu_backend_kind()`, the backend MLX resolved at runtime. `GpuBackendKind::None` reads as `Unsupported`. A build with no GPU backend resolves to `None`, and so does a ROCm build with no visible device.

- `BackendSupport` gains a `rocm` column and replaces both positional `[BackendStatus; 2]` arrays. None remain.
- Every family's ROCm column is `ROCM_FIRST_PASS = Partial`. The reasoning sits next to the const: the backend is experimental, fused kernels run as graph fallbacks until #1814, and few checkpoints have run on AMD. `Partial` keeps families runnable in the catalog without claiming Metal parity. Per-family refinement belongs to #1801.
- mlxcel-core gets `GpuBackendKind::ALL` and a compile-time guard. The guard's match has no wildcard, and each arm reads its own variant's slot of `ALL`. A new variant therefore fails to compile until it has an arm, and the arm fails to compile (`unconditional_panic` on `ALL[N]`) until `ALL` lists the variant. A catalog test walks `ALL`, so a listed kind with no registry column fails.
- `mlxcel arch --json` gains an additive `rocm` key under each family's `backends` object, documented in `docs/supported-models.md`. No HTTP response carries `BackendSupport`. The catalog still exposes only the `runnable_on_backend` boolean and its reason (`docs/webui/catalog.md` and `catalog.ko.md` now say it follows the runtime-resolved backend).

## Effect per build

- `--features metal` and `--features cuda`: unchanged. At the pinned MLX, `metal::is_available()` and `cu::is_available()` both return a compile-time `true`, so these builds always resolve `Metal` or `Cuda` and read the same column as before. The catalog contract fixture (`tests/fixtures/webui/examples/catalog.page.json`) and the five tests named in the issue are unchanged and pass on Metal.
- `--features rocm`: reads the new ROCm column, so runnable families show as runnable.
- macOS without the `metal` cargo feature: now reads the Metal column. `mlxcel-core/build.rs` builds MLX with Metal on for every macOS build, but the old chain also required `feature = "metal"`, so those builds reported nothing runnable. That includes the release workflow's `cargo build --release --target aarch64-apple-darwin --locked`, which uses default features. No tagged release is affected, because v0.7.0 predates the catalog.

## Verification (this host: Apple Silicon, Metal)

On the final tree:

- `cargo fmt --all -- --check`: pass.
- `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib server::webui::catalog`: 37 passed, 0 failed. This includes the five tests named in the issue and the four new `backend_tests`.
- `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib models::registry`: 10 passed, 0 failed. This includes `every_family_is_runnable_on_rocm` and the extended `registry_json_round_trips`, which pins the `backends` JSON for `qwen3`, `whisper` and `rt_detr_v2`.
- `cargo clippy -p mlxcel --lib --tests --features metal,accelerate -- -D warnings`: pass.

Revert checks (each change was restored afterwards):

- Dropping the `Rocm` arm from `backend_status_for` fails `each_backend_kind_reads_its_own_column` and `every_gpu_backend_kind_has_a_registry_column` (2 passed, 2 failed).
- Wiring `Rocm` to the CUDA column fails `each_backend_kind_reads_its_own_column`.
- Setting `ROCM_FIRST_PASS` to `Unsupported`, which reproduces the issue's defect in the registry, fails `every_family_is_runnable_on_rocm` and `registry_json_round_trips` (8 passed, 2 failed).
- Forcing the catalog to resolve `None`, which is what the old chain did on ROCm, fails 7 of the 37 catalog tests, among them the five issue tests and the live-wiring test. Forcing it to resolve `Rocm` passes all 37.
- Swapping two entries of `GpuBackendKind::ALL` stops mlxcel-core compiling (E0080). A standalone rustc build of the guard shows the missing-variant case: a fifth variant that is given an arm but left out of `ALL` fails to compile, and adding it to `ALL` makes the build compile again.

A search for other Metal/CUDA/else `cfg!` chains found none. The remaining `cfg!(feature = "cuda")` sites gate CUDA-only behavior, so nothing was filed.

## Real-binary verification (Metal)

Run by the epic #1910 orchestrator on M1 Ultra (macOS 27.0) with `mlxcel-server` and `mlxcel` built from this PR head (`988aa718`) under `--profile test-fast --features metal,accelerate`, serving the 224-entry local store with `--webui --models-dir models/mlx --no-models-autoload`.

- Every catalog page (224 items) was dumped from this binary and from a binary without this change (`414d45dc`, which does not touch the catalog). `supported`, `metadata.support` (`architecturally_supported`, `runnable_on_backend` and both reasons) and every `pre_load` capability are identical for all 224 entries: 201 supported and 209 runnable on both. The only raw difference was the `provider_ready` capabilities of the one model that was loaded during the baseline capture.
- `mlxcel arch --json`: all 181 `backends` objects carry `"rocm": "partial"` next to the unchanged `metal` and `cuda` values.

## Not verified on this host

These need the gfx1151 ROCm host or a CUDA host and were not run:

- `cargo test --workspace --profile test-fast --features rocm -p mlxcel --lib server::webui::catalog`
- `cargo test --workspace --profile test-fast --features rocm -p mlxcel --lib`
- `cargo clippy --workspace --all-targets --features rocm -- -D warnings`
- CUDA regression guard: `cargo test --workspace --profile test-fast --features cuda -p mlxcel --lib server::webui::catalog`
- Manual: on the gfx1151 host, start the server, open the WebUI model catalog, and confirm a qwen3 model is listed as runnable with its capabilities available.

In their place, this host ran the following. The pure mapping tests exercise `GpuBackendKind::Rocm` and `GpuBackendKind::None` against every column layout. The `ALL` walk fails for any kind without a column. `every_family_is_runnable_on_rocm` pins the registry side. The forced-`Rocm` catalog run passes all 37 catalog tests. The unchanged Metal contract fixture passes.

Closes #1886


inureyes:

### Intent
> The WebUI catalog reads the registry column for the GPU backend MLX resolved at runtime, and the registry gains a ROCm column, so ROCm builds stop reporting every architecture as not runnable.

### Findings Addressed
None at CRITICAL or HIGH; no commits added.

### Remaining Items
- [ ] (MEDIUM) The `GpuBackendKind::ALL` const guard rejects a missing arm, a misordered list and a repeat, but not a variant left out of `ALL`: with a fifth variant, writing only its arm (`=> 4`) compiles and the catalog's `ALL` walk never sees it (checked with standalone `rustc`). "keeps the list complete" overstates it. Enforcing variant, also checked: arms of a nested `const fn` return `GpuBackendKind::ALL[n]`, so the new arm's `ALL[4]` trips the deny-by-default `unconditional_panic` lint until `ALL` grows.
- [ ] (MEDIUM) The PR body (the future squash commit body) misstates the backend deltas. MLX's `cu::is_available()` is unconditionally `true` on a CUDA build, so CUDA always resolves `Cuda` and its catalog output is unchanged even without a GPU. The real delta is macOS without the `metal` cargo feature: MLX is still built with Metal there, so the catalog now reads the Metal column where the old chain reported nothing runnable. `release.yml` builds the shipped macOS binaries that way, so this PR also fixes that catalog on main.
- [ ] (LOW) `docs/supported-models.md:742` writes `lablup/mlxcel#1801`; the rest of the file uses bare `#NNN`.

### Verification
- [x] All stated requirements implemented (ROCm-host and CUDA-host items remain unverified here, as listed)
- [x] Tests pass: catalog 37/37, registry 10/10, clippy and fmt clean. With `runnable_on_backend` forced to `GpuBackendKind::Rocm`, catalog passes 37/37 including the five named tests; forced to `GpuBackendKind::None` (what the old chain did on ROCm), those five fail again.

inureyes:
Security and performance review of 44167ea4. No CRITICAL, HIGH or MEDIUM findings, so nothing was auto-fixed and no commit was added.

- Exposure: the catalog response shape is unchanged (the `runnable_on_backend` bool and its fixed reason string). `BackendSupport` is serialized only by `mlxcel arch --json`; no HTTP route emits it. No load path reads the catalog's `supported` flag, so `ROCM_FIRST_PASS = Partial` changes what the UI advertises, not what the server allows.
- Request cost: `runnable_on_backend` runs only on a catalog cache miss, and after the first call `gpu_backend_kind()` is a `OnceLock` read. Nothing is added per request.
- Probe trigger: router startup returns before `initialize_runtime_checked()`, so in router mode the first catalog cache miss can be the first probe in the process. It runs inside `spawn_blocking`, is memoized in Rust and C++, and takes no request input, so a client cannot repeat it. On Metal and CUDA it resolves from a compile-time constant; on ROCm it is one `hipGetDeviceCount`, which the first in-process model load runs anyway.
- LOW, not fixed: with the pinned MLX, `cu::is_available()` returns `true` unconditionally, so a CUDA build resolves `Cuda` even with no driver. The PR body's paragraph about a driverless CUDA host describes a state that cannot occur (CUDA catalog output is unchanged in every case), and "or available" in the `backend_status_for` doc comment holds only for ROCm.

On this Metal host: fmt pass, catalog tests 37 passed, registry tests 10 passed, clippy pass. Wiring `Rocm` to the CUDA column fails `each_backend_kind_reads_its_own_column` (3 passed, 1 failed), then restored. ROCm and CUDA host runs were not performed.

inureyes:
## PR finalization check

Verified against the requirements brief for issue #1886 on this Metal host (`988aa718`).

- Tests: `cargo test --profile test-fast --features metal,accelerate -p mlxcel --lib server::webui::catalog` (37 passed, 0 failed, includes the 4 new `backend_tests`) and `... --lib models::registry` (10 passed, 0 failed, includes `every_family_is_runnable_on_rocm` and the extended `registry_json_round_trips`).
- Lint/format: `cargo clippy -p mlxcel --lib --tests --features metal,accelerate -- -D warnings` and `cargo fmt --all -- --check` both pass.
- Docs: confirmed no other English or Korean doc (README.md, docs/installation.md's ROCm status table, the other `docs/webui/*.md`/`*.ko.md` files) describes the two-backend catalog contract or the `arch --json` backend fields; only `docs/supported-models.md` and `docs/webui/catalog.md`/`catalog.ko.md` line 29 needed the update already in this PR.
- No `[BackendStatus; 2]` remains anywhere, `recipes/registry/*`, `CHANGELOG.md`, and the named catalog fixture/tests are untouched, and the bare `#1886`/`#1801`/`#1814` references all resolve to real lablup/mlxcel issues.

Working tree is clean and matches `origin/fix/issue-1886-rocm-catalog-backend`; nothing was left to commit. ROCm-host acceptance items remain unverified here, as documented in the PR body.

PR #1922, opened 2026-09-19
refactor(webui): key screen copy in the catalog and drop window.confirm

inureyes:
Chat, Settings and Activity now resolve every user-facing string through `t(locale, key)`, and Chat's four `window.confirm` prompts are styled, localized `ConfirmDialog`s. Before this, the Korean Chat screen showed only its title in Korean, and that title was wrong (`채팅` instead of the catalog's `대화`).

- New `chatStrings` and `settingsStrings` modules and a rewritten `activityStrings` (same shape as `modelStrings`) feed `StringKey`. `models.next_profile.*` and `format.*` are inline catalog entries. Settings and Activity Korean moved verbatim; Chat copy has new translations. 262 entries generated from the catalog are appended to `tests/fixtures/webui/strings.json`; the existing 250 are byte-identical.
- `ConfirmDialog` (`primitives.tsx`, built from `Dialog` and `Button`) handles Edit on a turn, replace with saved history, Clear All and replace with imported history. Each answer is consumed once and dropped if stale: a newer history epoch, a started generation, or an edit whose conversation or turn id no longer matches. Declining saved history still disables saving, and a throw from the replacement shows the same banner as before. Clear All renders `settings.clear_history.confirm.body`. Focus returns to the trigger; the Chat heading is the fallback while clearing disables every control.
- `webui/src/i18n/drift.test.ts` walks `webui/src`, skipping tests and `i18n/`. It fails with `file:line` on native dialogs (`window.` or bare calls), `locale` compared with a locale literal in either order, and inline `words` helpers. It also checks the new modules against the fixture.
- Model pickers use `lifecycleLabel` (`alpha · Ready`), turn status is keyed (`chat.transcript.status.*`), and `newConversation(title)` now takes a localized title.
- `t()` substitutes only own properties of `values`.
- `docs/webui/api.yaml`: `StringCatalog.strings.maxItems` goes from 256 to 1024 because the fixture now holds 512 entries. This schema is fixture-only, so no `/ui-api/v1` endpoint changes.

## Validation

No cargo. `typecheck`, `lint`, `unit` (337 Vitest, 18 node), `build_bundle.py` plus `--verify`, `check_webui_contract.py`, `webui_contract_checks.py` and `check_binary_assets.py` pass on the final head. Playwright Chromium ran one spec per invocation on private port 4913 (`--strictPort`, no server reuse): chat, settings, activity, models and browser all pass. The three acceptance greps print nothing.

[...]

Issue #1902, opened 2026-09-16
refactor(webui): adopt every shared @lablup/ui-common component


inureyes:
Part of #1910.

Part of #1834.

## Problem / Background

The bundled WebUI depends on `@lablup/ui-common@0.1.0-alpha.19` (`webui/package.json:23`, which arrived with #1863) and uses 7 of the 17 components that package exports. Five of the remaining ten are duplicated by local code and five have no local counterpart at all. The goal recorded on 2026-09-16 is that the WebUI uses the shared components everywhere it can, so this issue migrates all ten.

## Current Behavior

`webui/node_modules/@lablup/ui-common/dist/index.d.ts` re-exports 17 components: Badge, BaseCard, Button, DataTable, Drawer, EmptyState, ErrorState, PageHeader, PageLayout, ProgressBar, Select, Skeleton, SmoothHeight, StatCard, StatusTag, Tabs, Tooltip. The Skeleton entry also exports SkeletonCard, SkeletonChart, SkeletonRow and SkeletonText, the StatCard entry exports `formatCompactNumber`, and `dist/hooks/index.d.ts` exports `usePrefersReducedMotion`. There is no Dialog and no Modal anywhere in that export surface.

Seven are in use: Button, DataTable, EmptyState, ProgressBar, StatusTag and Tabs through `webui/src/design-system/common-adapters.tsx:3-8,62`, and Select through `webui/src/design-system/common-select.tsx:3`. Those two files plus `webui/src/main.tsx:17`, which imports `styles/base.css`, are the only three files in the whole application that import `@lablup/ui-common`; every page composes the local `design-system` surface instead. That is why this migration is almost entirely confined to the design-system directory rather than spread across the feature screens: `webui/src/design-system/primitives.tsx` is 194 lines and `webui/src/design-system/shell.tsx` is 110 lines.

## Proposed Solution

Adopt each shared component behind the existing adapter seam (`common-adapters.tsx`, `common-select.tsx`) and keep re-exporting from `primitives.tsx`, so feature screens continue to import from `./design-system/primitives` and the diff stays in the design system.

[...]

PR #1923, opened 2026-09-19
feat(webui): adopt the ui-common theme system and add a glass theme

### LINKED ISSUE #1903: feat(webui): adopt the ui-common theme system and add a glass theme
Part of #1910.

Relates to epic #1834. The `@lablup/ui-common` dependency arrived with #1863; adopting the package's unused *components* is tracked in #1902 and is out of scope here.

## Problem / Background

Decision taken 2026-09-16: the bundled WebUI should take its theming from the `@lablup/ui-common` theme system instead of hand-configuring the token contract, and should carry a glass theme. `lablup/backend.ai-go` is the reference to study because its glass family already targets the design language epic #1834 set for mlxcel.

What ui-common actually provides, measured at the pinned `0.1.0-alpha.19` (`webui/package.json:18`): `dist/styles/base.css` declares the `--token-*` contract on `:root` (122 `--token-*` declarations; its own header describes it as the set every component in the package reads), and `dist/styles/themes/` ships exactly two files, `orange-light.css` and `orange-dark.css`, each overriding a subset of the contract (55 tokens in `orange-light.css`) under `[data-theme="orange-light"]` and `[data-theme="orange-dark"]`. The rule is written in those theme files' header: import at most one theme file alongside `base.css`, and components never import a theme file themselves. There is no glass theme in the package.

## Current Behavior

`webui/src/main.tsx:17` imports `@lablup/ui-common/styles/base.css` and no theme file at all. `webui/src/design-system/common-tokens.css` then assigns 122 `--token-*` values on a bare `:root`, deriving them from local `--color-*`, `--material-*`, `--radius-*` and `--shadow-*` tokens. So mlxcel already feeds the token contract, but by hand rather than through the theme mechanism, which is why no shipped or future theme file can currently take effect.

`webui/src/design-system/preferences.ts:101` writes `root.dataset.theme = preferences.theme`, where `ThemePreference` is `'system' | 'light' | 'dark'` (`preferences.ts:3`) and the default is `'system'` (`preferences.ts:18`). `webui/src/design-system/tokens.css` keys off `[data-theme="dark"]` at lines 47 and 75 and `[data-theme="light"]` at lines 48 and 51, with the `system` case carried by `@media (prefers-color-scheme: dark) { :root:not([data-theme="light"]) }` at lines 50-51.

`webui/src/design-system/glass-intensity.css` is 102 lines implementing the epic's glass intensity scale from 0 to 100 with default 35 (`preferences.ts:20`), driven by `:root[data-glass-intensity="N"]` and setting `--material-glass-alpha`, `--material-glass-blur` and `--material-glass-saturation`. The three local token files total 334 lines: `tokens.css` 106, `common-tokens.css` 126, `glass-intensity.css` 102.

`webui/index.html` is 13 lines and contains no early theme bootstrap, so nothing prevents a flash of the wrong theme before the app mounts.

## Highest risk: silent dead selectors

If mlxcel moves to family-scoped theme ids, all four theme selectors in `webui/src/design-system/tokens.css` (lines 47, 48, 51, 75) stop matching silently. Nothing fails to compile, no test necessarily fails, and dark mode simply stops applying. That is precisely the failure `lablup/backend.ai-go` built a CI gate for, so mlxcel should adopt an equivalent gate **in the same change** rather than after the first regression, and the gate should land before or with the id change and be wired into the existing `webui-bundle` job in `.github/workflows/ci.yml:390` the way the other repository checks are.

## Reference implementation to study (do not copy blindly)

- `data-theme` carries a combined identifier of family and scheme, not a bare mode: `src/themes/index.ts:49-61` enumerates twelve `ThemeId` values across six families (`orange`, `stained`, `glass`, `glass-legacy`, `reverie`, `bliss`), each as `<family>-light` and `<family>-dark`.
- The module also exports `ThemeFamily` (`index.ts:38`), `ColorScheme` with a `system` member, validators, system preference resolution, `getEffectiveColorScheme` (`index.ts:241`), `getThemeId` (`index.ts:251`), and a migration path from a legacy flat theme value.
- Theme tokens live in `src/themes/tokens/`, one file per theme id, twelve in total. `glass-light.css` is 425 lines and `glass-dark.css` is 427, each declaring 192 distinct `--token-*` names, with 132 declarations in the light file involving `backdrop-filter`, `rgba` or gradients.
- The glass file header states the theme is a frosted glass design inspired by Apple's Liquid Glass from macOS 27, using `backdrop-filter: blur(12px)`, semi-transparent overlays with subtle gradients, light reflection for depth, soft transparent borders, and fallbacks for browsers without `backdrop-filter`. That is the same design language epic #1834 set for mlxcel, which is why it is the right reference.
- All four glass token files (`glass-light`, `glass-dark`, `glass-legacy-light`, `glass-legacy-dark`) carry a `@media (prefers-reduced-transparency: reduce)` block.
- `index.html` prevents a flash by setting `data-theme`, `data-theme-family` and `data-color-scheme` on the document element from an inline script before the app boots (lines 182-184), resolving a `system` scheme through `matchMedia("(prefers-color-scheme: dark)")` (lines 175-176), with a three second fallback that reveals the body if the app fails to initialize (lines 191-200).
- `scripts/check-theme-selectors.mjs` is the CI gate, referenced to that repository's issue 4201, and it rejects selectors that can never match once ids are family-scoped. Its own documentation gives the example that `[data-theme="dark"]` matches nothing ever, and that the working forms are `[data-theme$="-dark"]` or the explicit `[data-theme="<family>-dark"]`. It also rejects reading the raw preference attribute and gating on `@media (prefers-color-scheme: dark)`, because the first can hold the literal string `system` and the second reflects the host OS rather than the applied theme. It has an escape hatch that requires a written reason on the preceding or same line, so an exemption is a sentence someone defends in review rather than a silent bypass. It runs as `pnpm check:theme-selectors` and as a step in that repository's quality-gates workflow.

## Scope (recommended sequence, not a mandate)

1. **Decide the theme id scheme first**, since everything else depends on it, and write down whether mlxcel adopts family-scoped ids or stays with bare modes and forgoes shipped theme files. Note that `applyAppearance` currently writes the raw preference, so `data-theme="system"` is written verbatim under the default and neither `tokens.css:47` nor `tokens.css:48` matches then; whatever scheme is chosen should separate the stored preference from the applied id.
2. **Land the theme-selector CI gate**, wired into the `webui-bundle` job.
3. **Move the hand-written `--token-*` configuration** in `common-tokens.css` into a theme file the mechanism can select, keeping the current appearance as the baseline so this step is behaviour-preserving and reviewable.
4. **Add the glass theme in light and dark**, wired to the existing glass intensity, reduce transparency and contrast controls.
5. **Add the early bootstrap to `webui/index.html`** so the theme is applied before the app mounts, reading the existing `mlxcel.webui.appearance` localStorage key (`preferences.ts:27`), and confirm it survives the embedded-bundle build rather than only the dev server.

**Out of scope:** adopting the unused ui-common *components* (tracked in #1902); any change to the model-actions, settings, or observation surfaces from epic #1834.

## Implementation Notes

- **Reuse**: `glass-intensity.css` and the `applyAppearance` attribute writer at `preferences.ts:100-108` already carry `data-material`, `data-glass-intensity`, `data-reduce-motion`, `data-reduce-transparency`, `data-high-contrast` and `data-backdrop-filter`; the glass theme must read those rather than introduce a parallel control surface. `expectAxeClean` at `webui/tests/browser-assertions.ts:51` already runs axe with `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa` and `wcag22aa` tags and is the hook for per-theme contrast coverage.
- **Keep, do not lose**: the epic's visual contract, glass on sidebar, toolbar and popovers while transcript, tables and forms get stable readable surfaces, motion 120 to 180 ms, no shimmer and no continuous backdrop animation.
- **Keep, do not lose**: the glass intensity control from 0 to 100 with default 35, reduce transparency and reduce motion overrides, and high contrast support, with contrast settings overriding decorative intensity.
- **Keep, do not lose**: an opaque fallback where `backdrop-filter` is unavailable (`supportsBackdropFilter()` at `preferences.ts:95-98` already reports this into `data-backdrop-filter`).
- **Keep, do not lose**: the manual pass the maintainer completed on 2026-09-16 against the installed artifact, covering keyboard order, Escape with focus restoration, native 200 percent zoom with no horizontal page scroll at 390 px, and readability with increased contrast. A theme change can break contrast and readability without breaking a test, so contrast must be checked across every shipped theme and scheme combination, and the 390 px and 200 percent cases re-verified.
- **Attribution**: if glass token values are derived from `lablup/backend.ai-go` rather than written fresh, record the derivation the way this repository requires for adapted code, per the Third-party attribution section of `CLAUDE.md` (per-file derivation line plus a top-level `NOTICE` entry).
- **Edge cases**: `data-theme` written as `system` under the default preference; storage blocked or full, where `preferences.ts` already falls back to in-memory values; a browser without `backdrop-filter`; `prefers-reduced-transparency: reduce` and `prefers-contrast: more` set simultaneously with a high decorative intensity, where contrast must win.

## Acceptance Criteria

- [ ] The app selects its theme through the ui-common mechanism, with at most one theme file imported alongside `base.css`, and no component importing a theme file.
- [ ] A glass theme exists in light and dark and honours glass intensity, reduced transparency, reduced motion and increased contrast.
- [ ] No CSS selector keys off a theme id that cannot match, enforced by a CI gate in the `webui-bundle` job rather than by review, and the gate fails on a deliberately introduced dead selector.
- [ ] No flash of the wrong theme before mount, verified against the installed artifact and not only the dev server.
- [ ] WCAG 2.2 AA contrast holds for every shipped theme and scheme combination, and the 390 px layout and 200 percent zoom behaviour is re-verified.
- [ ] Any values derived from `lablup/backend.ai-go` are attributed as the repository requires, with a `NOTICE` entry.
- [ ] Integrated into the real WebUI flow and shipped bundle, not left as an unreferenced stylesheet.

## Verification

```bash
pnpm --dir webui install --frozen-lockfile
pnpm --dir webui run lint
pnpm --dir webui run unit
pnpm --dir webui run check:theme-selectors   # new gate
pnpm --dir webui run browser
python3 scripts/webui/build_bundle.py --verify
```

A pass is: every command exits 0, the new gate fails when a dead selector such as `[data-theme="dark"]` is reintroduced (revert-and-watch-it-fail, not just a green run), and axe reports no violations for each shipped theme and scheme combination. Manual, against the built binary serving the embedded bundle rather than the dev server: load with a cold profile and confirm no flash of the wrong theme, then re-check keyboard order, Escape with focus restoration, 200 percent zoom with no horizontal page scroll at 390 px width, and readability with increased contrast.

## Technical Considerations

The ui-common package ships only the orange family today, so a glass theme has to live in this repository until (or unless) it is upstreamed. Whether the glass tokens belong in `webui/src/design-system/` or should be proposed back to `@lablup/ui-common` is worth deciding explicitly in this issue rather than by default.

inureyes:
Closes #1903. Part of #1910.

## What changes

- **Theme ids.** `data-theme` always holds `<family>-<scheme>`: `mlxcel-light`, `mlxcel-dark` (the approved baseline, still the default), `glass-light`, `glass-dark`. `webui/src/design-system/theme.ts` resolves it; the stored `themeFamily` and `colorScheme` (`system`/`light`/`dark`) land in `data-theme-family` and `data-color-scheme`. The old flat `theme` field is migrated, and `system` follows host changes live. Settings gets Theme (Standard, Glass) and Color scheme selects.
- **ui-common mechanism.** `main.tsx` imports one theme entry (`themes/index.css`) after `base.css`, and no ui-common theme file. The color, shadow and button `--token-*` values moved from `:root` into `[data-theme]` blocks unchanged; `common-tokens.css` keeps the theme-independent half.
- **Glass theme.** Translucent chrome with a light reflection, tinted buttons, highlighted primary and danger fills, near-opaque content surfaces, no per-button blur. It reads the existing intensity, material, transparency, backdrop support, contrast and motion switches plus the `prefers-*` queries; any transparency or contrast preference makes it opaque at any intensity. Values derive from lablup/backend.ai-go (derivation line, `NOTICE`); the tokens stay here because they bind to this product's semantic token layer.
- **No flash.** `webui/public/theme-bootstrap.js` is a classic render-blocking script in `<head>`, external because the CSP is `script-src 'self'`. `build_bundle.py` rejects a bundle where it is missing, duplicated, deferred in any way, or after the app module, and any inline script.
- **Gate.** `check:theme-selectors` (in the `webui-bundle` job and `make verify-webui-frontend`) evaluates every `data-theme`/`data-theme-family` selector against the shipped ids and rejects `data-color-scheme` selectors, `prefers-color-scheme` outside the resolver, stray theme imports and ids with no theme block; exemptions need a written reason. Against `main` it reports `tokens.css:47`, `:48`, `:50`, `:51`, `:75`.

## Found along the way

- Chrome blur never rendered in Chromium or Firefox (Lightning CSS drops a standard `backdrop-filter` that precedes its `-webkit-` twin), and the primary and danger fills were hard-coded over ui-common's rules. Both are one-line fixes in `components.css`, and both survive the merge with #1902's rewrite of that file, alongside its `:where(:not(:disabled))` hover rules.
- `features/chat/style.test.ts` passed vacuously (Vitest blanks CSS imports); it now reads from disk. `models.css` read two undefined tokens; inspector labels are now muted and the selected row highlighted.
- Review fixed rendered glass fills and a bootstrap check that let `type="module"` through; the security pass fixed the gate skipping its check through a symlink and two super-linear patterns.

## Validation

Rebased onto `fd33ff15` (#1919, #1921, #1922) with an unchanged source diff, then merged with main at `5616f86b` (#1924, #1902's component adoption); `app.tsx` keeps #1902's PageHeader Settings screen with the theme selects, and the bundle was regenerated. On `cdb80fe7`: typecheck, lint, the gate, unit (35 node:test, 470 Vitest), Chromium suite (88, including the glass button fill case and #1902's button hover cases; axe on all four ids), `build_bundle.py --verify`, contract, binary-asset and cross-repo checks. The engines smoke on Chromium, Firefox and WebKit passed before the rebase. Each new check was reverted to confirm it fails.

## Bundled-binary checks

Playwright Chromium against `mlxcel-server` built from `c26e3247` (before the rebase and the #1924 merge) with the embedded bundle and no model loaded.

1. **No flash: pass.** `theme-bootstrap.js` is served 200, `text/javascript`, `no-cache`, under the strict CSP, ahead of the app module. With the app module held, all nine cases (cold profile under light and dark hosts, each id stored under the opposite host, glass `system` under both hosts, legacy `{theme: dark}`) had the right `data-theme` and canvas with nothing mounted, and no CSP violation.
2. **Settings: pass.** Glass then Dark gives `glass-dark`, persists without the legacy field, is restored before mount on reload; `system` follows a host flip live.
3. **390x844: pass** in both families: 0 px overflow on all four routes; the sheet opens, navigates, closes on Escape and returns focus.
4. **200 percent zoom: not native** (headless Chromium cannot zoom; a half-width CSS viewport at deviceScaleFactor 2 stood in). 320 and 640 CSS px: 0 px overflow everywhere. 195 CSS px (390 px window): 125 px on every route from `body { min-width: 320px }`; main's bundle measures the same, so it predates this PR.
5. **Keyboard: pass** for `mlxcel-light` and `glass-dark`: 18 Tab stops in DOM order, each with a visible indicator; Escape restores focus from the command palette (button and Cmd+K), a Select popup and the sheet.
6. **Contrast: pass.** axe WCAG 2.2 AA on all four routes for all four ids at intensity 100: 0 violations normally, under `prefers-contrast: more` and with High contrast on; the chrome drops to `blur(0px)` with no gradient.
7. **Served-CSP spec: pass** (both variants).

## Merge with main

`origin/main` 31c04592 (#1920, verbatim display names) was merged into this branch as 83fc0580 by the epic #1910 orchestrator. Only the generated bundle conflicted; it was regenerated with `scripts/webui/build_bundle.py` (`--verify` digest 986294f2) and not hand-merged. On the merged tree: `typecheck`, `lint`, `check:theme-selectors` (152 files, 4 theme ids), `unit` (Vitest 473 in 51 files) and the Chromium Playwright suite on a private port (88 passed) all pass. The live-server script above was also re-run against a server built from cdb80fe7 (after #1924 was merged in): 23 of 25 cases pass, and the two failures are the pre-existing 195 CSS px overflow from `body { min-width: 320px }` described above.


inureyes:

### Intent
> Theme the WebUI through the ui-common `[data-theme]` mechanism with family-scoped ids, add a glass family in light and dark, apply it before mount, and gate dead theme selectors in CI.

### Findings Addressed
- [x] Glass primary and danger buttons never rendered their highlighted fill: `components.css` hard-coded `#245bd8` / `#c6352b` at the same specificity as ui-common's `.button--primary` and later in the bundle, so both glass schemes painted the mlxcel fill at rest and the glass hue only on hover. The two rules now read `--token-buttonPrimaryBg` / `--token-buttonDangerBg` (mlxcel resolves to the same colors). `theme.spec.ts` now asserts the rendered fill; with the hex values put back the new glass case fails (`background-image: none`). Fixed in ca0371d8. (HIGH)

### Remaining Items
- [ ] `scripts/webui/build_bundle.py:232` and `theme-bootstrap.test.ts:99` test `\b(type="module"|async|defer)\b`. The trailing `\b` after a quote never matches before a space or `>`, so a bootstrap tag with `type="module"` (deferred, can paint first) passes both guards, and the "before the app module" check then finds the bootstrap's own `type="module"`. Current markup is correct; the guard is weaker than it claims. Suggested: `\btype="module"|\basync\b|\bdefer\b`, and locate the app module by `src="./assets/`. (MEDIUM)
- [ ] `themes/glass.css:97-98` gives the danger button the primary's blue `--glass-button-shadow` (`rgba(0, 88, 185, 0.2)` in light); lablup/backend.ai-go uses a red-tinted danger shadow. Now visible since the fill renders. (LOW)
- [ ] `themes/glass.css:13` names upstream files by a bare repository path; the repository's upstream-reference rule asks for `https://github.com/lablup/backend.ai-go/blob/main/src/themes/tokens/glass-light.css` style URLs. (LOW)
- [ ] Glass `--token-buttonPrimaryBorder` / `--token-buttonDangerBorder` (glass edge) never render: `common-components.css:3` sets the inline button border. Decorative only; that file belongs to #1902. (LOW)
- [ ] The bootstrap parity matrix in `theme-bootstrap.test.ts` lists families by hand, so a third family added to `THEME_FAMILIES` but not to `theme-bootstrap.js` would not fail the unit gate. (LOW)
- [ ] Pending by design: no-flash on the embedded bundle per theme id, native 200 percent zoom and 390 px, keyboard order and Escape focus restoration, and increased-contrast readability against the installed binary.

### Verification
- [x] All stated requirements implemented (bundled-binary acceptance checks pending, as stated in the PR)
- [x] Integrated into project code flow (main.tsx entry, index.html bootstrap embedded by rust-embed, Settings controls, CI `webui-bundle` job)
- [x] Tests pass: typecheck, lint, `check:theme-selectors` (also fails on a reintroduced `[data-theme="dark"]`), unit (32 node:test, 401 Vitest), Chromium suite (60, up from 59 with the new rendered-fill case), `build_bundle.py --verify`, `check_binary_assets.py`, `check_webui_contract.py --self-test`

inureyes:
## Security and performance review (de46f4fe)

No CRITICAL or HIGH findings, so nothing was auto-fixed. Green: typecheck, lint, check:theme-selectors, unit (32 node:test, 425 Vitest), Chromium suite (60), `build_bundle.py --verify`, `check_binary_assets.py`, `check_webui_contract.py --self-test`. Not run: `tests/csp.spec.ts`, which needs a running server.

Clean: `theme-bootstrap.js`, fuzzed over 105 stored-value and `matchMedia` combinations (corrupt JSON, `__proto__` keys, non-string values, markup, a 5 MB payload, throwing storage and getters), never throws and writes only allow-listed values. No inline script or style; the bootstrap is same-origin under `script-src 'self'`. Glass adds no animation, and every opaque or contrast path removes the blur. The `system` listener is removed on cleanup.

Remaining:

- [ ] MEDIUM `webui/scripts/check-theme-selectors.mjs:277`: the entry guard compares `path.resolve(argv[1])` with the realpath in `import.meta.url`, so running the script by a symlinked absolute path exits 0 with no output even with a dead selector present (reproduced). `pnpm --dir webui run` is unaffected. Compare against `realpathSync(process.argv[1])`.
- [ ] LOW `check-theme-selectors.mjs:83`: `SCRIPT_IMPORT_RE` backtracks cubically on `import` followed by long whitespace (800 spaces: 275 ms); the attribute and `@media` patterns are quadratic. Repository input only.
- [ ] LOW gate gaps, none present today: uppercase attribute names, escapes or comments inside a selector, conditional `@import` and `<link media>` on `prefers-color-scheme`, `dataset['theme']`, backtick values; `.jsx`, `.cjs`, `.mts`, `.cts` files and dot-directories are not scanned.
- [ ] LOW `scripts/webui/build_bundle.py:232` and `theme-bootstrap.test.ts:103` accept a `nomodule` bootstrap tag, which browsers never run (reproduced against `assert_index_html`).
- [ ] LOW `themes/glass.css:105` blurs every ui-common secondary and icon button, transcript and table rows included; in WebKit each is its own backdrop layer during streaming scroll. Unmeasured; consider limiting control glass to the chrome.
- [ ] LOW (note) the `components.css` order swap turns on the chrome blur in Chromium and Firefox for the default family, so the sticky toolbar now re-blurs scrolled content there.

inureyes:
## PR finalization pass

Checked test coverage, documentation, and lint/format for #1903 (theme system). Everything already documented in the brief (design-system.md, ui-common.md, ux-contract.md, bundling.md/bundling.ko.md, NOTICE, README) was verified accurate and complete; no duplication added.

**Added:** two `preferences.test.ts` cases covering `clampIntensity`'s boundary behavior (negative, over 100, fractional, non-finite and non-numeric stored/direct values) through `loadAppearance` and `applyAppearance`. `glass-intensity.css` has one discrete `[data-glass-intensity="N"]` rule per integer 0-100, so an unclamped value would silently drop the glass blur and saturation with no visible error; that path had no prior test.

**Updated:** the Vitest count in `docs/webui/design-system.md` (Verification section) from 425 to 427, and `src/webui/assets/mlxcel-webui-manifest.json` (`source_digest_sha256` only, asset bytes unchanged) since the test file lives under `webui/src`. Also corrected stale validation counts in the PR description (32/401/59 to the current 35/427/60).

**Gates run, all green:** `pnpm --dir webui run typecheck`, `lint`, `check:theme-selectors`, `unit` (35 node:test, 427 Vitest across 40 files); the full Chromium suite via the isolated Playwright config (60 passed, port 4903); `python3 scripts/webui/build_bundle.py` and `--verify`; `check_webui_contract.py --self-test`; `check_binary_assets.py`; `check_cross_repo_refs.py`.

**Left open:** the bundled-binary acceptance checks (no-flash on the embedded bundle, native 200 percent zoom, 390 px, live contrast) remain pending an orchestrator-provided server, as already noted in the PR description.

Commit: c26e3247.

Issue #1918, opened 2026-09-18
feat(webui): redesign the model library as a dense, task-first list with row actions


inureyes:
Part of #1910.

## Problem / Background

Measured 2026-09-17 against a release build of main `0ef0a1a4` with 212 checkpoints in `models/mlx`: the Models page is 3,132 px tall at 1440x900 with four rows visible, and 5,936 px at 390x844. Each row is about 100 px because the name cell (`webui/src/features/models/screen.tsx:185-206`) stacks a ghost `Button` styled as a pill, a "Selected" label and a `source · quantization` line. The table has four columns (Model, Task, Lifecycle, Support / Completeness); there is no size column although `metadata.disk_bytes` is known for every entry, and the Support column repeats "Supported architecture / Complete files" on almost every row. Load, Use in Chat, Unload and Delete exist only inside the inspector (`webui/src/features/models/inspector.tsx:77-101`), which opens only through the name button (#1901), so the shortest path from a fresh page to chatting is: search, click the name pill, scroll the inspector to its bottom, click Load, click Use in Chat. The #1834 contract says the Models primary path is "Search/filter local library; inspect; Load; Use in Chat; Unload".

[...]

Issue #1914, opened 2026-09-18
feat(webui): app shell shows loaded models, uses PageLayout/PageHeader, and completes the global shortcuts


inureyes:
Part of #1910.

## Problem / Background
The toolbar pill in `webui/src/design-system/shell.tsx:77` prints `selectedModelLabel` from `webui/src/provider-surfaces.tsx:22-26`, which reads `snapshot.selectedModelId`, a browser-only preference. Verified 2026-09-17 on main `0ef0a1a4` with `qwen3-0.6b-4bit` Ready on the server: a fresh tab shows "No model selected" in the toolbar and "Select a model" in Chat, so nothing in the shell tells the operator that a model is loaded and holding memory. Epic #1834 required the top bar to identify the model and its actual state. The sidebar footer (`shell.tsx:104-107`) prints `router_pool · ready · v0.7.0 · seq 5` through `connection.footer.connected` (`webui/src/i18n/catalog.ts:197`), where `seq` is the internal SSE sequence from `snapshotSequence` (`provider-surfaces.tsx:176-178`).

Every route hand-rolls a `.screen-heading` block with an `.eyebrow` line ("Local library", "Local inference", "Issue #1843"): `webui/src/app.tsx:137` (Settings), `webui/src/features/models/screen.tsx:244-250`, `webui/src/features/chat/chat.tsx:131` (hard-coded English), `webui/src/features/activity/index.tsx:24`, `webui/src/gallery.tsx:14-18`, and `ProductConnectionSurface` (`provider-surfaces.tsx:83,109,123`). Page actions sit in a `.button-row` below the intro paragraph (`screen.tsx:292`) or a `.chat-toolbar` / `.activity-toolbar` div.

`shell.tsx:34-43` implements only Cmd/Ctrl+K and `?`. `docs/webui/ux-contract.md:19` also lists Cmd/Ctrl+N (new chat) and Cmd/Ctrl+Enter (send); `[`/`]` exist on the sidebar (`shell.tsx:54-59`). The help dialog (`app.tsx:100-103`) lists three items. `CommandPalette` (`app.tsx:146-155`) filters four route names and adds `gallery` whenever the hash contains "gallery".

## Proposed Solution
[...]

PR #1924, opened 2026-09-19
refactor(webui): adopt every shared @lablup/ui-common component

### LINKED ISSUE #1902: refactor(webui): adopt every shared @lablup/ui-common component

inureyes:
The bundled WebUI now uses all 17 `@lablup/ui-common@0.1.0-alpha.19` components instead of 7. The ten new ones sit behind the `webui/src/design-system/common-*.tsx` adapter seam and are re-exported from `primitives.tsx`, so feature screens still import only the design system. The local `Tooltip`, `Sheet` and `ErrorBanner` implementations are gone from `primitives.tsx`.

Count, measured the issue's way (`grep -rn '@lablup/ui-common' webui/src/` against the 17 names in `dist/index.d.ts`): 17 of 17. Package imports appear only in `common-adapters.tsx` (Button, StatusTag, ProgressBar, EmptyState, Tabs, DataTable), `common-select.tsx` (Select), `common-overlays.tsx` (Drawer, Tooltip), `common-feedback.tsx` (ErrorState, Skeleton), `common-layout.tsx` (BaseCard, PageHeader, PageLayout, SmoothHeight) and `common-data.tsx` (Badge, StatCard), plus the existing `styles/base.css` import in `main.tsx`.

## What each adapter keeps

- **Drawer** (compact nav below 960px, replaces the native-dialog `Sheet`): stable `onClose` (the package reruns its focus effect on every identity change, so each SSE-driven shell render moved focus); completes the first-open focus hand-off the package drops while the panel is still `visibility:hidden`; `<main>` is `inert` while it is open and an Escape that starts outside the panel still closes it (skipping IME composition and already-handled events), because the package only traps focus from inside its panel; the shortcut guard also matches `[role="dialog"][aria-modal="true"]`; left sheet at the approved `min(320px, 100vw - 32px)` via scoped CSS (two commented `!important`s outrank the package's `<768px` full width); localized title and close label; `data-testid="mobile-nav-sheet"` kept by a ref bridge; widening past 960px while open closes it and focuses the desktop sidebar (the old sheet stayed open with a `display:none` restore target).
- **ErrorState**, through the kept `ErrorBanner` name and props: the 23 call sites are unchanged, and #1913/#1914 rely on the name. A ref bridge removes the package's inner `role="alert" aria-live="polite"` so the banner stays the only live region (`status` for info, `alert` otherwise) and sets `role="none"` on the package h2 so no heading lands in sections that own one. `action` stays a ReactNode because `primaryAction` cannot express the chat anchor or icon Buttons. Tones map error/warning/info to danger/warning/accent.
- **Tooltip** (gallery): props are `{content, children}`. The package describes its wrapper, and only while open, so the control is described by an always-present hidden node; the wrapper adds no tab stop; inside `NativeModalContext` it renders in place instead of portalling out of the top layer. Escape now dismisses it (WCAG 1.4.13).
- **BaseCard** (`Card`, gallery cards; also inside StatCard): keeps `.surface-card`, passes `role` (article, or region for the named focusable table card), turns `hoverable` off, and restores the focus outline that BaseCard's `:focus-within` rule removes.
- **PageLayout** (shell content column, `wide`): drops the package's extra 0.5rem inline padding below 768px.
- **PageHeader** (Settings, gallery, the three sign-in/connection states): the eyebrow line is dropped there; its keys stay for #1914, which owns Models/Chat/Activity headings. A ref bridge restores the title and description test ids, and `focusFallback` puts `tabIndex=-1` plus `data-dialog-focus-fallback` on the Settings h1, so the dialog focus fallback stays a named heading.
- **StatCard** (Activity runtime summary and details), **Skeleton** (`LoadingStatus`: one visible, localized `role="status"` over decorative shapes in the Models table loading slot, since SkeletonRow/SkeletonText mount a status per row), **SmoothHeight** (Activity operations list, active only while an operation is in flight plus a 400ms settle window restarted on every transition, so a settled list is never height-pinned or focus-clipped), **Badge** (Models row: selected, source, quantization; rows at 1440 go from 100.5px to 79.5px).

## Recorded decisions and follow-up

- `usePrefersReducedMotion`: not adopted. It reads only the OS media query and cannot see the in-app `data-reduce-motion` toggle; CSS stays the only mechanism and its global rules also stop the shared components' motion (browser test: with the OS preference unset, the drawer's transitions are 1ms). StatCard `animate` stays off for the same reason.
- `formatCompactNumber`: not adopted. It falls back to a locale-less `toLocaleString()`; `format.ts` stays the single locale-aware formatter.
- Both are recorded at the seam (`common-data.tsx`) and in `docs/webui/ui-common.md`, which also records the follow-up to contribute `Dialog` (native top layer, focus trap, Escape with focus restoration) to `lablup/ui-common`; the same note sits above `ModalDialog`. Nothing was opened upstream.

## Tests first: failing before each swap, and naive reverts

Each test commit precedes its swap commit, and the shared-DOM assertion is last, so the red runs fail there unless noted. All Playwright runs used a private preview port (4902) so no sibling worktree's server was reused.

```
Tooltip    RED  vitest: expected null not to be null  (button.closest('.tooltip__wrapper'))
                pw: > 27 | expect(page.locator('.tooltip__wrapper').filter({ has: trigger })).toHaveCount(1)  0 elements
                pw Escape case: > 36 | expect.poll(() => shownTooltips(page)).toEqual([])  (old CSS tooltip is not dismissible)
           NAIVE no describedby clone: expected '' to be 'Tooltips are descriptive only'; toHaveAccessibleDescription failed
                 no modal fallback: expected <div id="tooltip-…"> to be null
                 default tabIndex 0: expected 0 to be less than 0; after Tab toBeFocused failed
ErrorState RED  vitest (4 cases): expected null not to be null  (.error-state--<tone>)
                pw: > 64 | expect(banner.locator('.error-state')).toHaveCount(1)  0 elements
           NAIVE inner live region kept: expected [<section>, …(1)] to deeply equal [<section>]
                 h2 kept as heading: expected ['Runtime', 'Observations are stale'] to deeply equal ['Runtime']
                 no compact CSS: > 62 | expect(layout.top).toBeLessThan(32)  Received: 100
BaseCard   RED  pw: > 86 | expect(page.locator('.surface-card.base-card')).toHaveCount(1)  0 elements
           NAIVE no role: > 73 | expect(articles).toHaveCount(2); hoverable: > 77 | toHaveCSS('transform', 'none'); no outline restore: > 85 | outlineStyle not 'none'
Drawer     RED  vitest (2): expected false to be true  (panel matches aside.drawer[role=dialog][aria-modal=true])
                pw (4): > 26 | expect(panel).toHaveJSProperty('tagName', 'ASIDE')  resolved to <dialog … class="ds-dialog ds-sheet">
                pw widening case: > 122 | expect(dialog).toBeHidden()  (old sheet stays open)
                pw pointer-press case: > 93 | focus left the drawer (Shift+Tab reached the page behind it)
           NAIVE per-render onClose: expected <button …> to be <a href="#chat" …>
                 guard only dialog[open]: expected true to be false  (command dialog opened)
                 no first-open hand-off: > 45 | expect(close).toBeFocused()
                 no left-sheet CSS: > 52 | expect(box.width).toBeLessThanOrEqual(320 + 1)
                 no closeLabel: > 45 | close button named 닫기 not found
                 no widening handler: > 122 | expect(dialog).toBeHidden()
                 no inert <main>: > 93 | focus left the drawer
                 no outside Escape: >  98 | expect(dialog).toBeHidden()  unexpected value "visible"
                 handled Escape still closes: vitest expected false to be true
PageLayout RED  pw (3): > 163 | expect(page.locator('.app-content.page-layout.page-layout--wide')).toHaveCount(1)  0 elements
           NAIVE package padding kept: > 157 | expect(Math.abs(metrics.left)).toBeLessThanOrEqual(1)  (390 and 390 at 200% text)
PageHeader RED  vitest (6): expected false to be true  (.page-header .page-header__title)
                pw (3): > 196 | expect(page.getByTestId('gallery-title')).toHaveClass(/page-header__title/)  resolved to <h1 data-testid="gallery-title">
           NAIVE no test-id bridge: Error: Missing settings-title
                 no focus fallback: expected <button …> to be <h1 class="page-header__title" …>
                 no title CSS: > 186 | fontWeight toBeGreaterThanOrEqual(600)
StatCard   RED  vitest: expected [null, null] to deeply equal ['Active requests: 2 requests', …]
                pw (390, 1440): > 132 | locator('.activity-metric.stat-card') toHaveCount(4)  0 elements
           NAIVE nowrap/ellipsis: > 125 | clipped toEqual([]); uppercase: > 126; hover lift: > 129 | toHaveCSS('transform', 'none')
Skeleton   RED  vitest (en, ko): expected 0 to be greater than 0  (.skeleton inside the status)
                pw (390, 1440): > 158 | expect(status.locator('.skeleton').first()).toBeVisible()  not found
           NAIVE non-decorative shapes: expected [ …(4) ] to have a length of 1 but got 4
SmoothHeight RED vitest: expected null not to be null
                pw (390, 1440): > 193 | locator('.smooth-height.smooth-height--active') toHaveCount(1)  0 elements
           NAIVE always active: expected true to be false (settled list pinned); no settle window: expected false to be true
                 window not restarted on false-true-false: expected false to be true
Badge      RED  vitest: expected [] to deeply equal ['Selected', 'cache', '4bit']
                pw (390, 1440): > 226 | cell.locator('.badge:not(.status-tag)') toHaveText(['models_dir', '4bit'])  0 elements
           NAIVE no whitespace: expected 'alphaSelectedcache4bit' not to contain 'cache4bit'
```

## Also fixed here

`.ds-button:hover` applied to disabled buttons and swapped a disabled primary fill for the 8% selection tint behind white text. Dropping the eyebrow moved the sign-in Connect button up, so after sign-in the pointer rests on the still-disabled Models "Add" button and the existing 1440 signed-in axe check failed about one run in ten; the rule is broken on main too. The hover rules now use `:hover:where(:not(:disabled))`, which excludes disabled buttons without raising specificity, so the package's primary and danger hover fills still win on enabled buttons (a first version at higher specificity made enabled hover text 1.16:1; the test for that case failed before the correction).

Known and left: at 390px the loading or empty Models table scrolls horizontally with nothing focusable inside (`scrollable-region-focusable`, also on main); the Skeleton test tolerates exactly that node and the table belongs to #1918. The served-CSP spec exercises the gallery only, so SmoothHeight and Skeleton reach the real CSP only through their CSSOM height/width writes, which strict `style-src 'self'` permits.

## Validation

- Branch is current with main at fd33ff15 (#1919, #1921, #1922) through a merge commit rather than a rebase, so the test-first commit hashes cited above stay valid. Conflicts with #1922 were resolved so its catalog keys survive inside the StatCard and SmoothHeight swaps and its `ConfirmDialog` stays next to the removed `Sheet`/`Tooltip`; the bundle was rebuilt, not hand-merged.
- `pnpm --dir webui run typecheck` and `lint`: pass. `unit`: 18 node:test and 359 Vitest tests (47 files) pass, including #1922's `src/i18n/drift.test.ts`.
- Playwright chromium at the merged head: ui-common-shell 12, ui-common-components 14, browser 19, activity 3, chat 2, models 3, settings 3, all pass.
- `python3 scripts/webui/build_bundle.py` then `--verify`: clean, regenerated `src/webui/assets/` committed; initial JS 152.7 kB gzip against the 200 KiB budget. `check_webui_contract.py` (49 fixtures) and `check_binary_assets.py`: pass.
- CI at be87e27b: all checks pass on run attempt 2. Attempt 1 failed only `src/features/chat/confirmations.test.tsx` "refocuses the Clear All trigger once clearing settles if focus is still on body" (from #1922): its `setTimeout(0)` focus claim can run before React commits the re-enabled button. It passed 15 of 15 isolated and 4 of 4 full-suite runs locally, and this PR does not touch that flow; recorded as a flaky #1922 test, not relaxed here.
- Review: implementation review fixed the hover specificity and the drawer's non-inert background; the security and performance review found nothing at CRITICAL, HIGH or MEDIUM; three LOW hardening items (settle window restart, handled-Escape guard, adapter-scoped layout whitelist) were applied test-first.

## Live acceptance against the embedded bundle

Run against `mlxcel-server --webui --models-dir models/mlx --no-models-autoload --settings --props --metrics --slots` built from be87e27b (test-fast profile, embedded bundle), signed in with a session key, no model loaded, nothing loaded, unloaded or downloaded. Headless Chromium, driven by a local uncommitted spec.

- [x] PASS, 390 px nav: the menu button opens the drawer with focus on its close button and `<main>` inert; Tab cycles home, Models, Chat, Activity, Settings, Close and wraps back to home without leaving the drawer; Shift+Tab from Close reaches Settings then Activity; Escape closes it and focus returns to the menu button; choosing Activity closes it with focus on the menu button; after a pointer press on the drawer heading, Shift+Tab goes to the document and then to the drawer's last link, never to the page behind. No horizontal overflow with the drawer closed or open.
- [x] PASS, 200 percent zoom, emulated (headless Chromium cannot drive native browser zoom): a 780x844 window at 200 percent, emulated as a 390x422 CSS px viewport at deviceScaleFactor 2. Models, Chat, Activity, Settings and `#gallery` have 0 px horizontal overflow with the nav closed and open; the drawer is 320 px and every nav link stays inside the viewport. Not a native-zoom observation.
- [ ] Not met, pre-existing and routed to #1914: a 390x844 window at 200 percent, emulated as 195x422 CSS px at deviceScaleFactor 2, overflows by exactly 125 px on every route with the nav closed or open. The cause is `body { min-width: 320px }` on main (320 - 195), not this PR; the drawer itself fits (163 px, links inside the viewport).
- [x] PASS, real-server `webui/tests/csp.spec.ts` via `MLXCEL_WEBUI_CSP_URL=http://127.0.0.1:<port>/webui/` and `playwright.csp.config.ts`: both cases (light 1440, dark 390) pass with zero CSP violations and no external requests, including the new steps that open the drawer and show the tooltip.
- [x] PASS, Settings and banners: on Settings, with focus on the document, Cmd/Ctrl+K then Escape lands focus on the Settings title (`h1.page-header__title`, `settings-title`). Both Settings info banners are `role="status"` and the gallery warning banner is `role="alert"`, with no nested live region and the package title set to `role="none"` (no extra heading).
- [x] Extra: axe (WCAG 2.2 AA tags) on signed-in Models, Activity and Settings at 390 px with the nav open reports no violations.

Part of #1910

Closes #1902


inureyes:

### Intent
> Adopt the remaining ten `@lablup/ui-common` components behind the adapter seam (17 of 17) without silent accessibility regressions.

### Findings Addressed
- [x] Enabled primary and danger buttons lost their hover fill (HIGH). e6e38ee9 raised `.ds-button:hover` to `:hover:not(:disabled)`, the same (0,3,0) specificity as the package's `.button--primary:hover:not(:disabled)`, and product CSS loads later, so hovering showed white text on the 8% selection tint (axe: 1.16:1 in light). `:where(:not(:disabled))` restores the original specificity and still excludes disabled buttons. Test 827d3a60 (red on the old CSS), fix eb57bd8e.
- [x] The open nav drawer was not modal for mixed pointer and keyboard use (HIGH). The package traps Tab and handles Escape only inside its panel and leaves the page interactive. After a pointer press on the drawer heading, Shift+Tab reached the gallery's Open dialog button behind the drawer, and Escape did nothing; the native sheet prevented both. `<main>` is now `inert` while the drawer is open, and the adapter closes on an Escape that starts outside the panel. Test 682c424e (red), fix 1bd39dba; each mitigation, reverted alone, fails its own assertion. Docs 68a3671c, bundle 81415621.

### Remaining Items
- [ ] `.activity-metric dt/dd` rules in `activity.css` no longer match anything after the StatCard swap (LOW). Left alone to keep #1913's rebase simple.
- [ ] `ModalProps.position` and `className` are unused now that Sheet is gone (LOW). The brief limits `primitives.tsx` edits.
- [ ] Moving the pointer onto the tooltip content still dismisses it (WCAG 1.4.13 hoverable, LOW). This was already true on main.
- [ ] The PR body's Playwright counts (ui-common-components 13, ui-common-shell 11) are now 14 and 12. Its Drawer and "Also fixed here" notes do not yet mention these two fixes.

### Verification
- [x] Tests pass (typecheck, lint, 336 Vitest; Playwright on the private port: shell 12, components 14, browser 19, models 3, settings 2, chat 2, activity 2; bundle `--verify`, contract and binary-asset checks)

inureyes:
## Security and Performance Review (81415621)

No CRITICAL, HIGH or MEDIUM findings, so nothing was changed.

### Checked
- CSP: every new inline geometry goes through CSSOM (React `style` props, and SmoothHeight's `element.style.height`). The committed bundle and the ui-common dist contain no `setAttribute("style")`, `<style>` creation, `insertRule`, `cssText` or `adoptedStyleSheets`. `dangerouslySetInnerHTML`/`innerHTML` counts match main (React DOM internals only). The four new `expectSafeLayout` entries each pin an exact property set and value pattern.
- DOM injection: the ref bridges only write constant roles, test ids and the focus-fallback marker. Server strings (error bodies, model source and quantization, metric names) reach the DOM as React text or React-set attributes (StatCard `aria-label`).
- Top layer and focus: the Tooltip keeps in-place markup under `NativeModalContext`. The command and help dialogs render outside the inert `<main>`, and no native dialog can be open at the same time as the drawer, so the outside-panel Escape handler cannot misfire. It skips IME composition, and its listener and rAF loop (at most 11 frames) are released when the drawer closes.
- Bundle: `--verify` is clean. Initial JS is 152.73 kB gzip against the 200 KiB budget (+4.6 kB versus main). No new URL, font, `@import` or `url()`.

### Remaining (LOW, not fixed)
- [ ] `.smooth-height` and `.skeleton` whitelist selectors are bare library classes. Scoping them to `.ds-smooth-height` and `.ds-loading__shapes` would keep an unadapted future use from passing.
- [ ] `csp.spec.ts` covers only the gallery, so the served CSP never sees SmoothHeight (Activity) or Skeleton (Models loading). Both write through CSSOM, so this is a coverage gap, not a violation. The real-server run is still pending.
- [ ] SmoothHeight settle window: if `animate` goes false, true, false within 400 ms, the first timer ends the window early. The effect is cosmetic.

Gates rerun: typecheck, lint, unit (336 Vitest), and on port 4902 `ui-common-shell` 12 and `ui-common-components` 14 all pass.

PR #1926, opened 2026-09-19
fix(webui): open the model inspector from anywhere in a library row

### LINKED ISSUE #1901: fix(webui): make the whole model library row open the inspector
Part of #1910.

## Problem / Background

In the bundled WebUI model library, the only way to open the inspector panel for a model is clicking the button that renders the model name in the first table cell; its accessible name is "Inspect {model name}" (`webui/src/features/models/screen.tsx:190-198`, string key `models.library.inspect` at `webui/src/features/models/strings.ts:106-109`). Clicking anywhere else in the table row does nothing.

Load, Use in Chat, Unload and Delete all live in that inspector panel (`webui/src/features/models/inspector.tsx:78-99`), so a user who does not discover the name button concludes that loading works but unloading is missing. The maintainer hit exactly this. An automated probe written for this session hit it too, by clicking the `<tr>` rather than the name button. The existing end-to-end coverage never exposes the gap because every spec drives the name button by role and accessible name (`webui/tests/models.spec.ts:178`, `webui/tests/models.spec.ts:241`, `webui/tests/router-real.harness.ts:321`).

## Current Behavior

The `DataTable` call site at `webui/src/features/models/screen.tsx:413-427` passes `columns`, `rows`, `getRowKey`, `ariaLabel`, `testId`, `rowClassName`, `loading`, `loadingState` and `emptyState`, and no row-activation prop, so the row is inert and only the nested name button calls `actions.selectModel(entry.identity.id)`.

The underlying functionality is correct and was verified end to end against the running bundled server. After loading `qwen3-0.6b-4bit`, the inspector shows Load visible and disabled (already loaded), Use in Chat enabled, Unload visible and enabled, and Delete absent because the entry's source is `models_dir` rather than the managed cache, which matches the cache-only deletion contract (`canDelete` requires `entry.identity.source === 'cache'`, `webui/src/features/models/policy.ts:40-51`). Driving unload through `POST /ui-api/v1/model-actions` moved the lifecycle from `unloading` to `unloaded` with `worker_exit_observed: true`, which satisfies the epic's requirement that an unload awaits worker shutdown before reporting unloaded (`src/server/router_lifecycle.rs:102`).

So this is a discoverability and affordance problem, not a missing feature.

## Proposed Solution

Make the lifecycle actions reachable without first having to discover the name button: either the row itself becomes an affordance for opening the inspector, or the actions become reachable some other way from the library. This issue deliberately does not prescribe one design. Whatever is chosen keeps keyboard operability and the existing accessible naming intact, so the specs that target `Inspect {name}` by role keep passing unchanged.

The shared table already supports the row-affordance option without a fork: `@lablup/ui-common`'s `DataTable` accepts `onRowClick(row)`, documented as rendering "a clickable affordance (cursor: pointer + keyboard binding)", and `isRowClickable(row, index)` to narrow it to rows that can act on it. Reuse those props rather than hand-rolling an `onClick` on `<tr>`, which would produce a click target with no keyboard binding and no announced role. A row-level handler must not swallow or double-fire the nested name button's activation.

## Scope

**In scope:** `webui/src/features/models/screen.tsx` (the `DataTable` call site and the name column), `webui/src/features/models/models.css` if a hover or focus affordance is needed, and the specs `webui/tests/models.spec.ts` and `webui/src/features/models/screen.test.tsx`.

**Out of scope:** the inspector's contents and action policy (`inspector.tsx`, `policy.ts`), the `/ui-api/v1` contract, and PR #1898. It was agreed to handle this after the first pass of epic #1834 rather than inside PR #1898.

## Implementation Notes

- **Reuse**: `onRowClick`, `isRowClickable` and the already-used `rowClassName` on the shared `DataTable`; `actions.selectModel(entry.identity.id)` is the single existing selection entry point.
- **Constraints**: the UX contract states that selecting a model never loads it (`models.library.subtitle`: "Inspect local checkpoints. Selecting never loads a model."), so a row affordance opens the inspector and nothing else. Selection also drives the `models-selected` row class, which must keep working.
- **Edge cases**: rows in a busy or failed lifecycle state still open the inspector; dragging to select text inside a cell must not be treated as a row activation; the name button stays a real button with its existing `aria-label`, so screen reader users are not forced onto the row.
- **Error handling**: unchanged. No new network call or state transition is introduced.

## Acceptance Criteria

- [x] A user who never clicks the model name can reach Load, Use in Chat and Unload for a catalog entry.
- [x] Keyboard users reach that same target with Tab and activate it with Enter or Space, with visible focus.
- [x] The accessible name "Inspect {model name}" still resolves, and `webui/tests/models.spec.ts:178` plus `webui/tests/router-real.harness.ts:321` pass unchanged.
- [x] A new spec in `webui/tests/models.spec.ts` activates the row itself rather than the name button and asserts the inspector opens for that entry; reverting the source change makes it fail.
- [x] The 25-row pagination and filter assertions at `webui/src/features/models/screen.test.tsx:199-201` still pass.
- [x] Verified against the running bundled server rather than only mocks: the new path opens the inspector, and Unload still reports `worker_exit_observed: true`.

## Verification

```bash
pnpm --dir webui run lint
pnpm --dir webui run unit
pnpm --dir webui run browser
make verify-webui-frontend
```

A pass means all five commands exit 0 with the new row-activation spec among the executed browser tests. Manual pass: with a real checkpoint loaded in the installed artifact, open the model library, activate a row by mouse and again by keyboard, confirm the inspector opens for the correct entry, and confirm Unload still drives the lifecycle to `unloaded`.

## Technical Considerations

Part of epic #1834; found while validating the integration gate #1848 (PR #1898). Agreed with the maintainer to handle after the epic's first pass rather than inside PR #1898.

inureyes:
In the Models library a click anywhere in a row, not only on the name button, now opens the inspector for that entry, so Load, Use in Chat, Unload and Delete are reachable without discovering the name button. Selecting still only calls `selectModel` and never loads; clicking the row that is already selected does nothing, because `selectModel` always aborts the live event stream and forces a full snapshot even for the same id.

The behavior is an opt-in seam in the shared `DataTable` adapter, `activateRowPrimary`, for #1918 to build on. A body-row click focuses and clicks the row's primary control (the element with `ROW_PRIMARY_CLASS`, here the "Inspect {name}" button), so pointer and keyboard run the same handler. Clicks on another interactive element in the row, clicks that end a text-selection drag, and rows without an enabled primary control (including the loading and empty rows) are ignored. Rows keep their native `row` role and gain no tab stop; while the name button has `:focus-visible` the whole row is outlined.

## Why not ui-common `onRowClick`

The issue suggested alpha.19's `onRowClick` / `isRowClickable`. Measured on the Models page, they render every body `<tr>` as `role="button"` with `tabIndex=0`, which fails axe `aria-required-children` (critical: the rowgroup holds buttons, not rows) and `nested-interactive` (serious: the name button sits inside the row button), doubles the tab stops per row, and names the row `Inspect alpha cache 4bit chat, ...`, so the `router-real.harness.ts` locator `new RegExp('Inspect .*' + name)` matches two elements and its `.click()` throws. The adapter's props type omits both props so they cannot come back by accident; `docs/webui/ui-common.md` records this.

## Changes during review

- Re-selecting the selected row is a no-op (security review: each stray click cost a stream reconnect, a full snapshot and the Activity history).
- After merging #1923 the selected row paints its selection fill, and the accent "Selected" badge on it failed WCAG AA contrast: 4.49:1 in mlxcel-light and 3.73 to 4.49:1 in every theme under high contrast, because ui-common paints `--token-colorPrimary` text on a fixed translucent blue fill. The Badge adapter now draws the accent tone as body text inside a `--color-focus` outline. A new test checks a selected and a hovered row for contrast in all four themes with high contrast off and on; without the CSS change it fails in five of the eight variants.

## Test-first evidence

The row-activation tests against the unmodified source (origin/main 31c04592):

```
  ✓  1 [chromium] › tests/models.spec.ts:164:1 › library journey observes operations before load/chat/unload/cache removal (924ms)
  ✓  2 [chromium] › tests/models.spec.ts:218:3 › Models 1440 light: bounded inventory, keyboard inspection, no search POST, a11y (713ms)
  ✓  3 [chromium] › tests/models.spec.ts:218:3 › Models 390 dark: bounded inventory, keyboard inspection, no search POST, a11y (688ms)
  ✘  4 [chromium] › tests/models.spec.ts:270:3 › Models row activation › clicking a non-name cell opens that entry and reaches Load, Use in Chat and Unload (5.2s)
  ✘  5 [chromium] › tests/models.spec.ts:294:3 › Models row activation › keyboard Tab reaches the next row with a visible row outline, and Space or Enter opens it (315ms)
  ✘  6 [chromium] › tests/models.spec.ts:323:3 › Models row activation › dragging to select text inside a cell does not activate the row (5.5s)
  ✘  7 [chromium] › tests/models.spec.ts:351:3 › Models row activation › rows keep table semantics, each entry keeps one Inspect button, and the page stays axe clean (5.3s)

    Error: expect(locator).toBeVisible() failed
    Locator: getByRole('complementary', { name: 'Model details' })
    Error: element(s) not found

  4 failed
  3 passed (23.0s)
```

Reverting only `screen.tsx` on the final branch fails all five row-activation tests (the four above plus the contrast test) while the three name-button tests still pass. Every adapter guard was removed in turn and each removal fails a unit test; removing only the selection guard fails the drag test, which ends with a control click on the same point so it cannot pass by missing the row.

## Validation

- `typecheck`, `lint`, `unit` (51 Vitest files, 484 tests, plus 35 node tests including the i18n drift and theme-selector checks), `check:theme-selectors`.
- Playwright Chromium, one spec per run on an isolated port: models 8/8, theme 23/23, browser 28/28, ui-common-components 14/14, ui-common-shell 12/12, settings 3/3, chat 2/2, activity 3/3.
- Bundle regenerated from the merged source; `build_bundle.py --verify`, `check_webui_contract.py`, `check_binary_assets.py` and `check_cross_repo_refs.py` pass. `browser:all` (multi-engine) is left to CI.

## Real-binary acceptance

Run at head e7f3db03 against the bundled server (`--webui --models-dir models/mlx --no-models-autoload --settings --props --metrics --slots`, 224 catalog entries), Chromium at 1440x900, with the library filtered to `qwen3-0.6b-4bit` (two rows: that `models_dir` entry and the `mlx-community/Qwen3-0.6B-4bit` cache entry). All pass:

- [x] Clicking the Lifecycle cell of `qwen3-0.6b-4bit` opens the inspector for that entry; the row gets `models-selected`, its selection fill and the "Selected" badge; no POST is sent. Clicking the other row switches the inspector to it.
- [x] Clicking the already-selected row five more times sends no bootstrap, catalog or events request (0 / 0 / 0). As a control, switching to a different row sends 1 bootstrap, 5 catalog and 1 events reconnect.
- [x] Keyboard: from the search field, 5 Tabs reach the first row's name button (`:focus-visible`); the row shows a 2px solid focus-colored outline and has no `role` or `tabindex`. Space opens that row; Tab moves to the next row (outlined) and Enter opens it.
- [x] From the inspector opened by a row click: Load (lifecycle `loading` then `ready`), Use in Chat (enabled, navigates to `#chat`), then back on Models, Unload through the confirmation dialog (`unloading` then `unloaded`). `GET /ui-api/v1/catalog?q=qwen3-0.6b-4bit` then reports `state: unloaded`, `worker_exit_observed: true`. Exactly two POSTs were sent, both to `/ui-api/v1/model-actions`. The model is left unloaded.
- [x] Drag-selecting text in a Task cell of an unselected row selects "chat, rerank," and leaves the inspector unchanged; a plain click at the same point then opens that row.
- [x] At 390x844 a row click opens the inspector with no horizontal page scroll.

## Real-router harness

Run by the epic #1910 orchestrator on this head (`e7f3db03`): `cargo test --profile test-fast --features metal,accelerate --lib server::router_server::router_webui_playwright_harness_tests::real_router_browser_harness -- --ignored --exact --nocapture` (the `make verify-webui-integration-fake` target) passed. Its Playwright log shows the spec ran rather than skipped: `tests/router-real.harness.ts` "secured Rust router drives download, load, chat, Stop/drain, unload, remove and negative security cases" passed in 5.9 s, including the unchanged `Inspect {name}` locator at `router-real.harness.ts:321`.

Closes #1901

Part of #1910


inureyes:

### Intent
> Open the Models inspector from anywhere in a library row through an opt-in `activateRowPrimary` seam on the shared `DataTable` adapter, keeping native row semantics and the name button as the keyboard target.

### Findings Addressed
- [x] No CRITICAL or HIGH findings; no code changes were needed.

### Remaining Items
- [ ] Two adapter guards survive mutation: dropping `event.currentTarget.contains(row)`, or widening `if (nested && row.contains(nested))` to `if (nested)`, keeps the adapter and screen unit tests green (38/38). The second matters for #1918: a table inside a `tabindex="0"` scroll region would silently lose row activation. The PR body claim that each guard removal is caught by a unit test needs softening or a focusable-container test. (MEDIUM), left: auto-fix covered CRITICAL and HIGH only
- [ ] Pre-existing since #1891: `.models-library .models-selected` uses `var(--surface-selected, var(--surface-hover))` and neither variable is defined, so the selected row computes to `rgba(0, 0, 0, 0)` with no hover fill; only the "Selected" badge marks it. (LOW), out of scope here; candidate for #1918
- [ ] The first click of a word-select double-click in a cell activates the row; only drag selection is guarded. Harmless because activation only selects. (LOW)

### Verification
- [x] All stated requirements implemented (real-binary acceptance still pending with the orchestrator)
- [x] Tests pass: typecheck, lint, unit (370 Vitest plus node tests), `models.spec.ts` 7/7 on an isolated port, bundle `--verify`, webui contract and binary-asset checks. A WebKit and Firefox probe also confirmed pointer activation, the drag-selection guard and a painted row outline on keyboard focus.

inureyes:

No CRITICAL or HIGH findings; no code changes were made.

### Checked
- Row activation only reaches the clicked row's own `ds-row-primary` element, which on Models is the "Inspect {name}" button calling `actions.selectModel(id)`. Load, Unload, Delete and the confirmation dialogs render outside the delegated wrapper. A probe on a 25-row page sent zero POSTs across 85 row clicks.
- CSP: no inline script, inline style or `style` prop is added, and `index.html` changes only the two hashed asset names. `build_bundle.py --verify` passes, so the committed bundle is exactly what the source builds.
- Handler cost on 25 rows: under 0.1 ms per click, the same as clicking the name button directly.

### Remaining items
- [ ] MEDIUM: clicking the already-selected row is not free. `selectModel` always runs `selectionChanged()`, which aborts the live `/events` stream and forces a full snapshot, and the reducer clears `runtimeHistory` even when the id is unchanged. With the stream held open, 2.5 s idle sent no bootstrap, catalog or events requests; 5 clicks on the selected row sent 5 of each. The name button already behaved this way, but the whole row is now the target, so stray clicks reset Activity history and reconnect the stream. Suggested fix: skip `selectModel` when the id equals the current selection. Left because auto-fix covered CRITICAL and HIGH only.
- [ ] LOW: the `ROW_INTERACTIVE` comment says it mirrors the package guard, but it uses `a[href]` where alpha.19 uses `a`. Harmless on Models, which has no anchors.

PR #1930, opened 2026-09-19
feat(webui): redesign the model library as a dense list with row actions

### LINKED ISSUE #1918: feat(webui): redesign the model library as a dense, task-first list with row actions

inureyes:
Closes #1918. Part of #1910.

The Models library becomes a dense list where each row is the unit of work: Load, Use in Chat and Unload are reachable from the row, and internal state is no longer copy.

## What changes

- **Rows** are one 33 px line with a fine pointer (28 px row controls; touch keeps 32 px): name (truncated, full text in `title`), size, quantization or weight dtype, task badges in declared order, state, and actions. A row shows Use in Chat when chat is available, Unload on any Ready row, and Load otherwise (disabled while it cannot load). Every row has an "Inspect {name}" icon button (the #1901 row primary) and cache rows a danger Delete. Row and inspector share one `openAction(kind, entry)` path, so the stale, capacity and eviction confirmations are unchanged; a row loads with that entry's own profile (`loadProfileFor`). Focus follows a row action to the control that replaces it (Load after Unload).
- **Sorting** moved to the Name, Size, Quantization and State headers in controlled mode, pinned to ready, then transitions, then the rest. The Sort select and its strings are gone. Source and Task filters show labels.
- **Inspector**: a pane at 1100 px and wider, a right drawer (`role="dialog"`, "Model details", the shared Drawer with `width="medium" side="end"`, the same variant Chat's settings drawer uses) below it. Overview and capability badges up top; ids, reasons, profile, removal notes, disabled server actions, roots and the docs link sit in a Details disclosure that renders nothing until opened. In the drawer, Tab stays inside the panel (the Details summary included) and Escape in a confirmation closes only the confirmation. The row Load goes through `load-action.ts` (from #1931), shared with Chat.
- **Removed**: Library operations, the provider-summary paragraph, the disabled-actions list and the permanent roots disclosure. Kept: download pseudo-rows (`op:` keys), Load with eviction after a capacity conflict, and the `models-pending` status line. Roots are the empty state's action.
- **Server**: nullable `dtype` in `CatalogMetadata` from `config.json` `torch_dtype`/`dtype` (top level, then `text_config`), normalized to `bf16`/`fp16`/`fp32`/`fp64`; only recognized tensor dtypes are reported, anything else is null. Required and nullable in `api.yaml`; DTO, fixtures and `catalog.md`/`.ko.md` updated. No schema-version bump.

## Drawer reconciliation with #1931

Both PRs added an end-anchored Drawer. The adapter keeps one API, main's `width`/`side` modifier classes over the constant inline sheet width, and one Tab wrap computed at key-press time: it wraps at the panel's first and last usable stops, counts a Details summary, skips a closed details' contents, and stops every Tab inside the panel in the capture phase so the package's stale handler cannot wrap early. It keeps main's IME Escape guard and main's rule that an Escape inside an open native dialog belongs to that dialog, plus this PR's fixes: a drawer confirmation's Escape leaves the drawer open, and a Tab pressed with focus on no element enters the panel. Chat's and Models' drawer tests all pass.

## Deviations from the issue

- Unload appears on any Ready row, not only a chat-capable one: an embedding or rerank model would otherwise have no row path to unload.
- The #1914 shell did not move `connection.authenticated.detail` into the sidebar footer. This PR puts it inside the footer's Connection details disclosure (desktop copy only, so the test id names one element).
- ui-common alpha.19 DataTable sorts rows itself even in controlled mode, so columns use a `sortComparator` carrying the lifecycle pin instead of a bare `sortValueAccessor`.
- `SkeletonRow` mounts one `role="status"` per row, breaking #1902's single-status contract; the loading slot keeps `LoadingStatus` with 8 decorative rows.
- Below 1100 px the drawer opens only on Inspect, row activation, or a chip/palette request, not from a remembered selection.
- Columns give way as the list narrows (tasks below 60rem, size and quantization below 44rem, state below 32rem); the name cell then carries state and, for downloads, progress.
- No screenshot baselines or SHA256 table exist since #1904, so that item does not apply.
- Also fixed: the 390 px `scrollable-region-focusable` violation (an overflowing table box becomes a named focusable region) and bidi isolation of names in dialog sentences.

## Validation

Rust test first, before the fix:

```
test server::webui::catalog::tests::dtype_reports_the_declared_weight_dtype_and_null_when_unknown ... FAILED
assertion `left == right` failed: bf16
  left: None
 right: Some(String("bf16"))
```

After: `... ok`; `server::webui` (79 tests), clippy and fmt pass.

After merging main at `18fda739` (#1928 Settings, #1931 Chat): typecheck, lint, unit (35 node + 664 Vitest), `check:theme-selectors`, Chromium Playwright (119, including `chat.spec.ts`), `build_bundle.py --verify`, `check_webui_contract.py` and `check_binary_assets.py` pass. The row spec drives load, chat and unload through `models-row-*` only; the drawer focus specs fail with their fixes reverted.

Page measurements, mock API, 120 entries:

| Viewport | Height before | Height after | Words before | Words after | Rows in view before | after |
|---|---|---|---|---|---|---|
| 1440x900 | 2514 px | 1176 px | 346 | 281 | 5 | 19 |
| 390x844 | 4190 px | 2083 px | 346 | 128 | 0 | 5 |

Real-binary acceptance (load, Use in Chat, unload from a row with `qwen3-0.6b-4bit`, `worker_exit_observed: true`, real-catalog heights and word counts): **pending**, run against a server built from this PR head.

Known and left as LOW: the page behind the drawer is not `inert` (Tab and Escape are handled), and row button labels interpolate names without bidi isolation, because the pinned `Inspect {name}` accessible names would change.


inureyes:

### Intent
> Make the Models library a dense list where Load, Use in Chat and Unload run from the row, move internal state behind a Details disclosure, and report the declared weight dtype in the catalog.

### Findings Addressed
- [x] A Ready model without chat (embedding, rerank, transcription) showed a disabled Load and no Unload on its row; Unload now follows the lifecycle, and focus after a row Load lands on Unload for such a model (HIGH, a176036b)
- [x] Download pseudo-rows lost their progress bar and state below 44rem / 32rem, where the Size and State columns hide and no inspector exists to recover them; the name cell now carries both as a fallback (HIGH, a176036b)
- [x] Escape in a confirmation opened from the inspector drawer also closed the drawer and dropped focus on the page heading (HIGH, a176036b)
- [x] The drawer's Tab trap (alpha.19, built once on open, no `summary`) made the Details disclosure unreachable by keyboard below 1100 px; the adapter now computes the cycle at key-press time (HIGH, a176036b)

### Remaining Items
- [ ] At exactly 1100 px the pane renders but `models.css` still collapses the layout to one column (`max-width: 1100px` vs `min-width: 1100px`) (LOW)
- [ ] `Badge` gained an unused `title` prop (LOW)
- [ ] The page behind the Models drawer is not `inert`; the new trap keeps Tab inside, but Tab from `<body>` after a pointer press on the drawer heading can reach the page (MEDIUM)
- [ ] Real-binary load, Use in Chat and unload from a row: pending, run separately against this head

### Verification
- [x] All stated requirements implemented (real-binary acceptance pending)
- [x] Tests pass: typecheck, eslint, unit (35 node + 556 Vitest), check:theme-selectors, all 114 Chromium Playwright cases, contract and binary-asset checks, `build_bundle.py --verify`, `server::webui` Rust tests, clippy and fmt

inureyes:

Reviewed at `29e8c3a9`. No CRITICAL or HIGH findings, so nothing was auto-fixed.

### Verified
- `normalize_dtype` accepts only `[a-z0-9_]{1,32}` and never panics. Its output matches the `api.yaml` pattern, and the client validates against that same schema.
- No `dangerouslySetInnerHTML`, eval, style attributes or inline scripts were added, and the bundle's `innerHTML`/`eval`/`setAttribute("style")` counts match main. The docs link is static and uses `rel="noreferrer"`.
- `loadProfileFor` uses `Object.hasOwn` on the validated profile document. A row load of entry X submits X's own profile, the same one the inspector shows.
- The Drawer listeners are removed when it closes. No hidden drawer can trap Tab: the Models drawer unmounts at 1100 px and wider, and the nav sheet closes when the window widens. The ResizeObserver disconnects correctly and only watches nodes that stay mounted.
- Only Inspect is a row primary. Unload, Delete and eviction still require their confirmations.

### Remaining (LOW)
- [ ] Row labels and the Inspect/Delete tooltips put `display_name` in without isolation (`screen.tsx:307-346`). Adding `isolate()` would change the pinned accessible names.
- [ ] `dtype` passes any bounded identifier through, so a downloaded config can show a misleading label when `quantization` is null (`catalog_metadata_config.rs:130`). An allowlist would close this.
- [ ] `inventory()` re-sorts the catalog on every snapshot change (`screen.tsx:100`), measured at 1.2 ms per call for 1,000 entries. A `useMemo` keyed on the catalog, filter and sort would avoid it.
- [ ] `useWideInspector` creates a MediaQueryList on every render (`inspect-request.ts:40`).
- [ ] An Inspect at 1100 px or wider leaves `drawerOpen` set, so narrowing the window later opens the drawer over the list.

### Gates
`server::webui` tests (79), clippy, fmt, typecheck, eslint, unit (35 node + 557 Vitest), theme selectors, `models.spec.ts` (14), `ui-common-components.spec.ts` (14), `build_bundle.py --verify`, contract and binary-asset checks all pass.

PR #1936, opened 2026-09-20
fix : handle missing sequence entries

saitejagirada:
During the per-sequence decode sampling step
```rust
  let seq = self.active_batch.get_mut(seq_id).unwrap();
```
If a sequence was aborted or timed out during the forward pass or lookahead teardown, self.active_batch.get_mut(seq_id) returns None. Calling .unwrap() panics on the continuous batching thread, crashing the batch scheduler.
Fix Action: Safely guard against missing sequence entries


inureyes:
## Hardening review

Thanks for the fix. I merged current `main` into the branch and pushed one follow-up commit (`2459e8d0`).

**Finding:** `decode_single_step` had two `active_batch.get_mut(seq_id).unwrap()` calls. This PR guarded the first (before sampling); the second, taken right after the sampled token is evaluated, was still an unwrap. It now uses the same warn-and-return guard, so the function has no unchecked lookups left.

**Reachability note:** on current `main` neither unwrap can fire. The scheduler owns `active_batch` through `&mut self` on a single thread, and nothing between the entry check and these lookups removes a sequence (the one removal on this path, `eval_failures_exhausted`, is followed by an immediate `return`). The guards are defensive hardening against future edits, not a fix for an observed crash, so this is labeled `priority:low`.

**Security / performance:** no new input surface; the guards add one branch on an already-borrowed map lookup per decode step, no measurable cost.

**Validation (GB10, CUDA build):**
- `cargo clippy --release --features cuda --lib -- -D warnings`: clean
- `cargo fmt --check`: clean
- `cargo test --release --features cuda --lib server::batch -- --test-threads=1`: 453 passed, 0 failed, 7 ignored

